Issue #21, opened 2026-04-17
feat: fork abandonment — validators should detect and recover from minority forks


jeffrey-l-turner:
> **Migrated from F1R3FLY-io/f1r3node#459** — originally filed 2026-03-29 by @spreston8.
> This issue was moved as part of the f1r3node → f1r3node-rust deprecation (EPOCH-005).
> Original: https://github.com/F1R3FLY-io/f1r3node/issues/459

## Summary

When a validator ends up on a minority fork (e.g., its block is rejected by the majority due to ReplayCostMismatch or other validation failure), it has no mechanism to detect this and rejoin the majority chain. The validator continues building on its own fork indefinitely, permanently losing its stake weight from consensus.

## Current Behavior

1. V1 creates a block that V2/V3 reject (for any reason — replay mismatch, invalid transaction, etc.)
2. V1 doesn't receive any signal that its block was rejected
3. V1 continues building on top of the rejected block
4. V2/V3 reject all subsequent V1 blocks (state divergence — "Missing mergeable entry")
5. V1 is permanently expelled from consensus with no recovery path
6. At `fault-tolerance-threshold = 0.67`, V2/V3 can finalize without V1 — the network survives but V1 is permanently dead
7. V1 logs "Requesting tips update as newest latest message is more than 60s old" indefinitely

## Expected Behavior

V1 should detect that it's on a minority fork and recover:

1. V1 observes that its blocks are not being referenced in other validators' justifications
2. After a configurable timeout/threshold, V1 recognizes it's on a dead fork
3. V1 rolls back to the last commonly-agreed block (LCA with the majority chain)
4. V1 rebuilds its hot store state from that point and resumes proposing on the majority chain
5. V1's stake weight returns to consensus

## Production Precedent

Every production BFT consensus protocol has a version of this:

- **Tendermint/CometBFT**: State sync — validators catch up by replaying from the latest committed block
- **HotStuff (Diem/Aptos)**: View change protocol forces convergence after leader failure
- **Ethereum Casper FFG + LMD-GHOST**: Fork choice rule follows heaviest chain — validators naturally switch

## Design Considerations

- **Detection**: V1 needs to recognize it's on a minority fork. Possible signals: LFB not advancing while peers are active, own blocks not appearing in peer justifications, receiving valid blocks from peers that build on a different fork
- **Rollback**: V1 needs to reset its RSpace state to the LCA point. This requires the history store to support rollback to a prior root hash
[...]

Issue #5, opened 2026-04-17
inner contract has different variable value than outer


jeffrey-l-turner:
> **Migrated from F1R3FLY-io/f1r3node#37** — originally filed 2025-03-17 by @IhorR-DevBrother.
> This issue was moved as part of the f1r3node → f1r3node-rust deprecation (EPOCH-005).
> Original: https://github.com/F1R3FLY-io/f1r3node/issues/37

Sometimes `loop` has different `payload` value than `service_id-notify-listeners`.
```
@"service_id-listeners"!({})|
contract @"service_id-notify-listeners"(@payload) = {
    new loop, stdOut(`rho:io:stdout`) in {
        stdOut!(["before loop", payload])|
        contract loop(@listeners) = {
            stdOut!(["loop", payload, *loop])
        }|
        for(@listeners <<- @"service_id-listeners") {
            stdOut!(["outer", payload, *loop])|
            loop!("whatever")
        }
    }
}
```
Here is log where payload value got changed:
```
["before loop", "2"]
["outer", "2", Unforgeable(0x7c0f90ef0694d79f30b3f7ea605455f38ae5450e892f5224c5bc38894117389c)]
["loop", "1", Unforgeable(0x7c0f90ef0694d79f30b3f7ea605455f38ae5450e892f5224c5bc38894117389c)]
```

Nama21yo:
I've checked this issue and tested the closure capture behavior in the current codebase. The bug does not reproduce aka each contract instance correctly captures its own payload value:
```
["before loop", 1]  → ["outer", 1]  → ["loop", 1]
["before loop", 2]  → ["outer", 2]  → ["loop", 2]
```

PR #27, opened 2026-04-24
Fix minority fork recovery and add related tests

Nama21yo:
## What changed

  This PR addresses two issues:

  - Fixes validator recovery from minority forks: #21
  - Adds regression coverage for inner contract payload capture behavior: #5

  ### For #21
  - Added a recovery path for validators that get stuck on a stale minority fork.
  - When a validator detects that its own latest message is stale and not finalized, it now leaves `Running` and re-enters the existing initialization flow.
  - Recovery requests `ApprovedBlock` from connected peers directly, so the node can rebuild from the majority chain instead of staying on a dead fork.
  - Updated the engine wiring and test harness so this path is exercised the same way in production and tests.

  ### For #5
  - Investigated the reported inner-contract variable capture issue.
  - The bug was not reproduced in the current codebase.
  - Added compiler/runtime regression tests to lock in the expected behavior and guard against future regressions.

  ## Why it changed

  ### #21
  A validator that produced a block rejected by the majority could keep building on its own fork  forever. The node detected staleness, but it had no recovery action beyond requesting fork-choice tips. This left the validator permanently outside consensus.

  This PR gives the validator a recovery path back to the majority chain.

  ### #5
  The report described an inner contract observing a different payload value than the outer contract. Since that behavior did not reproduce in this branch, the safest change was to add targeted regression tests and preserve the current semantics.

  ## How it was verified

  Ran targeted tests for the changed behavior:

  ```bash
  cargo test -p casper stale_validator_should_transition_to_initializing_and_request_approved_block
  cargo test -p casper validator_on_minority_fork_should_rejoin_majority_chain_after_stale_detection
  ```

  ## Tests included

  - Engine-level recovery for stale validators
  - 3-validator integration test for minority-fork recovery
  - Compiler/runtime regression tests for inner contract payload capture behavior

  ## Follow-up work
  - If issue #5 is observed again, capture the exact failing deploy/context so the runtime path can be reproduced directly.

### DIFF
--- a/casper/src/rust/engine/casper_launch.rs
+++ b/casper/src/rust/engine/casper_launch.rs
@@ -30,9 +30,9 @@ use crate::rust::engine::engine::{
     record_direct_to_running_init_metrics, transition_to_initializing, transition_to_running,
 };
 use crate::rust::engine::engine_cell::EngineCell;
-use crate::rust::engine::running::RunningRecoveryContext;
 use crate::rust::engine::genesis_ceremony_master::GenesisCeremonyMaster;
 use crate::rust::engine::genesis_validator::GenesisValidator;
+use crate::rust::engine::running::RunningRecoveryContext;
 use crate::rust::errors::CasperError;
 use crate::rust::estimator::Estimator;
 use crate::rust::multi_parent_casper_impl::MultiParentCasperImpl;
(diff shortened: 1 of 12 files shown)


Nama21yo:
@jeffrey-l-turner and @spreston8 could you review the PR.

Nama21yo:
Hey @spreston8 can you run the ci once again.

Nama21yo:
Superseded by #58 targeting staging.

PR #3, opened 2026-04-15
migration plan

jeffrey-l-turner:
Automated PR for multi-agent review.

PR #29, opened 2026-04-30
feat: where-clause guards on receives and matches, match-as-expression, first-class if (hard fork)

metaweta:
## Summary

This update requires a hard fork. It changes the protocol buffers, which results in a difference in the hash of blocks.  Clients will need to roll this out in a coordinated way.

This change primarily adds a `where` clause to receives:
```
for (a <- x & b <- y where a + b > 10) { ... }
```
and to match blocks:
```
match <expr> {
  <ptrn1> where <cond1> => <body1>
  ...
  <ptrnn> where <condn> => <bodyn>
}
```

For some reason, Claude first added `where` clauses to each bind
```
for (a <- x where a > 5 & b <- y where b > 10) { ... }
```
and then refactored it, so there's a protobuf field `BindPattern.condition` that got added then dropped.

During the process of creating this change, I discovered that `if` statements were implemented wrong.  They were being desugared as
```
⟦if (<cond>) { <then-case> } else { <else-case> }⟧ =
match <cond> {
  true => <then-case>
  false => <else-case>
}
```
The desugaring works fine for boolean processes, but anything else falls through silently.  This change adds `if` as a fundamental construct that throws a runtime error if `<cond>` is not a boolean process.

I also extended the grammar so we can use`match` blocks as boolean expressions when every RHS is a boolean expression:
```
match <process> {
  <ptrn1> => <bool_expr_1>
  ...
  <ptrnn> => <bool_expr_n>
}
``` 

Find Claude's summary below, followed by suggestions where human review would be helpful.

### Claude's summary

Adds three coordinated Rholang language extensions plus the runtime work to support them:

- **Receive-side `where` guards** on `for`-comprehensions: `for (@x <- chan where x > 0) { ... }` only commits when the spatial pattern matches AND the guard is true. Guard-fail is indistinguishable from a spatial mismatch (atomic).
- **Match-case `where` guards** with fall-through: a non-`true` guard (false, non-bool, error) causes the case to fall through to the next case.
- **Match-as-boolean-expression**: a `match` block whose every case body is a single pure-bool expression compiles to `EMatchExpr` (an `Expr`) so it composes inside `if` conditions, `where` guards, and other expression positions.

Also two prerequisite runtime changes:
- **`if` is now a first-class IR construct** (`If` proto message + `Par.conditionals`) with synchronous `IfConditionTypeError` on non-bool conditions, replacing the previous Match-desugaring with silent no-op.
- **`rho-pure-eval` workspace crate**: a side-effect-free Expr evaluator that both `rholang` and `rspace++` can depend on without creating a cycle. Drives every guard.

This is a hard fork. Six commits change consensus-relevant IR fields (`If`, `Receive.condition`, `MatchCase.guard`, `EMatchExpr`, `BindPattern.condition` (added then dropped), `TaggedContinuation.guard`), and the empty-state and fixed-execution hashes are bumped twice over the series.

## Per-commit walkthrough

**`3115371` docs(plans): add design plan for where-clauses and match guards.** Initial design doc covering the three language extensions plus four prerequisite/architectural decisions: `;` desugars to nested `for`s (already implemented in the normalizer, requiring docs alignment); `if` becomes first-class IR with synchronous type error; receive guards live inside the rspace matcher (Option 3) via a shared `rho-pure-eval` crate; and (initially) a syntactic sublanguage check on guard predicates.

**`1b94fe9` docs(rholang): align ; semantics docs with actual nested-for behavior.** The Rust normalizer at `p_input_normalizer.rs:48-59` already desugars `;`-separated receipts into nested `for`-comprehensions, but two docs and the dining-philosophers example claimed `;` was an atomic-join operator. Aligns the docs and rewrites `tut-philosophers.rho` to use `&` (the actual atomic-join operator). Other `.rho` files using `;` have independent channels so they produce equivalent results either way; left unchanged.

**`a3eb0ed` docs(plans): drop syntactic sublanguage check for guards and if conditions.** Investigation showed `eval_expr` only walks `par.exprs`, so `Send`/`Receive`/`New` in a guard are inert at runtime regardless. Since `rho-pure-eval` mirrors that Expr-only walk, the syntactic forbidden-list is redundant — any program that puts a `Send` in a guard would have it inert anyway. Removes the sublanguage check from the plan and replaces it with documentation of the asymmetric runtime treatment of non-bool guards (receive = match-fail, match-case = fall-through, `if` = synchronous error).

**`1740ca8` feat(rholang): make if a first-class IR construct.** *Consensus-affecting.* New `If` proto message + `Par.conditionals` field; `p_if_normalizer` constructs `If` directly instead of desugaring to `Match { true => P; false => Q }`; new `eval_if` reducer arm raises `IfConditionTypeError` synchronously on non-bool conditions. Threading the new field through ~17 files: `models` (PartialEq/Hash, Score, sort matcher, test fixtures), `rholang::matcher::sub_pars`, `registry_bootstrap`, `Substitute<If>`, etc. Side effects in conditions stay inert (`eval_expr` walks only `par.exprs`); any `Par` with non-empty `sends`/`receives`/etc. now fails fast with `IfConditionTypeError` rather than silently no-op'ing.

**`87dad55` feat(rholang): adapt to rholang-rs `where`-clause AST schema.** Pins `rholang-parser` at the local `feat/where-clauses` branch via a workspace `[patch]` override (commit `988edd6`). The new AST shapes (`Receipt { binds, guard }`, `Case { ..., guard }`) deref to the previous types so most call sites keep working unchanged; the changes are limited to `p_input_normalizer.rs` parameter-type adjustment plus `guard: None` insertions at ~24 `Case` literal sites in normalizer code and test fixtures. No semantic change yet — the parser still emits `guard: None` for everything; Phase 5 wires payloads through.

**`d89e72b` feat(rho-pure-eval): add shared pure expression evaluator crate.** Adds the `rho-pure-eval` workspace crate with `eval(par, env) -> Par` over a de Bruijn `Env<A>`. Implements ground values, `EVar` resolution, bool ops, comparison ops, integer arithmetic with wrapping semantics. Returns `EvalError::UnsupportedExpression` for the bits not yet needed (`EMethodBody`, `EMatchesBody`, `EPercentPercentBody`, `EPlusPlusBody`, `EMinusMinusBody`, `EPathmapBody`, `EZipperBody`). Process-level Par fields are carried through unchanged so "side effects in guards are inert" holds. `Env<A>` is moved out of `rholang::interpreter::env` into `rho_pure_eval::Env`; the rholang module becomes a thin re-export, so existing call sites are unaffected. 28 unit tests.

**`e2a1876` docs(plans): update progress through Phase 4.** Records Phase 2/3/4 commit hashes in §8. Adjusts §3.9 to call out that the rho-pure-eval crate landed as an independently-written initial subset rather than a wholesale extraction of `eval_expr` from `reduce.rs`; that extraction will land as Phases 5/6 surface a need.

**`8d540d2` feat(rholang): normalize where-guards into Receive.condition and MatchCase.guard.** *Consensus-affecting.* New proto fields `Receive.condition = 8` and `MatchCase.guard = 4` (both `Par`, empty = no guard). `p_match_normalizer` and `p_input_normalizer` populate them from the `where` payloads now plumbed through the rholang-rs CPS parser. Sort matchers (`receive_sort_matcher`, `match_sort_matcher`) include the optional guard's score so canonical hashing stays deterministic. The `;`-desugar uses a new `alloc_for_with_guards` so each receipt's guard is preserved on its own nested `for` (the previous version silently dropped guards on all but the last receipt of a sequential `for`). 6 new normalizer tests.

**`e66de77` feat(rholang): match-case fall-through + EMatchExpr.** *Consensus-affecting.* Two related runtime changes. (1) `eval_match` evaluates `MatchCase.guard` (when `Some`) via `rho_pure_eval` against the case's extended env; fires the body iff the result is `GBool(true)`, falls through otherwise. (2) New `EMatchExpr` `Expr` variant (proto tag 38) with full plumbing: `ExprSortMatcher` arm with `Score::EMATCH_EXPR=123`, `has_locally_free`, `get_type`, `pretty_printer`, `web_api`. The normalizer classifies a `Match`: when every case body is a single pure-bool `Expr` (recursively), emits `EMatchExpr` instead of `Match`. New `eval_match_expr` does spatial matching but evaluates the chosen body via `eval_expr` to return a value Par. Side effect: `if (match x { 1 => true _ => false })` now works because the condition's match becomes an `EMatchExpr` in `condition.exprs` and `extract_bool` recognizes the resulting `GBool`.

**`196c0d6` fix(tests): bump empty-state hash and add guard:None to model tests.** Two fallout fixes from Phases 2/5/6 surfaced when the pre-push test suite ran. The `empty_state_hash_should_be_the_same_as_hard_coded_cached_value` test compares the live hash against `RuntimeManager::empty_state_hash_fixed`; bumped because the new IR fields shifted the bootstrap-registry encoding. And `models/tests/par_sort_matcher_test.rs` had ~24 `MatchCase` literals that needed `guard: None` after Phase 5's mandatory-field addition.

**`9c61002` fix(casper-test): bump fixed-rholang post-state hash for hard fork.** Same root cause as `196c0d6`: the fixed-Rholang-term execution hash changed because the new IR fields shifted the canonical encoding of resulting Pars. The test's "without_hard_fork" name now refers to the post-fork baseline.

**`dcee971` chore(deps): bump rholang-parser pin to d5dfe2c.** Drops the local workspace `[patch]` override and updates the per-crate rev pins (`rholang`, `node`, `rspace++`) from `d25f953a` to `d5dfe2c` — the tip of `feat/where-clauses` on rholang-rs, which has both the AST schema (`988edd6`) and the CPS-parser plumbing for guards.

**`2c33567` feat(rspace-matcher): receive guards evaluate inside the matcher (Option 3).** *Consensus-affecting.* Phase 7. `BindPattern.condition` proto field added. `eval_receive` substitutes `Receive.condition` once at depth=1 and replicates it into every `BindPattern`. `Matcher::Match::get` evaluates the per-bind guard via `rho_pure_eval` after spatial match succeeds; returns `None` (guard-fail) on anything-not-`GBool(true)`. Atomicity is preserved: guard-fail looks like a spatial mismatch from outside the matcher. Documented limitation: cross-channel guards in multi-bind joins fail per-bind because each invocation of `Match::get` only sees one bind's variables. Empty-state hash and fixed-execution hash both bump. 3 new runtime tests in `reduce_spec.rs`.

**`48d0280` docs(rholang): tutorial + examples for `where` clauses.** Phase 8. New `## where clauses (guards)` section in `rholangtut.md` covering all three features; new "Receive guards with where" subsection in `08-channels-and-concurrency.md`. Three runnable examples in `/examples/` (`where_receive_guard.rho`, `where_match_fallthrough.rho`, `where_match_as_expression.rho`) plus a parse/normalize sanity test (`where_examples_compile.rs`) so the docs stay in sync. Plan §7.12 expanded to document the multi-bind cross-channel-guard limitation in detail with a path forward.

**`9c4345e` docs(plans): describe Phase 9 — multi-bind cross-channel guards.** Restructures plan §7.12 from a "known limitation" note into a concrete design + step-by-step implementation: move the guard from `BindPattern.condition` to `TaggedContinuation.guard`, add a default-true `check_commit` hook on rspace's `Match` trait, wire it into `extract_first_match` (and the replay mirror) right after the per-channel all-match test. Guard-fail falls through to the existing rollback-and-retry path so atomicity is preserved.

**`532ba07` feat(rspace,rholang): cross-channel where-clause guards (Phase 9).** *Consensus-affecting.* The implementation. Drops `BindPattern.condition`, adds `TaggedContinuation.guard`. Extends `Match<P, A>` → `Match<P, A, K>` with `check_commit(&K, &[A]) -> bool` defaulting to true. Wires the hook into both `extract_first_match` (produce-driven path) and `locked_consume` (consume-driven path). `Matcher::check_commit` concatenates every bind's matched pars in receive order and runs the guard through `rho_pure_eval`. `eval_receive` substitutes once and threads the guard through to `TaggedContinuation`. Hashes bump again. 3 new `cross_bind_guard_*` runtime tests; existing single-bind guard tests still pass.

**`f0c97ed` test(rholang): cover user-facing multi-bind where syntax (Phase 9).** The Phase 9 runtime tests construct the IR directly. This commit adds a parse-and-normalize test plus a parse-and-evaluate test that go through the parser for `for (@x <- a & @y <- b where x + y > 10) { ... }` to confirm the user-facing syntax works end-to-end (the `where` clause attaches to the receipt; binds atomic-joined with `&`).

## Where to focus human review

These are the bits where a thoughtful second pair of eyes would catch the most.

1. **Hash-bump correctness, twice (`196c0d6`, `9c61002`, `2c33567`, `532ba07`).** Three of the consensus-affecting commits update `empty_state_hash_fixed` and the fixed-execution checkpoint. The hashes are derived deterministically from the encoded bootstrap registry and a fixed Rholang term; any future-incompatible IR change anywhere in the IR walk would change them. Reviewers should sanity-check that *only* the intended fields shifted the encoding (no accidental drift from unrelated `Par` reordering, AlwaysEqual changes, etc.) and that the recorded final values match what a from-scratch run produces locally.

2. **Match trait widening from `Match<P, A>` to `Match<P, A, K>` (`532ba07`, `rspace++/src/rspace/match.rs`).** Adding a generic parameter ripples through every `dyn Match<...>` site (rspace.rs, replay_rspace.rs, reporting_rspace.rs, rholang_cli.rs, casper test fixtures, etc.). The default `check_commit` is `true`, so any rspace consumer not implementing it preserves prior behaviour. Worth checking: out-of-tree consumers of rspace++ (if any exist) need a one-line addition to their `impl Match<...>`.

3. **`extract_first_match` and `locked_consume` rollback paths (`532ba07`, `rspace++/src/rspace/space_matcher.rs:121`, `rspace++/src/rspace/rspace.rs:629`).** The two paths into the matcher have to behave identically when the guard rejects: leave data in place, install the wk. The implementation rolls back the `channel_to_index_data` cache in the produce path and just falls into the `_` branch in the consume path. The replay mirror (`replay_rspace.rs`) intentionally was *not* changed — the recorded comm-event drives replay determinism, and the matcher's `check_commit` is a deterministic function of the same inputs. A reviewer with replay context should confirm that's actually safe.

4. **De Bruijn ordering in `Matcher::check_commit` (`532ba07`, `rholang/src/rust/interpreter/matcher/match.rs`).** The cross-channel guard sees concatenated bound vars in receive-bind order: bind 0's vars first, then bind 1's, etc. `rho_pure_eval`'s `Env::put` extends the env so that `BoundVar(0)` resolves to the most recent push, which means the guard for `for (@x <- a & @y <- b where x + y > 10)` reads as `BoundVar(1) + BoundVar(0) > 10`. The Phase 9 tests cover the happy path; a reviewer should verify the ordering matches what the parser/normalizer assigns — particularly under edge cases like remainder-binds (`@x ... @rest`), three-or-more atomic joins, and nested `;`-desugared receipts where each nested `for` has its own (independent) guard.

5. **`EMatchExpr` classification rules in `p_match_normalizer` (`e66de77`).** The normalizer decides whether a `match` becomes `Match` (process) or `EMatchExpr` (expression) by recursing into every case body and checking whether it's a "pure-bool Expr". The set of allowed Exprs is hand-coded (`GBool`, `ENot`, `EAnd`, `EOr`, comparisons, `EMatches`, recursive `EMatchExpr`). If a future Expr variant is added but missed here, the classifier silently picks the process-level path — the program still works, but loses composability inside `if`/`where`. A reviewer should sanity-check the allow-list against the current `Expr` variant list and consider whether to fail-closed (reject unknown) instead.

6. **`rho-pure-eval` arithmetic semantics (`d89e72b`).** `EPlus`/`EMinus`/`EMult` use Rust wrapping arithmetic to match `reduce.rs`. `EDiv`/`EMod` raise `DivisionByZero`. The match between pure-eval and the full reducer matters for deterministic guard outcomes and for replay. Worth a careful read: every binary-arithmetic arm should match the reducer's behaviour byte-for-byte, including i64 overflow, BigInt promotion (or lack thereof), and FixedPoint handling. The 28 unit tests cover the arms but not all overflow corners.

7. **`if`-as-IR error semantics (`1740ca8`).** Pre-Phase-2, `if (foo)` where `foo` was non-bool silently no-op'd because the desugared `Match { true => ... false => ... }` matched neither arm. Post-Phase-2 the same program raises `IfConditionTypeError`. There is exactly one `.rho` file in the corpus that uses `if` (`examples/dupe.rho`); it has a properly bool condition. But on-chain deployed contracts may have non-bool `if` conditions that silently no-op'd before and now fail at runtime. A reviewer should sanity-check that this is the intended semantics (the plan says yes; worth confirming against any deployed contract corpus).

## Test plan

- [ ] Manually run `cargo test --workspace` and confirm all green (already verified locally: 143 rholang lib + 299 reduce_spec + 5+ integration suites + 342 casper main + 41 rspace).
- [ ] Verify the empty-state hash matches a from-scratch local computation (no replay).
- [ ] Run the three `examples/where_*.rho` files via `rholang-cli` and confirm expected output.
- [ ] On a follow-up node bringup, confirm `rho:execution:abort` and other system processes still work (their patterns plumb through `BindPattern` and `TaggedContinuation`).
- [ ] Replay an existing chain segment from a node running `master` against this branch and confirm the replay does not commit anything new (sanity check that the IR shifts don't accidentally accept old blocks under the new schema).


spreston8:
Can you open a PR in `rholang-rs` so those changes can get reviewed and merged into `master` there? Then we can pin the crate in here to the `master` commit.

spreston8:
# PR #29 review summary — `feat/where-clauses-and-match-guards`

Repo: `F1R3FLY-io/f1r3node-rust`
Branch: `feat/where-clauses-and-match-guards` at `f0c97ed`
Base: `dev` at `d33e5234`
Stats: 77 files, +4039/-418, hard fork

> **This is the first round of review.** Subsequent rounds are likely as the cross-repo `rholang-rs` PR opens, fixes land, and follow-up surfaces (e.g., a deeper `rho-pure-eval` ↔ `reduce.rs` parity audit, replay-determinism stress tests, integration-test results) come into scope. Treat this as a starting point, not an exhaustive list.

## TL;DR

| | |
|---|---|
| Test suite | **2128 passed, 0 failed, 31 ignored** (`cargo test --workspace --release`, no warnings) |
| Empty-state hash live↔fixed test | ✅ passes (`empty_state_hash_should_be_the_same_as_hard_coded_cached_value`) |
| Must-fix before merge | **2 HIGH** in `rho-pure-eval` |
| Strongly recommended | 6 Medium (architectural / arithmetic / cross-repo) |
| Cross-repo coordination | **rholang-rs `feat/where-clauses` has no open PR** |

The two HIGH items are small, bounded fixes (~5–10 lines each plus tests). Everything else is incremental.

---

## Must fix before merge

### HIGH-1 — `i64::MIN / -1` and unary `-i64::MIN` panic in `rho-pure-eval`

**File:** `rho-pure-eval/src/eval.rs:408-425` (`int_div_or_mod`) and `:108-116` (`ENegBody` arm)

`int_div_or_mod` checks for division by zero but not for `i64::MIN / -1` overflow. Native `/` on i64 with these operands panics in debug and is undefined behavior in release. The reducer explicitly handles this case:

```rust
// reduce.rs:1637-1641 — reducer does this
if lhs == i64::MIN && rhs == -1 {
    return Err(InterpreterError::ReduceError(
        "Arithmetic overflow in division".to_string(),
    ));
}
```

Same problem in `ENegBody` at `eval.rs:113` — uses `-i` (panics on `i64::MIN`); reducer uses `checked_neg` (`reduce.rs:1505`).

**Why it matters:** the matcher's `check_commit` invokes pure-eval. A receive guard like `where x / -1 == 0` where `x` resolves to `i64::MIN` panics inside the matcher. The panic propagates up through `unsafe impl Send` boundaries — at minimum, you get a poisoned mutex and an unrelated error elsewhere; at worst, node crash. **Attacker can trigger by deploying a contract with such a guard.**

**Fix (small):**
```rust
// In int_div_or_mod, add before the GInt/GInt arm:
(ExprInstance::GInt(&i64::MIN), ExprInstance::GInt(-1)) if op == "/" =>
    Err(EvalError::Overflow { op }),
// In ENegBody arm:
ExprInstance::GInt(i) => i.checked_neg()
    .map(par_with_int)
    .ok_or(EvalError::Overflow { op: "-" }),
```

Plus regression tests with `i64::MIN`.

---

### HIGH-2 — NaN equality differs between `rho-pure-eval` and reducer

**File:** `rho-pure-eval/src/eval.rs:357-367` (`eq_binop`) vs `reduce.rs:2043-2074` (`EEqBody` / `ENeqBody`)

Pure-eval uses `Par::PartialEq` directly:
```rust
let eq = lhs == rhs;
```

Reducer implements IEEE 754 NaN semantics:
```rust
let result = if par_contains_nan_double(&sv1) || par_contains_nan_double(&sv2) {
    false  // NaN == anything is always false
} else {
    sv1 == sv2
};
```

`GDouble` is `fixed64` in proto (raw bits storage). Two NaNs with the same bits compare byte-equal under `Par::PartialEq` → pure-eval returns true; reducer returns false.

**Why it matters:** same `where x == x` (where `x` is NaN) returns different answers in `Match` context (uses pure-eval) vs `EMatchExpr` context (uses full reducer). Same syntax, different semantics depending on classification. Cross-validator: each validator runs the same code path so consensus holds, but user-facing semantics are inconsistent.

**Fix (small):** mirror the reducer's NaN check inside `eq_binop`. ~5 lines.

Add regression tests covering NaN equality and inequality.

---

## Strongly recommended before merge

### M1 — Asymmetric guard evaluators (architectural)

`MatchCase.guard` is the same field, but it's evaluated by **two different evaluators** depending on classification context:

| Context | Evaluator | Operator surface | Error mode |
|---------|-----------|-------------------|------------|
| `Match` (process) — `eval_match` line 1102-1108 | `rho_pure_eval::eval` | GInt only for arithmetic; GInt/GString/GBool for comparison | silent fall-through |
| `EMatchExpr` (expression) — `eval_match_expr` line 1347-1352 | `self.eval_expr` (full reducer) | GInt, GDouble, GBigInt, GBigRat, GFixedPoint | propagated via `?` |

**Concrete divergences** the user can observe:

- `where bigint_a + bigint_b > 0` works in `EMatchExpr`, silently falls through in `Match`.
- `where x.length() > 0` (method call) works in `EMatchExpr`, falls through in `Match` (`EMethodBody` is unsupported in pure-eval).
- A bug in the guard (division by zero, type mismatch) errors in `EMatchExpr`, silent fall-through in `Match`.
- Connective patterns (`ConnBool`, `ConnInt`) work in `EMatchExpr`, silently fail to match in pure-eval's `simple_pattern_match` because it does structural equality only (`eval.rs:261-267`).

**The plan §3.5 said "guards are in the same sublanguage as receive guards" — implying both should use pure-eval.** The implementation deviates: `eval_match_expr` uses the full reducer.

**Long-term fix:** pick one evaluator everywhere. Either grow pure-eval toward full `eval_expr` parity (substantial work, plan §3.9 acknowledges this trajectory), or use the full reducer in both contexts (gives up the side-effect-free type-system guarantee).

**Short-term mitigation:** document the supported subset in user-facing docs; consider a parser-time warning when a guard contains operators outside pure-eval's surface.

---

### M2 — `eval_match` doesn't collapse `Some(empty Par)` to "no guard"

**File:** `reduce.rs:1102-1108` (`eval_match`) and `reduce.rs:1347-1352` (`eval_match_expr`)

The codebase has **four guard-evaluation sites** with inconsistent handling of `Some(Par::default())`:

| Site | Treats `Some(empty)` as "no guard"? |
|------|--------------------------------------|
| `eval_receive` (`reduce.rs:969-974`) | ✅ yes (`Some(c) if c != &Par::default()`) |
| `Matcher::check_commit` (`matcher/match.rs:75`) | ✅ yes (`is_empty_par` check) |
| `eval_match` | ❌ no (eval, then `extract_bool != Some(true)` → fall-through) |
| `eval_match_expr` | ❌ no (same pattern) |

**Concrete divergence:** `match x { y where Nil => p _ => q }` evaluates `q` (fall-through). `for (...) where Nil { p }` fires `p`. Same `where Nil` syntax, different behavior.

The triggering path is user-written `where Nil` — `p_match_normalizer.rs:91` produces `Some(Par::default())` when a user writes any `where` clause whose body normalizes to empty Par.

**Fix:** add the same short-circuit to `eval_match` and `eval_match_expr`:
```rust
let guard_passes = match &single_case.guard {
    Some(g) if g != &Par::default() => match rho_pure_eval::eval(g, &case_env) { ... },
    _ => true,
};
```

---

### M3 — Sort matcher leaks `None` vs `Some(empty Par)` to the wire

**Files:** `models/src/rust/rholang/sorter/{match_sort_matcher.rs, receive_sort_matcher.rs, expr_sort_matcher.rs}`

All three sort matchers handle the new optional Par fields with this pattern:

```rust
let guard_par = match_case.guard.clone().unwrap_or_default();
let sorted_guard = ParSortMatcher::sort_match(&guard_par);

// Output term:
guard: match_case.guard.as_ref().map(|_| sorted_guard.term.clone()),

// Score:
score: Tree::Node(vec![..., sorted_guard.score]),
```

Score-tree determinism is preserved (same score for `None` and `Some(empty)`). But the **output term preserves the distinction**:
- `MatchCase { guard: None }` → field absent on wire
- `MatchCase { guard: Some(empty Par) }` → tag + length-0 on wire (~2 bytes)

**Practical risk today:** only user-written `where Nil` triggers `Some(empty)`, and all validators see the same source → same representation → no actual divergence. But fragile against future construction-site changes.

**Fix (defensive):** collapse at sort matcher boundary:
```rust
guard: match_case.guard.as_ref()
    .filter(|p| *p != &Par::default())
    .map(|_| sorted_guard.term.clone()),
```
Or enforce `None`-only at every construction site.

---

### M4 — `rholang-parser` pin is a feature-branch SHA with no open upstream PR

**Files:** `rholang/Cargo.toml:46`, `rspace++/Cargo.toml:45`, `node/Cargo.toml:45`

All three pin to `rev = "d5dfe2c"` (tip of `feat/where-clauses` on `F1R3FLY-io/rholang-rs`).

**As of this review:**
- `feat/where-clauses` exists on rholang-rs remote with two commits (`988edd6` AST, `d5dfe2c` CPS plumbing).
- **No open PR** for that branch on rholang-rs.
- Cargo.lock pins to full SHA `d5dfe2caf9c787afd15a1a919c7132f757efa0da`.

**Risks:**
1. Reproducibility — if anyone force-pushes / rebases / deletes the branch before it's merged, the SHA becomes orphan and is eventually GC'd. Builds break.
2. Upstream review — the parser changes haven't been reviewed via a GitHub PR.
3. Coordination — if rholang-rs maintainers want changes, those land on `feat/where-clauses` with a new SHA, requiring a pin bump here.

**Recommended workflow before merge:**
1. Open a PR on rholang-rs for `feat/where-clauses` → master.
2. Get it reviewed and squash-merged.
3. Bump the three Cargo.toml pins here to the merged-master SHA.
4. Re-run CI; merge this PR.

---

### M5 — Arithmetic overflow handling diverges (wrapping vs checked)

**File:** `rho-pure-eval/src/eval.rs:141-148` vs `reduce.rs::EMultBody`/`EPlusBody`/`EMinusBody`

Pure-eval uses `wrapping_*`:
```rust
ExprInstance::EPlusBody(EPlus { p1, p2 }) =>
    int_binop("+", ..., i64::wrapping_add),
ExprInstance::EMinusBody(EMinus { p1, p2 }) =>
    int_binop("-", ..., i64::wrapping_sub),
ExprInstance::EMultBody(EMult { p1, p2 }) =>
    int_binop("*", ..., i64::wrapping_mul),
```

Reducer uses `checked_*` (`reduce.rs:1559` for mult, similar for plus/minus) and errors on overflow:
```rust
let result = lhs.checked_mul(rhs).ok_or_else(|| {
    InterpreterError::ReduceError("Arithmetic overflow in multiplication".to_string())
})?;
```

**Concrete divergence:**
- `where x + 1 > 0` where `x = i64::MAX`:
  - In `Match` context (pure-eval): `i64::MAX.wrapping_add(1)` returns `i64::MIN`. Then `i64::MIN > 0` is `false`. Guard fails silently with no overflow signal.
  - In `EMatchExpr` context (full reducer): errors with "Arithmetic overflow in addition" — error propagated via `?`.

Same `MatchCase.guard` field, different runtime answers for overflow inputs. Within-context: each is deterministic. Cross-context: silently inconsistent.

This compounds with M1 (asymmetric evaluators) — the reducer correctly errors; pure-eval silently produces a "wrong" answer that downstream guards then build on.

**Fix:** mirror the reducer's `checked_*` approach in pure-eval:
```rust
ExprInstance::EPlusBody(EPlus { p1, p2 }) =>
    int_binop_checked("+", ..., i64::checked_add),
// ...
fn int_binop_checked<F>(op: ..., f: F) -> Result<Par, EvalError>
where F: Fn(i64, i64) -> Option<i64> {
    ...
    match (&v1, &v2) {
        (ExprInstance::GInt(i1), ExprInstance::GInt(i2)) => f(*i1, *i2)
            .map(par_with_int)
            .ok_or(EvalError::Overflow { op }),
        ...
    }
}
```

Bundle with HIGH-1's regression tests (`i64::MIN`/`i64::MAX` boundary cases).

---

### M6 — `ReplayRSpace` does call `check_commit` (PR description framing is misleading)

**File:** `rspace++/src/rspace/replay_rspace.rs:145` (impl SpaceMatcher)

The PR description's "where to focus" item #3 says:
> "The replay mirror was *not* changed — the recorded comm-event drives replay determinism, and the matcher's `check_commit` is a deterministic function of the same inputs."

The first half is technically false. `ReplayRSpace` implements `SpaceMatcher`, which provides `extract_first_match` (defined in `space_matcher.rs:121`). That method calls `matcher.check_commit(...)` at line 152. So replay DOES exercise the new code path.

**Determinism still holds** because:
1. Replay's `fetch_matching_continuations` filters by `comm.consume == wc.source` (no shuffle) — narrows to exactly the originally-matched continuation.
2. Replay's `fetch_matching_data` filters by `comm.produces.contains(&datum.source)` — narrows to exactly the originally-matched data.
3. `check_commit` is deterministic given the same `(continuation.guard, matched_data)`.

So replay reaches the same answer as the original execution. **But the framing in the PR description should be updated** to acknowledge the inheritance and document the determinism preconditions explicitly:
- `TaggedContinuation.guard` is stable in storage.
- COMM-event-driven filtering of candidates and data prevents shuffle-induced divergence.
- `rho_pure_eval::eval` is deterministic (modulo HIGH-1's panic case).

**Concrete vector for replay divergence: HIGH-1.** If a historical guard contains `i64::MIN / -1`, original execution may have crashed; replay panics in the same place. Same outcome (no commit), but ugly. Fixing HIGH-1 closes this off.

---

## Test gaps to close

The author claimed in the PR description:
> "Manually run `cargo test --workspace` and confirm all green (already verified locally...)"

Confirmed: 2128 passed, 0 failed.

But these explicit test gaps remain:

### T1 — Pure-eval edge cases

`rho-pure-eval/src/tests.rs` has 28 unit tests but NO coverage of:
- `i64::MIN / -1` (HIGH-1)
- NaN equality and inequality (HIGH-2)
- `i64::MIN` unary negation (HIGH-1 bundle)
- Arithmetic overflow (`i64::MAX + 1` etc.)

Add these alongside the HIGH fixes.

### T2 — Cross-bind De Bruijn ordering test is too weak

`reduce_spec.rs::cross_bind_guard_should_commit_when_combined_predicate_holds` and the end-to-end `where_examples_compile.rs::where_multi_bind_atomic_join_evaluates` both use **commutative** guards (`x + y > 10`). If the parser/normalizer assigned indices in reverse order, the test would still pass because addition is commutative.

**Recommended:** add a non-commutative cross-bind test like:
```
for (@x <- a & @y <- b where x > 100 && y < 100) { stdout!(("ok", x, y)) }
// data: a!(200) | b!(50)
// correct: x=200, y=50 → fires
// reversed indices: x=50, y=200 → does NOT fire
```
Capture stdout output to verify the captured values match the producer's values, not just `errors.is_empty()`.

Same for: remainder-binds (`@x ... @rest`), three-or-more atomic joins, nested `;`-desugared receipts (each with independent guards).

---

## Cosmetic cleanup (bundle into one follow-up commit)

These don't block merge but are easy to fix and worth doing. None affect runtime behavior.

### C1 — Stale doc comments in `rho-pure-eval`

**`rho-pure-eval/src/lib.rs:17-19`** claims `EMatchExpr` unsupported, but `eval.rs:179` actually handles it via `eval_match_expr`. Drop the `EMatchExpr` mention; clarify that the unsupported set is `EMethodBody`, `EMatchesBody`, `EPercentPercentBody`, `EPlusPlusBody`, `EMinusMinusBody`, `EPathmapBody`, `EZipperBody`.

**`rho-pure-eval/src/eval.rs:188-190`** says "Phase 7 will move spatial matching into a place this crate can call." Phase 7 already shipped in this PR. Drop the sentence.

**`rho-pure-eval/src/env.rs:3`**: `//! Mirrors the original \`coop.rchain.rholang.interpreter.Env\` from Scala.` Drop the Scala reference; describe what `Env<A>` does (HashMap-backed, level + shift, De Bruijn lookup).

### C2 — `rholang/src/rust/interpreter/env.rs:3-7` comment is wrong on two counts

Says "rspace++ (which evaluates guards inside the matcher in Phase 6)". Two issues:
- Phase 7 (matcher integration), not Phase 6.
- The matcher impl that calls `rho_pure_eval::eval` lives in `rholang::interpreter::matcher::match.rs`, not in rspace++. rspace++ defines the trait but doesn't depend on rho-pure-eval (verified via Cargo.toml — only rholang has the dependency).

### C3 — `BindPattern.condition` field tag 4 not reserved

**File:** `models/src/main/protobuf/RhoTypes.proto:136-140`

Phase 7 added `BindPattern.condition = 4`; Phase 9 dropped it. Current `BindPattern`:
```protobuf
message BindPattern {
  repeated Par patterns = 1;
  Var remainder = 2;
  int32 freeCount = 3;
}
```

No `condition` field, no `reserved 4;` declaration. Proto3 best practice when dropping a field is to reserve the tag so future code can't accidentally re-use it with different semantics. Defensive — not load-bearing today.

### C4 — `EMatchExpr` classification excludes `EVarBody` and `EMethodBody` at top level

**File:** `rholang/src/rust/interpreter/compiler/normalizer/processes/p_match_normalizer.rs:159-194`

`par_is_pure_bool_expr` doesn't recognize:
- `EVarBody` — a bound variable that resolves to a bool can't be statically known. Excluding is conservative-correct; consequence is `match x { y => y _ => false }` (where `y` is a bound bool var) emits process-level `Match`, not `EMatchExpr`. User has to wrap: `match x { y => y == true _ => false }`.
- `EMethodBody` — a method call that returns bool can't be statically known. Same workaround.

This is a usability nit, not a bug. Worth a code comment explaining the rationale so future readers don't try to "fix" it by adding these variants (which would require runtime type checking the result).

### C5 — `.unwrap()` → `.expect()` in `reduce.rs`

7 `.unwrap()` calls in PR-added code on protobuf `no_box=true` fields. All safe by construction. Replace with `.expect("normalizer post-condition")` or `.expect("protobuf no_box invariant")` for diagnostics. Locations: `eval_match` line 1115; `eval_if` lines 1143, 1150, 1154; `eval_match_expr` lines 1326, 1354.

---

## Author-flagged items — review verdicts

The PR description called out 7 review focus areas. Verdicts:

| # | Area | Verdict |
|---|------|---------|
| 1 | Hash-bump correctness | ✅ Live↔fixed test passes (verified). Mechanism sound. |
| 2 | `Match<P,A>` → `Match<P,A,K>` widening | ✅ Clean within workspace. Out-of-tree consumers need to add the `K` parameter — worth a CHANGELOG note. |
| 3 | `extract_first_match` / `locked_consume` rollback paths | ✅ Both behave correctly on guard reject. **Replay framing needs correction (see M6).** |
| 4 | De Bruijn ordering | ✅ Math correct. **Test coverage too weak (see T2).** |
| 5 | `EMatchExpr` classification rules | ✅ Allow-list is conservative-correct (fail-open on unknown variants). Two usability footnotes (see C4). |
| 6 | `rho-pure-eval` arithmetic semantics | ❌ **HIGH-1 (panic on `i64::MIN / -1`); HIGH-2 (NaN equality); M5 (wrapping vs checked overflow).** Plus M1 (numeric type asymmetry vs reducer). |
| 7 | `if`-as-IR error semantics | ✅ Corpus clean (3 `.rho` files, all bool conditions). **Production-chain sweep is operator-side** — each operator must verify their on-chain contracts before upgrading. |

---

## Summary checklist

**Before merge:**
- [ ] HIGH-1: Add `i64::MIN / -1` and `-i64::MIN` checks to pure-eval; add tests.
- [ ] HIGH-2: NaN-aware `eq_binop` in pure-eval; add tests.
- [ ] M5: Mirror reducer's `checked_*` arithmetic in pure-eval; tests for `i64::MAX + 1` and similar.
- [ ] M6: Update PR description focus #3 to acknowledge replay inherits `check_commit`.
- [ ] M4: Open + merge rholang-rs PR for `feat/where-clauses`; bump pins to merged SHA.
- [ ] T2: Add non-commutative cross-bind De Bruijn test.

**Strongly recommended (decide explicitly even if deferring):**
- [ ] M1: Decide on architectural fix for Match vs EMatchExpr asymmetry. At minimum, document the supported subset.
- [ ] M2: Add `Some(empty Par)` collapse to `eval_match` / `eval_match_expr` for consistency with `eval_receive` / `check_commit`.
- [ ] M3: Defensive collapse in sort matchers OR enforce None-only at construction.

**Bundle into one cleanup commit:**
- [ ] C1–C5: stale doc comments, classification footnote, `.unwrap()` → `.expect()`, BindPattern reserved tag.

**Operator-side action:**
- [ ] Sweep on-chain contracts for `if (non_bool)` patterns before upgrading.

**For follow-up PR:**
- [ ] T1: Pure-eval edge-case tests (bundle with HIGH fixes).

---

PR #76, opened 2026-06-11
fix: resolve HTTP API stall under sustained polling — RwLock + multi-thread runtime + spawn_blocking

### LINKED ISSUE #73: Multi-threaded HTTP runtime for observer API
## Problem
The HTTP API server (:40403) becomes completely unresponsive under sustained polling (1 req/s from a testbed monitor). TCP connections are accepted (kernel-level), but the HTTP handler never sends a response — clients receive 0 bytes and timeout.

### Root Cause
1. Single-threaded HTTP runtime

node/src/rust/runtime/servers_instances.rs (~line 280):

```rust
let http_server_handle = tokio::task::spawn_blocking(move || {
    let rt = tokio::runtime::Builder::new_current_thread()
        .enable_all()
        .build()?;
    rt.block_on(async move {
        axum::serve(listener, http_router).await
    })
});
```

The entire HTTP server runs on a single OS thread. If that thread blocks, all HTTP connections stall.

semyon-dev:
## Summary

The HTTP API server becomes completely unresponsive under sustained polling. TCP connections are accepted (kernel-level), but the HTTP handler never sends a response — clients receive 0 bytes and timeout.

Closes #73

## Root Cause

Three compounding issues:

1. **Single-threaded HTTP runtime** — `servers_instances.rs` used `tokio::runtime::Builder::new_current_thread()` for both HTTP and Admin HTTP servers. A single blocking call stalls every connection.

2. **`global_lock` was `Mutex<()>`** — `BlockDagKeyValueStorage::global_lock` serialized all reads and writes. Every call to `get_representation()` (the hot path for all read endpoints) acquired an exclusive lock, even though it only reads data.

3. **Blocking handlers on async workers** — All HTTP handlers called into `BlockDagKeyValueStorage` (which acquires `global_lock` and does synchronous LMDB I/O) directly on async worker threads, blocking them.

## Changes

- **`block-storage`**: `global_lock` changed from `Mutex<()>` to `RwLock<()>`. `get_representation()` uses `.read()` (concurrent readers), write-path methods use `.write()`. This allows HTTP read handlers to proceed in parallel while no write is in progress.
- **`node/runtime`**: HTTP runtime changed from `new_current_thread()` to `new_multi_thread().worker_threads(4)`. A single blocking call no longer stalls all connections.
- **`node/web`**: all HTTP handlers route through a reusable `offload()` helper that wraps `tokio::task::spawn_blocking` + `block_on`. Blocking DAG/LMDB operations are moved off async worker threads onto the blocking thread pool.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -295,6 +295,12 @@ jobs:
     # the branch into the canonical repo (e.g. staging).
     if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
+    # Gates every CI run on the ephemeral-launch environment so OCI_* +
+    # RELEASE_PAT secrets (which live env-scoped) flow to the job after
+    # required-reviewer approval. Applies to internal pushes/PRs and fork
+    # PRs alike — fork PRs need the environment gate for any secret access
+    # at all; the friction on internal runs is the accepted trade-off.
+    environment: ephemeral-launch
     permissions:
       actions: write # needed for gh api ... actions/runners/registration-token
     timeout-minutes: 10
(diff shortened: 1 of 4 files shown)


spreston8:
LGTM. Just need to wait for CI to be resolved on our side.

semyon-dev:
I opened new PR into staging branch without conflicts: https://github.com/F1R3FLY-io/f1r3node-rust/pull/113

Issue #36, opened 2026-05-18
Improve API error handling: structured JSON and proper status codes (4xx/5xx)


Timekiller7:
### What did you expect to see?

The `/api/epoch/rewards` endpoint (and other endpoints that run exploratory deploys) should return errors in a structured, predictable form:

- a meaningful HTTP status code that reflects the nature of the failure (`4xx` for caller-fixable problems, `5xx` for node-side problems);
- a machine-readable `error` field clients can switch on;
- a human-readable `message` that does not expose internal interpreter implementation details.

Specifically, when a Rholang execution failure occurs (arithmetic overflow, division by zero, out of phlo, etc.), the API should return 500 status (if caused by network desync, for ex) with a clear `error_kind`, so a client can tell apart "your term cannot be executed against current state" from "the node itself is broken".

### What did you see instead?

The endpoint returns `400 Bad Request` for **every** kind of failure, with a plain-text body that leaks the full internal error chain — including the `BugFoundError` wrap that can misrepresent errors.

Example response from the `/api/epoch/rewards` endpoint on a desynced observer:

```bash
Error: HTTP 400 Bad Request Something went wrong: Interpreter error: Bug found: Unexpected error while capturing results from Rholang: Interpreter error: Reduce error: Arithmetic overflow in multiplication
```

The error is caused by server-side state (network desync), not by an invalid client request, yet the API returns 400 Bad Request. Also, the error message is too large and exposes internal implementation details.

### What version of F1R3FLY Node are you using?
rust/dev, tag: rust-v0.4.15

Timekiller7:
I’ve picked this up and I’m currently working on it

Timekiller7:
implemented in #78

Issue #35, opened 2026-05-18
No validator public key validation in the /api/validator endpoint


Timekiller7:
<img width="1175" height="230" alt="Image" src="https://github.com/user-attachments/assets/6e33cce5-5c7f-4f24-ba07-43dc659235b9" />

### What did you expect to see?
When calling the `GET /api/validator/{some_invalid_pub_key}` endpoint with an invalid validator public key, the API should return a `400` or `404` status code with an error message indicating that the validator public key is invalid. This would confirm that the validator public key is properly validated

### What did you see instead?
The API returns a `200` status code, and the invalid public key is treated as a valid one. The response includes information such as whether it is bonded and how much stake it has

### Steps to reproduce the bug
Call the observer node API endpoint:
`GET /api/validator/{some_invalid_address}`

### What version of F1R3FLY Node are you using?
rust/dev, tag: rust-v0.4.15

Timekiller7:
I’ve picked this up and I’m currently working on it

Timekiller7:
implemented in #78

PR #78, opened 2026-06-11
Fix API issues: inconsistency, validation, error handling 

Timekiller7:
Fix of issues #35 and #36


spreston8:
Took a pass through the diff against `staging` and traced the error paths through `block_api.rs`, `shared_handlers.rs`, and the gRPC surface. Direction's good and the classifier design is nice (the `InterpreterError`/`CasperError` matches are exhaustive with no `_` arm, so new variants force a conscious mapping). A few findings, mostly spots where the documented contract doesn't hold yet at the call site:

- **Exploratory deploy by block hash, unknown block** — `block_api.rs:1680` and `:1694` still `Err(eyre!("Can not find block {:?}", block_hash))`. Docs promise 404 `block_not_found` for `/api/explore-deploy-by-block-hash`; actual = 500 `unknown_error` (and a status regression from the old 400). Two-line fix with the PR's own `BlockNotFoundError`.
- **POST /api/deploy validation chain** — `block_api.rs:416-481` is `Result<(), String>` → `eyre!(string)`: read-only-node rejection, wrong `shardId`, forbidden system key, `phlo_price` below minimum, expired deploy. All five are client errors; all return 500 `unknown_error`. Docs promise 400 for invalid field values on the most-used endpoint. Bitter detail: the expired case constructs `DeployExpiredError` and immediately `.to_string()`s it (`:468-474`) — the comment says "for consistent error classification" but the stringification is exactly what defeats the classifier. (Chain predates the PR, but the PR documents the 400 contract; converting it to typed errors is small and uses the PR's own pattern.)
- **POST /api/data-at-name-by-block-hash** — `get_data_at_par` (`block_api.rs:1738`) uses plain `eyre!("Invalid block hash")` (docs: 400 `invalid_hash`; actual: 500), and `:1740` calls `block_store().get_unsafe(...)` which double-`.expect()`s (`key_value_block_store.rs:86`) — a syntactically-valid but unknown block hash panics the handler. Docs claim 404 `block_not_found`. Pre-existing panic, but it's now a documented-as-404 endpoint; should become `get` + `BlockNotFoundError`.
- **Negative `nameQty` is a remote unbounded-allocation DoS** (pre-existing, newly documented) — `preview_private_names` (`block_api.rs:1467`): `let safe_qty = std::cmp::min(name_qty, 1024) as usize;`. For client-supplied `nameQty: -1` (i32, straight from JSON): `min(-1, 1024) = -1`, and `-1 as usize` = `usize::MAX` → `(0..usize::MAX).map(|_| rand.next()…).collect::<Vec<_>>()` — unbounded CPU+memory on a public, unauthenticated endpoint this PR now advertises. One-line fix: `name_qty.clamp(0, 1024) as usize`, plus a regression test for `nameQty: -1` → 400 or empty.
- **Error-kind enum is incomplete vs the "stable, switch on it" contract** — the doc-comment on `ApiErrorResponse` says the `error` kinds are "stable across node versions — safe to switch on", and api-reference.md repeats the enum — but `classify_casper_error` emits 6 kinds absent from both lists: `kv_store_error`, `history_error`, `system_runtime_error`, `stream_error`, `lock_error`, `other_error`. Also emitted but undocumented: `endpoint_not_found`, `method_not_allowed` (router fallbacks) and `invalid_hex` (trace_handler — which should arguably be `invalid_hash` for consistency with the rest of the API). Either document all emitted kinds or collapse the rare internal ones.
- **Most of the annotation work is invisible in the served OpenAPI** — `web_api_docs.rs:8-44` registers only `status`, `deploy`, `explore-deploy`, `explore-deploy-by-block-hash`, `blocks`, `block` (+ `propose` on admin) in `paths(...)`. Every other endpoint this PR re-annotated — balance, registry, validators, validator, bond-status, epoch, epoch/rewards, estimate-cost, find-deploy, is-finalized, deploy-finalization-status, prepare-deploy, data-at-name-by-block-hash, last-finalized-block, blocks-by-height/depth, reporting/trace — is absent from both OpenApi derives, so none of those annotations reach swagger-ui / openapi.json. Pre-existing gap, but it strands the PR's main documentation effort; fix is one line per route in `paths(...)`.
- **Admin `/propose` annotation is aspirational** (verified against code) — read-only node → `eyre!("Propose error: read-only node.")` (`admin_web_api.rs:51`) → 500 `unknown_error` (annotation is silent on this; the old one said 403). "No new deploys"/constraint failures surface from `BlockAPI::create_block` as plain strings → 500 `unknown_error`, not the annotated 400 `unknown_error`. Either type the propose failure modes (read-only + no-new-deploys are client-class) or correct the annotation and the api-reference.md propose table.
- **Warn-spam** — `AppError::into_response` logs `tracing::warn!("API error: {:#}", ...)` for every error including routine client 4xx; a misbehaving poller turns the log into noise. Suggest warn for 5xx / debug for 4xx, and a stable `target:` consistent with the #67 logging standardization.
- **Transport asymmetry on the #35 fix** — gRPC `bond_status` (`deploy_grpc_service_v1.rs:732`) passes `request.public_key` bytes straight to `BlockAPI::bond_status`, so the gRPC surface still has the original #35 behavior (garbage in → `isBonded:false`). Pushing the 65-byte/curve check down into `BlockAPI::bond_status` (bytes variant) would fix both transports at once.
- **Note (out of scope, pre-existing)** — `play_exploratory_par` swallows execution errors into `Ok(Vec::new())` with only an error log, so the bonds/active-validator readers can't surface interpreter failures. Same silent-default class worth a follow-up.

Doc-side cleanups these imply: complete the error-kind enum in both the `ApiErrorResponse` doc-comment and api-reference.md (the 6 casper kinds + `endpoint_not_found`/`method_not_allowed`) and rename `invalid_hex` → `invalid_hash`; fix the prepare-deploy annotation (`Use seqNumber as validAfterBlockNumber` → it's the validator's next sequence number; api-reference.md already says this, so the two disagree); correct the `/propose` table/annotation to match actual statuses; register the re-annotated routes in `web_api_docs.rs` `paths(...)` so the new status tables actually reach the served OpenAPI. The status tables for the four endpoints in the first findings become accurate once those call-site conversions land — worth re-checking each against the converted code.

spreston8 (approved):
Re-reviewed `d3ac9602..5bd06ae6` against the earlier findings — the revision resolves both HIGH items (the client-error paths now return the documented 400/404 via typed errors, including eliminating the `get_unsafe` panic on `/api/data-at-name-by-block-hash`, and `nameQty` is now clamped) plus the error-kind enum completeness, the warn/debug level split, the OpenAPI `paths()` registration, gRPC `bond_status` pubkey parity, and the prepare-deploy doc correction. Nicely done — approving.

Two small follow-ups, non-blocking:

- **`/propose` annotation vs behavior:** `NoNewDeploysError` maps to **409** in `classify_error` (and `api-reference.md` documents 409), but the utoipa annotation in `admin_web_api_routes.rs` has no 409 entry and instead lists `no_new_deploys` under 500. Add the 409 row and drop it from the 500 line so the OpenAPI matches the code.
- **Internal names leak into client messages:** `validate_and_decode_pubkey` puts function-name prefixes into the client-facing `message` (`"decode: …"`, `"validate_secp256k1_bytes: …"`). Minor re-introduction of the internal-detail leak #36 set out to remove — drop the prefixes for a clean message.

PR #86, opened 2026-06-25
feat(api): make http body limit configurable via config

semyon-dev:
Makes the HTTP API request body size limit configurable via a new api-server.http-max-body-bytes setting, bringing parity with the existing grpc-max-recv-message-size knob.


semyon-dev:
i am reopened PR from staging branch https://github.com/F1R3FLY-io/f1r3node-rust/pull/98

PR #91, opened 2026-06-30
feat: adopt rholang-rs method-call + agent block sugar

metaweta:
## Summary

Adopts the two parse-time desugarings landed upstream in rholang-rs:

- **Method-call sugar** (rholang-rs PR #93, FIP 2025-08-20 §"Method-call sugar"):
  ```
  x!y(args).             ↔  x!?("y", args).
  x!y(args); P           ↔  x!?("y", args); P
  for (z <- x!y(args))   ↔  for (z <- x!?("y", args))
  ```

- **Agent block sugar** (rholang-rs PR #94, FIPs 2025-08-20 + 2026-01-28):
  ```
  agent fooCtor {
    constructor(<ctorPtrns>) { Pc } |
    method foo(<ptrns>) { P } | ... |
    default(...@args) { Q } |
    private method ... |          (optional)
    private default ...           (required if any private method)
  }
  ```
  desugars to a persistent for-comprehension that opens `new this, private`, runs the constructor body, a parallel public-dispatch for-comprehension matching on the method-name tuple, optionally a parallel private-dispatch, and replies `r!(bundle+{*this})`.

## What this PR is — and isn't

Both desugarings happen entirely at parse time in the upstream parser. The downstream normalizer needs **no new code** — the parser produces ordinary `SendSync` / `ForComprehension` / `New` / `Match` / `Par` / `Send` / `Bundle` constructs that the existing normalizer already handles.

So this PR is just:
- A pin bump (3 `Cargo.toml` files: `c163755` → `0202e79`)
- A new integration test file (`rholang/tests/sugar_regression.rs`) with 10 end-to-end equivalence tests

No code changes in `rholang/src/`.

## Pin choice note

`0202e79` is the head of [rholang-rs#95](https://github.com/F1R3FLY-io/rholang-rs/pull/95) (branch `pr/un-reserve-agent-keywords`), **not master**.

PR #94 as merged into upstream master (`e0d0c3e`) inadvertently reserved `agent`/`constructor`/`method`/`default` as global keywords, which broke parsing of existing Rholang code that uses these as identifiers — notably `casper/src/main/resources/SystemVault.rho`'s `contract _create(@vaultAddress, constructor, retCh) = { ... }`. Pre-push hook caught this; `compute_parents_post_state_regression_spec` failed in release mode.

[rholang-rs#95](https://github.com/F1R3FLY-io/rholang-rs/pull/95) un-reserves the four words and relies on tree-sitter's GLR context-disambiguation (same approach the original PR already took for `private`). Verified locally: with the new pin, the previously-failing 3 casper regression tests now pass.

**When rholang-rs#95 merges**, the pin in this PR should be re-bumped to the resulting master commit. The current SHA is a stopgap so this consumer-side PR can land in parallel.

## Test plan

- [x] `cargo build -p rholang` — clean
- [x] `cargo test -p rholang --lib` — 142 pass, no regressions
- [x] `cargo test -p rholang --test sugar_regression` — 10 pass (new file)
- [x] `cargo test --release -p casper --test compute_parents_post_state_regression_spec` — 3 pass (these were the regression-spec tests that flagged the upstream keyword-reservation bug)
- [x] Mutation-verified the regression tests have real discriminating power (changing a method name in a dispatch arm fails the affected tests)

Test coverage in the new file:

**Method-call sugar (6 cases)**:
- terminator: `x!y(1, 2).`
- no-args: `x!y().`
- sequential: `x!set(42); Nil`
- for-source: `for (@z <- x!get()) { Nil }`
- for-source with args: `for (@z <- x!compute(1, 2, 3)) { Nil }`
- nested: `x!y(a!z(1).).`

**Agent block sugar (4 cases)**:
- minimal: ctor + default only
- with methods: Cell shape
- with private: Counter shape (mixed pub + priv)
- private state: `@[*private, *stateToken]` compound key

Each case compiles both the sugared and hand-written-desugared forms through the full `Compiler::source_to_adt` pipeline and asserts the resulting `Par`s are identical.

## Stack

- Upstream rholang-rs PR #93 (merged) — method-call sugar
- Upstream rholang-rs PR #94 (merged) — agent block sugar
- Upstream rholang-rs PR #95 (open) — keyword-reservation regression fix; **this PR's pin currently points at #95's branch SHA, not master**
- **This PR** — downstream adoption

## Out of scope

- The follow-up FIPS doc PR (PR-A) clarifying the spec text for `default`-always-required and `private`-always-bound, which surfaced during implementation.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,9 +144,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.102"
+version = "1.0.103"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7f202df86484c868dbad7eaa557ef785d5c66295e41b460ef922eca0723b842c"
+checksum = "2a4385e2e34eb35d6b3efe798b9eb88096925d87726c0798709bf56d9ed84af3"
 
 [[package]]
 name = "approx"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5066,7 +5066,7 @@ dependencies = [
 [[package]]
 name = "rholang-parser"
 version = "0.1.0"
-source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=c163755#c1637550a844253f8fe781fc07e57fe5e5aabe7c"
+source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=0202e79#0202e79f24e060456f6f885dc1fd77a89f9ad1a1"
 dependencies = [
  "bitvec",
  "nonempty-collections",
@@ -5081,7 +5081,7 @@ dependencies = [
 [[package]]
 name = "rholang-tree-sitter"
 version = "0.1.2"
-source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=c163755#c1637550a844253f8fe781fc07e57fe5e5aabe7c"
+source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=0202e79#0202e79f24e060456f6f885dc1fd77a89f9ad1a1"
 dependencies = [
  "cc",
  "tree-sitter-language",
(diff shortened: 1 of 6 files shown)


metaweta:
Dylon approved privately:
https://discord.com/channels/@me/1362123632393978140/1521649159231242321

PR #98, opened 2026-07-07
feat(api): make http body limit configurable via config

semyon-dev:
Makes the HTTP API request body size limit configurable via a new api-server.http-max-body-bytes setting, bringing parity with the existing grpc-max-recv-message-size knob.


semyon-dev:
@spreston8 could we merge it?

PR #120, opened 2026-07-14
feat(fileio): native syscall layer for File I/O FIP (Phase 1)

metaweta:
## Summary

Phase 1 of the [File I/O FIP](https://github.com/F1R3FLY-io/FIPS/blob/main/fileio/under-review/2026-02-06-File-IO/2026-02-06-File-IO.md) rollout: the **native syscall layer** described in the FIP. A set of thin system processes that bridge `tokio::fs` calls to Rholang messages, plus supporting helpers (fd table, mode-string parser, response encoder, stat record shape, path quarantine). No agent layer, no powerbox, no static config wiring — those are later phases and require the FS agent classes on top of these primitives.

The URNs registered here are **internal** (`rho:io:fs:native:1.0.0/*`); user-facing Rholang code will go through the `Fs` agent under `rho:io:fs:1.*` in a follow-up PR.

## Layout

New `rholang/src/rust/interpreter/io/` module:

- `handle_table.rs` — `FileHandleTable` = `Arc<RwLock<HashMap<i64, Arc<FileHandle>>>>` with monotonically-increasing fd counter (never reused, for replay determinism). Wraps `tokio::fs::File` in a `tokio::sync::Mutex` because the read/write methods need `&mut self`.
- `mode.rs` — fopen-style mode-string parser (`open_options_for`). Handles all nine strings the FIP §"With the config file" table enumerates, plus `"wbx"` synonym. Unknown modes return `None`; the caller translates to `FSERR_BAD_ARG`.
- `response.rs` — `[true, ...values]` / `[false, code, msg]` helpers plus twelve `FSERR_*` string constants matching the FIP §"Standard error codes" subsection. Plus a `from_io_error` mapping `NotFound` / `AlreadyExists` / `PermissionDenied` to the FIP-named codes.
- `stat.rs` — encodes the FIP TODO 5 record shape. `mode_string(u32) -> "rwxr-xr-x"`, `stat_record(name, meta) -> Map` — one source of truth for both `nativeStat` and `nativeEntries`. `owner`/`group` deferred to the `nativeChown` slice.
- `path.rs` — `canonicalize_and_quarantine(root, rel)` for the FIP's quarantine promise. Handles the "creating a fresh file inside root" case by walking up to the deepest existing ancestor and resolving the non-existent tail lexically, rejecting any `..` that would climb above the ancestor. Not a substitute for openat/renameat (TOCTOU races against a concurrent writer are out of scope per the FIP).

## Primitives registered

16 primitives across five commits, all under `rho:io:fs:native:1.0.0/*`, `FixedChannels` bytes 38-55, `BodyRefs` 37-54.

| URN | Signature |
|---|---|
| `.../open` | `(path, mode) -> [true, fd] \| [false, code, msg]` |
| `.../close` | `(fd) -> [true] \| ...` |
| `.../read` | `(fd, n) -> [true, bytes] \| ...` (loops until buffer full or EOF) |
| `.../write` | `(fd, bytes) -> [true, nWritten] \| ...` (uses `write_all`) |
| `.../seek` | `(fd, offset, whence)` where `whence ∈ {"set","cur","end"}` |
| `.../tell` | `(fd) -> [true, pos]` (via `seek(Current(0))`) |
| `.../size` | `(fd) -> [true, size]` (via `metadata().len()`) |
| `.../truncate` | `(fd, n) -> [true]` (via `set_len`) |
| `.../flush` | `(fd) -> [true]` (via `sync_all` — data + metadata; the FIP promises "durable write") |
| `.../stat` | `(path) -> [true, record]` (via `symlink_metadata`, doesn't follow symlinks) |
| `.../entries` | `(dirPath) -> [true, [record, ...]]` (sorted lexicographically) |
| `.../exists` | `(path) -> [true, bool]` (via `tokio::fs::try_exists`, follows symlinks) |
| `.../rename` | `(from, to) -> [true] \| ...` (`FSERR_CROSS_DEVICE` on EXDEV/18) |
| `.../copyFile` | `(from, to) -> [true, nBytes]` |
| `.../removeFile` | `(path) -> [true]` |
| `.../removeDir` | `(path, recursive: Bool) -> [true]` |
| `.../chmod` | `(path, modeBits: Int) -> [true]` (Int 0..=0o777; symbolic-string parsing lives in the Rholang agent layer) |
| `.../quarantine` | `(rootPath, relPath) -> [true, canonPath] \| [false, FSERR_QUARANTINE, ...]` |

## Not in this PR (later phases)

- **`nativeChown`** — needs an NSS-lookup design decision: `libc::getpwnam_r` via `unsafe`, or add `nix` / `users` crate as a direct dep. Once landed, also extends `io::stat::stat_record` with `owner`/`group` fields.
- **Non-deterministic-replay wrapping** — for the state-observing primitives (`nativeRead`, `nativeStat`, `nativeEntries`, `nativeExists`, `nativeChown`, the pickers), wrap in the `FailedNonDeterministicCall` path (`dispatch.rs:92-95`) so the lead node's results replay deterministically to followers. The bookkeeping primitives (`open`, `close`, `seek`, `tell`, writes, truncate, flush, chmod, rename, copy, remove, quarantine) are deterministic and don't need the wrap.
- **Picker primitives** (`rho:sys:ui:1.*`) — `nativePickFile` / `nativePickDir` shelling to `osascript` / `zenity` / `kdialog`. Behind a feature flag. Available only under Use Cases 2 & 3 (interactive `.rho` / REPL).
- **Cost wiring** — per-op functions in `accounting/costs.rs` per the FIP §"Cost accounting" table. Each native handler calls `CostManager::charge()` at entry.
- **Rholang agent layer** — `Fs` / `File` / `Dir` / `ReadOnlyStream` / `AppendOnlyStream` / `LockToken` agent classes. Wraps the native primitives, calls `nativeQuarantine` before every path-taking native, shapes results into the `try/catch` tuple form.
- **Powerbox + static provisioning** — CLI flags / config-file entries materialize into native fd allocations at boot; powerbox contract seeds the FS agents into the versioned registry. Will copy the GAP-2 `rho:deploy:data` authentication pattern for any deployer-scoped capabilities.

## Verified

- `cargo build -p rholang` clean.
- `cargo test -p rholang --lib io::` — 23 pass across `handle_table`, `mode`, `response`, `stat`, `path`.
- `cargo test --release -p casper --test mod util::rholang::runtime_spec::empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes — the new URNs are system-process registrations (Rust handlers), not tuplespace-installed contracts, so the empty state hash is unchanged.
- Full per-crate pre-push hook (release-mode tests across `shared`, `models`, `graphz`, `block-storage`, `rspace_plus_plus`, `comm`, `rholang`, `casper`, `node`, plus `fmt` / `clippy` / `deny`) — all 11 checks pass.

## Test plan

- [ ] CI runs the workspace release-mode tests; verify all crates pass.
- [ ] `cargo test -p rholang --lib io::` locally verifies the 23 io-submodule tests.
- [ ] The FS agent PR that follows will exercise the primitives end-to-end from Rholang.

## Stack

Depends on [PR #112](https://github.com/F1R3FLY-io/f1r3node-rust/pull/112) — merged into `dev`. This branch is based on the resulting `dev` tip, so the versioned registry substrate and the GAP-2 authentication pattern are already available; the follow-up FS agent layer will copy them.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/120"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786642671&installation_id=145946300&pr_number=120&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F120&signature=cc4d331c07b0a451a4a9a8763e71baddb6f53f753814f26b8ef717f6570916da"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


metaweta:
## Update — security-review hardening

Followed up on this PR with commit [`cddec6ea`](https://github.com/F1R3FLY-io/f1r3node-rust/pull/120/commits/cddec6ea) addressing the five items from the security review of the initial commits.

### 1. Hide `rho:io:fs:native:*` URNs from `urn_map` — CRITICAL

The native primitives are unauthenticated arbitrary-host-FS operations. Publishing their URNs in `urn_map` would let any deploy on the network do:

```rholang
new open(`rho:io:fs:native:1.0.0/open`),
    read(`rho:io:fs:native:1.0.0/read`), ret in {
  open!("/etc/passwd", "r", *ret) |
  for (@[true, fd] <- ret) {
    new rd in { read!(fd, 100000, *rd) | ... }
  }
}
```

and exfiltrate host files (or overwrite them, chmod them, etc.). New `is_internal_urn` predicate in `rho_runtime.rs` filters them out of `setup_maps_and_refs` before insertion. Handlers stay bound on their fixed channels via `introduce_system_process`; the FS-agent Rholang (future PR) will reach them via per-deploy `NormalizerEnv` injections populated at genesis, not `urn_map` — see the implementation plan §"URN visibility" for the full design.

The predicate is prefix-based (`urn.starts_with("rho:io:fs:native:")`) rather than a per-`Definition` flag so introducing a new internal primitive is a URN-naming decision, not a boolean someone might forget to set.

### 2. Extend `non_deterministic_ops()` for all 18 fileio body refs — CRITICAL for consensus

Every primitive that observes or mutates host state is now routed through the `FailedNonDeterministicCall` path in `dispatch.rs`. Without this, a consensus deploy that touched the filesystem would diverge between nodes (two nodes read `/etc/passwd`, get different bytes, disagree on the block outcome). `NATIVE_CLOSE` is included for defense-in-depth even though closing a known fd is deterministic given deterministic prior state — the fd itself came from a non-det `NATIVE_OPEN`, so routing the close through the same replay log keeps the fd-table timeline coherent across nodes.

### 3. Cap `nativeRead` at 64 MiB per call

`MAX_READ_BYTES = 64 * 1024 * 1024`. `native_read(fd, i64::MAX)` would otherwise `vec![0u8; usize::MAX]` and OOM the node. Rejected with `FSERR_BAD_ARG` before allocation. Callers wanting larger reads issue multiple calls or go through the future agent-layer `lines`/`text` methods (which iterate internally).

### 4. Cap `nativeTruncate` at 16 GiB per call

`MAX_TRUNCATE_BYTES = 16 * 1024 * 1024 * 1024`. On sparse-file filesystems (ext4, apfs) `set_len(n)` is O(1) at any n, so `nativeTruncate(fd, i64::MAX)` would report success; on non-sparse filesystems the same call fills the disk. 16 GiB is a defensive per-call ceiling; per-deploy quotas belong in the powerbox layer once that exists.

### 5. Unify symlink semantics between `nativeStat` and `nativeExists`

Previously `nativeStat` used `symlink_metadata` (does NOT follow symlinks) but `nativeExists` used `tokio::fs::try_exists` (DOES follow). Adjacent primitives disagreeing on the same path is a footgun: `exists(path) && stat(path)` could give contradictory answers on a dangling symlink or a symlink to an in-quarantine target. `nativeExists` now uses `symlink_metadata` too, translating `Err(NotFound)` → `Ok(false)` and other errors → `FSERR_IO`. A symlink present at `path` reports `true` (consistent with `nativeStat` reporting `kind: "symlink"`). Callers wanting follow-symlink semantics open the file via `nativeOpen` (which uses `OpenOptions::open` and does follow) and observe via the fd.

## Tests added

Four new unit tests across `rho_runtime` and `system_processes` modules:

- `rho_runtime::tests::internal_urns_are_the_fileio_native_prefix_and_nothing_else` — asserts the prefix predicate matches every native URN we care about and rejects every public/internal-non-fileio URN.
- `rho_runtime::tests::setup_maps_and_refs_hides_native_fileio_urns_from_urn_map` — drives `setup_maps_and_refs` and asserts no `rho:io:fs:native:*` key leaked into the resulting map. Positive-sanity check that `rho:io:stdout` is still present.
- `system_processes::tests::non_deterministic_ops_covers_every_fileio_body_ref` — asserts the full 18-entry fileio set is in `non_deterministic_ops()`. Adding a new NATIVE_* BodyRef in the future will break this test intentionally, forcing a determinism decision.
- `system_processes::tests::non_deterministic_ops_still_covers_prior_ai_and_grpc_refs` — regression guard that the fileio addition didn't drop the pre-existing entries (GPT4, DALLE3, TEXT_TO_AUDIO, OLLAMA_*, GRPC_TELL, CHROMA_QUERY).

The handler-level cap/symlink changes (items 3, 4, 5) aren't unit-tested here — they'd require standing up a full `SystemProcesses` and are more naturally exercised by end-to-end tests once the FS-agent Rholang layer lands.

## Verified

- `cargo build -p rholang` clean.
- `cargo test -p rholang --lib` — 219 pass (215 prior + 4 new).
- `cargo test -p rholang --lib io::` — 23 io-submodule tests still pass (unchanged).
- `empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes — filtering URNs out of `urn_map` doesn't affect tuplespace state (urn_map isn't tuplespace-persisted; it's runtime-scoped metadata).
- Full pre-push hook (fmt + clippy + deny + release-mode tests across all 11 crates) passed on this exact commit set. The one clippy nit from the first commit attempt (an unnecessary `&mut` on a test call) is fixed in the committed version.

## Cross-references

- Security review that prompted these changes: this thread's earlier code-review analysis.
- Implementation plan documenting the URN-visibility model + non-det list + caps: `fileio/under-review/2026-02-06-File-IO/implementation-plan.md` in the FIPs repo (§"Native syscall layer" subsections "URN visibility", "Non-determinism wrapping", "Input caps", "Symlink semantics").
- rholang-rs [PR #96](https://github.com/F1R3FLY-io/rholang-rs/pull/96) is the sibling PR landing the `try/catch` sugar the future FS-agent layer will use.

dylon (approved):
There are no blockers so I've approved this PR, but claude had the following non-blocking recommendations:

  ## Review — PR #120 (native primitives)

  **Verdict: approve.** Merging is safe: `is_internal_urn` keeps every `rho:io:fs:native:*`
  URN out of `urn_map`, no `Fs` agent exists, so no deploy can reach these handlers today.
  The findings below are gating on the `Fs`-exposure milestone, not on this merge.

  Verified correct (so nobody re-checks): `FixedChannels` bytes 38–55 and `BodyRefs` 37–54
  are collision-free against live `dev` (max prior: byte 37, ref 36; test channels at
  101–108). All eight FIP mode strings translate correctly. `RhoMap::create_par` routes
  through `SortedParMap`, so stat records are canonically ordered — no Par-encoding
  nondeterminism. `check_prefix` uses `Path::starts_with`, which is component-wise, so
  there is no `/srv/data` vs `/srv/data2` prefix confusion. `flush` correctly uses
  `sync_all` rather than `AsyncWriteExt::flush`. Every handler's `arity` matches its
  destructuring width.

  ### Must fix before the `Fs` agent is user-reachable

  1. **`native_open` has no file-type guard, and no `O_NONBLOCK`.** Opening a **FIFO** for
     reading blocks until a writer appears; `tokio::fs` dispatches to the blocking pool, so
     one FIFO inside a provisioned tree pins a blocking-pool thread for the life of the
     runtime, and repeated calls exhaust the pool. Opening a **directory** `O_RDONLY`
     succeeds on Linux, so `openFile(<a dir>)` hands back a usable-looking fd whose reads
     all fail `EISDIR`→`FSERR_IO` while occupying a table slot. `nativeEntries` reports both
     as `kind: "other"`, so a `Dir` holder can find them. Please `symlink_metadata` first and
     reject non-regular files with `FSERR_UNSUPPORTED`.
  2. **Type-mismatched arguments abort the deploy instead of returning an error tuple.**
     Every handler does `RhoNumber::unapply(..) else { return Err(illegal_argument_error(..)) }`.
     That becomes `IllegalArgumentError` → `NonDeterministicProcessFailure` (`reduce.rs:353-365`),
     killing the deploy. Since the agents forward caller values unchecked, `file!read("abc")`
     or `dir!removeDir("x", "yes")` is user-reachable. The FIP's whole try/catch contract says
     this should be `[false, FSERR_BAD_ARG, msg]` — §Standard error codes names `FSERR_BAD_ARG`
     for exactly "argument was malformed".
  3. **`native_quarantine` returns `canonical.to_string_lossy().into_owned()`.** A non-UTF-8
     path is silently mangled to U+FFFD, so the path that was *checked* is not the path
     later *used*. Please `to_str().ok_or(BadArg)`.
  4. **`native_entries` follows symlinks and aborts on one bad entry.** It uses
     `entry.metadata()` (follows) while `native_stat`/`native_exists` use `symlink_metadata`
     (doesn't) — and `native_exists`'s own doc argues at length for uniform non-follow
     semantics "to avoid that footgun". Worse, `Err(e) => { err = Some(e); break; }` means a
     single dangling symlink or `EACCES` entry fails the entire listing instead of the one row.
  5. **No cost accounting on any native.** `dispatch.rs` has zero `charge` calls. The FIP defers
     cost accounting, but the consequence is unmetered I/O once `Fs` ships — `nativeRead(fd, 64MiB)`
     in a loop, `nativeEntries` on a huge directory, and especially **`nativeFlush` = `fsync`** in
     a tight loop, which is a synchronous disk barrier. Please treat metering as a hard gate.

  ### Should fix

  6. **`ctime` is birth time, not POSIX `ctime`.** `stat.rs` uses `meta.created()`, which is btime
     and returns `Err(Unsupported)` on filesystems without statx btime — so `epoch_secs` yields
     `None` and the key is **omitted**. The field is mislabelled (FIP `ctime` is `st_ctime`, what
     `ls -lc` shows) and the record *shape* becomes host-dependent. Benign under the oracle model,
     a divergence source the moment `consensus-static-files` is exercised. Use `MetadataExt::ctime()`.
  7. **`kind: "symlink"` contradicts FIP line 456** ("`kind` therefore never reports `"symlink"`").
     Defensible at the native layer, but nothing downstream maps it yet.
  8. **`resolve_lexical` bounds `..` against the deepest *existing* ancestor, not `root`.** With
     root `/r` and `/r/sub` existing, `"newdir/../../x.txt"` resolves to `/r/x.txt` — inside the
     root — but is rejected. Fails closed, so usability not security.
  9. **`native_read` preallocates `vec![0u8; n]`** up to 64 MiB regardless of file size; `read(fd, 64MiB)`
     on a 10-byte file allocates 64 MiB, and concurrent deploys multiply it. Clamp to `min(n, size - pos)`.
  10. **`native_entries` is unbounded** — a directory with a million entries builds a million ~9-key
      `Par` maps in one `spawn_blocking`.
  11. **`native_size` holds the per-file mutex** while its doc says it is "Independent of the file
      position, so a caller can query size without disturbing an in-progress positional read/write."
      It does block one.
  12. **`mode.rs` accepts `"wbx"`**, documented as "per §'With the config file' table". That table has
      eight entries and no `"wbx"`. Drop it or amend the FIP.
  13. **`native_open`'s canonicalize fallback** stores the raw caller string in `canonical_path`, a field
      documented as existing "so a future path-quarantine layer has something to check against".
  14. **PR body drift** (these become the merge commit): it says `exists` uses `try_exists` and follows
      symlinks (the code uses `symlink_metadata` and the code doc is right), and that the bookkeeping
      primitives "don't need the [replay] wrap" (they all got one in #137).
  15. `native_remove_file`'s doc claims `FSERR_PERM` on a directory; Linux returns `EISDIR` → `FSERR_IO`.
  16. The lychee `--timeout 60` CI change is unrelated scope creep.

metaweta:
Closing as obsolete.

The File I/O FIP has been rewritten from method-centric to stream-centric — every method whose result size isn't bounded by a caller-supplied argument now returns a `Stream` (or a specialization: `CharStream`, `ByteStream`, `EntryStream`, `LineStream`), and every unbounded-input write method takes a `Stream`. The redesigned spec is in `F1R3FLY-io/FIPS#21` (merged 2026-07-24) and lives at `under-review/2026-02-06-File-IO/2026-02-06-File-IO.md`.

That's a structural change to the surface this PR implements, not a tweak. The native-primitive shape, the agent-block shape, the caching story, the concurrency model, the error-code set, and the stdio wrappers all need to be reworked to match the new spec. Continuing this PR would layer additional review debt onto a foundation the FIP no longer endorses.

The whole stack that built on this PR (#130, #131, #132, #137, #139, #140, #141, #144, #145) is being closed for the same reason. Implementation work will restart against the streams-first spec on new branches.

Branch (`feat/fileio-native-primitives`) is retained for reference; feel free to delete once you've mined anything useful from it.

PR #122, opened 2026-07-14
Fix/dem merge recovery; merge prior to #118

jeffrey-l-turner:
---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/122"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786663854&installation_id=145946300&pr_number=122&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F122&signature=e4615b11fe891c0074736ef84b90d888d1a28c51585347c73f2618d56491b769"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**
- anthropic (claude-opus-4-5-20251101): Needs Review (confidence: 0.75)
- bedrock (us.amazon.nova-pro-v1:0): Needs Review (confidence: 0.85)
- xai (grok-4.5): Needs Review (confidence: 0.78)

---

### Summary

This is a large and complex PR focused on fixing merge recovery and deploy handling in the Casper consensus implementation. The changes add significant new functionality including deploy admission policies, finality lag backpressure, floor/frontier caching for DAG blocks, and improved deploy recovery mechanisms. While the code appears well-structured with good logging, there are concerns about the slash deploy construction potentially using the wrong public key, the BFS ancestry check's height pruning logic, and the change from assertion to warning for DAG integrity checks. The PR would benefit from human review of the slashing logic correctness and the DAG ancestry algorithm's edge cases. The large number of new constants and complex state machines for deploy admission should be reviewed for configurability and testability.

This is a large, consensus-critical port of merge recovery and finalized-floor machinery (floor/frontier memoization, multi-parent canonical_won_sigs, deploy-inclusion leadership, overflow-checked IntegerAdd, LMDB registration of new indexes). The direction is sound and several production crash/starvation fixes are clearly motivated, but the change set is too broad and still carries acknowledged expected-red coverage for under-load convergence and inclusion leadership. The highest-priority concrete defect is ParentsPostStateCacheKey omitting the justification snapshot used to derive the floor, which can return a wrong cached merge base. Secondary concerns are a hot-path performance regression in invalid_latest_messages_from_hashes, fragile string-matched retry classification, and heavy per-propose DAG walks. Recommend human review of cache keying and admission-policy edge cases before merge; do not treat remaining red suite items as optional if they affect multi-writer recovery.

The code change introduces several new functions and structs, primarily focused on improving deploy management and recovery mechanisms. While the changes seem logically sound, there are potential race conditions in caching mechanisms and inconsistent use of `await` that need to be addressed. Additionally, adding documentation to new code would enhance maintainability.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential race condition in caching</b> (Reported by: bedrock)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 110)

The `get_cached_floor` and `put_cached_floor` functions use a plain memoization store without any synchronization. This could lead to race conditions if multiple threads attempt to access or modify the cache simultaneously.

**Suggestion:**
```
Consider using a thread-safe data structure or adding synchronization mechanisms to ensure safe concurrent access.
```

</details>

<details>
<summary><b>:orange_circle: Major: invalid_latest_messages_from_hashes regresses to per-validator lookups</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 394)

The previous implementation filtered latest_message_hashes against an in-memory invalid_block_hashes set (work proportional to |invalid| ∩ |latest|). The new loop calls lookup() for every latest message. The nearby comment still claims steady-state work is proportional to |invalid_latest_messages|, which is no longer true. On large validator sets this is an avoidable hot-path cost (fork-choice / slashing prep).

**Suggestion:**
```
Restore set-based filtering using invalid_blocks()/invalid_blocks_set (or a HashSet of invalid hashes), and only fall back to metadata.lookup when the set is missing or needs revalidation.
```

</details>

<details>
<summary><b>:orange_circle: Major: BFS ancestry check may have correctness issues with height pruning</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 510)

The `is_dag_ancestor` function uses height-based pruning (`if self.block_number_unsafe(&current)? <= stop_height`) but this could incorrectly prune valid paths in DAGs where a block at height N could have a parent at height N-2 or lower that eventually leads to the ancestor. The comment claims this is safe but the logic may miss cases where intermediate blocks are at or below the ancestor height.

**Suggestion:**
```
Verify that the DAG structure guarantees strictly increasing heights along all parent paths, or adjust the pruning condition to use `<` instead of `<=` to avoid pruning the ancestor itself.
```

</details>

<details>
<summary><b>:orange_circle: Major: Retry classification depends on fragile error-string matching</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 192)

is_retryable_single_value_batch_error / next_checkpoint_retry_limit gate batch-size reduction on substring matches against HistoryError/RuntimeError messages ("single-value cell", "IntegerAdd single-value invariant", etc.). Any wording change, localization, or alternate error path will silently disable retry shrinking and can leave the proposer stuck or over-batching.

**Suggestion:**
```
Introduce typed error variants (or structured error codes) for single-value overfill / IntegerAdd invariant failures and match on those instead of free-form strings.
```

</details>

<details>
<summary><b>:orange_circle: Major: Build slash deploy uses proposer's public key instead of offender's</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 1370)

The `build_slash_deploy` function creates a `SlashDeploy` with `pk: proposer_public_key.clone()` but the slashing deploy should reference the offender/equivocator's public key, not the proposer's. This appears to be setting the wrong validator identity in the slash deploy.

**Suggestion:**
```
Review whether `pk` in `SlashDeploy` should be the offender's public key (from `invalid_block_hash` lookup) rather than the proposer's public key.
```

</details>

<details>
<summary><b>:orange_circle: Major: ParentsPostStateCacheKey omits justification snapshot</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 88)

Floor derivation in compute_parents_post_state is a pure function of the frozen latest_messages/justifications, but ParentsPostStateCacheKey only keys on sorted_parent_hashes, snapshot_lfb_hash, and disable_late_block_filtering. Two merges with the same parents and LFB but different justification snapshots can share a cache entry and return the wrong floor base / rejected set. That can cause incorrect pre-state packaging on propose (later rejected on replay) or cross-validation pollution on a shared RuntimeManager.

**Suggestion:**
```
Include a deterministic fingerprint of latest_messages (e.g. sorted (validator, latest_block_hash) pairs or a hash of that map) in ParentsPostStateCacheKey, or drop/bypass the cache when justifications are not identical to the key context.
```

</details>

<details>
<summary><b>:orange_circle: Major: Large number of magic constants without configuration</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 57)

Many hardcoded constants like `DEPLOY_INCLUSION_LEASE_BLOCKS`, `FINALITY_LAG_SOFT_BACKPRESSURE_BLOCKS`, etc. are defined at module level. While documented, these should ideally be configurable or at least grouped in a configuration struct.

---

block_creator.rs gained a dense cluster of caps, leader election, staleness, fallback, and backpressure helpers. Correctness of non-leader starvation fixes depends on interactions among many constants (lease blocks/ms, lag soft/hard, fallback caps). PR notes still list expected-red cases (under-load convergence, inclusion-leadership, recovery-cycle retries).

**Suggestion:**
```
Consider grouping related constants into a configuration struct that could be made runtime-configurable in the future.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: invalid_blocks_map re-lookups and re-filters already-invalid set</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 411)

invalid_blocks_map iterates invalid_blocks_set then re-looks up each hash and drops entries unless metadata.invalid is true. If the set and metadata ever diverge, the map silently shrinks; if they are always in sync, the extra lookups are pure overhead. Behavior change vs the previous direct map from the set is subtle and under-documented.

**Suggestion:**
```
Either trust invalid_blocks_set as authoritative (map directly) or document/assert the revalidation invariant and log when a set member is not invalid in metadata.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Height-map contiguity assert downgraded to warn</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_metadata_store.rs` (line 388)

A non-contiguous height_map previously panics as a hard invariant; it now only warns and continues. That improves crash resilience but can hide storage corruption or insert bugs until later, harder-to-diagnose failures (wrong topo, floor walks, etc.).

**Suggestion:**
```
Keep the warn for production, but add a metric/counter and consider a debug-assert or test-only hard fail so CI still catches contiguity bugs.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Non-contiguous height map changed from assertion to warning</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_metadata_store.rs` (line 391)

The change from `assert!` to `tracing::warn!` means the node will continue operating with a potentially corrupt DAG state. While this prevents crashes, it could mask serious data integrity issues that should be investigated.

**Suggestion:**
```
Consider adding a metric to track this condition and potentially triggering an alert or recovery procedure rather than silently continuing.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent use of `await`</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1674)

The `compute_parents_post_state` function is called with `await`, but the result is not awaited. This could lead to unexpected behavior if the function is asynchronous.

**Suggestion:**
```
Ensure that all asynchronous functions are properly awaited to avoid potential issues.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Complex struct definitions lack documentation</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 61)

New structs like `DeployAdmissionPolicy`, `BranchDeployInfo`, `DeployInclusionProgress`, etc. lack doc comments explaining their purpose and field meanings.

**Suggestion:**
```
Add doc comments to these structs explaining their role in the deploy admission flow.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Potential integer underflow in retry limit calculation</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 230)

The function `next_single_value_retry_limit` uses `current / 2` which is safe, but the calling code should handle the case where this returns `Some(0)` vs `None` consistently.

**Suggestion:**
```
Verify that `Some(1)` (the minimum) is the correct floor value and document this behavior.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Multiple full DAG walks on every block create</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 861)

create() now runs several ancestor walks per proposal (scope_has_unfinalized_user_deploys, self_chain_has_unfinalized_user_deploys, deploy_inclusion_progress / branch_unfinalized_user_deploy_info, finalized_ancestor_deploy_sigs, canonical_won_sigs, parent_frontier_extends_lfb). Under load with deep unfinalized branches this multiplies store I/O on the propose path.

**Suggestion:**
```
Share a single bounded ancestor scan result (or cache by dag_generation + parent frontier) across these helpers for one create() call.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Multiple redundant deploy storage scans</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 1085)

Functions like `fresh_local_deploy_stats` and `in_scope_local_deploy_stats` each call `deploy_storage.lock().read_all()` and scan through deploys separately. When called together in `create()`, this results in multiple full scans of deploy storage.

**Suggestion:**
```
Consider combining these into a single scan that computes all needed statistics in one pass.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Newly-bonded LM placeholder semantics need a short invariant note</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1025)

Newly bonded validators without a latest-message slot are registered to height-0 min hash (else the inserting block). This is intentional for reverse-height insert and justification_follows, but the fallback to block_hash when height 0 is missing reintroduces the old behavior silently.

**Suggestion:**
```
Document the invariant (genesis always at height 0 after init) and prefer hard error if height 0 is empty post-genesis rather than falling back to the inserting block.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Missing documentation for new functions</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 167)

Several new functions and structs have been added without documentation. Adding comments or docstrings would improve code readability and maintainability.

**Suggestion:**
```
Add documentation to all new functions and structs to explain their purpose and usage.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: IntegerAdd overflow-checked combine is a good fail-closed fix</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/merger/merging_logic.rs` (line 48)

combine_mergeable_value now uses checked_add for IntegerAdd and rejects overflow instead of wrapping. That closes a silent launder path past non-negative apply gates. No issue; call out as positive.

**Suggestion:**
```
Ensure all call sites treat None as hard reject (not skip/zero) and that conflict_set_merger / dag_merger paths are covered by the new unit/proptest tests.
```

</details>
---

<details>

#### anthropic (claude-opus-4-5-20251101)

This is a large and complex PR focused on fixing merge recovery and deploy handling in the Casper consensus implementation. The changes add significant new functionality including deploy admission policies, finality lag backpressure, floor/frontier caching for DAG blocks, and improved deploy recovery mechanisms. While the code appears well-structured with good logging, there are concerns about the slash deploy construction potentially using the wrong public key, the BFS ancestry check's height pruning logic, and the change from assertion to warning for DAG integrity checks. The PR would benefit from human review of the slashing logic correctness and the DAG ancestry algorithm's edge cases. The large number of new constants and complex state machines for deploy admission should be reviewed for configurability and testability.

The code change introduces several new functions and structs, primarily focused on improving deploy management and recovery mechanisms. While the changes seem logically sound, there are potential race conditions in caching mechanisms and inconsistent use of `await` that need to be addressed. Additionally, adding documentation to new code would enhance maintainability.

#### xai (grok-4.5)

This is a large, consensus-critical port of merge recovery and finalized-floor machinery (floor/frontier memoization, multi-parent canonical_won_sigs, deploy-inclusion leadership, overflow-checked IntegerAdd, LMDB registration of new indexes). The direction is sound and several production crash/starvation fixes are clearly motivated, but the change set is too broad and still carries acknowledged expected-red coverage for under-load convergence and inclusion leadership. The highest-priority concrete defect is ParentsPostStateCacheKey omitting the justification snapshot used to derive the floor, which can return a wrong cached merge base. Secondary concerns are a hot-path performance regression in invalid_latest_messages_from_hashes, fragile string-matched retry classification, and heavy per-propose DAG walks. Recommend human review of cache keying and admission-policy edge cases before merge; do not treat remaining red suite items as optional if they affect multi-writer recovery.
</details>

---

casper/src/rust/rholang/runtime.rs
```diff
@@ -1178,6 +1178,65 @@ impl RuntimeOps {
...
+    /// the caller falls back to its local configuration in that case.
+    pub async fn get_fault_tolerance_threshold_ppm(
+        &mut self,
+        start_hash: &StateHash,
+    ) -> Result<Option<i64>, CasperError> {
+        let ppm_pars = self
+            .play_exploratory_par(Self::fault_tolerance_ppm_query_par().clone(), start_hash)
+            .await?;
+
+        if ppm_pars.is_empty() {
```
machieke, line 1235: treats an empty getFaultToleranceThresholdPpm query as “old chain, fall back to local config”, but the helper underneath also converts runtime.inj failures into Ok(Vec::new()) at runtime.rs:891. That means a real query execution failure on a chain that does have the getter will silently flow into the local-threshold fallback at casper.rs:233, reintroducing node-local finalized-floor divergence. For this consensus parameter, fallback should only happen after a successful query with no result; injection/runtime errors should fail startup.

machieke (changes requested):

casper/src/rust/rholang/runtime.rs
jeffrey-l-turner: agreed -- adding unit test for this also

jeffrey-l-turner: Confirmed and fixed in c77972ff — you were right that the lenient helper's `Ok(Vec::new())`-on-error conflated "query execution failed" with "getter absent", which would have routed a transient failure into the local-threshold fallback and re-opened node-local floor divergence.

**Fix:** the helper is split into `play_exploratory_par_with_mode(strict)`. The existing lenient wrapper keeps its behavior for display/API callers (bonds, active validators) and is now documented as never-for-consensus. `get_fault_tolerance_threshold_ppm` uses the new `play_exploratory_par_strict`, which propagates injection failures as errors — since `hash_set_casper` `?`-propagates, a failed read now fails node startup. `None`/local-fallback happens only after a **successful** query with no result, exactly as you prescribed.

**Regression test:** `strict_exploratory_query_propagates_execution_failure` (casper/tests/util/rholang/runtime_manager_test.rs) pins the contrast on the same deterministically-failing term (`1 / 0` at reduce time): the lenient path degrades it to an empty result, the strict path returns `Err`. Companion `fault_tolerance_threshold_ppm_round_trips_through_genesis` still passes through the strict path, covering the happy case.

jltatbeach (approved):

PR #129, opened 2026-07-16
Sealed floor merge squashed

dylon:
Merges the latest chages from `dev` into the feature branch for PR #90 and cleans up all discovered issues. Consecutive commits were squashed by author.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/129"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786833217&installation_id=145946300&pr_number=129&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F129&signature=bbdb7637b4a1c336086f33e39629f5b9558fc2583139a3c9e3bda9e53f41e52b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

| xai | `grok-4.5` | Timeout Error | 0 |

---

### Summary

The code change appears to be well-structured with several improvements and optimizations. However, there are some performance considerations and style inconsistencies that should be addressed. Additionally, adding more detailed documentation would enhance maintainability.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential performance bottleneck</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/lfs_tuple_space_requester.rs` (line 566)

The use of `async_stream::stream!` and the nested loops may lead to performance issues under high load. Consider optimizing the stream processing or using a more efficient data structure.

**Suggestion:**
```
Review the stream processing logic and consider using a more efficient data structure or parallel processing where applicable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent comment style</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 1140)

The comments use a mix of markdown and plain text styles. Consistency in comment style improves readability.

**Suggestion:**
```
Standardize the comment style to either markdown or plain text throughout the file.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Lack of detailed documentation</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/finality/floor.rs` (line 1057)

The function `derive_floor_selects_highest_sound_candidate_over_chain` lacks detailed documentation explaining its purpose and usage.

**Suggestion:**
```
Add detailed documentation to explain the function's purpose, parameters, and return values.
```

</details>
---

<details>

The code change appears to be well-structured with several improvements and optimizations. However, there are some performance considerations and style inconsistencies that should be addressed. Additionally, adding more detailed documentation would enhance maintainability.

#### xai (grok-4.5)

> **ERROR TIMEOUT**: Empty response from provider
</details>

---

jeffrey-l-turner:
## Review triage: performance finding assessed as FALSE POSITIVE — resolved

> **Major — Potential performance bottleneck** (`casper/src/rust/engine/lfs_tuple_space_requester.rs:566`, reported by bedrock/Nova):
> "The use of `async_stream::stream!` and the nested loops may lead to performance issues under high load. Consider optimizing the stream processing or using a more efficient data structure or parallel processing where applicable."

**Determination: false positive. No action required. This finding is closed.**

Rationale, from reading the code at the flagged site:

1. **No nested loops exist.** The code is a single `loop` around a `tokio::select!` with three arms (state-chunk responses, request triggers, idle-timeout backoff). All arms await; nothing busy-waits. The `biased` ordering is a deliberate, commented anti-starvation choice.
2. **"High load" does not apply to this path.** This is the last-finalized-state (LFS) tuple-space downloader — it runs once at node bootstrap to fetch the approved block's state from peers. Throughput is gated by network round-trips and LMDB trie import I/O, not loop control flow; the `async_stream` generator poll is noise by comparison.
3. **Parallel processing would be incorrect here.** State-part import carries parents-first trie ordering invariants. The Scala original this ports (fs2 request/response streams run `concurrently`, with `evalMap(_ => st.get)`) has the same sequential-import semantics, which the `select!` arms reproduce faithfully — the request/response interleaving already provides the concurrency the design permits.
4. The only per-event cost of note is the `st.lock()` + `state.clone()` per yield, inherited from the Scala `st.get`-per-event semantics and bounded by the outstanding-paths set — immaterial on a bootstrap-only path.

Context on review confidence: this finding came from a single-voter review round (the other two providers abstained on API errors), and is hedged in its own wording ("may lead").

jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

| xai | `grok-4.5` | Needs Review | 0.72 |

---

### Summary

This is a large (251-file, partially truncated) squashed merge touching consensus-critical code: fault-tolerance-threshold adoption is made a hard requirement, the proposer's active-validator check is redefined in terms of parent-block bonds, the MakeMint genesis contract's decr protocol gains a method tag (an arity change), and LFS sync gains error propagation plus a 600s no-progress deadline. The bulk of the visible additions are high-quality formal-verification-backed property tests (fork-choice determinism, floor maximality, ancestry closure, equivocation convergence) and a runtime order-dependence guard in the merger. No security vulnerabilities were found in the reviewed portion, but several changes have consensus/backward-compatibility implications that warrant human judgment: the hard failure on missing on-chain threshold can brick legacy chains, the bonds-vs-active-set semantic change may admit proposals that peers reject, and the MakeMint arity change silently deadlocks any un-updated caller. Because the diff is heavily truncated (only ~5% shown), a complete review of the remaining ~240 files is still required.

The visible portion contains useful validation and determinism work, but the new LFS timeout mechanism does not actually impose a hard no-progress deadline when a peer continues sending non-progressing responses, allowing startup to remain blocked indefinitely. The merge diagnostic also has build-dependent fatal behavior that contradicts its documented semantics and adds potentially substantial hot-path memory overhead. These issues should be addressed before merging; additionally, this assessment covers only the provided, truncated portion of the 251-file change.

This is a large sealed-floor merge with substantial consensus-adjacent changes: on-chain fault-tolerance adoption is hardened to a single fail-closed path, LFS sync becomes deadline-aware and error-propagating, deploy finalization gains a known-block fallback, and several formal-verification gaps are pinned with property tests (ancestry, deploy-support promotion, floor maximality, equivocation unbonded-window fix, merge order-dependence guard). The highest-risk items for human review are the proposer eligibility/stake-check rewrites and the deploy-finalization known-block fallback, because both can change node behavior under realistic edge conditions. Overall the direction looks careful and well-tested, but the PR is too broad and consensus-sensitive to approve from the truncated diff alone.

The code change appears to be a merge of the latest changes from the `dev` branch into the feature branch for PR #90, with some cleanup and squashing of consecutive commits. The changes include updates to configuration files, workflow scripts, and various Rust source files. While the changes seem to be mostly functional and safe to merge, there are some performance and style issues that could be improved. Overall, the code looks good, but some minor adjustments could enhance readability and performance.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Missing latest-message mapped to not_bonded() — misleading and may block first propose</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/last_finalized_height_constraint_checker.rs` (line 21)

When a validator has no latest message, the checker now returns `CheckProposeConstraintsResult::not_bonded()` instead of an error. A freshly bonded validator that has never produced a block (and whose latest message hasn't been initialized) would be reported as 'not bonded' and skipped from proposing — if latest messages are only seeded on first block, this could prevent such a validator from ever proposing. The `not_bonded` variant is also semantically wrong for 'no latest message yet'.

---

When a validator has no latest message, the height checker returns CheckProposeConstraintsResult::not_bonded() rather than an error. That softens a previously hard failure into a propose skip. If this path is hit for a bonded validator after restart/state loss, diagnostics become less precise and may hide storage/index issues.

**Suggestion:**
```
Confirm this is the desired UX for first-time proposers only. If so, distinguish 'never proposed' from 'latest message lookup unexpectedly empty' and log/metric the unexpected case at warn/error.
```

</details>

<details>
<summary><b>:orange_circle: Major: MakeMint decr protocol arity change — all call sites and existing purses must match</b> (Reported by: anthropic)</summary>

**File:** `casper/src/main/resources/MakeMint.rho` (line 46)

The purse decr contract changed from `contract thisPurse(@providedDecr, @amount, return)` (3-tuple) to `contract thisPurse(@"decr", @providedDecr, @amount, return)` (4-tuple with method tag). One internal caller (`@src!("decr", ...)` in the split/transfer path) was updated, but any other contract (RevVault, PoS, test contracts, or user code) still sending the old 3-tuple to a purse channel will now silently never match, leaving the caller's `for (@decrSuccess <- result)` blocked forever (a silent deadlock, not an error). The truncated diff makes it impossible to verify all call sites were updated. Also, this changes genesis-contract semantics — nodes with different MakeMint.rho versions in the same shard would diverge.

**Suggestion:**
```
Audit every `!(...decr...)` / purse-decr invocation across all .rho resources and tests, and confirm this contract change is coordinated with a genesis/protocol version bump.
```

</details>

<details>
<summary><b>:orange_circle: Major: known_block_hash fallback can bypass deploy-index consistency checks</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 618)

resolve_with_known_block first tries the deploy index; on Unknown it falls back to an externally supplied block hash. If callers can supply an arbitrary known_block_hash, this may report finalization status for a deploy/block association that is not indexed, weakening the previous index-as-source-of-truth invariant and potentially masking corruption.

**Suggestion:**
```
Restrict the fallback to trusted internal callers only, and still validate that the deploy signature is present in the known block body (which run_prelude_from_block does) plus any additional authorization that the hash is from a legitimate observed path.
```

</details>

<details>
<summary><b>:orange_circle: Major: Active-validator check replaced with parent-bonds check — semantic change</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/proposer.rs` (line 569)

ProductionActiveValidatorChecker previously checked `casper_snapshot.on_chain_state.active_validators`; it now checks whether the validator appears in the FIRST parent block's bonds list. 'Bonded in parent state' and 'active validator' are not the same set (e.g., a validator can be bonded but excluded from the active/epoch committee, or newly bonded but not yet active). This could allow proposals from validators that the on-chain active set would reject, causing those blocks to be rejected by peers during validation. Additionally, `casper_snapshot.parents.first()` with `.unwrap_or(false)` means an empty parent set (genesis-adjacent snapshots) always resolves to not_bonded, potentially blocking legitimate proposals.

**Suggestion:**
```
Confirm this semantic change is intentional and consistent with block validation's committee check. If the goal is to check the parent tip's committee, document why bonds ⇒ eligible-to-propose, and handle the empty-parents case explicitly (e.g., fall back to genesis bonds).
```

</details>

<details>
<summary><b>:orange_circle: Major: Active-validator check now uses parent bonds instead of on-chain active set</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/proposer.rs` (line 570)

ProductionActiveValidatorChecker switched from casper_snapshot.on_chain_state.active_validators to parent.body.state.bonds. Bonded and active are not the same set in PoS (bonded validators can be inactive/jailed/unbonding). This can allow inactive validators to propose or block active ones depending on parent bond lag, and it is a consensus-relevant policy change that needs explicit justification against the PoS model.

**Suggestion:**
```
Confirm this is intentional and matches the formal/protocol definition of propose eligibility. Prefer reading the same on-chain active-validator source used elsewhere, or document why parent bonds are the authoritative committee for proposal constraints.
```

</details>

<details>
<summary><b>:orange_circle: Major: Stake/synchrony checker no longer receives runtime/block-store context</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/proposer.rs` (line 587)

ProductionStakeChecker dropped RuntimeManager and KeyValueBlockStore and now only holds the validator identity. If synchrony_constraint_checker::check previously needed state/history lookups, this can silently weaken or change the synchrony gate. The diff alone does not prove behavioral equivalence.

**Suggestion:**
```
Verify the new synchrony_constraint_checker::check signature and implementation still enforce the same stake/synchrony constraints, and add/adjust tests for under-stake and lagging-validator rejection.
```

</details>

<details>
<summary><b>:orange_circle: Major: Hard failure when on-chain fault-tolerance threshold is absent breaks legacy chains</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/casper.rs` (line 209)

`hash_set_casper` previously tolerated a chain whose genesis predates the fault-tolerance-threshold parameter (logged a warning and used the locally-derived ppm). The new code calls `read_on_chain_fault_tolerance_threshold_ppm` which per the comment now FAILS the node when the value is absent/out-of-range. Any existing shard whose genesis lacks the parameter will be unable to start any node after upgrading — a backward-compatibility break for running networks, not just a consistency improvement.

**Suggestion:**
```
Confirm no deployed chain has a genesis without this parameter, or gate the hard failure behind a protocol-version/feature check with a documented migration path.
```

</details>

<details>
<summary><b>:orange_circle: Major: LFS send failures are now non-fatal and rely solely on resend/deadline</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 389)

request_next switched from try_join_all (fail-fast) to join_all with warn-and-continue. Combined with overall_deadline, this is more resilient, but a permanently broken peer path can now stall until the 600s deadline instead of failing fast. The same pattern is applied in lfs_tuple_space_requester.

**Suggestion:**
```
Ensure callers surface the returned error handle and that CI/integration tests cover partial-send failure, zero-progress timeout, and successful recovery via resend. Consider a max-resend or consecutive-failure budget in addition to wall-clock deadline.
```

</details>

<details>
<summary><b>:orange_circle: Major: Forward-horizon deadline is not enforced during continuous traffic</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 578)

As in the tuple-space requester, the deadline is evaluated only from the idle-timeout branch. Response activity can continually reset the idle timer even when no new paths become Done. Consequently, duplicate or useless peer responses can keep this stream alive forever, bypassing `overall_deadline` and blocking startup.

**Suggestion:**
```
Maintain a separate progress-deadline sleep in the select loop and reset it exclusively when `done_count` increases. The retry/backoff idle timer should remain independent from the hard no-progress deadline.
```

</details>

<details>
<summary><b>:orange_circle: Major: Potential performance bottleneck</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/lfs_tuple_space_requester.rs` (line 525)

The function `stream` uses a loop with exponential backoff, which could lead to performance issues if the loop runs for a long time. Consider adding a maximum number of retries or a timeout to limit the execution time.

**Suggestion:**
```
Add a maximum number of retries or a timeout to the loop.
```

</details>

<details>
<summary><b>:orange_circle: Major: Progress deadline can be bypassed by non-progress responses</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/engine/lfs_tuple_space_requester.rs` (line 699)

The overall no-progress deadline is checked only when `idle_timeout` fires. However, the response-processing path resets that idle timer whenever a response is received, while `last_progress` advances only when `done_count` increases. A peer that continually sends valid but duplicate or otherwise non-progressing responses can therefore prevent the timeout branch from ever running, leaving initialization stuck indefinitely despite the new 600-second deadline. This is an availability risk because peer-controlled traffic can keep a node from transitioning to Running.

**Suggestion:**
```
Add an independent deadline timer to the `tokio::select!`, based on `last_progress + overall_deadline`, and reset that timer only when `done_count` increases. Do not rely on the response-idle timer to enforce a progress deadline.
```

</details>

<details>
<summary><b>:orange_circle: Major: debug_assert!(false) contradicts documented 'non-fatal' guard behavior</b> (Reported by: anthropic, openai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 578)

The guard is documented as 'detection only … non-fatal', but `debug_assert!(false, ...)` will panic in debug builds (including tests and any debug-profile deployments) before the tracing/metrics path runs. This makes the guard's behavior differ between build profiles, and in debug builds a guard trip aborts the merge entirely.

---

The surrounding documentation repeatedly describes the order-dependence guard as detection-only and non-fatal, but `debug_assert!(false, ...)` turns a detected violation into a process panic in debug or debug-assertion-enabled builds. Since this path processes consensus-derived state, a malformed or adversarially constructed merge could terminate such a node instead of merely emitting the advertised log and metric.

**Suggestion:**
```
Either drop the debug_assert and rely on the error log + metric, or explicitly document that a trip is intended to be fatal under debug builds.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Ancestry property test is O(n^2) BFS over generated DAGs</b> (Reported by: xai)</summary>

**File:** `block-storage/tests/block_dag_storage_test.rs` (line 301)

is_dag_ancestor_matches_reflexive_transitive_closure_over_parents compares every pair with a fresh stack BFS. With proptest and up to 10 blocks this is likely fine, but it can become expensive if the generator bounds grow.

**Suggestion:**
```
Keep current bounds or precompute full ancestor sets once per DAG to avoid repeated BFS if the case count increases.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: try_join_all → join_all downgrades chunk-send failures to warnings</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 392)

Both LFS requesters now swallow individual chunk-send errors (`join_all` + warn) instead of failing fast. Persistent, non-transient failures (serialization bugs, closed transport) will now only surface after the full 600s no-progress deadline expires, delaying failure detection by up to 10 minutes and hiding the root-cause error (only the generic 'no progress' RuntimeError is surfaced).

**Suggestion:**
```
Preserve the last send error and include it in the deadline error message; consider failing fast when ALL sends in a fan-out fail.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent comment style</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 1140)

The comments in the file use a mix of uppercase and lowercase letters, which can make the code harder to read. Consider using a consistent style for comments.

**Suggestion:**
```
Use a consistent style for comments, such as all uppercase or all lowercase.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent use of whitespace</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/finality/floor.rs` (line 1057)

The code uses inconsistent whitespace around operators and after commas, which can make the code harder to read. Consider using a consistent style for whitespace.

**Suggestion:**
```
Use a consistent style for whitespace, such as always using a space around operators and after commas.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: OrderDependenceGuard adds always-on allocation overhead to the merge hot path</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 368)

The Finding-A guard clones every datum byte-vector of every survivor's `datums_changes` (added and removed) into nested HashMaps on every merge, purely for detection. For large conflict sets / large state changes this is O(total datum bytes) extra allocation and hashing on a consensus-critical hot path, even though the guard is expected to (almost) never trip.

**Suggestion:**
```
Store hashes of datums instead of full byte clones, or gate the guard behind a debug/diagnostics feature flag or a cheap pre-check (e.g., only run when a channel is touched by ≥2 survivors).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Diagnostic guard duplicates all merged datum payloads</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 477)

`OrderDependenceGuard::observe` clones each distinct added and removed datum into additional nested hash maps for the entire merge. This adds hashing work and potentially duplicates an amount of memory proportional to the complete survivor payload on a hot consensus path, even in release builds. Large merges may therefore experience a significant transient memory and CPU increase solely for diagnostics.

**Suggestion:**
```
Consider tracking fixed-size datum digests with collision-safe verification on matches, gating the expensive diagnostic behind configuration, or integrating duplicate detection into existing conflict analysis so full datum byte vectors do not need to be retained twice.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Order-dependence guard is non-fatal in production</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 565)

The Finding-A guard correctly detects multi-survivor duplicate datums on non-mergeable channels, but only debug_assert!s and emits a metric/error log. In release builds a true invariant violation will not stop merge, so association-dependent roots could still be produced if the conflict checker regresses.

**Suggestion:**
```
If this invariant is consensus-critical, consider failing merge (or at least escalating to a hard error path) when first_offender is Some, not only counting metrics.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: MakeMint decr API shape changed</b> (Reported by: xai)</summary>

**File:** `casper/src/main/resources/MakeMint.rho` (line 44)

Purse decr now requires an explicit "decr" method tag and =*thisDecr exact-match. This is a rholang contract interface change and may break callers still using the old arity/shape.

**Suggestion:**
```
Ensure all in-repo callers were updated (the transfer path was) and note the compatibility break in release notes if external contracts depend on MakeMint.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unnecessary allocation in deploy-id lookup</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 166)

`dag.lookup_by_deploy_id(&sig.to_vec())` allocates a Vec just to take a reference to it. If the API accepts `&[u8]`, pass `sig` directly; if it requires `&Vec<u8>`, consider changing its signature.

**Suggestion:**
```
Pass `sig` directly or adjust `lookup_by_deploy_id` to accept `&[u8]`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: LFS_SYNC_DEADLINE hardcoded at 600s</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/initializing.rs` (line 597)

The overall LFS sync deadline is a compile-time constant. Large tuple-space states over slow links may legitimately need longer; operators have no knob. The adjacent request timeout comment even mentions configurability as a goal ('Keep LFS retry cadence configurable').

**Suggestion:**
```
Expose the deadline via node configuration or an environment variable with the 600s default.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Load-bearing 'DO NOT re-add purge' invariant documented only in a comment</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 12)

The module doc explains that the finalization-time rejected-deploy purge must stay removed and points to a slow (~308s) integration test as its only guard. A comment plus one heavy test is fragile protection for a correctness invariant that a well-meaning refactor could reintroduce.

**Suggestion:**
```
Consider adding a fast unit-level assertion (e.g., that FinalizationContext has no rejected_deploy_buffer handle) or a lint/architecture test that fails quickly if the purge is reintroduced.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Large historical design comments are valuable but dense</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 12)

The PR adds extensive formal-methods and anti-regression commentary across finalization, fork-choice, equivocation, and merge code. This is useful for future reviewers, but some modules are becoming harder to scan.

**Suggestion:**
```
Keep the critical 'do not reintroduce' notes in-code; consider moving long FV narratives to docs/theory with short pointers from the implementation.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test name no longer matches asserted behavior</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/equivocation_detector.rs` (line 1004)

`tier0_polluted_record_falsely_neglects_honest_block` now asserts the OPPOSITE (the record is never polluted and the honest block is NOT falsely flagged). The name describes the pre-fix bug, not the post-fix property being pinned.

**Suggestion:**
```
Rename to something like `tier0_unbonded_observation_does_not_pollute_record_or_flag_honest_block`.
```

</details>
---

<details>

This is a large (251-file, partially truncated) squashed merge touching consensus-critical code: fault-tolerance-threshold adoption is made a hard requirement, the proposer's active-validator check is redefined in terms of parent-block bonds, the MakeMint genesis contract's decr protocol gains a method tag (an arity change), and LFS sync gains error propagation plus a 600s no-progress deadline. The bulk of the visible additions are high-quality formal-verification-backed property tests (fork-choice determinism, floor maximality, ancestry closure, equivocation convergence) and a runtime order-dependence guard in the merger. No security vulnerabilities were found in the reviewed portion, but several changes have consensus/backward-compatibility implications that warrant human judgment: the hard failure on missing on-chain threshold can brick legacy chains, the bonds-vs-active-set semantic change may admit proposals that peers reject, and the MakeMint arity change silently deadlocks any un-updated caller. Because the diff is heavily truncated (only ~5% shown), a complete review of the remaining ~240 files is still required.

The code change appears to be a merge of the latest changes from the `dev` branch into the feature branch for PR #90, with some cleanup and squashing of consecutive commits. The changes include updates to configuration files, workflow scripts, and various Rust source files. While the changes seem to be mostly functional and safe to merge, there are some performance and style issues that could be improved. Overall, the code looks good, but some minor adjustments could enhance readability and performance.

#### openai (gpt-5.6-sol)

The visible portion contains useful validation and determinism work, but the new LFS timeout mechanism does not actually impose a hard no-progress deadline when a peer continues sending non-progressing responses, allowing startup to remain blocked indefinitely. The merge diagnostic also has build-dependent fatal behavior that contradicts its documented semantics and adds potentially substantial hot-path memory overhead. These issues should be addressed before merging; additionally, this assessment covers only the provided, truncated portion of the 251-file change.

#### xai (grok-4.5)

This is a large sealed-floor merge with substantial consensus-adjacent changes: on-chain fault-tolerance adoption is hardened to a single fail-closed path, LFS sync becomes deadline-aware and error-propagating, deploy finalization gains a known-block fallback, and several formal-verification gaps are pinned with property tests (ancestry, deploy-support promotion, floor maximality, equivocation unbonded-window fix, merge order-dependence guard). The highest-risk items for human review are the proposer eligibility/stake-check rewrites and the deploy-finalization known-block fallback, because both can change node behavior under realistic edge conditions. Overall the direction looks careful and well-tested, but the PR is too broad and consensus-sensitive to approve from the truncated diff alone.
</details>

---

jeffrey-l-turner:
## Review Finding Triage — 2 findings resolved (false positives)

Follow-up triage of the multi-agent review findings. Both findings below were verified against the actual control flow and are **closed as false positives**.

---

### 1. ❌ False positive — "known_block_hash fallback can bypass deploy-index consistency checks"

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (reported at ~618, actual fallback at 621–641) · **Reported by:** Grok · **Status: closed**

The fallback cannot bypass the deploy-index consistency check, because the check runs *before* the fallback is consulted and its error propagates past it:

1. **The consistency check fires first and short-circuits.** `run_prelude(dag, block_store, sig)?` runs unconditionally at the top of `resolve_with_known_block`. If the sig *is* in the deploy index but the indexed block's body doesn't list it, that returns `Err(DeployFinalizationCorruption)` and the `?` propagates immediately — the `known_block_hash` match arm is never reached. Whenever there is an index entry to be consistent with, the corruption path wins.
2. **The fallback only fires when there is nothing to check.** It runs only on `PreludeOutcome::Unknown`: sig absent from the deploy index, or indexed block absent from the block store. Both are deliberately typed as *absence* (previously resolved to `pending_unknown`), not corruption — and that typing predates this change. There is no index entry to compare against in that branch.
3. **The fallback applies the same body-containment validation.** `run_prelude_from_block` on the supplied hash requires the sig to appear in that block's `body.deploys`; otherwise it returns the same `DeployFinalizationCorruption` (downcast to `pending_unknown` at the API layer in `block_api.rs`). A bogus hash cannot manufacture an `Active` state.
4. **The hash is not caller-controlled input.** Both call sites (`web_api.rs`, `deploy_grpc_service_v1.rs`) pass the hash the node itself just resolved via `BlockAPI::find_deploy` (deploy index or the recent-blocks scan). Closing the index-lag race — `findDeploy` returns a block while the status API says `pending_unknown` for the same sig — is the purpose of the fallback.

Net: the change strictly *adds* a resolution path for the index-lag race; every corruption path is intact.

---

### 2. ❌ False positive (as a defect) — "LFS send failures are now non-fatal and rely solely on resend/deadline"

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs:389` · **Reported by:** Grok · **Status: closed**

The observation is factually accurate, but the change is the fix, not the regression:

1. **The old behavior was the bug-shaped one.** Pre-change, the fan-out used `try_join_all`, so a *single* failed chunk send terminated the entire forward-horizon stream — during initial bootstrap, exactly where transient send failures are routine. Worse, the old code then **discarded the recorded error** (`let _ = last_error;`), so callers only saw `is_finished() == false` with no context.
2. **The new failure model is bounded and the error is now surfaced.** Failed sends are counted and warned, the affected paths remain pending, and the idle-timeout arm re-requests them with exponential backoff. Zero progress for `LFS_SYNC_DEADLINE` (600s) records a typed `CasperError` and terminates — and the stream now *returns* the error handle instead of dropping it.
3. **Safety is fully preserved at the consumer.** `initializing.rs` has three independent hard-fail checks before the node can transition to Running: the surfaced deadline error, an unfinished final state ("incomplete sync; N chunk paths still pending"), and a missing final state. There is no path where lost sends silently yield a Running node with an incomplete horizon.

Residual (non-blocking) note: a permanently unreachable peer now takes up to 600s to detect instead of failing immediately; an all-sends-failed early-exit could shorten that. That is a tuning refinement, not a defect, and trades against the old behavior where one transient failure aborted bootstrap outright.

---

*Triage: claude-session (agent-assisted verification against the code on `sealed-floor-merge-squashed`).*

PR #130, opened 2026-07-16
feat(fileio): nativeChown + owner/group in stat records

metaweta:
## Summary

Fills out the last Phase-1 native primitive from the [File I/O FIP](https://github.com/F1R3FLY-io/FIPS/blob/main/fileio/under-review/2026-02-06-File-IO/2026-02-06-File-IO.md) §"Permission management" (TODO 8): `nativeChown` plus the paired `owner`/`group` fields on the `stat`/`entries` record shape (TODO 5).

Stacked on [PR #120](https://github.com/F1R3FLY-io/f1r3node-rust/pull/120). GitHub will auto-retarget this to `dev` once #120 merges.

### Layer 1 — NSS lookup helpers (new module `io/nss.rs`)

Direct `libc` FFI wrapped in a safe API:

- `resolve_uid(name: &str) -> Option<u32>` — via `getpwnam_r`.
- `resolve_gid(name: &str) -> Option<u32>` — via `getgrnam_r`.
- `user_name(uid: u32) -> Option<String>` — via `getpwuid_r`.
- `group_name(gid: u32) -> Option<String>` — via `getgrgid_r`.

Buffer starts at 512 B and doubles up to 64 KiB on `ERANGE`, per the POSIX thread-safe pattern. Missing users/groups return `None` rather than surfacing an error — matches the FIP convention that irresolvable fields are omitted rather than fabricated. Interior nulls, non-UTF-8 names, and other malformed inputs all funnel to `None`.

`libc` is added as a direct dep of the `rholang` crate (was already available transitively via `console` / `native-tls` / etc.); pinned via workspace `libc = "0.2"`.

### Layer 2 — Handler + registration

- **`nativeChown(path: String, owner: String, group: String) -> [true] | [false, code, msg]`**. Empty string for either `owner` or `group` means "don't change" (passed through to `std::os::unix::fs::chown` as `None`). Unknown names return `FSERR_BAD_ARG` — consistent with the FIP wording that numeric uid/gid isn't exposed at the Rholang surface, so the string is the only handle we accept.
- The chown syscall is synchronous (`std::os::unix::fs::chown` has no async twin); wrapped in `tokio::task::spawn_blocking` so tokio worker threads stay responsive.
- Non-Unix hosts return `FSERR_UNSUPPORTED` (same shape as `native_chmod`).
- `FixedChannels::native_chown = byte 56`, `BodyRefs::NATIVE_CHOWN = 55`. New `Definition` row in `std_system_processes()`. Added to `non_deterministic_ops()` (chown outcome depends on host NSS + permission state, which can differ across nodes).

### Layer 3 — stat/entries records

`io/stat.rs`'s `stat_record` now emits `owner: String` and `group: String` fields on Unix, using `MetadataExt::{uid,gid}` + the new NSS reverse-lookups. Fields are omitted when NSS reverse-lookup fails (uid orphaned from `/etc/passwd`, backend down), matching how `mtime`/`ctime`/`atime` are omitted when unavailable. Non-Unix hosts omit both fields entirely.

Module docstring updated to list the two new fields.

## Not in this PR (deferred to later phases)

- End-to-end tests of `nativeChown` through the runtime — would require standing up a full `SystemProcesses`. Naturally exercised by the FS-agent Rholang tests once that layer lands.
- Consensus-mode filtering of `owner`/`group`/timestamps in the agent layer (the FIP says these are oracular-only). The native layer emits everything it can compute; stripping is the agent's job.
- FS-agent Rholang wrapping `nativeChown` into a user-facing `chmod`/`chown` method — depends on rholang-rs [PR #96](https://github.com/F1R3FLY-io/rholang-rs/pull/96) (try/catch sugar) merging first.

## Tests

- **`io::nss` (6 tests)** — round-trip lookups for `root` / gid 0, and the four failure paths (unknown user/group/uid, interior null in name).
- **`io::stat` (+1 test)** — `stat_record_includes_owner_and_group_on_unix` asserts both fields are present for a file owned by the test uid/gid.
- **`system_processes::tests::non_deterministic_ops_covers_every_fileio_body_ref` (updated)** — the fileio non-det-ops list is now 19 entries (was 18) with `NATIVE_CHOWN` added.

Total: 30 io-submodule tests pass (23 prior + 7 new).

## Verified

- [x] `cargo build -p rholang` clean.
- [x] `cargo test -p rholang --lib io::` — 30 pass.
- [x] `cargo test -p rholang --lib` — 226 total pass (222 prior + 4 new).
- [x] `empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes — adding native primitives doesn't affect tuplespace state.
- [x] Full pre-push hook (`fmt` / `clippy` / `deny` + release-mode tests across all 11 crates) passed on this exact commit set.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/130"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786836379&installation_id=145946300&pr_number=130&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F130&signature=c75f4337e997f104d5758a385a76246414d41627a3502412a1ec074e994ec701"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5095,6 +5095,7 @@ dependencies = [
  "itertools 0.14.0",
  "k256",
  "lazy_static",
+ "libc",
  "metrics 0.23.1",
  "models",
  "num-bigint",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ num-traits = "0.2"
 num-integer = "0.1"
 uuid = { version = "1", features = ["v4"] }
 clap = { version = "4.5", features = ["derive"] }
+libc = "0.2"
 lazy_static = "1.5"
 indexmap = "2.8"
 lz4_flex = "0.11"
--- a/rholang/Cargo.toml
+++ b/rholang/Cargo.toml
@@ -44,6 +44,12 @@ dotenv = "0.15.0"
 indexmap = { workspace = true }
 rayon = { workspace = true }
 tempfile = { workspace = true }
+# libc is available transitively (via console, native-tls, etc.), but
+# needed as a direct dep for the getpwnam_r / getgrnam_r / getpwuid_r
+# / getgrgid_r NSS lookups used by the File I/O `nativeChown` and by
+# the stat/entries record's owner/group fields (FIP 2026-02-06 §Entries
+# record shape TODO 8).
+libc = { workspace = true }
 clap = { workspace = true }
 rholang-parser = { workspace = true }
 validated = { workspace = true }
--- a/rholang/src/rust/interpreter/io/mod.rs
+++ b/rholang/src/rust/interpreter/io/mod.rs
@@ -12,6 +12,8 @@
 
 pub mod handle_table;
 pub mod mode;
+#[cfg(unix)]
+pub mod nss;
 pub mod path;
 pub mod response;
 pub mod stat;
--- a/rholang/src/rust/interpreter/io/nss.rs
+++ b/rholang/src/rust/interpreter/io/nss.rs
@@ -1,0 +1,206 @@
+//! Name Service Switch lookups for user/group names ↔ uid/gid.
+//!
+//! Wraps the POSIX thread-safe `getpwnam_r` / `getgrnam_r` /
+//! `getpwuid_r` / `getgrgid_r` calls in a small safe API. Used by
+//! the File I/O layer for `nativeChown` (name → id) and for the
+//! `stat`/`entries` record's `owner`/`group` fields (id → name).
+//!
+//! Missing users/groups return `None` rather than surfacing an
+//! error, matching the FIP §Entries record shape convention:
+//! fields that can't be resolved are simply omitted.
+//!
+//! The buffer for the strings that the `passwd` / `group` struct
+//! fields point into starts at 512 bytes and doubles up to 64 KiB
+//! on ERANGE, per the POSIX pattern. Systems with unusual /etc/passwd
+//! entries (very long GECOS fields, huge group member lists) would
+//! trip the ceiling; that's rare enough in practice that we prefer
+//! bounded memory over unbounded growth.
+
+use std::ffi::{CStr, CString};
+
+const INITIAL_BUF: usize = 512;
+const MAX_BUF: usize = 64 * 1024;
+
+/// Resolve a user name to a uid. `None` if the user does not
+/// exist, if `name` contains an interior null byte, or if the
+/// lookup fails for any other reason.
+pub fn resolve_uid(name: &str) -> Option<u32> {
+    let c_name = CString::new(name).ok()?;
+    let mut buf = vec![0u8; INITIAL_BUF];
+    loop {
+        // SAFETY: getpwnam_r requires a zero-initialized `passwd`
+        // and a mutable pointer for the result. We give it both.
+        // The strings inside `pwd` point into `buf`, which we keep
+        // alive until we've copied out the uid.
+        let mut pwd: libc::passwd = unsafe { std::mem::zeroed() };
+        let mut result: *mut libc::passwd = std::ptr::null_mut();
+        let rc = unsafe {
+            libc::getpwnam_r(
+                c_name.as_ptr(),
+                &mut pwd,
+                buf.as_mut_ptr() as *mut libc::c_char,
+                buf.len(),
+                &mut result,
+            )
+        };
+        if rc == 0 {
+            // POSIX: rc==0 with result==NULL means "not found".
+            if result.is_null() {
+                return None;
+            }
+            return Some(pwd.pw_uid);
+        }
+        if rc == libc::ERANGE && buf.len() < MAX_BUF {
+            buf.resize(buf.len() * 2, 0);
+            continue;
+        }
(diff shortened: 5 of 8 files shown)


metaweta:
## Update — M1 hardening from the security review

Followed up with commit [`bb146833`](https://github.com/F1R3FLY-io/f1r3node-rust/pull/130/commits/bb146833) addressing **M1** from the security analysis: NSS lookups were running on the tokio event thread, which starves other Rholang tasks on hosts using LDAP / NIS / sssd (where `getpwnam_r` can block for seconds under directory-server load or unreachability).

### Three handler edits

- **`native_chown`** — merged `nss::resolve_uid` / `resolve_gid` and the `chown` syscall into a single `spawn_blocking`. Closure returns `Result<(), Box<Par>>` (boxed to keep the Ok path a single word and silence `clippy::result_large_err` on the ~250-byte `Par`).
- **`native_stat`** — wrapped the single `stat_record` call in `spawn_blocking` after the async `symlink_metadata` fetch. `stat_record` does NSS reverse-lookup via `user_name`/`group_name` on Unix, so it needs the same off-thread treatment.
- **`native_entries`** — wrapped the sort + per-entry `stat_record` batch in one `spawn_blocking`. One blocking task per directory listing is cheaper than one per entry and matches the shape of the operation (already-collected data → CPU-bound transformation).

### Behavior

- **On `/etc/passwd`-only hosts** (most dev + CI): unchanged. The extra `spawn_blocking` round-trip is tens of microseconds.
- **On hosts with slow NSS backends** (LDAP, NIS, sssd): the tokio worker returns immediately; slow calls run on the blocking pool without starving other Rholang tasks.

### Tests

Existing 30 `io::` tests still pass — they exercise `nss` and `stat_record` directly (sync), so the async-handler changes don't need new coverage. The future FS-agent end-to-end tests will exercise the full path once that layer lands.

### Remaining security-review items

- **L1** (symlink semantics divergence between `native_stat`/no-follow and `native_chown`/follow) — doc-only, will note in the follow-up FS-agent PR that quarantines paths.
- **L2** (no NSS caching) — can be a follow-up PR; not blocking.
- **L3** (owner/group info disclosure inside the powerbox tree) — data-classification consideration for the future agent-layer PR.

dylon (approved):
claude found no blockers but had the following recommendations:

  ## Review — PR #130 (nativeChown + owner/group)

  **Verdict: approve.** Same reachability argument as #120. The FFI is genuinely well done:
  correct POSIX `_r` reentrant variants, sound `unsafe` blocks with SAFETY comments that
  actually match what the code does, ERANGE doubling with a 64 KiB ceiling, and
  `spawn_blocking` around every blocking NSS call — including the good call to push
  `stat_record` and the whole `entries` batch off the reactor thread.

  ### Must fix

  1. **A transient NSS failure is indistinguishable from "user does not exist."** In
     `resolve_uid`/`resolve_gid`, any `rc != 0` that isn't `ERANGE` falls through to
     `return None` — the same value returned for `rc == 0 && result.is_null()`. So an
     LDAP/NIS/sssd outage (EIO, EAGAIN) surfaces as
     `FSERR_BAD_ARG: unknown user "alice"`, telling the caller their *input* was malformed
     when the directory service was down. The FIP has `FSERR_IO` for precisely this. The
     same collapse silently drops `owner`/`group` from every `stat`/`entries` record during
     an outage, making the record shape depend on transient infrastructure state — the same
     variable-shape hazard as the `ctime` issue in #120. Please distinguish not-found from
     lookup-failed and map the latter to `FSERR_IO`.
  2. **`chmod` and `chown` follow symlinks.** `tokio::fs::set_permissions` and
     `std::os::unix::fs::chown` both resolve the final symlink. Since the FIP's "directories
     must be trees" rule is only checked at launch (see #131), a symlink appearing afterwards
     redirects the permission change outside the sandbox. Prefer `lchown` /
     `fchmodat(AT_SYMLINK_NOFOLLOW)`, or operate on an `O_NOFOLLOW` fd.

  ### Note

  3. `chown` to another user needs `CAP_CHOWN`. Nodes running as root in containers will
     therefore let a deploy rewrite ownership anywhere inside a granted tree — worth an
     explicit line in the operator docs when the powerbox lands.
  4. The unmetered-syscall gate from #120 item 5 applies here too.

metaweta:

PR #131, opened 2026-07-17
feat(fileio): static-config + CLI-flag surface for the powerbox

metaweta:
## Summary

Adds the boot-time configuration surface for the [File I/O FIP](https://github.com/F1R3FLY-io/FIPS/blob/main/fileio/under-review/2026-02-06-File-IO/2026-02-06-File-IO.md) §"With the config file" and §"With flags". Parses, validates, and stores the four lists of pre-authorized files and directories the deployment ships with at boot; the FS-agent Rholang layer (later PR) consumes this at genesis and hands out capabilities to the initial deployment.

**Nothing here opens fds** — that's the provisioning step, which lives with the future FS-agent bootstrap PR. This PR is scoped to config schema + CLI flags + validation so downstream work has a well-typed structure to consume.

Stacked on [PR #130](https://github.com/F1R3FLY-io/f1r3node-rust/pull/130) (nativeChown). GitHub will auto-retarget this to `dev` once #130 (and its own base #120) merge.

## Config schema

New module `node/src/rust/configuration/file_io.rs`:

- **`FileIo`** — top-level section on `NodeConf` with four `Vec<...>` fields:
  - `oracle_static_files: Vec<StaticFileEntry>`
  - `oracle_static_dirs: Vec<StaticDirEntry>`
  - `consensus_static_files: Vec<StaticFileEntry>` (accepted and stored but not exercised until a follow-up FIP defines multi-node consensus-mode file-state semantics)
  - `consensus_static_dirs: Vec<StaticDirEntry>` (same status)

  All default empty via `#[serde(default)]`.

- **`StaticFileEntry { path: PathBuf, mode: String }`** — mode is one of the nine fopen strings per FIP §"With the config file" table (`"r"`, `"w"`, `"a"`, `"r+"`, `"w+"`, `"a+"`, `"wx"`, `"w+x"`, `"wbx"`); serde default `"r"`.

- **`StaticDirEntry { path: PathBuf, mode: String }`** — mode restricted to `"r"` or `"rw"` per FIP §"The `openDir` method"; serde default `"r"` (default-restrictive).

- Constructors `StaticFileEntry::from_flag` / `StaticDirEntry::from_flag` encode the CLI-flag defaults, which **differ for dirs**: flag entries default to `"rw"` (default-permissive) per FIP, since the CLI form is intended for interactive/dev use where broader default authority is convenient. Server deployments should prefer the config-file form and opt in to `"rw"` explicitly per-dir.

## CLI flags

Added to `RunOptions`:

- `--oracle-static-file PATH` (repeatable)
- `--oracle-static-dir PATH` (repeatable)
- `--consensus-static-file PATH` (repeatable)
- `--consensus-static-dir PATH` (repeatable)

Each occurrence appends one entry to the corresponding `file_io.*` list. The `override_config_values` merge in `config_mapper.rs` is `Vec::extend` rather than `try_override_value` — unlike scalar flags, list flags are **additive** to the config-file block, not replacements. `#[derive(Default)]` added to `RunOptions` so targeted tests can construct a mostly-empty options struct.

## Validation

`file_io::validate(&FileIo) -> Result<(), Vec<FileIoConfigError>>` (called at future config-load time; not wired to boot flow yet) reports every failure at once so operators can fix all misconfigurations in one edit pass. Enforces:

- **Path exists** — via `symlink_metadata` so a symlink at the leaf is reported as a traversal, not as NotFound on the target.
- **No symlink traversal** — the path canonicalizes to itself. FIP §"With the config file" promises "no symbolic or hard links" for static-provisioned entries; enforced here at load time.
- **File-mode string is valid** — reuses `rholang::rust::interpreter::io::mode::open_options_for` as the source of truth.
- **Dir-mode string is `"r"` or `"rw"`**.

`FileIoConfigError` implements `Display` with an actionable per-variant message identifying which of the four sections the offending entry came from.

## `defaults.conf`

New `file-io { ... }` block with all four lists initialized empty, plus an extensive comment block documenting the mode-string options, the config-vs-flag default asymmetry for dirs, the consensus-side entries' deferred-follow-up status, and the symlink-rejection promise.

## Tests

- **`file_io::tests`** (12 tests): defaults are empty; each entry type's default mode; validate accepts a real file with each of the nine fopen modes; each validation failure mode (missing path, invalid file mode, invalid dir mode, symlink traversal Unix-only); multiple errors surfaced at once.
- **`config_mapper::tests::file_io_cli_flags_append_to_config_entries`** (1 test): loads `NodeConf` from `EMBEDDED_DEFAULTS` (proving the new HOCON block parses), seeds one entry per list, runs `override_config_values` with a `RunOptions` whose only populated fields are the four file-io CLI flags, asserts each flag occurrence lands as an additional entry with the correct per-surface default mode.

## Not in this PR

- **Provisioning**: opening the fds via `FileHandleTable::insert` and building the injection bundle for the FS-agent's genesis deploy. That's the FS-agent bootstrap PR.
- **Wiring `validate` into the boot flow**: currently the function exists and is unit-tested but no boot code calls it. That happens with the provisioning step, which needs to run validation just before opening the fds.
- **Any use of the `consensus-*` entries**: accepted and stored per the FIP but not exercised by this FIP.

## Verified

- [x] `cargo build -p node` clean.
- [x] `cargo test -p node --lib configuration::` — 28 pass, including the new tests and the existing `embedded_defaults_deserialize_into_node_conf` (proves the new `file-io` HOCON block parses cleanly).
- [x] `cargo test -p node --lib` — 123 total pass.
- [x] `empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes — this PR is node-crate-only, doesn't touch rholang system processes or the tuplespace.
- [x] Full pre-push hook (`fmt` / `clippy` / `deny` + release-mode tests across all 11 crates) passed on this exact commit set.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/131"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786894722&installation_id=145946300&pr_number=131&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F131&signature=a31d7865f5303c8474bb5394d95a8c5f910f5fffba82ed560e80769fe000a54e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3593,6 +3593,7 @@ dependencies = [
  "serial_test",
  "shared",
  "sysinfo",
+ "tempfile",
  "thiserror 2.0.18",
  "tokio",
  "tokio-stream",
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -83,6 +83,7 @@ rspace_plus_plus = { path = "../rspace++" }
 casper = { path = "../casper", features = ["test-utils"] }
 tower = { workspace = true }
 serial_test = { workspace = true }
+tempfile = { workspace = true }
 
 [build-dependencies]
 tonic-prost-build = { workspace = true }
--- a/node/src/main/resources/defaults.conf
+++ b/node/src/main/resources/defaults.conf
@@ -784,3 +784,38 @@ logging {
     retention = 14
   }
 }
+
+# File I/O static provisioning (FIP 2026-02-06 File-I/O §"With the
+# config file"). Empty by default -- deployments that want the FS
+# agent to hand out capabilities at boot populate these lists.
+#
+# Each entry names an absolute path plus an fopen-style mode string:
+#   { path = "/srv/data/log.txt", mode = "a+" }
+#
+# File modes are the nine fopen-style strings the FIP §"With the
+# config file" table enumerates:
+#   "r", "w", "a", "r+", "w+", "a+", "wx", "w+x", "wbx"
+# File-entry default: "r" (read-only).
+#
+# Dir modes are restricted to:
+#   "r"  -- permits read-side methods only.
+#   "rw" -- permits the full mutation surface plus `path: true` on openDir.
+# Dir-entry default: "r" (default-restrictive) for entries authored
+# in this config block. The equivalent --oracle-static-dir CLI flag
+# defaults to "rw" (default-permissive) instead, since flags are
+# aimed at interactive/dev use where broader default authority is
+# convenient. Server deployments should prefer this config block.
+#
+# `consensus-*` entries are accepted and stored but not exercised
+# by this FIP; a follow-up FIP defines multi-node consensus-mode
+# file-state semantics.
+#
+# Validation runs at config-load time: every path must exist and
+# must NOT traverse a symbolic link. Failures are reported in one
+# batch so operators can fix everything in one edit pass.
+file-io {
+  oracle-static-files = []
+  oracle-static-dirs = []
+  consensus-static-files = []
+  consensus-static-dirs = []
+}
--- a/node/src/rust/configuration/commandline/config_mapper.rs
+++ b/node/src/rust/configuration/commandline/config_mapper.rs
@@ -374,6 +374,35 @@ impl ConfigMapper<Options> for NodeConf {
                 &mut self.casper.heartbeat_conf.advanced.deploy_recovery_max_lag,
                 run.heartbeat_advanced_deploy_recovery_max_lag,
             );
+
+            // File I/O static provisioning (FIP 2026-02-06 §"With
+            // flags"). CLI-flag entries APPEND to whatever the
+            // config file's `[file-io]` block declared -- flags are
+            // additive rather than overriding, because they name
+            // individual entries not the whole list. The wrapping
+            // `from_flag` constructors apply the per-surface
+            // defaults ("r" for files, "rw" for dirs).
+            use crate::rust::configuration::file_io::{StaticDirEntry, StaticFileEntry};
+            self.file_io.oracle_static_files.extend(
+                run.oracle_static_file
+                    .into_iter()
+                    .map(StaticFileEntry::from_flag),
+            );
+            self.file_io.oracle_static_dirs.extend(
+                run.oracle_static_dir
+                    .into_iter()
+                    .map(StaticDirEntry::from_flag),
+            );
+            self.file_io.consensus_static_files.extend(
+                run.consensus_static_file
+                    .into_iter()
+                    .map(StaticFileEntry::from_flag),
+            );
+            self.file_io.consensus_static_dirs.extend(
+                run.consensus_static_dir
+                    .into_iter()
+                    .map(StaticDirEntry::from_flag),
+            );
         }
     }
 
@@ -678,6 +707,10 @@ mod tests {
                 heartbeat_advanced_deploy_recovery_max_lag: Some(333),
                 synchrony_finalized_baseline_enabled: Some(false),
                 synchrony_finalized_baseline_max_distance: Some(666666),
+                oracle_static_file: Vec::new(),
+                oracle_static_dir: Vec::new(),
+                consensus_static_file: Vec::new(),
+                consensus_static_dir: Vec::new(),
             })),
         };
 
@@ -822,6 +855,7 @@ mod tests {
                 deployer_private_key: None,
             },
             openai: Default::default(),
+            file_io: Default::default(),
         };
 
         // Apply CLI options to the configuration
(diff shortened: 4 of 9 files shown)


metaweta:
## Update — M1 + M2 hardening from the security review

Followed up with commit [`3cbd5633`](https://github.com/F1R3FLY-io/f1r3node-rust/pull/131/commits/3cbd5633) addressing both Medium findings from the security review.

### M1 — `validate()` now runs at boot

Previously `file_io::validate` existed but was only called from tests. A running node would accept invalid modes, missing paths, and symlink-traversing paths silently — errors would only surface when the (future) FS-agent tried to consume the config at first `openFile`, buried in an unrelated stack trace.

Wired into the existing `configuration::validate_config` at boot-load time. Failure is fatal: `eyre::bail!` with all collected errors joined into a multi-line message so operators see every misconfiguration in one edit pass rather than one at a time.

### M2 — lexical normalization before symlink check

The old `canonicalize(path) != path` check fired on four purely-lexical differences that had nothing to do with symlinks:

- Relative paths (input `data/x` vs. canonical `/cwd/data/x`)
- Trailing slashes (`/tmp/dir/` vs `/tmp/dir`)
- `.` / `..` components (`/tmp/./foo` vs `/tmp/foo`)
- OS-managed root prefixes (macOS `/tmp` → `/private/tmp`)

New `lexically_normalize` helper walks the input components without following symlinks — prepends cwd if relative, drops `CurDir`, pops on `ParentDir`, preserves separators and root prefixes verbatim. The check is now `canonicalize == lexically_normalize`. If they agree, no symlink was traversed; if they differ, at least one component was a symlink (canonical form resolved through it, lexical form didn't). The four purely-lexical false positives no longer fire.

**macOS `/tmp` note**: still rejected on the strict reading of the FIP's "no symbolic or hard links" wording (the OS-managed `/tmp` IS a symlink). Operators should either pre-canonicalize paths in the config or spell out `/private/tmp/…` directly. The `SymlinkTraversal` error message already includes the canonical form, so the fix is copy-paste.

### Nine new tests (21 total in `configuration::file_io::`, 132 in node lib)

- **Seven `normalize_*`** covering absolute unchanged, relative + cwd prepend, `CurDir` drop, `ParentDir` pop, trailing slash, root `..` no-op, multiple leading slashes.
- **`validate_accepts_relative_path_without_symlink`** — direct regression guard for the M2 relative-path false positive. Skipped if cwd is itself under a symlink so we don't fight the test environment.
- **`validate_accepts_path_with_current_and_parent_dir_components`** — the `.` / `..` case.

### Remaining review items (deferred)

- **L1** (TOCTOU between validation and FS-agent provisioning) — the FS-agent PR should re-run `validate()` immediately before opening fds.
- **L2** (paths visible in ledger via error messages) — design note for the FS-agent PR.
- **L3** (no cap on list size) — nice-to-have, not blocking.
- **L4** (non-UTF-8 CLI paths pass through) — nice-to-have; ultimate check lands in the Rholang layer.

dylon (changes requested):
claude found 1 blocker and made a few recommendations:

  ## Review — PR #131 (static config + CLI flags)

  **Verdict: blocked.** The batched multi-error reporting and the flag-vs-config default
  asymmetry both match the FIP exactly, and validation *is* wired into the boot path via
  `builder::validate_config` (the PR body says otherwise — it undersells itself). But the
  FIP's central safety precondition isn't implemented.

  ### Blocking

  1. **FIP line 80 — "no symbolic or hard links … the node should fail to launch" — is not
     implemented.** `validate_path` checks only that *the configured path itself* doesn't
     traverse a symlink. It never walks the tree and never inspects `nlink`. In-tree symlinks
     pointing outward are caught later by `canonicalize_and_quarantine`, so those are
     defence-in-depth — but **hard links have no second line of defence**. `ln /etc/shadow
     /srv/data/notes.txt` canonicalizes to an in-root path, passes `check_prefix`, and grants
     read (and under `"rw"`, write) to the aliased inode. Please walk each configured dir at
     load time rejecting `is_symlink()` and (Unix) `nlink() > 1` on regular files, reported
     through the existing `Vec<FileIoConfigError>`. If an O(tree) boot walk is unacceptable
     for large trees, that's a FIP amendment — not a silent omission.

  ### Must fix

  2. **Relative paths are resolved against the process cwd.** `lexically_normalize` falls back
     to `std::env::current_dir()`, and `validate_accepts_relative_path_without_symlink`
     enshrines it. So `--oracle-static-dir data` grants a *different tree* depending on where
     `rnode` was launched from. For a capability grant that's a real hazard; the FIP's examples
     are all absolute. Please reject relative paths outright.
  3. **`"wx"` / `"w+x"` config entries validate but can never open.** `validate_path` *requires*
     the path to exist; `create_new(true)` *requires* it not to. The two checks are mutually
     contradictory, and `validate_accepts_all_nine_fopen_modes` asserts the contradiction by
     looping every mode against an existing temp file. Every such entry is guaranteed to fail
     at provisioning after passing config validation.
  4. **The schema drops the relative-name namespace the FIP's `openFile` depends on.** FIP config
     keys are *logical names* (`"relative/path"`) that Rholang later passes to
     `fs!openFile("relative/path")` (lines 67–78, 263–280); the flag form carries the same name
     plus an optional mode. This PR stores four flat `{path, mode}` lists with no name, and the
     flag is a bare `PathBuf` — so `--oracle-static-file` cannot express `mode: "r+"` even though
     the FIP's own example does. The `Fs` PR will hit this immediately. Either restore the map
     form or raise a FIP amendment.

  ### Should fix

  5. **A symlink anywhere in the absolute prefix makes the node refuse to boot.** Comparing
     `canonicalize` against `lexically_normalize` rejects macOS `/tmp` → `/private/tmp`,
     container bind mounts, and `/home` → `/var/home` on several distros. The FIP mandates
     fail-to-launch for symlinks *within the provisioned tree*; this is broader. The comment in
     `validate_accepts_real_file_with_valid_mode` ("Wait — the two DO differ in that case") reads
     as unresolved. At minimum, make the error tell the operator to supply the canonical path.
  6. No duplicate/overlap detection: the same path in both the config block and a flag yields two
     entries with different modes, i.e. two capabilities on one tree.
  7. `"wbx"` propagates into the error text and tests — see #120.
  8. PR body says `validate` is not wired to the boot flow; it is.

metaweta:

PR #132, opened 2026-07-17
feat(fileio): NormalizerEnv injection bundle for the FS-agent bootstrap

metaweta:
## Summary

Adds `crate::interpreter::io::injections::fileio_native_urns()`, the `HashMap<String, Par>` that the FS-agent's genesis deploy will pass to `Compiler::source_to_adt_with_normalizer_env` to obtain the fixed-channel `Par`s for the 19 `rho:io:fs:native:1.0.0/*` URNs.

Why a bundle instead of extending `urn_map`: the fileio native URNs are deliberately hidden from the shared `urn_map` (see `rho_runtime::is_internal_urn`) so user code cannot bind them via `new x(URN)`. `NormalizerEnv` is the supported way to bind hidden URNs on a per-deploy basis — the env is copied into each `new` node's `injections` `BTreeMap` at normalization time, and `eval_new` falls back to injections when a URN misses `urn_map`.

Values are bare `GPrivate` unforgeables (what `FixedChannels::native_*()` returns) — **not** `Bundle`-wrapped. `eval_new`'s injection path (`reduce.rs:1297-1325`) only accepts `GUnforgeable` or `Expression` Pars; bare is the right authority level for the FS-agent, which owns the channel and never leaks the name.

## Changes

- `rholang/src/rust/interpreter/io/injections.rs` (new) — `fileio_native_urns()` returning 19 URN → `Par` bindings + 4 unit tests.
- `rholang/src/rust/interpreter/io/mod.rs` — `pub mod injections;`.
- `rholang/src/rust/interpreter/reduce.rs` — replaces the stale "Injections (from normalizer) are not used currently" TODO with an accurate comment explaining the injection path and the fileio use case.

## Tests

Four new tests in `io::injections`:

- **`bundle_covers_every_fileio_native_urn`** — asserts all 19 expected URNs are present and no extras. Drift-detection: if someone renames or drops a URN in `std_system_processes()` without updating the bundle, this fires before the FS-agent hits `"No value set for {urn}"` at runtime.
- **`every_value_is_a_bare_gprivate_unforgeable`** — each `Par` must have zero exprs, exactly one unforgeable, and that unforgeable must be `GPrivate`. Catches accidental `Bundle` wrapping or `Expression` shapes that `eval_new`'s injection path rejects.
- **`injections_land_in_the_new_node_after_compilation`** — end-to-end plumbing: compiles a fragment binding one fileio URN with the bundle in the env, asserts the resulting `New`'s `injections` `BTreeMap` carries the URN mapped to the same fixed-channel `Par` we injected. This is the exact hop the FS-agent's genesis deploy will rely on at boot.
- **`empty_env_omits_the_url_from_injections`** — negative twin: compiling with an empty env leaves the URN in the `New`'s `uri` list but nothing in `injections`, which at runtime produces `"No value set for {urn}"`. Documented here so a future caller who forgets to pass the bundle understands the failure mode.

## Test plan

- [x] `cargo test -p rholang --lib io::` — 34 pass (30 prior + 4 new)
- [x] `empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes (no genesis-contract changes)

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/132"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786910903&installation_id=145946300&pr_number=132&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F132&signature=fe91407815383b7cae3c0cd732723f3c51aa54dd51c22e0aefdc385c25fcbf9f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/rholang/src/rust/interpreter/io/injections.rs
+++ b/rholang/src/rust/interpreter/io/injections.rs
@@ -1,0 +1,277 @@
+//! `NormalizerEnv` bundles for the File I/O native primitives.
+//!
+//! The `rho:io:fs:native:1.0.0/*` URNs are filtered out of the
+//! runtime's shared `urn_map` (see
+//! `rho_runtime::is_internal_urn`), so user code can't bind them
+//! via `new x(URN)`. The genesis-time FS-agent deploy still
+//! needs a way to obtain the fixed-channel `Par` for each
+//! primitive so its Rholang can dispatch to the handlers. The
+//! per-deploy `NormalizerEnv` (a `HashMap<String, Par>` accepted
+//! by `Compiler::source_to_adt_with_normalizer_env` /
+//! `RhoRuntime::evaluate_with_env_and_phlo`) is that channel.
+//!
+//! During compilation the env is copied into every `new` node's
+//! `injections` `BTreeMap`. At `eval_new` time (`reduce.rs`),
+//! URNs that miss `urn_map` fall through to `new.injections` --
+//! meaning the fileio native URNs, which are absent from the
+//! former, are resolved from the latter. See
+//! `implementation-plan.md` §"Bootstrap path for the FS agent"
+//! for the full design.
+//!
+//! Injection values are the bare `GPrivate` `Par`s
+//! `FixedChannels::native_*()` returns. They are NOT
+//! bundle-wrapped: `eval_new`'s injection path
+//! (`reduce.rs:1297-1325`) accepts only `GUnforgeable` or
+//! `Expression` Pars and errors on `Bundle`. Bare unforgeable is
+//! the right authority level for the FS-agent, which owns the
+//! channel and never leaks the name; user code has no path to
+//! obtain the private name because the URN isn't in `urn_map`.
+
+use std::collections::HashMap;
+
+use models::rhoapi::Par;
+
+use crate::rust::interpreter::system_processes::FixedChannels;
+
+/// Return the `NormalizerEnv` bundle for the File I/O native
+/// primitives: the seventeen `rho:io:fs:native:1.0.0/*` URNs
+/// registered in `std_system_processes()`, each mapped to its
+/// fixed-channel `Par` (a `GUnforgeable::GPrivate` under the
+/// hood).
+///
+/// The FS-agent's genesis deploy passes this map (merged with
+/// whatever other genesis bindings it needs) to
+/// `Compiler::source_to_adt_with_normalizer_env` so its
+/// `new nOpen(`rho:io:fs:native:1.0.0/open`), ...` binding
+/// resolves at compile-into-`injections` time and at runtime via
+/// `eval_new`'s injection fallback.
+///
+/// Consumers must NOT publish this map on any user-reachable
+/// surface. Handing the map to user Rholang would immediately
+/// grant that Rholang unauthenticated arbitrary-host-FS
+/// authority.
+pub fn fileio_native_urns() -> HashMap<String, Par> {
+    let mut m = HashMap::with_capacity(17);
+    m.insert(
+        "rho:io:fs:native:1.0.0/open".to_string(),
+        FixedChannels::native_open(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/close".to_string(),
+        FixedChannels::native_close(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/read".to_string(),
+        FixedChannels::native_read(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/write".to_string(),
+        FixedChannels::native_write(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/seek".to_string(),
+        FixedChannels::native_seek(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/tell".to_string(),
+        FixedChannels::native_tell(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/size".to_string(),
+        FixedChannels::native_size(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/truncate".to_string(),
+        FixedChannels::native_truncate(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/flush".to_string(),
+        FixedChannels::native_flush(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/stat".to_string(),
+        FixedChannels::native_stat(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/entries".to_string(),
+        FixedChannels::native_entries(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/exists".to_string(),
+        FixedChannels::native_exists(),
+    );
+    m.insert(
+        "rho:io:fs:native:1.0.0/rename".to_string(),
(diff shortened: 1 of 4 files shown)


metaweta:
## Update — two Lows from the security review

Follow-up commit [`1c83f6e4`](https://github.com/F1R3FLY-io/f1r3node-rust/commit/1c83f6e4) addresses both actionable Low findings from the security review.

### L-A — narrow the raw bundle to `pub(crate)` with a controlled wrapper

`fileio_native_urns()` was `pub`, so every crate depending on `rholang` could reach the raw `HashMap<String, Par>` that grants unauthenticated arbitrary-host-FS authority. The FS-agent's genesis deploy (Stage D, not yet written) is the only legitimate consumer, and it doesn't need the raw map — it just needs its source string compiled with the map spliced into the normalizer env.

- `fileio_native_urns()` → `pub(crate)`. Accidental cross-crate reach is now a compile error rather than a code-review rule.
- New `pub fn compile_fileio_genesis_source(source, extra_env)` in the same module. Merges the fileio bundle with the caller's `extra_env`, calls `Compiler::source_to_adt_with_normalizer_env`, returns only the compiled `Par`. The raw bundle never crosses the crate boundary.
- Collisions between `extra_env` and the fileio URN namespace are rejected — the fileio URNs are a fixed vocabulary, so a caller attempting to shadow one is a programming error, not a legitimate override.

### L-B — runtime negative tests

The unit tests in `interpreter::io::injections` prove the *compile-time* shape (bundle → New's `injections`). The two new integration tests in `rholang/tests/injection_runtime_spec.rs` close the loop at the *runtime* end via `RhoRuntimeImpl::evaluate_with_env`:

- **`user_deploy_cannot_bind_hidden_fileio_urn_with_empty_env`** — compiles `new x(`rho:io:fs:native:1.0.0/open`) in { Nil }` with an empty env, evaluates it, and asserts eval fails with "No value set for rho:io:fs:native:1.0.0/open". This is the end-to-end proof of the ocap story: user code cannot obtain the fixed-channel Par for a fileio primitive, because the URN is filtered from `urn_map` by `is_internal_urn` *and* nothing populates the injection fallback for it.
- **`bundle_wrapped_injection_is_rejected_at_runtime`** — supplies a `Bundle`-wrapped Par via the injection env and asserts eval fails with "invalid injection". Explains why `fileio_native_urns()` returns bare `GPrivate` Pars rather than the more common `Bundle`-wrapped form used elsewhere in `urn_map`. Previously only implied by reading `reduce.rs:1297-1325`.

### Wrapper unit tests

Three additional unit tests in `io::injections::tests` (10 total in the module):

- `wrapper_compiles_fileio_source_and_injects_bundle` — happy path with empty `extra_env`.
- `wrapper_rejects_extra_env_shadowing_a_fileio_urn` — proves the collision guard fires.
- `wrapper_preserves_non_colliding_extra_env` — proves the merge preserves non-colliding caller-supplied bindings.

### Cosmetic Lows fixed in the original commit

- `HashMap::with_capacity(17)` → `19` (matches the 19 URNs).
- Docstring "seventeen" → "nineteen".

### Deferred to the Stage-D FS-agent PR (per reviewer)

- Confirm the genesis call site consumes `compile_fileio_genesis_source` through a system-deploy path (`evaluate_system_source` or a genesis-only helper), not `RhoRuntimeOps::process_deploy` where a subsequent user deploy could hit the same env.
- Add end-to-end test: user deploy with `new x(`rho:io:fs:native:1.0.0/open`) in { x!("/etc/passwd", "r", ack) }` against a full `RhoRuntimeImpl`, assert non-empty `errors` and no `tokio::fs::open` side effect. (The runtime negative test in this PR proves the plumbing; the Stage-D version proves it in-context with the FS-agent's actual boot sequence.)

dylon:
This is still a draft PR, but claude had the following to say about it:

  ## Review — PR #132 (NormalizerEnv injection bundle)

  **Verdict: blocked** (and still marked draft while seven PRs sit on top of it).

  This is the ocap linchpin of the stack and the core mechanism is right. Hiding the native
  URNs from `urn_map` via `is_internal_urn` is the correct approach; the prefix filter is the
  right shape; `setup_maps_and_refs_hides_native_fileio_urns_from_urn_map` is a good invariant
  test; and `user_deploy_cannot_bind_hidden_fileio_urn_with_empty_env` proves the property
  end-to-end rather than by inspection. I traced `eval_new` (`reduce.rs:1293-1350`) and confirm
  the `urn_map` → `new.injections` fallback is the only resolution path, and that deploys carry
  source normalized node-side via `normalizer_env_from_deploy`, so injections cannot arrive
  from the wire. Two things undercut it.

  ### Blocking

  1. **`compile_fileio_genesis_source` is `pub` and takes an arbitrary `&str`.** The module
     docstring justifies `pub(crate)` on `fileio_native_urns()` so that "a downstream crate
     accidentally reaching for [it] will then be a compile error, not a code-review invariant"
     — but the `pub` wrapper compiles *whatever source it is given* with the full native-FS
     bundle injected, handing the same authority straight back. The `pub(crate)` buys nothing.
     Please have the wrapper close over the embedded agent sources (`include_str!` of
     `file.rho` / `dir.rho` / the future `fs.rho`) — e.g. `compile_fs_genesis(extra_env)` —
     so no call site can supply source.
  2. **Injected channels are bare `GPrivate` (read + write); every other system URN is
     `bundle+` write-only.** `Definition::to_urn_map` (`system_processes.rs:447-455`) wraps
     every public channel in `Bundle { write_flag: true, read_flag: false }`.
     `fileio_native_urns()` returns the bare name. Since `introduce_system_process` installs
     the handler as a *persistent* continuation via `space.install(..)`, a holder of the bare
     name can install a competing `for (x <- nOpen)` consume, and RSpace will match sends
     nondeterministically against either the system handler or the intruder — an
     interception/denial primitive on the FS syscall channel.

     Not exploitable today (the genesis FS agent is the only holder, one trust domain), but it
     is strictly weaker than every other system URN, and any future sub-agent delegation
     inherits read authority it should never have. The PR justifies bare-`GPrivate` purely
     because `eval_new`'s injection path rejects `Bundle` — and that's accurate
     (`reduce.rs:1296-1325` accepts only `RhoUnforgeable` or `RhoExpression`). But that means a
     *platform limitation is silently downgrading the authority model*. The fix belongs in
     `eval_new`: accept a `Bundle` Par as a third case, then inject `bundle+`-wrapped channels
     exactly as `to_urn_map` does. `bundle_wrapped_injection_is_rejected_at_runtime` then flips
     from documenting the limitation to guarding the fix.

  ### Process

  3. Please mark ready for review — #137/#139/#140/#141 are all blocked behind a draft.

metaweta:

PR #114, opened 2026-07-13
Fix/merge recovery finalization

machieke:
## Summary

This PR brings the sealed-floor merge recovery stack onto `dev` and fixes the final deploy-inclusion/finality failure modes found during local shard testing.

The branch now handles three related problems:

- merge-rejected deploys must be observable, recoverable, and unable to wedge finalization;
- non-leader validators should support finality instead of competing to package the same user deploys;
- deploys accepted into non-final branches must not starve from canonical inclusion when empty/support branches finalize first.

## Key Changes

- Bases multi-parent merge and validation on the finalized floor.
- Makes merge rejections explicit and recoverable through local rejected-deploy buffering.
- Prevents rejected/recovered deploys from blocking finalization progress.
- Adds deploy-inclusion leadership so ordinary user-deploy packaging is coordinated.
- Replaces the binary non-leader gate with bounded/adaptive fallback admission.
- Tracks deploy-inclusion progress by deploy signatures, not only by latest deploy-carrying block, so recycled in-scope work does not look like fresh inclusion progress.
- Adds finality-lag backpressure and adaptive fresh fallback caps.
- Makes non-final in-scope deploys recoverable until they are canonical/finalized.
- Fixes expiry accounting so `pool>0 valid=0 blockExpired=0 selected=0` cannot persist for expired in-scope deploys.
- Biases canonical support toward non-final deploy-carrying branches, preferring the branch with the most unfinalized deploy signatures, so empty sibling branches do not consistently finalize over deploy-heavy work.
- Adds metrics and diagnostics for admission, fallback, backpressure, and recovery behavior.
- Fixes stale invalid-justification validation.
- Adds regression coverage for deploy leadership, fallback admission, recycled signatures, in-scope recovery, expiry cleanup, and deploy-support parent selection.

## Validation

Ran local Rust validation:

- `cargo fmt --package casper --package node`
- `git diff --check`
- `cargo test -p casper rust::engine::multi_parent_casper::snapshot::tests -- --nocapture`
- `cargo test -p casper rust::blocks::proposer::block_creator::tests -- --nocapture`
- `cargo test -p casper --test mod blocks::block_creator`
- `cargo test -p casper --lib`

PR #133, opened 2026-07-18
fix(casper,node): complete PR #114 port — finalization scope, byte-bounded admission, DAG-width backpressure

jeffrey-l-turner:
## Summary

Ports the three commits from `fix/merge-recovery-finalization` (PR #114) that the
integrated line does not already carry, per the commit-level absorption analysis in
`docs/discoveries/2026-07-18-merge-recovery-finalization-gaps.md` (included in this PR)
and the [context comment on PR #114](https://github.com/F1R3FLY-io/f1r3node-rust/pull/114#issuecomment-5011006362).

The other five head commits of #114 were verified as already absorbed (equivalent or
superseded implementations with equal-or-better test coverage) and are **not** re-ported.
This PR supersedes #114's unique content; #114's final disposition (close with reference
vs. keep as history) remains the maintainer's call.

Base: the integrated sealed-floor line (`sealed-floor-merge-squashed` after the dev
`32a45d02` and master `35ce65f1` merges).

## Port 1 — `e3e3ae90` (from #114's `9ec1096e`): merge-recovery finalization scope

**Closes a proof↔code divergence.** The FV campaign's artifacts already prove the fixed
behavior — `finalized_floor/theories/Selection.v` maps its ancestry relation to
`is_dag_ancestor`, and `DeployLifecycle.tla` models the buffer purge (`PurgeRejectedBuf=TRUE`)
— but the code still implemented the pre-fix predicates. This port aligns them:

- `block_in_floor_merge_scope`: `is_in_main_chain` → `is_dag_ancestor` in
  `compute_parents_post_state`'s visible-block filter, so a finalized below-floor
  co-parent branch outside the floor's main-parent spine is correctly dropped from merge
  scope (the `ComputedPreStateMismatch` / floor-divergence surface), + backstop test
- Recoverable-buffer gating: `rejected_buffer_has_recoverable_deploys` /
  `local_rejected_buffer_has_recoverable_deploys` replace bare `non_empty()` — buffers
  holding only canonical-won or expired sigs no longer trigger futile recovery blocks
- Expired rejected-buffer purge from deploy storage + buffer in
  `prepare_user_deploys_with_policy`
- `RecoveryDeferred` propose outcome: legitimate non-leader recovery deferral is
  non-bug (debug, not warn) across proposer / block_api / heartbeat

## Port 2 — `c965107b` (from #114's `6012c6a3`): encoded-bytes admission bound

Closes the size-dimension gap flagged in `docs/validation/fv-campaign-gap-analysis.md:47`
(the count-cap raise to 128 had been absorbed without the byte machinery that motivated it):

- Unified `select_deploys_for_block(deploys, cap, reserve_tail, byte_budget)` replaces
  count-only selection; `deploy_encoded_len` via prost `encoded_len`
- `USER_DEPLOY_BYTE_PROPOSAL_BUDGET` = 2 MiB (512 KiB under finality backpressure),
  threaded as a shared decrementing budget across retry → ordinary → in-scope-recovery
- Progress carve-out: one oversize deploy is still admitted when it alone exceeds the budget
- `PreparedUserDeploys` byte accounting + 5 admission byte metrics

## Port 3 — `43c4dc79` (from #114's `6981b37a`): DAG-width backpressure

The only mechanism bounding unfinalized-DAG **width** (existing heartbeat caps bound only
LFB **lag** — a wide fan at similar height passes them):

- `EmptyFrontierPressure` + `empty_frontier_pressure()`: a validator that recently
  self-proposed, with no deploy pressure, stops emitting empty frontier-follow /
  stale-recovery / convergence blocks once unfinalized-DAG width exceeds the cap;
  `!empty_frontier_backpressure` gates all six propose-decision paths, pending-deploy
  proposing deliberately ungated
- Config knob `empty-frontier-max-unfinalized-blocks = 64` (defaults.conf, casper_conf,
  CLI option, config-mapper + HOCON override)
- `prune_dag_covered_parents` in `compute_snapshot`: collapses the parent set to a single
  deploy-free parent that DAG-covers all others
- `compute_parents_post_state` fast path narrowed (`is_in_main_chain` → `is_dag_ancestor`
  + empty-deploys guard); regression-spec assertion deliberately reversed (a DAG-covering
  empty parent now legitimately skips the merge)

## Deliberate port deltas (inherited context NOT taken)

- **Chase-cap policy kept from this line**: `6981b37a` narrowed the deploy-recovery
  frontier-chase cap (`min` with 2); this line's soak-tested policy deliberately widens it.
  Only the width-backpressure mechanism was ported (noted in a code comment at the site).
- **No `lag_recovery_leader` gate on convergence recovery** — inherited context from
  #114's lineage, not part of `6981b37a`'s diff.
- Spec code updated to current APIs (`create_genesis_block` by shared ref, no useless
  `.into()`).

## Verification

- `cargo check` + `clippy -D warnings` clean for casper + node including test targets
- Heartbeat module 11/11 (both width-backpressure tests + `RecoveryDeferred`)
- Snapshot module 11/11 (incl. `parent_selection_prunes_dag_covered_parents`)
- Node configuration 15/15 (incl. HOCON/CLI override of the new knob)
- Regression spec 4/4 (incl. the full-runtime fast-path pin, ~7s)
- Port-1 backstops 9/9 (incl. `floor_scope_drops_finalized_off_main_dag_ancestors`)
- Both byte-budget tests + 50-test block_creator sweep + config-cap integration tests

## FV follow-ups (tracked, not in this PR)

The byte-bound and width-bound have no formal artifacts on either lineage; whether to
extend `DeployLifecycle.tla` (admission bounds) and add a width-pressure model, or accept
unit-test-only coverage with an explicit note in `fv-campaign-gap-analysis.md`, is an open
decision recorded in the discovery doc.

---
Corrections applied versus your draft: title reflects all three ports rather than only 9ec1096e; the stale "Excluded" paragraph is replaced by the port-delta section (those exclusions were temporary sequencing within port 1, later restored in port 3); the verification list covers all three commits' evidence; and the trailing empty bullet is gone. Ready to use for gh pr create against dev — want me to open it?

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/133"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786967055&installation_id=145946300&pr_number=133&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F133&signature=eae108d5931015e625d94bc4b32b57db6a89fc9297fc9577d7592b929a5a90da"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -97,6 +97,33 @@ Three crates have `build.rs` for protobuf code generation:
 - `rust-toolchain.toml` — nightly channel pin
 - `Cross.toml` — cross-compilation for amd64/arm64
 
+### Recommended Claude Code local settings
+Add to `.claude/settings.local.json` (personal, not committed) when working
+in this repo with Claude Code:
+
+```json
+{
+  "fileCheckpointingEnabled": false,
+  "env": {
+    "BASH_DEFAULT_TIMEOUT_MS": "1200000",
+    "BASH_MAX_TIMEOUT_MS": "1800000"
+  }
+}
+```
+
+- `fileCheckpointingEnabled: false` — Claude Code's checkpointing/rewind
+  feature runs `git stash` + `git reset --hard` against the workspace repo
+  around tool events, taking real `.git/index.lock` locks that collide with
+  concurrent git commands ("Unable to create index.lock"; see
+  anthropics/claude-code#68315). Disabling it trades away `/rewind` file
+  restore in this repo.
+- Bash timeouts raised to 20/30 min — the pre-push hook runs the full
+  release test suite for all 11 crates (~9 min, longer on cold caches),
+  which exceeds the default 10-minute window and gets a `git push` killed
+  mid-gate when run through Claude Code.
+
+Both settings take effect at the next session start.
+
 ## Network Ports
 | Port | Service |
 |------|---------|
--- a/casper/src/rust/api/block_api.rs
+++ b/casper/src/rust/api/block_api.rs
@@ -73,6 +73,9 @@ fn recoverable_propose_failure_message(status: &ProposeStatus) -> Option<String>
         ProposeStatus::Failure(ProposeFailure::NoNewDeploys) => {
             Some("No new deploys to propose.".to_string())
         }
+        ProposeStatus::Failure(ProposeFailure::RecoveryDeferred) => {
+            Some("Rejected deploy recovery deferred to selected leader.".to_string())
+        }
         ProposeStatus::Failure(ProposeFailure::CheckConstraintsFailure(
             CheckProposeConstraintsFailure::NotEnoughNewBlocks,
         )) => Some("No new blocks from peers yet; synchronize with network first.".to_string()),
--- a/casper/src/rust/blocks/proposer/block_creator.rs
+++ b/casper/src/rust/blocks/proposer/block_creator.rs
@@ -25,6 +25,7 @@ use models::rust::casper::protocol::casper_message::{
 };
 use models::rust::validator::Validator;
 use prost::bytes::Bytes;
+use prost::Message;
 use rholang::rust::interpreter::system_processes::BlockData;
 use rspace_plus_plus::rspace::errors::HistoryError;
 use tracing;
@@ -62,6 +63,9 @@ pub struct PreparedUserDeploys {
     pub selected_in_scope_recovery_count: usize,
     pub selected_in_scope_recovery_sigs: HashSet<Bytes>,
     pub already_in_scope_count: usize,
+    pub selected_user_deploy_bytes: usize,
+    pub deferred_user_deploy_bytes: usize,
+    pub byte_cap_hit: bool,
 }
 
 #[derive(Clone, Copy, Debug, PartialEq, Eq)]
@@ -120,6 +124,16 @@ struct FreshAdmissionFallback {
     backpressure: bool,
 }
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct DeploySelection {
+    deploys: HashSet<Signed<DeployData>>,
+    strategy: &'static str,
+    selected_bytes: usize,
+    deferred_bytes: usize,
+    count_capped: bool,
+    byte_capped: bool,
+}
+
 #[derive(Clone, Copy, Debug, Default, PartialEq, Eq)]
 struct FinalityLagStats {
     dag_tip: i64,
@@ -136,6 +150,8 @@ struct FinalityLagStats {
 /// already in place.
 const DEPLOY_SELECTION_RESERVE_TAIL_ENABLED: bool = true;
 const ORDINARY_DEPLOY_PROPOSAL_CAP: usize = 128;
+const USER_DEPLOY_BYTE_PROPOSAL_BUDGET: usize = 2 * 1024 * 1024;
+const USER_DEPLOY_BACKPRESSURE_BYTE_PROPOSAL_BUDGET: usize = 512 * 1024;
 const RETRY_DEPLOY_REPROPOSAL_CAP: usize = 32;
 const NON_LEADER_FALLBACK_ORDINARY_DEPLOY_CAP: usize = 8;
 const NON_LEADER_FALLBACK_MIN_ORDINARY_DEPLOY_CAP: usize = 4;
@@ -167,40 +183,80 @@ fn ordered_user_deploys(deploys: &HashSet<Signed<DeployData>>) -> Vec<Signed<Dep
     ordered
 }
 
+#[cfg(test)]
 fn select_recovered_deploys_for_block(
     deploys: &HashSet<Signed<DeployData>>,
     _block_number: i64,
     cap: usize,
-) -> HashSet<Signed<DeployData>> {
-    let ordered = ordered_user_deploys(deploys);
-    if ordered.is_empty() || cap == 0 {
-        return HashSet::new();
-    }
-    ordered.into_iter().take(cap).collect()
+) -> DeploySelection {
+    select_deploys_for_block(deploys, cap, false, USER_DEPLOY_BYTE_PROPOSAL_BUDGET)
 }
 
-fn select_ordinary_deploys_for_block(
+fn deploy_encoded_len(deploy: &Signed<DeployData>) -> usize {
+    DeployData::to_proto(deploy.clone()).encoded_len()
+}
+
+fn select_deploys_for_block(
     deploys: &HashSet<Signed<DeployData>>,
     cap: usize,
     reserve_tail: bool,
-) -> (HashSet<Signed<DeployData>>, &'static str) {
+    byte_budget: usize,
+) -> DeploySelection {
+    let total_bytes: usize = deploys.iter().map(deploy_encoded_len).sum();
     let ordered = ordered_user_deploys(deploys);
-    if ordered.is_empty() || cap == 0 {
-        return (HashSet::new(), "none");
+    if ordered.is_empty() || cap == 0 || byte_budget == 0 {
+        return DeploySelection {
+            deploys: HashSet::new(),
+            strategy: "none",
+            selected_bytes: 0,
+            deferred_bytes: total_bytes,
+            count_capped: !ordered.is_empty(),
+            byte_capped: byte_budget == 0 && !ordered.is_empty(),
+        };
     }
-    if ordered.len() <= cap {
-        return (ordered.into_iter().collect(), "uncapped");
-    }
-    if reserve_tail && DEPLOY_SELECTION_RESERVE_TAIL_ENABLED && cap > 1 {
+    let count_capped = ordered.len() > cap;
+    let (candidates, strategy): (Vec<Signed<DeployData>>, &'static str) = if ordered.len() <= cap {
+        (ordered, "uncapped")
+    } else if reserve_tail && DEPLOY_SELECTION_RESERVE_TAIL_ENABLED && cap > 1 {
         let oldest_take = cap.saturating_sub(1);
-        let mut picked: HashSet<Signed<DeployData>> =
+        let mut candidates: Vec<Signed<DeployData>> =
             ordered.iter().take(oldest_take).cloned().collect();
-        if let Some(newest) = ordered.iter().last().cloned() {
-            picked.insert(newest);
-        }
-        (picked, "oldest-plus-newest")
+        candidates.extend(ordered.iter().last().cloned());
+        (candidates, "oldest-plus-newest")
     } else {
         (ordered.into_iter().take(cap).collect(), "oldest-only")
+    };
+    let mut selected = HashSet::new();
+    let mut selected_bytes = 0usize;
+    let mut byte_capped = false;
+    for deploy in candidates {
+        let deploy_bytes = deploy_encoded_len(&deploy);
+        let next_bytes = selected_bytes.saturating_add(deploy_bytes);
+        if next_bytes <= byte_budget || selected.is_empty() {
+            selected_bytes = next_bytes;
+            selected.insert(deploy);
+        } else {
+            byte_capped = true;
+        }
+    }
+    if selected_bytes > byte_budget {
+        byte_capped = true;
+    }
+    DeploySelection {
+        deploys: selected,
+        strategy,
+        selected_bytes,
+        deferred_bytes: total_bytes.saturating_sub(selected_bytes),
+        count_capped,
+        byte_capped,
+    }
+}
+
+fn user_deploy_byte_budget(admission_policy: DeployAdmissionPolicy) -> usize {
+    if admission_policy.backpressure {
+        USER_DEPLOY_BACKPRESSURE_BYTE_PROPOSAL_BUDGET
+    } else {
+        USER_DEPLOY_BYTE_PROPOSAL_BUDGET
     }
 }
 
@@ -627,46 +683,68 @@ async fn prepare_user_deploys_with_policy(
         HashSet::new()
     };
     let retry_candidate_count = retry_candidates.len();
-    let selected_retries = select_recovered_deploys_for_block(
+    let total_byte_budget = user_deploy_byte_budget(admission_policy);
+    let mut remaining_byte_budget = total_byte_budget;
+    let retry_selection = select_deploys_for_block(
         &retry_candidates,
-        block_number,
         RETRY_DEPLOY_REPROPOSAL_CAP,
+        false,
+        remaining_byte_budget,
     );
-    let retry_capped = retry_candidate_count > selected_retries.len();
+    remaining_byte_budget = remaining_byte_budget.saturating_sub(retry_selection.selected_bytes);
+    let retry_capped = retry_selection.count_capped || retry_selection.byte_capped;
     if retry_capped {
-        let deferred_retries = retry_candidate_count.saturating_sub(selected_retries.len());
+        let deferred_retries = retry_candidate_count.saturating_sub(retry_selection.deploys.len());
         tracing::info!(
             target: "f1r3fly.casper.recovery",
-            "Retry deploy selection capped for block #{}: selected={}, deferred={}, cap={}",
+            "Retry deploy selection capped for block #{}: selected={}, deferred={}, cap={}, selected_bytes={}, deferred_bytes={}, byte_budget={}",
             block_number,
-            selected_retries.len(),
+            retry_selection.deploys.len(),
             deferred_retries,
-            RETRY_DEPLOY_REPROPOSAL_CAP
+            RETRY_DEPLOY_REPROPOSAL_CAP,
+            retry_selection.selected_bytes,
+            retry_selection.deferred_bytes,
+            total_byte_budget
         );
     }
-    let (selected_ordinary, ordinary_selection_strategy) = select_ordinary_deploys_for_block(
+    let ordinary_selection = select_deploys_for_block(
         &ordinary_candidates,
         ordinary_cap,
         admission_policy.reserve_tail,
+        remaining_byte_budget,
     );
-    let ordinary_capped = ordinary_candidates.len() > selected_ordinary.len();
-    let (selected_in_scope_recovery, in_scope_recovery_selection_strategy) =
-        select_ordinary_deploys_for_block(
-            &in_scope_recovery_candidates,
-            in_scope_recovery_cap,
-            false,
-        );
+    remaining_byte_budget = remaining_byte_budget.saturating_sub(ordinary_selection.selected_bytes);
+    let ordinary_capped = ordinary_selection.count_capped || ordinary_selection.byte_capped;
+    let in_scope_recovery_selection = select_deploys_for_block(
+        &in_scope_recovery_candidates,
+        in_scope_recovery_cap,
+        false,
+        remaining_byte_budget,
+    );
     let in_scope_recovery_capped =
-        in_scope_recovery_candidates.len() > selected_in_scope_recovery.len();
+        in_scope_recovery_selection.count_capped || in_scope_recovery_selection.byte_capped;
+    let selected_in_scope_recovery = in_scope_recovery_selection.deploys;
     let selected_in_scope_recovery_sigs: HashSet<Bytes> = selected_in_scope_recovery
         .iter()
         .map(|deploy| deploy.sig.clone())
         .collect();
-    let selected: HashSet<Signed<DeployData>> = selected_retries
+    let selected: HashSet<Signed<DeployData>> = retry_selection
+        .deploys
         .into_iter()
-        .chain(selected_ordinary.into_iter())
+        .chain(ordinary_selection.deploys.into_iter())
         .chain(selected_in_scope_recovery.into_iter())
         .collect();
+    let selected_user_deploy_bytes = retry_selection
+        .selected_bytes
+        .saturating_add(ordinary_selection.selected_bytes)
+        .saturating_add(in_scope_recovery_selection.selected_bytes);
+    let deferred_user_deploy_bytes = retry_selection
+        .deferred_bytes
+        .saturating_add(ordinary_selection.deferred_bytes)
+        .saturating_add(in_scope_recovery_selection.deferred_bytes);
+    let byte_cap_hit = retry_selection.byte_capped
+        || ordinary_selection.byte_capped
+        || in_scope_recovery_selection.byte_capped;
     let selected_retry_count = selected
         .iter()
         .filter(|deploy| is_retry_candidate(deploy))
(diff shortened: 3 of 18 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :mag: Needs Review (66.6700% agreement)

**Reviewed by:**

| anthropic | `claude-fable-5` | Feedback Provided | 0.65 |
| openai | `gpt-5.6-sol` | Abstain | 0 |

---

### Summary

This is a carefully documented, well-tested port of three mechanisms (finalization merge scope, byte-bounded deploy admission, DAG-width backpressure) with strong test coverage including backstop and regression tests, and it aligns code with already-proven FV artifacts. No security issues were found. The main concerns are logic-level: the byte-budget selection loop's skip-and-continue behavior can indefinitely starve large deploys (the carve-out only fires on an empty selection), and the recoverability predicate is duplicated with divergent filters between snapshot.rs and block_creator.rs, risking inconsistent recovery gating. There are also performance regressions worth watching — the recovery gate went from O(1) non_empty() to block-store ancestry scans on every propose/snapshot, and deploy sizes are re-encoded with clones multiple times per selection. The floor-scope predicate change (>= to >, excluding the floor itself) goes beyond the advertised main-chain→DAG-ancestor swap and deserves an explicit comment even though tests assert it. All issues are addressable; the change is functional and safe to merge after considering the starvation and predicate-consistency feedback.

This port closes real proof↔code and admission-dimension gaps, and the added RecoveryDeferred path, recoverable-buffer gating, byte-budget selection, and floor-scope DAG-ancestor tests are directionally right. Before merge, a human should re-check three consensus-sensitive behaviors: deploy-bearing DAG-cover fast-path short-circuiting, parent-set compaction via prune_dag_covered_parents, and the duplicated recoverable-buffer predicates that can disagree between snapshot recovery_context and block-creator admission. The encoded-length helper also needs a cheaper implementation for the propose hot path. No direct security vulnerability stands out in the visible diff, but the behavioral risk on merge/parent selection is high enough that this should not be treated as a pure approve.

The PR introduces significant changes related to byte budget enforcement and DAG pressure management. While the changes seem logically sound, they require thorough testing to ensure correctness. Additionally, some minor style and documentation improvements can be made to enhance code readability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Recoverability predicate diverges between snapshot and block-creator implementations</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 254)

local_rejected_buffer_has_recoverable_deploys (snapshot.rs) filters candidates with `valid_after_block_number < current_block_number && valid_after_block_number > earliest_block_number && !is_expired_at(...)`, while rejected_buffer_has_recoverable_deploys (block_creator.rs) additionally honors `rejected_in_scope` (allowing below-earliest deploys that are rejected in scope) and excludes clean in-scope deploys. The two gates can disagree: compute_snapshot may report recovery_backlog=false while block creation would find recoverable candidates (or vice versa), producing inconsistent recovery-context decisions across the propose pipeline.

---

local_rejected_buffer_has_recoverable_deploys filters with valid_after_block_number < current && > earliest && !expired. rejected_buffer_has_recoverable_deploys in block_creator.rs uses not_future_deploy / not_expired_deploy plus in-scope clean/rejected handling. These can disagree on boundary VABs and in-scope clean buffer entries, so snapshot recovery_context and block-creator admission policy may diverge for the same buffer contents.

**Suggestion:**
```
Extract a single shared predicate/function (e.g., in interpreter_util) parameterized by the available scope information, and have both call sites use it, or document explicitly why the snapshot-level check is intentionally narrower.
```

</details>

<details>
<summary><b>:orange_circle: Major: Byte budget logic</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 167)

The logic for handling the byte budget seems complex and could be prone to errors. Ensure that the byte budget is correctly calculated and enforced.

**Suggestion:**
```
Consider adding more unit tests to cover edge cases for byte budget enforcement.
```

</details>

<details>
<summary><b>:orange_circle: Major: deploy_encoded_len clones full deploy on every size check</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 196)

deploy_encoded_len calls DeployData::to_proto(deploy.clone()).encoded_len() for every candidate, and select_deploys_for_block also sums encoded length over the full input set before selection. Under backlog pressure this repeatedly clones and re-encodes the same deploys (retry + ordinary + recovery passes), which can dominate propose latency and allocate large temporary buffers. The progress carve-out for one oversize deploy is correct, but the cost model is expensive for the hot path this change is meant to protect.

**Suggestion:**
```
Cache encoded length per deploy once (e.g. map sig -> len), or compute length from references without cloning Signed<DeployData>. At minimum memoize within a single prepare_user_deploys_with_policy call so the three selection passes and deferred-byte accounting do not re-encode.
```

</details>

<details>
<summary><b>:orange_circle: Major: Byte-budget skip-and-continue can permanently starve large deploys</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 232)

In select_deploys_for_block, when a deploy exceeds the remaining byte budget it is skipped (byte_capped = true) but the loop continues, so newer, smaller deploys after it in the ordered list can still be admitted. Since selection is oldest-first, an old large deploy will be repeatedly passed over on every proposal as long as smaller deploys keep arriving and filling the budget — the one-oversize carve-out only applies when `selected.is_empty()`, which will rarely be the case for a busy validator. This breaks the oldest-first fairness invariant and can indefinitely defer a large deploy, contradicting the progress guarantee claimed in the PR description.

**Suggestion:**
```
Either break out of the loop on the first deploy that doesn't fit (prefix-preserving selection), or track skipped deploys and prioritize them in subsequent rounds (e.g., admit the oldest skipped oversize deploy alone in a dedicated proposal when it has been deferred N times).
```

</details>

<details>
<summary><b>:orange_circle: Major: rejected_buffer_has_recoverable_deploys replaces O(1) non_empty() with a block-store scan on every create()</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2548)

The previous gate was `buffer.non_empty()`. The new gate reads the entire rejected buffer, and if any candidates survive filtering, calls interpreter_util::canonical_won_sigs which walks the block store from the parents down to scan_floor. The same pattern now also runs in compute_snapshot (local_rejected_buffer_has_recoverable_deploys), so a persistently non-empty buffer causes two ancestry scans per propose attempt. Additionally, scan_floor is computed as `min(min valid_after, earliest_block_number)`, which always deepens the scan to at least the full deploy lifespan rather than shortening it when all candidates are recent.

**Suggestion:**
```
Cache canonical_won results keyed on (parent set, scan_floor) as done for other per-snapshot scans, and reconsider `.map(|h| h.min(earliest_block_number))` — if the intent is to never scan shallower than the oldest candidate, `max` semantics or per-candidate floors would avoid unconditionally scanning the full lifespan.
```

</details>

<details>
<summary><b>:orange_circle: Major: prune_dag_covered_parents may collapse useful multi-parent tips</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 191)

Any deploy-free parent that DAG-covers all others replaces the entire parent set. That reduces width, but also drops co-parents that may still be required for justification/liveness semantics or for carrying non-deploy state. Combined with the broader compute_parents_post_state fast path, this makes parent compaction a consensus-behavior change, not just a backpressure heuristic.

**Suggestion:**
```
Confirm against the finalization/justification model that collapsing to a covering empty parent never omits required justifications. Consider pruning only under EmptyFrontierPressure, or only when all covered parents are also deploy-free and already finalized/represented.
```

</details>

<details>
<summary><b>:orange_circle: Major: Fast-path now skips merge for deploy-bearing covering parents</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 881)

compute_parents_post_state switched from is_in_main_chain to is_dag_ancestor and intentionally dropped the empty-deploys guard. If a covering parent did not actually incorporate a secondary parent's effects (bug, partial seal, or unexpected DAG shape), the fast path returns the cover post-state without merging and can silently drop deploy effects. The new regression covers the happy path where cover already contains side effects, but not the mismatch case.

**Suggestion:**
```
Keep the DAG-ancestor fast path only when cover post-state is known to include all covered parents' effects (e.g. empty secondary parents, or an explicit merged-parents attestation). Add a negative regression where cover is an ancestor but state diverges, expecting merge rather than blind short-circuit.
```

</details>

<details>
<summary><b>:orange_circle: Major: count_capped mislabeled when byte_budget == 0</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 205)

In the early-return branch of select_deploys_for_block, `count_capped: !ordered.is_empty()` is set even when the return was triggered solely by `byte_budget == 0` with a non-zero cap. Callers OR count_capped and byte_capped together for cap_hit, so behavior is unaffected, but metrics/log lines that distinguish the two cap dimensions will report a spurious count cap.

---

When byte_budget is 0 and there are candidates, the early return sets count_capped = !ordered.is_empty() and byte_capped = true. Callers treat retry/ordinary/recovery as capped if either flag is set, so metrics/logs can say count-capped when only the byte budget was exhausted by a prior pass.

**Suggestion:**
```
Set `count_capped: cap == 0 && !ordered.is_empty()` (or `ordered.len() > cap`) in the early-return branch.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Struct naming</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 120)

The struct `FreshAdmissionFallback` could be renamed to something more descriptive, like `DeployAdmissionStrategy`.

**Suggestion:**
```
Rename `FreshAdmissionFallback` to `DeployAdmissionStrategy`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Missing documentation</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 167)

Add documentation to the constants and functions related to byte budget enforcement to improve code readability.

**Suggestion:**
```
Add inline comments or docstrings to explain the purpose and usage of each constant and function.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: deploy_encoded_len clones and re-encodes each deploy multiple times</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 195)

deploy_encoded_len does `DeployData::to_proto(deploy.clone()).encoded_len()`, requiring a full clone plus proto conversion. In select_deploys_for_block every deploy is measured twice (once for total_bytes, once in the selection loop), and the function is invoked for three selection rounds (retry/ordinary/in-scope-recovery). With 128+ candidates carrying multi-hundred-KiB payloads this is nontrivial per-proposal overhead.

**Suggestion:**
```
Compute encoded lengths once into a Vec/HashMap keyed by sig at the top of select_deploys_for_block, and/or add an encoded_len path that borrows instead of cloning.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Nested lock acquisition: deploy_storage guard held while locking rejected_deploy_buffer</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 396)

In prepare_user_deploys_with_policy's expired-purge path, `deploy_storage_guard` (parking_lot::Mutex) is held while acquiring the `rejected_deploy_buffer` std::Mutex. If any other code path acquires these locks in the opposite order (buffer first, then storage), this introduces a deadlock hazard. Even if no such path exists today, the ordering constraint is implicit and fragile.

**Suggestion:**
```
Release the storage guard before locking the buffer (the two removals are independent), or document the required lock ordering at both lock declarations.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Shared byte budget accounting can overstate deferred bytes</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 683)

deferred_user_deploy_bytes sums deferred_bytes from retry, ordinary, and in-scope selections. Each selection computes deferred as total_bytes - selected_bytes for its own candidate set, so overlapping classification or later passes seeing a reduced budget can inflate deferred metrics and make byte_cap_hit harder to interpret under load.

**Suggestion:**
```
Compute deferred bytes once from the final selected set against the union of candidates, or track remaining candidates explicitly after each pass.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: recovery_backlog now forced false when parent list is empty</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 422)

Previously recovery_backlog checked buffer.non_empty() unconditionally. The refactor moved the backlog check into the `!sorted_parents_list.is_empty()` branch, so with an empty parent list (edge case, e.g., early bootstrap) a non-empty recoverable buffer no longer sets recovery_context. Likely benign, but it's a silent behavior change worth confirming.

**Suggestion:**
```
Confirm the empty-parents case cannot occur once past genesis, or run local_rejected_buffer_has_recoverable_deploys with an empty parent_hashes slice (which the tests show is supported) in that branch.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: block_in_floor_merge_scope changes >= to > and excludes the floor itself; error defaults differ between call sites</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 67)

The old predicate kept blocks with `block_number >= floor_block_number`; the new one keeps only `> floor_block_number`, so blocks at exactly floor height (including the floor itself) are now dropped unless they are not DAG ancestors of the floor. The backstop test asserts this is intentional, but it is a semantic change beyond the advertised is_in_main_chain→is_dag_ancestor swap and deserves an explicit comment. Also note the two call sites use opposite error defaults (`unwrap_or(false)` in include_visible_ancestor vs `unwrap_or(true)` in the retain filter) — pre-existing asymmetry, but now both routed through the same helper it stands out and merits a justifying comment since lookup_unsafe errors are the common failure mode for missing metadata.

**Suggestion:**
```
Add a comment at block_in_floor_merge_scope documenting the deliberate exclusion of floor-height DAG ancestors (including the floor) and why the two call sites choose different error defaults.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unrelated tooling documentation bundled into a consensus-logic PR</b> (Reported by: anthropic, xai)</summary>

**File:** `CLAUDE.md` (line 100)

The CLAUDE.md changes (Claude Code local settings, bash timeouts) are developer-tooling documentation unrelated to the finalization/admission/backpressure ports. Bundling them makes the PR's review scope and revert surface larger than necessary.

---

The CLAUDE.md checkpoint/timeout guidance is unrelated to finalization scope, byte admission, or DAG-width backpressure and adds review noise on an already large consensus change.

**Suggestion:**
```
Split the CLAUDE.md change into a separate docs/tooling PR.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Byte budgets hard-coded as constants rather than configurable</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 152)

USER_DEPLOY_BYTE_PROPOSAL_BUDGET (2 MiB) and the backpressure variant (512 KiB) are compile-time constants, unlike the new empty-frontier width cap which got a full config knob (defaults.conf, CLI, HOCON). Operators tuning block sizes for their network transport will need a rebuild.

**Suggestion:**
```
Consider promoting the byte budgets to casper_conf knobs (with these values as defaults) in a follow-up, mirroring the empty-frontier-max-unfinalized-blocks plumbing.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: prune_dag_covered_parents is O(n²) in DAG-ancestor queries</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 191)

For each empty-deploy candidate parent, an is_dag_ancestor query is issued against every other parent. Parent counts are typically small so this is acceptable, but each is_dag_ancestor call can itself be a non-trivial DAG traversal; on wide parent sets under the exact recovery conditions this feature targets, this could add measurable latency to compute_snapshot.

**Suggestion:**
```
Short-circuit by iterating candidates in descending block_number order (only the max-height parent can cover all others), reducing to at most one covering check per snapshot.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Port note disagrees with PR description on empty-deploys guard</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 884)

The PR summary says Port 3 narrows the fast path with is_dag_ancestor + empty-deploys guard, but the code comment says the empty-deploys guard is deliberately NOT taken. Reviewers need a single source of truth for the intended semantics.

**Suggestion:**
```
Update the PR description and discovery doc to match the code, and record why the guard was rejected so future ports do not reintroduce it accidentally.
```

</details>
---

<details>

This is a carefully documented, well-tested port of three mechanisms (finalization merge scope, byte-bounded deploy admission, DAG-width backpressure) with strong test coverage including backstop and regression tests, and it aligns code with already-proven FV artifacts. No security issues were found. The main concerns are logic-level: the byte-budget selection loop's skip-and-continue behavior can indefinitely starve large deploys (the carve-out only fires on an empty selection), and the recoverability predicate is duplicated with divergent filters between snapshot.rs and block_creator.rs, risking inconsistent recovery gating. There are also performance regressions worth watching — the recovery gate went from O(1) non_empty() to block-store ancestry scans on every propose/snapshot, and deploy sizes are re-encoded with clones multiple times per selection. The floor-scope predicate change (>= to >, excluding the floor itself) goes beyond the advertised main-chain→DAG-ancestor swap and deserves an explicit comment even though tests assert it. All issues are addressable; the change is functional and safe to merge after considering the starvation and predicate-consistency feedback.

The PR introduces significant changes related to byte budget enforcement and DAG pressure management. While the changes seem logically sound, they require thorough testing to ensure correctness. Additionally, some minor style and documentation improvements can be made to enhance code readability.

#### openai (gpt-5.6-sol)

> **ABSTAIN**: Empty response from API

#### xai (grok-4.5)

This port closes real proof↔code and admission-dimension gaps, and the added RecoveryDeferred path, recoverable-buffer gating, byte-budget selection, and floor-scope DAG-ancestor tests are directionally right. Before merge, a human should re-check three consensus-sensitive behaviors: deploy-bearing DAG-cover fast-path short-circuiting, parent-set compaction via prune_dag_covered_parents, and the duplicated recoverable-buffer predicates that can disagree between snapshot recovery_context and block-creator admission. The encoded-length helper also needs a cheaper implementation for the propose hot path. No direct security vulnerability stands out in the visible diff, but the behavioral risk on merge/parent selection is high enough that this should not be treated as a pure approve.
</details>

---

jeffrey-l-turner:
## Multi-review finding resolutions — 4a2c8340

Disposition of the review's critical and major findings. Fixes landed in `4a2c8340`; triage rationale below for the rest.

### 🔴 Critical: recoverability predicate divergence (`snapshot.rs` vs `block_creator.rs`) — **documented as deliberate design (not a defect)**

The two filters are *born* different, inherited verbatim from PR #114's `9ec1096e`, because they run at different pipeline stages: the snapshot copy executes **inside `compute_snapshot`, before the snapshot's `deploys_in_scope`/`rejected_in_scope` sets exist**, so it can only approximate with the block-number window; the block-creator copy runs after snapshot completion and refines with the scope sets. Disagreements are benign by construction — the snapshot copy only drives the parent-narrowing heuristic, never admission: true-there/false-here costs one narrowed-parent propose that declines recovery; false-there/true-here skips the narrowing while recovery still admits at block creation. `4a2c8340` adds cross-referencing doc comments on both functions explaining exactly this and warning against "harmonizing" them. Both also gained the O(1) `non_empty()` fast path (see below).

### ✅ Fixed in 4a2c8340

- **`deploy_encoded_len` full-deploy clone per size check** — added `DeployData::to_proto_ref(&Signed<DeployData>)` in models (existing `to_proto` delegates to it; zero drift), and `select_deploys_for_block` now computes each deploy's encoded length **once per call** via a sig-keyed map (previously twice, with a full `Signed` clone each time).
- **`count_capped` mislabel when `byte_budget == 0`** — the early return now sets `count_capped: cap == 0 && !ordered.is_empty()`, so a pure byte-budget-zero case reports `byte_capped` only.
- **O(1) → block-store-scan regression in `rejected_buffer_has_recoverable_deploys`** — the O(1) `non_empty()` check now guards `read_all()` in both predicate copies, restoring cheap steady-state `create()` for the common empty-buffer case; the canonical-won scan runs only when recovery work actually exists, and its `scan_floor` bound is documented.

### 📝 Documented (no code change warranted)

- **Byte-budget skip-and-continue starvation** — starvation is finite, not permanent: a budget-skipped deploy keeps its `(valid_after, timestamp, sig)` position while everything ahead of it is selected, included, and removed, and new deploys always sort behind it — so it strictly advances to the front, where the `selected.is_empty()` carve-out admits it alone even if it exceeds the whole budget. Now stated as a liveness comment on the loop.
- **`prune_dag_covered_parents` "may collapse useful tips"** — no tip content can be lost: collapse only happens when a single deploy-free parent DAG-covers *every* other candidate, meaning the others are already in its ancestry and their effects are in its post-state; sibling tips that don't cover each other always survive (`covers_all` fails). Now stated as a doc comment on the function.
- **Fast path skips merge for deploy-bearing covering parents** (`interpreter_util.rs`) — intentional, argued in the `33e117dd` commit and its in-code comment: a covering parent's post-state already carries its ancestry's effects, deploys included; forcing a merge instead produced the degenerate ancestor-merge the conflict-set merger rejects (the two release-suite failures that commit fixed). Flagged for human sign-off as the reviewers requested.

Verification for `4a2c8340`: full workspace + tests compile, fmt clean, clippy `-D warnings` clean (models, casper), byte-budget/backlog tests + 61-test module sweep pass.

### CI status note

- `43c4dc79`'s CI failure (`get_listening_name_data_response_should_work_across_a_chain`) is the pre-fix state repaired by `33e117dd`.
- `c2f8da1c`'s only genuine failure was `test_deploy_phlo_price_too_small` timing out a 10s inclusion window on the **amd64-docker slot only** (passed on amd64-subprocess, arm64-docker, arm64-subprocess; the red "Integration Tests (arm64/amd64)" entries are result-propagation gates, not test failures). The run on `4a2c8340` is the deciding data point for flake vs. real.

Issue #53, opened 2026-05-20
API `estimate-cost` endpoint returns much more lower value than the real one


Timekiller7:
### What did you expect to see?

When calling `POST /api/estimate-cost` with a Rholang term, the returned `cost` should closely match the actual phlogiston cost charged when the same term is submitted as a real deploy via `POST /api/deploy`. The estimate must be an upper bound — or at minimum a tight approximation — of the real cost, not a lower bound.

### What did you see instead?

The endpoint consistently returns a value **lower** than the real on-chain cost. Tested on token transfers (100 tokens):

| Run | `estimate-cost` returned | Actual cost | Underestimate |
|-----|--------------------------|-------------------------------------|---------------|
| 1   | 185,676                  | 254,985                             | ~27% |
| 2   | 107,239                  | 175,000                             | ~39% |

This makes the estimate unsafe to use directly as `phloLimit`. Clients are forced to apply an arbitrary safety multiplier, which defeats the purpose of an "estimate" endpoint.

Note that the gap is not a fixed offset; it varies between runs of the same kind of operation (~27% vs ~39% above), suggesting estimation diverges from execution in a non-deterministic or context-dependent way rather than just missing a constant overhead.

### What version of F1R3FLY Node are you using?
staging, rust-v0.4.15, ff76624ba2a309a63c7db12630c31c91473516c4

Timekiller7:
took issue

jeffrey-l-turner:
Closing as fixed and shipped to `master` (promoted via PR #213). Fix chain: PR #134 ("Estimate cost fix", merged to dev 2026-08-07) plus the follow-on exploratory-deploy/estimate-cost commits `cba7e1c6` (pass deployer pubkey to exploratory deploy for accurate estimate-cost), `8aa228b5`, `825fb298` (fix empty `sig` visibility to Rholang so cost estimates no longer diverge from real deploys), and `ddc55b88` (estimate-cost response reports which identity produced the estimate). Reopen if a remaining divergence case surfaces.

PR #134, opened 2026-07-19
Estimate cost fix

Timekiller7:
Fix of issue #53

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/134"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787086899&installation_id=145946300&pr_number=134&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F134&signature=4c85dadba0f2a28dcf0241655aaf71e68bf1740c2d61c27c6a3c3877c6e123ca"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


spreston8:
Reviewed at `bb83f452` against `dev`, reading the live files. The diagnosis is right and non-obvious — running exploratory deploys under a process-wide random key meant identity-dependent terms took a cheaper path (vault lookup for a RevAddress that doesn't exist short-circuits), which explains both #53 symptoms: the systematic underestimate and why the gap varied between runs rather than being a fixed offset. I also confirmed the hand-built `DeployData` matches `source_deploy`'s defaults exactly (`phlo_price: 1`, `valid_after_block_number: 0`, `shard_id: ""`, `expiration_timestamp: None`), so the change really does isolate identity, and `estimate_cost_with_deployer_matches_real_deploy_cost` passes here in 8.57s.

Requested changes below.

**Medium**

1. **The fix is opt-in, so the default response still has the #53 problem.** `deployer` is optional and every existing caller passes `None`, so `POST /api/estimate-cost` keeps returning the ephemeral-identity value unless a client knows to send the new field — and nothing in the response distinguishes that value from an accurate one. The issue asked for an upper bound or tight approximation and specifically called the old number "unsafe to use directly as `phloLimit`"; after this PR that number is still what a caller gets by default. The field docs and `api-reference.md` are honest about it, which helps, but a client that doesn't read them hits exactly the trap that produced the issue. Could the response carry which identity was used, so callers can tell they're holding the unreliable variant?

2. **The empty `sig` is visible to Rholang, so the comment isn't quite right.** The new branch sets `sig: Bytes::new()` with "sig is intentionally empty: exploratory deploys are never signature-verified. pk is only used to derive the deployer identity". The verification half is true, but `sig` is also what backs `rho:rchain:deployId` — `system_processes.rs:515` does `deploy_id: template.sig.to_vec()`. So a term reading `rho:rchain:deployId` sees empty bytes on the new path, where the `None` path (`source_deploy` → `Signed::create`) gives it a real 64-byte signature. Net effect: passing a deployer improves identity fidelity but degrades deployId fidelity — the same estimate-vs-execution divergence #53 is about. The new test doesn't catch it because its term uses `rho:system:deployerId` rather than `deployId`. Correcting the comment is the minimum; a deterministic 64-byte placeholder, or an explicit note about the residual gap, would be better.

**Low**

3. Building `Signed` as a struct literal sidesteps both of its invariant-preserving constructors (`create` signs, `from_signed_data` verifies and returns `Option`). The fields are `pub` so it compiles, but afterwards nothing at the type level separates a signed `Signed` from an unsigned one. Would a named constructor in the crypto crate — `Signed::unverified` or similar — be worth it, so the intent is explicit and greppable?

4. The gRPC `ExploratoryDeploy` path passes `None` (`deploy_grpc_service_v1.rs`), so gRPC clients keep the old behaviour with no way to opt in. No objection if that's deliberate — worth saying so.

Timekiller7:
ok, thanks, I' ll fix it

Timekiller7:
@spreston8, regarding Low 3 naming.
I used a create_unbound fn instead of a Signed::unverified constructor. unverified suggests "wrap given bytes without checking them," but this method signs data itself just with a placeholder key while carrying the caller's pk. So the signature is unbound from pk by construction, and the name points at the invariant that's deliberately dropped rather than just "verification skipped." Happy to change / add something if needed.

PR is ready for review

PR #137, opened 2026-07-21
feat(fileio): non-deterministic-replay guards for the 19 native syscalls

metaweta:
## Summary

Every fileio native primitive was already registered in `non_deterministic_ops()` (so the dispatcher captures its output on the lead node's play), but the 19 handlers themselves ignored `is_replay` / `previous_output` and re-ran the syscall on followers — observing whatever the follower's disk state happened to be at that moment. On any node whose disk diverges from the lead's (observer nodes without the same static-config directory, a lead that mutated its own tree between play and replay, etc.), replay would diverge from the log's recorded output and consensus would break.

This PR adds the canonical replay guard to all 19 handlers, matching the shape used by `gpt4` / `grpc_tell` / the other pre-existing non-deterministic system processes.

## Change

For each of the 19 handlers in `system_processes.rs`:

1. Change the destructuring from `(produce, _, _, args)` to `(produce, is_replay, previous_output, args)`.
2. After the arg-shape and type-check validation blocks, but before the syscall, insert:

    ```rust
    if is_replay {
        produce(&previous_output, ack).await?;
        return Ok(previous_output);
    }
    ```

Placement rationale: after validation so followers observe the same arg-type failures the lead did (deterministic given deterministic args); before the syscall so followers never touch the host filesystem.

**Handlers covered** — every fileio native, without exception:

- Fd lifecycle: `native_open`, `native_close`, `native_flush`
- Positional: `native_read`, `native_write`, `native_seek`, `native_tell`, `native_size`, `native_truncate`
- Filesystem queries: `native_stat`, `native_entries`, `native_exists`, `native_quarantine`
- Filesystem mutations: `native_rename`, `native_copy_file`, `native_remove_file`, `native_remove_dir`, `native_chmod`, `native_chown`

Bookkeeping-only handlers (`seek`, `tell`, `close`, `flush`) go through the guard too — their local behavior looks deterministic but the fd they operate on came from a `native_open` whose fd allocation IS non-deterministic (fd table timeline coherence). Matches the "defense-in-depth" design rationale already recorded in `non_deterministic_ops()`.

## Tests

New integration suite `rholang/tests/fileio_replay_spec.rs`. The `play_mutate_replay` harness runs the standard evaluate → checkpoint → mutate → rig → re-evaluate → `check_replay_data` sequence, but **mutates the disk between play and replay** so any handler that cheated (ran the syscall on the follower) would produce output derived from the mutated state, mismatching the log and tripping `check_replay_data`.

Three concrete cases:

- **`exists_replays_false_after_file_is_created_between_runs`** — play sees "doesn't exist"; between runs the file is created; replay must still emit `[true, false]` from the log, not the fresh syscall's `[true, true]`.
- **`exists_replays_true_after_file_is_deleted_between_runs`** — symmetric: play sees "exists", file is deleted, replay must still emit `[true, true]`.
- **`stat_replays_original_metadata_after_file_is_modified`** — play captures a stat record; the file is overwritten between runs (size + mtime change); replay must emit the original record.

**Regression-proof verified in-workflow**: git-stashed the handler changes, re-ran the suite, watched all 3 fail with `RSpaceError(BugFoundError("Unused COMM event: replayData multimap has 2 elements left"))` — exactly the signal we wanted. Restored the guards, all 3 pass.

## Test plan

- [x] `cargo test -p casper --test mod empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes (no genesis-contract changes)

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/137"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787252504&installation_model_id=426883&pr_number=137&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F137&signature=3d67ef37ed390dae6d904a48f0f38d70fd6ba48f9d7ca4b59693e27a7819d2c4"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/rholang/src/rust/interpreter/io/handle_table.rs
+++ b/rholang/src/rust/interpreter/io/handle_table.rs
@@ -90,6 +90,36 @@ impl FileHandleTable {
     pub async fn remove(&self, fd: i64) -> Option<Arc<FileHandle>> {
         self.inner.write().await.remove(&fd)
     }
+
+    /// Capture the current `next_fd` counter so a later
+    /// [`truncate_to`](Self::truncate_to) can identify "fds
+    /// inserted since this snapshot" by monotonicity: any fd
+    /// >= the snapshot was allocated after it. Used by the
+    /// runtime's `evaluate_with_env_and_phlo` to tie fd
+    /// allocation to the same soft-checkpoint boundary that
+    /// wraps rspace state, so a deploy that errors after
+    /// `nativeOpen` doesn't leave zombie fds in the table.
+    pub fn snapshot_next_fd(&self) -> i64 { self.next_fd.load(Ordering::SeqCst) }
+
+    /// Remove every entry with fd >= `snapshot` and reset the
+    /// `next_fd` counter to `snapshot`. Used on the error path
+    /// of `evaluate_with_env_and_phlo` to roll back fd
+    /// allocations from a deploy whose rspace state is being
+    /// reverted -- otherwise the OS-level file descriptors
+    /// held by `tokio::fs::File` inside `Arc<FileHandle>`
+    /// would stay open until the runtime is dropped, giving
+    /// a hostile deploy an easy path to exhaust the process's
+    /// `ulimit -n`.
+    ///
+    /// Reusing fd numbers across deploys is safe: any produce
+    /// event that mentioned a reverted fd was itself reverted
+    /// by the same soft-checkpoint restore, so no surviving
+    /// log entry refers to it.
+    pub async fn truncate_to(&self, snapshot: i64) {
+        let mut map = self.inner.write().await;
+        map.retain(|fd, _| *fd < snapshot);
+        self.next_fd.store(snapshot, Ordering::SeqCst);
+    }
 }
 
 impl Default for FileHandleTable {
--- a/rholang/src/rust/interpreter/rho_runtime.rs
+++ b/rholang/src/rust/interpreter/rho_runtime.rs
@@ -258,6 +258,15 @@ pub struct RhoRuntimeImpl {
     pub invalid_blocks_param: InvalidBlocks,
     pub deploy_data_ref: Arc<tokio::sync::RwLock<DeployData>>,
     pub merge_chs: Arc<tokio::sync::RwLock<HashMap<Par, MergeType>>>,
+    /// The runtime's single File-I/O open-fd table, shared with
+    /// every fileio native handler via the dispatch-table build.
+    /// Exposed on the runtime (not just buried inside dispatch
+    /// closures) so `evaluate_with_env_and_phlo` can snapshot
+    /// `next_fd` before evaluate and truncate on error paths,
+    /// tying fd lifetime to the same soft-checkpoint boundary
+    /// that rolls rspace state. Tests can also inspect it
+    /// directly to verify fd table state after a deploy.
+    pub file_handles: crate::rust::interpreter::io::handle_table::FileHandleTable,
 }
 
 impl RhoRuntimeImpl {
@@ -268,6 +277,7 @@ impl RhoRuntimeImpl {
         invalid_blocks_param: InvalidBlocks,
         deploy_data_ref: Arc<tokio::sync::RwLock<DeployData>>,
         merge_chs: Arc<tokio::sync::RwLock<HashMap<Par, MergeType>>>,
+        file_handles: crate::rust::interpreter::io::handle_table::FileHandleTable,
     ) -> RhoRuntimeImpl {
         RhoRuntimeImpl {
             reducer,
@@ -276,6 +286,7 @@ impl RhoRuntimeImpl {
             invalid_blocks_param,
             deploy_data_ref,
             merge_chs,
+            file_handles,
         }
     }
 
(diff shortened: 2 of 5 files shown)


metaweta:
## Update — fd table lifecycle: sharing + M1 rollback

Fix landed in commit `0ac2a4df`. Two related bugs, both caught while investigating M1 from the security review.

### Prerequisite bug: fd table wasn't actually shared

While preparing the M1 fix, I noticed `ProcessContext::create` was called once per `Definition` in the dispatch-table build loop, and each call constructed a fresh `SystemProcesses` with `file_handles: FileHandleTable::new()`. Each of the 19 fileio native handlers thus closed over its OWN table — `nativeOpen` inserted `fd=N` into one table, `nativeRead(N)` looked up in a different (empty) one and returned `[false, FSERR_CLOSED, "fd N is not open"]`.

**The entire native primitive layer was non-functional through the runtime.** The only reason no existing test caught it: no integration test had actually exercised `open → read` through the dispatcher. Unit tests in `interpreter::io::handle_table` use the table directly; PR #132's `injection_runtime_spec` and #137's `fileio_replay_spec` reach only arg-validation branches.

Fix: thread a single `FileHandleTable` through `create_rho_env` → `setup_reducer` → `dispatch_table_creator` → `ProcessContext::create` → `SystemProcesses::create`. Cloning `FileHandleTable` shares its inner `Arc<RwLock<HashMap>>`. The table is also stashed on `RhoRuntimeImpl` (returned as the 5th element of `create_rho_env`'s tuple) so the runtime can manage its lifecycle.

### M1 fix: rollback around `evaluate_with_env_and_phlo`

`FileHandleTable` gains `snapshot_next_fd(&self) -> i64` and `truncate_to(&self, snapshot: i64)` (removes entries with `fd >= snapshot`, resets `next_fd`). `RhoRuntimeImpl` overrides the trait default `evaluate_with_env_and_phlo` to snapshot before evaluate and truncate on both error paths (Ok-with-errors and hard Err).

Fd reuse across deploys is safe: any produce event that mentioned a reverted fd was itself reverted by the rspace restore, so no surviving log entry refers to it. Not a consensus issue — `FileHandleTable` isn't part of the state hash.

### Tests

New `rholang/tests/fileio_lifecycle_spec.rs` covering both invariants:

- **`open_then_read_shares_fd_table`** — opens a file, reads 5 bytes from the returned fd, asserts the read returns `[true, GByteArray([104, 101, 108, 108, 111])]` (ASCII "hello"). Regression guard for the sharing bug.
- **`successful_deploy_leaves_fds_in_table`** — baseline: after a clean deploy that opens two files, `next_fd` has advanced by 2.
- **`erroring_deploy_rolls_back_fd_allocations`** — opens a file, then errors via `1 / 0`. After the deploy, `file_handles.snapshot_next_fd()` must equal the pre-deploy value.

**Regression-proof verified in-workflow**: temporarily disabled just the two `truncate_to` calls in the M1 override, re-ran the suite: `erroring_deploy_rolls_back_fd_allocations` fails with `left: 2, right: 1` — the M1 fix is exercised. The sharing test continued to pass — confirming the two fixes are independently verified. Restored the truncate calls; all 3 pass.

### Verified

- 233 `rholang` lib tests pass (unchanged)
- 3 lifecycle + 3 replay + 2 injection + 37 io-lib integration tests all pass
- Empty state hash unchanged
- Pre-commit + pre-push clean

dylon (changes requested):
claude found 3 blockers and made a few recommendations:

  ## Review — PR #137 (non-deterministic replay guards + fd rollback)

  **Verdict: blocked.** The replay half is right; the fd-lifecycle half is aimed at a method
  deploys never call.

  The guards themselves are correct and uniformly applied to all 19 handlers, byte-identical
  in shape to the pre-existing `gpt4`/`dalle3` guards, and placed consistently — after argument
  validation (so followers observe the same type failures) and before the syscall (so followers
  never touch their disk). The `play → mutate-the-disk → replay → check_replay_data` harness is
  a genuinely good test design, and verifying it by stashing the guards and watching all three
  fail with `Unused COMM event` is exactly the right way to prove a regression test regresses.

  ### Blocking

  1. **The fd rollback never runs on the deploy path.** It overrides
     `RhoRuntimeImpl::evaluate_with_env_and_phlo`, but deploys don't call it:
     `casper/src/rust/rholang/runtime.rs:1049` (`RuntimeOps::evaluate`) calls
     `self.runtime.evaluate(term, cost, normalizer_env_from_deploy(deploy), rng)` directly, and
     `runtime.rs:926` uses `self.runtime.inj(..)`. `RhoRuntimeImpl::evaluate` (line 288) goes
     straight to `inj_attempt` with no soft checkpoint. `rg -n "evaluate_with_env_and_phlo"
     casper/ node/` finds no production call site — the override is reachable only from the
     REPL and tests. `erroring_deploy_rolls_back_fd_allocations` calls it directly, so it passes
     while production is unprotected. Please hook the snapshot/truncate into the casper path, or
     move it into `RhoRuntimeImpl::evaluate` so every caller inherits it.
  2. **No cap on open handles.** On the success path fds are never released —
     `successful_deploy_leaves_fds_in_table` asserts exactly that — and there is no table bound,
     per-deploy quota, or idle eviction. `spawn_runtime` builds a fresh table per `compute_state`,
     so leaks are per-block rather than per-node, but a block of open-only deploys still walks the
     process toward `ulimit -n`, and the long-lived REPL runtime has no such bound. The FIP already
     defines `FSERR_QUOTA_EXCEEDED` for precisely this and it currently has zero emit sites.
  3. **`truncate_to` breaks the never-reuse-fd invariant the design rests on.** It does
     `map.retain(|fd, _| *fd < snapshot)` then `next_fd.store(snapshot)`. The safety argument
     ("any produce event that mentioned a reverted fd was itself reverted by the same
     soft-checkpoint restore") holds only under strictly serialized evaluation. With two
     evaluations sharing one runtime, B's rollback rewinds the counter while A still holds fd 5;
     C then opens and is *issued fd 5*, and A's next read hits C's file. `handle_table.rs`
     documents and tests the opposite invariant (`removed_fd_is_not_reissued`).

  ### Design note

  4. Because the table is per-spawned-runtime, FIP §Use case 4 — "clients … maintain access to
     files between deployments" — is not implementable as currently structured. Worth deciding
     where the table should live before the powerbox lands.
  5. Worth stating explicitly somewhere: with these guards, followers replay captured tuples and
     never touch their own disk, so the implementation realises FIP §Use case **5a** (proposer's
     filesystem is authoritative, others check processing not data) but not **5b** ("the shard …
     can check the consistency of the oracle"). Nothing lets a validator check the proposer's
     claimed results. That's a real security boundary and shouldn't have to be inferred.

metaweta:

PR #93, opened 2026-06-30
Add PeTTa/MeTTa smart contract execution integration

rmgaray:
Add PeTTa (Prolog-based MeTTa) integration for executing MeTTa smart contracts from Rholang via the rho:petta:execute system contract URN.

PR #139, opened 2026-07-22
feat(fileio): minimal File agent + rholang-parser bump

metaweta:
## Summary

First rung of Phase 2: the minimal `File` agent wrapping the native fd surface, plus a rholang-parser pin bump that brings in the try/catch sugar Phase 2 needs.

## Commits

- **`03172f5e`** — `chore(deps): bump rholang-parser to master tip (7588a3a)`. Pin was at `527af8a` (after PRs #93/#94/#95 landed) but before PR #96 (try/catch sugar). Bumping to `7588a3a` (PR #96 merge) brings in try/catch sugar, hygiene fix (`2f247dc`), and SR/SM cleanup (`eb85eff`). Empty state hash unchanged; no genesis-installed `.rho` uses the newly-recognized syntax.
- **`684c9ff4`** — `feat(fileio): minimal File agent (read/size/close/default)`. See below.

## File agent

`rholang/src/rust/interpreter/io/agents/file.rho` — the `agent File { constructor / method read / method size / method close / default }` block. Uses agent-block sugar (rholang-rs PR #94), `!?` sync-send + try/catch sugar (PR #96), and `NormalizerEnv` injection to reach the native primitives — all FIP-style, no workarounds.

Factored as a standalone block (not wrapped in its own `new`) so the same string serves two purposes:
- Integration tests wrap it in a `new`-scope that binds the native URNs and drives the agent directly.
- The Phase-2 `Fs` agent's genesis deploy (future PR) will embed it inside its own scope, never exposing the URN bindings outward.

Rust plumbing: `rholang/src/rust/interpreter/io/agents.rs` exposes `pub const FILE_AGENT_SRC: &str = include_str!("agents/file.rho");`. Registered via `pub mod agents;` in `io/mod.rs`.

### Method surface

| Method | Signature | Delegates to |
|--------|-----------|--------------|
| `read(n)` | `[true, bytes]` or `[false, code, msg]` | `nativeRead(fd, n)` |
| `size()` | `[true, n]` or error tuple | `nativeSize(fd)` |
| `close()` | `[true]` or error tuple | `nativeClose(fd)` |
| `default(...@args)` | `[false, "FSERR_UNSUPPORTED", ...]` | — |

### Per-instance state

Constructor uses `@fd` (process pattern), receives the fd Int directly, stashes on `@(*this, "fd")`. Methods peek (`<<-`) non-consumingly. Standard Rholang per-instance-state pattern.

### Default-arm reply idiom

The agent-block sugar desugars `default(...@args) { body }` to `_ => body`. This is correct: the outer `for(...@args <= this)` already binds `args` in scope for every match arm, so the default body has access to the full incoming args list. Its first element is the caller's return channel as a Process (from send's auto-deref of the reply-channel Name). The default arm destructures it and re-quotes with `@` to reply:

```rholang
default(...@args) {
  match args {
    [returnProc, ...rest] => {
      @returnProc!([false, "FSERR_UNSUPPORTED",
                    ("File agent has no method:", rest)])
    }
  }
}
```

## Native-handler ack-first swap (folded into the same commit)

Every fileio native handler in `system_processes.rs` swapped from `let [args..., ack] = args.as_slice()` (ack LAST) to `let [ack, args...] = args.as_slice()` (ack FIRST). Variable names preserved; only positions moved.

**Why**: `!?` sync-send and try/catch's SendReceive source both PREPEND the ephemeral return channel (verified in `p_send_sync_normalizer.rs:35-39` and `p_input_normalizer.rs:148,164`). Before the swap, `nRead!?(fd, n)` inside a `try/catch` sent `(*tmp, fd, n)`, the handler read `fd_par = *tmp` (a Name), failed `RhoNumber::unapply`, and returned `IllegalArgumentError`. After the swap, the sugar composes directly and File.rho methods use the FIP-style `try @<pat> <- nX!?(args) { ok } catch @[code, msg] { err }` form verbatim — half the lines vs. an alternative hand-rolled send-then-for-then-match cascade.

Callers updated: 6 raw-send call sites across three integration test files reordered to ack-first.

## Tests

4 integration tests in `rholang/tests/fileio_file_agent_spec.rs`. Together they exercise the full stack top-to-bottom: agent-block sugar → dispatch → NormalizerEnv injection → native handlers → replay guards → shared `FileHandleTable`. Any regression in any layer surfaces here.

- **`file_agent_read_returns_file_contents`** — opens `"hello, file agent"`, calls `read(17)`, asserts the returned `GByteArray` is the expected ASCII bytes.
- **`file_agent_size_returns_byte_length`** — `size()` returns `GInt(10)` for a 10-byte file.
- **`file_agent_close_then_read_reports_closed`** — after `close()`, a subsequent `read()` surfaces `FSERR_CLOSED` from the native handler. Also validates that fd-table state persists across the agent's method invocations.
- **`file_agent_unknown_method_returns_unsupported`** — unknown method hits the default arm; asserts the reply is `[false, "FSERR_UNSUPPORTED", ...]` with the offending method name in the payload.

## Test plan

- [x] `cargo test -p rholang --test fileio_file_agent_spec` — 4 pass
- [x] `cargo test -p rholang --lib` — 233 pass (unchanged)
- [x] `cargo test -p casper --test mod empty_state_hash_should_be_the_same_as_hard_coded_cached_value` passes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5136,7 +5136,7 @@ dependencies = [
 [[package]]
 name = "rholang-parser"
 version = "0.1.0"
-source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=527af8a#527af8afb37f6cad470516fe316076f120b9c353"
+source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=7588a3a#7588a3ab09e244a148166549527865de179de9b2"
 dependencies = [
  "bitvec",
  "nonempty-collections",
@@ -5151,7 +5151,7 @@ dependencies = [
 [[package]]
 name = "rholang-tree-sitter"
 version = "0.1.2"
-source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=527af8a#527af8afb37f6cad470516fe316076f120b9c353"
+source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=7588a3a#7588a3ab09e244a148166549527865de179de9b2"
 dependencies = [
  "cc",
  "tree-sitter-language",
@@ -5160,7 +5160,7 @@ dependencies = [
 [[package]]
 name = "rholang-tree-sitter-proc-macro"
 version = "0.1.2"
-source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=527af8a#527af8afb37f6cad470516fe316076f120b9c353"
+source = "git+https://github.com/F1R3FLY-io/rholang-rs?rev=7588a3a#7588a3ab09e244a148166549527865de179de9b2"
 dependencies = [
  "anyhow",
  "quote",
@@ -6245,7 +6245,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "32497e9a4c7b38532efcdebeef879707aa9f794296a4f0244f6f69e9bc8574bd"
 dependencies = [
  "fastrand",
- "getrandom 0.4.3",
+ "getrandom 0.3.4",
  "once_cell",
  "rustix",
  "windows-sys 0.61.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,7 +78,7 @@ utoipa = "5"
 validated = "1.0"
 serial_test = "3.1"
 metrics-util = "0.17"
-rholang-parser = { git = "https://github.com/F1R3FLY-io/rholang-rs", rev = "527af8a" }
+rholang-parser = { git = "https://github.com/F1R3FLY-io/rholang-rs", rev = "7588a3a" }
 reqwest = "0.12"
 
 [profile.dev]
--- a/rholang/src/rust/interpreter/io/agents.rs
+++ b/rholang/src/rust/interpreter/io/agents.rs
@@ -1,0 +1,25 @@
+//! Rholang agent-class sources for the File I/O FIP.
+//!
+//! Each constant is the raw text of an `agent` block (per rholang-rs
+//! PR #94's sugar) that expects its enclosing scope to bind the
+//! native-URN names it references. That factoring lets one string
+//! serve two purposes:
+//!
+//! 1. **Integration tests** wrap the block in a `new`-scope that
+//!    binds the native URNs and directly exercises the agent's
+//!    methods, isolating the agent-layer semantics from the rest
+//!    of the FS-agent stack.
+//! 2. **The `Fs` agent's genesis deploy** (Phase 2) embeds the
+//!    same block inside its own `new`-scope, which binds the
+//!    native URNs via `NormalizerEnv` injection and never exposes
+//!    them outward. User code reaches these agents only through
+//!    the `Fs` methods that construct them.
+//!
+//! Native URNs each `.rho` file expects:
+//!
+//! - `file.rho` — `nRead`, `nSize`, `nClose` (extended as the
+//!   `File` surface grows in follow-up PRs).
+
+/// Source of the `File` agent block. Expects `File`, `nRead`,
+/// `nSize`, `nClose` to be bound in the enclosing scope.
+pub const FILE_AGENT_SRC: &str = include_str!("agents/file.rho");
--- a/rholang/src/rust/interpreter/io/agents/file.rho
+++ b/rholang/src/rust/interpreter/io/agents/file.rho
@@ -1,0 +1,109 @@
+// File agent — wraps a native fd with the FIP `File` method surface.
+//
+// This is the MINIMAL surface: `read`, `size`, `close`, `default`.
+// Extended in follow-up PRs with `write`, `seek`, `tell`, `truncate`,
+// `flush`, `readAt`, `writeAt`, `lockRange`, `chmod`, `chown`, plus
+// line-based `text` / `lines` / `mapReduceLines*`.
+//
+// This file is an `agent` block that expects the enclosing scope
+// to bind:
+//   - `File`               — the agent's constructor channel
+//   - `nRead`, `nSize`,
+//     `nClose`             — fixed-channel Pars for the native
+//                            primitives, obtained via `NormalizerEnv`
+//                            injection at compile time (see
+//                            `interpreter::io::injections`).
+//
+// Constructor calling convention: the `@fd` formal is a process
+// pattern, so the caller sends the fd as a plain Process arg
+// (e.g. `File!(*replyChan, fd)`). The constructor stashes it on
+// a per-instance subchannel keyed on `*this`, and every method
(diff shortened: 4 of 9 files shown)


metaweta:
## Update — native handlers swapped to ack-first destructuring

Landed as commit `72d93375`. Followed up on the convention-mismatch limitation I flagged in the previous update.

### The mismatch (re-diagnosed)

`p_send_sync_normalizer.rs:35-39` and `p_input_normalizer.rs:148,164` both PREPEND the ephemeral return channel — `!?` sends `(*tmp, a, b)`, and try/catch's SendReceive source does the same. Agent-block dispatch matches `[*return, "method", ...formals]` (return first). So sugar ↔ agent-block composes correctly; my earlier "method-call sugar orders args wrong" claim was wrong.

The actual mismatch was between the sugar (return FIRST) and the fileio native handlers (ack LAST): `nRead!?(fd, n)` sent `(*tmp, fd, n)`, `native_read` destructured `let [fd_par, n_par, ack] = args.as_slice()`, read `fd_par = *tmp` (a Name), failed `RhoNumber::unapply`, and returned the `IllegalArgumentError` that forced the send-then-for-then-match workaround.

### The fix

Every fileio native handler now uses `let [ack, arg0, ..., argN] = args.as_slice()` — ack FIRST — matching the sugar convention. Variable names preserved; only positions moved.

Mechanical change to 19 handlers in `system_processes.rs`. Callers updated:
- `file.rho` — `read`/`size`/`close` rewritten to the FIP-style `try @<pat> <- nX!?(args) { ok } catch @[code, msg] { err }` form. Roughly half the lines vs. the send-then-for-then-match workaround; matches FIP §"Error syntax" examples verbatim.
- 6 raw-send call sites across 3 integration test files reordered to ack-first.

### Rationale for option (a) over option (b)

The prior commit's limitations section listed two fix paths:
- (a) Swap handlers to ack-first (this commit).
- (b) Leave handlers alone; agent-layer Rholang uses raw sends only, never the sync-send sugar against native URNs.

Option (b) would pollute every Phase 2+ agent-layer file with hand-rolled send-then-for cascades that are less legible than the FIP examples. Option (a) is a mechanical 19-line change on the Rust side that makes all downstream Rholang authoring match the FIP verbatim. Worth the swap.

### Test coverage unchanged

- `fileio_file_agent_spec` — 4 pass (now via `!?`+try/catch sugar rather than the workaround).
- `fileio_lifecycle_spec` — 3 pass.
- `fileio_replay_spec` — 3 pass.
- `injection_runtime_spec` — 2 pass.
- 37 io-lib unit tests unchanged.
- Empty state hash unchanged.
- Pre-push clean (11 slices).

### Remaining limitation

`default(...@args) { body }` still desugars to `_ => body` without binding `return`/`args`, so the default arm can't reply — unknown methods drop silently. That's a rholang-rs sugar issue, orthogonal to this fix. Separate follow-up.

metaweta:
## Force-pushed: squashed into 2 clean commits

Previous history had 4 commits capturing the working process (initial File agent → ack-first swap → default-arm reply idiom). Now squashed to reflect the final state:

- `03172f5e` — parser bump (unchanged).
- `684c9ff4` — File agent (folds in the ack-first swap and the default-arm reply idiom).

PR body updated to match. Previous comments on this PR reference the pre-squash commit SHAs (`72d93375`, `0a46d87a`, `323a6cba`) — those are gone from the branch but the discussion context still holds: the ack-first swap and default-arm idiom are now folded into `684c9ff4`.

**Ready for review.**

dylon (approved):
claude found no blockers but did make a few recommendations:

  ## Review — PR #139 (minimal File agent + rholang-parser bump)

  **Verdict: approve.**

  The agent conforms to the Agents FIP: `default` clause present (line 44's requirement),
  dispatch arm shape `[*return, "name", ...patterns]`, `bundle+{*this}` returned, `return!`
  used freely in method bodies. Instance state uses peek (`<<-`) rather than consume, so
  concurrent method calls don't serialize — correct.

  I checked the parser bump specifically since it changes the grammar for the whole workspace
  and no CI has run on it: `527af8a..7588a3a` is exactly rholang-rs #96, with `finally` dropped
  in `519a929` — which matches the FIP's own "No `finally` clause" section and its reasoning
  about asynchronous branch bodies. I also checked whether `try`/`catch` becoming syntax breaks
  existing sources: the grammar does **not** globally reserve them (grammar.js:214), and the only
  `.rho` file in this repo matching is `Registry.rho`, where both hits are inside comments. No
  regression.

  ### Should fix

  1. **Instance state is keyed `@(*this, "fd")` with a public string.** The Agents FIP (line 46)
     and the parser corpus (`agent_private_state.rho`) both use `@[*private, *stateToken]` — two
     unforgeables. I checked and this is *not* currently exploitable: clients hold
     `bundle+{*this}`, and a Bundle-wrapped Par is structurally distinct from the bare `GPrivate`,
     so `@(*b, "fd")` names a different channel. But a guessable string key plus a bundled-out
     first component is strictly weaker for no gain, and `private` is supported by the parser
     today (`agent_with_private.rho`) — so this is a choice, not a limitation. Neither agent in
     the stack declares any private member, so the Private Methods FIP is currently unexercised.
  2. **The `default` arm's `match` is itself non-exhaustive.** `match args { [returnProc, ...rest] => ... }`
     has no arm for an empty `args` (a zero-argument send to the agent), so that message is
     silently dropped by a stuck match. Benign in practice — with no arguments there's no return
     channel to reply on — but it re-introduces the exact totality gap the FIP's mandatory
     `default` clause exists to close.
  3. **`default` returns a tuple where the FIP error shape expects a message string:**
     `[false, "FSERR_UNSUPPORTED", ("File agent has no method:", rest)]`.
  4. **The ack-first ABI change isn't reflected in the handler docs.** Moving ack to the front is
     *correct* — it's what `!?`/SendReceive generates per Agents FIP line 37 — but it silently
     rewrote the ABI introduced in #120/#130/#137, and every per-handler doc comment still reads
     `/// nativeOpen(path, mode) -> …`. Also worth a line in `io/mod.rs`: fileio is now the only
     system-process family that is ack-first; `verify_signature_contract`, `hash_contract`, and
     the rest are all ack-last. (I verified all 19 handlers were reordered and every `arity` still
     matches its destructuring width.)
  5. **`Cargo.lock` downgrades `getrandom` 0.4.3 → 0.3.4** as a side effect — confirm intentional.
  6. **Coordinate the parser pin.** `feature/cost-accounted-rho` pins `rholang-parser` at `527af8a`;
     this moves the workspace to `7588a3a`. Please sync so the two lines don't diverge on the grammar.
  7. No agent-level replay test exists — #137's replay tests cover raw natives only, and the agent
     layer introduces additional COMM events.

metaweta:

PR #140, opened 2026-07-22
feat(fileio): extend File agent with write/seek/tell/truncate/flush

metaweta:
## Summary

Continues Phase 2 by adding the next natural chunk of native-primitive-wrapping methods to the `File` agent (stacked on #139). Each new method follows the pattern established by `read`/`size`/`close` — `try @<pat> <- nX!?(args) { ok } catch @[code, msg] { err }`, backed by the ack-first native handlers.

## Methods added

| Method | Signature | Delegates to | Notes |
|--------|-----------|--------------|-------|
| `write(bytes)` | `[true, nWritten]` or `[false, code, msg]` | `nativeWrite(fd, bytes)` | Requires writeable fd; otherwise native returns `FSERR_PERM`. |
| `seek(offset, whence)` | `[true, newPos]` or error tuple | `nativeSeek(fd, offset, whence)` | `whence ∈ {"set","cur","end"}` |
| `tell()` | `[true, pos]` or error tuple | `nativeTell(fd)` | |
| `truncate(n)` | `[true]` or error tuple | `nativeTruncate(fd, n)` | Requires writeable fd. Native caps at 16 GiB per call. |
| `flush()` | `[true]` or error tuple | `nativeFlush(fd)` | Uses `sync_all` (data + metadata). |

## Not shipped this PR (deferred)

- **`readAt`/`writeAt`** — naive seek-then-read/write composition is racy; needs range-lock infrastructure (Phase 4).
- **`chmod`/`chown`** — need Rholang mode-string parser (planned alongside `Dir` agent).
- **`text` / `lines` / `mapReduceLines*`** — need multi-call streaming logic; separate follow-up.

## Test harness

`file_agent_env()` extended with the 5 new URN → Par mappings. `wrap` split into `wrap_with_mode(body, path, mode)` (for writeable-fd tests using `"w+"`/`"r+"`) plus a read-only convenience wrapper preserving the existing `"r"` default. All 4 previously-shipped tests continue to route through the read-only convenience form and pass unchanged.

## Tests

4 new integration tests, each exercising a chain of method calls to prove fd-table state persists across multiple invocations through the agent layer:

- **`file_agent_write_then_readback_roundtrip`** — opens `"w+"`, writes `.hexToBytes()` of "hello", seeks to 0, reads 5 bytes, asserts the returned `GByteArray` is the expected ASCII bytes. Four-hop chain: write → seek → read → sink.
- **`file_agent_seek_and_tell_report_position`** — `seek(3, "set")` then `tell()` → pos=3; then `seek(2, "cur")` from there, `tell()` → pos=5. Exercises both `"set"` and `"cur"` whence.
- **`file_agent_truncate_shrinks_file`** — opens `"r+"` on a 10-byte file, calls `truncate(3)`, then `size()` reports 3, and the OS-level file size on disk is also 3.
- **`file_agent_flush_returns_success`** — write + flush; asserts the flush success tuple.

## Sharp edge encountered

`"hello".toByteArray()` protobuf-serializes the string as a Rholang `Par` — correct for signing/hashing, wrong for on-disk content. The write test initially wrote 5 protobuf-encoding bytes instead of ASCII bytes. Fixed by using `"68656c6c6f".hexToBytes()` (same pattern as the crypto system-processes tests). Documented inline in the test.

## Test plan

- [x] `cargo test -p rholang --test fileio_file_agent_spec` — 8 pass (4 prior + 4 new)
- [x] `cargo test -p rholang --test injection_runtime_spec` — 2 pass (unchanged)
- [x] `cargo test -p rholang --test fileio_lifecycle_spec` — 3 pass (unchanged)
- [x] `cargo test -p rholang --test fileio_replay_spec` — 3 pass (unchanged)

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/140"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787297203&installation_model_id=426883&pr_number=140&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F140&signature=3bcfdafcc2a292ed907d2e8a230e2ac096b42445000c262fe46a235ee0717e68"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/rholang/src/rust/interpreter/io/agents/file.rho
+++ b/rholang/src/rust/interpreter/io/agents/file.rho
@@ -1,15 +1,18 @@
 // File agent — wraps a native fd with the FIP `File` method surface.
 //
-// This is the MINIMAL surface: `read`, `size`, `close`, `default`.
-// Extended in follow-up PRs with `write`, `seek`, `tell`, `truncate`,
-// `flush`, `readAt`, `writeAt`, `lockRange`, `chmod`, `chown`, plus
-// line-based `text` / `lines` / `mapReduceLines*`.
+// Currently ships: `read`, `write`, `seek`, `tell`, `size`,
+// `truncate`, `flush`, `close`, `default`. Extended in follow-up
+// PRs with `readAt`, `writeAt`, `lockRange` (all need range-lock
+// infrastructure, Phase 4), `chmod`, `chown` (need mode-string
+// parser), plus line-based `text` / `lines` / `mapReduceLines*`.
 //
 // This file is an `agent` block that expects the enclosing scope
 // to bind:
 //   - `File`               — the agent's constructor channel
-//   - `nRead`, `nSize`,
-//     `nClose`             — fixed-channel Pars for the native
+//   - `nRead`, `nWrite`,
+//     `nSeek`, `nTell`,
+//     `nSize`, `nTruncate`,
+//     `nFlush`, `nClose`   — fixed-channel Pars for the native
 //                            primitives, obtained via `NormalizerEnv`
 //                            injection at compile time (see
 //                            `interpreter::io::injections`).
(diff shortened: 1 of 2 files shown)


dylon (approved):
claude found no blockers but did make a few recommendations:

  ## Review — PR #140 (File positional methods)

  **Verdict: approve.** Clean and well tested — the truncate test asserts on-disk size rather
  than just the reply, and the write→seek→read round-trip genuinely exercises fd-table state
  across four method invocations. Signatures match FIP §Positional.

  ### Hard gate before `Fs.openFile` becomes user-reachable

  1. **The positional surface ships without any of its concurrency control.** FIP §Positional
     (lines 406–432) specifies range locks (multiple readers / single writer), a line-vs-byte
     mutex, `{"wait": true}`, and explicit `lockRange(offset, length, mode)` returning a
     `bundle+` token with `release()` — with `FSERR_BUSY` on conflict. None of it exists, and
     `FSERR_BUSY` has zero emit sites in the whole stack. The per-`FileHandle`
     `tokio::sync::Mutex` serializes individual syscalls but not `seek`+`read` *pairs*, and the
     agent reads its fd with a non-consuming peek, so two Rholang processes sharing a `File`
     agent interleave into torn reads with no error surfaced. The tests avoid this only by
     chaining sequentially. Fine as a staging point — but this must land before user exposure.

  ### Should fix

  2. **The stored open mode is never consulted.** `FileHandle.mode` is recorded and documented as
     "Kept for later use in enforcing per-method mode requirements (e.g. `chmod` needs write
     mode)", but no method reads it. So `write` on an `"r"` fd fails at the OS layer with `EBADF`
     → `FSERR_IO` rather than the FIP's `FSERR_PERM`/`FSERR_UNSUPPORTED`, and the FIP's
     mode-as-attenuation story is OS-enforced only. FIP line 612 requires `chmod`/`chown` to
     *check* write-capable mode.
  3. **`seek` + `write` on an append-mode fd silently ignores the seek.** FIP F8 says `write`
     "Writes at the current position"; the mode table says `"a"` means "writes go to the end".
     The implementation follows the mode table (which I think is the right reading) — worth one
     line of doc so callers aren't surprised.
  4. Argument type-checking: `truncate("abc")` / `seek(0, 42)` abort the deploy rather than
     returning `[false, FSERR_BAD_ARG, …]` — see #120 item 2.

metaweta:

PR #141, opened 2026-07-22
feat(fileio): Dir agent (entries/stat/exists)

metaweta:
## Summary

Continues Phase 2 with the first cut of the `Dir` agent class: three query methods (`entries`, `stat`, `exists`) plus the shared `default` reply idiom. Establishes the "quarantine-per-call" pattern for path-taking methods that later PRs will follow for the mutating methods.

## What's shipped

`rholang/src/rust/interpreter/io/agents/dir.rho` — `agent Dir { constructor / method entries / method stat / method exists / default }`. Rust plumbing: `pub const DIR_AGENT_SRC` in `agents.rs`.

### Method surface

| Method | Signature | Delegates to | Quarantine? |
|--------|-----------|--------------|-------------|
| `entries()` | `[true, entryList]` or error | `nativeEntries(root)` | No (root itself) |
| `stat(relPath)` | `[true, statRecord]` or error | `nativeStat(canonPath)` | Yes |
| `exists(relPath)` | `[true, bool]` or error | `nativeExists(canonPath)` | Yes |
| `default(...@args)` | `[false, "FSERR_UNSUPPORTED", ...]` | — | — |

### Per-instance state

Constructor takes an already-canonicalized absolute root path (String) as `@rootPath`, stashes it on `@(*this, "root")`. Methods peek non-consumingly, same pattern as `File`'s fd stash.

### Quarantine pattern

`entries()` skips quarantine because it operates on the root itself, which was canonicalized + escape-checked at Dir construction time (per `nativeQuarantine`'s documented precondition). The `Fs` agent that constructs Dir instances is responsible for that precondition.

Path-taking methods (`stat`/`exists`) route through `nativeQuarantine(root, relPath)` first. If quarantine fails (typically `FSERR_QUARANTINE` for a `..` escape or symlink traversal), the tuple is forwarded verbatim — the underlying native is not called.

## Deferred to follow-up PRs

- **Mutating methods**: `removeFile`, `removeDir`, `rename`, `copyFile` — natural extension of the quarantine pattern.
- **Agent-composition methods**: `openFile`, `openDir` — need to build `File`/`Dir` instances from the returned fd/rootpath and hand back a bundled name.
- **`chmod`/`chown`** — need the mode-string parser.

## Tests

Six integration tests in `rholang/tests/fileio_dir_agent_spec.rs`:

- **`dir_agent_entries_lists_root_children`** — `entries()` on a dir with `alpha.txt`+`beta.txt` returns both names in the sorted list.
- **`dir_agent_stat_reports_child_metadata`** — `stat("target.txt")` returns a stat record with name and size fields.
- **`dir_agent_exists_true_for_present_child`** — `exists("here.txt")` returns `[true, true]`.
- **`dir_agent_exists_false_for_absent_child`** — `exists("nope.txt")` returns `[true, false]` (outer success distinguishes "query succeeded, answer is no" from "query failed").
- **`dir_agent_stat_rejects_parent_escape`** — `stat("../outside.txt")` is caught by `nativeQuarantine`, returns `[false, "FSERR_QUARANTINE", ...]`. Also creates the target-of-escape file to prove the rejection is on principle, not because the target doesn't exist.
- **`dir_agent_unknown_method_returns_unsupported`** — same default-arm reply idiom as `File`.

Temp dirs are canonicalized via `Path::canonicalize` before constructor invocation so macOS's `/tmp` → `/private/tmp` translation doesn't cause `nativeQuarantine` to see two different absolute forms.

## Test plan

- [x] `cargo test -p rholang --test fileio_dir_agent_spec` — 6 pass
- [x] `cargo test -p rholang --test fileio_file_agent_spec` — 8 pass (unchanged)

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/141"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787301068&installation_model_id=426883&pr_number=141&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F141&signature=e59edee04bbb9e0833dba26ee235bb19e9dd7f7c0be636629f89403eb783b6fa"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/rholang/src/rust/interpreter/io/agents.rs
+++ b/rholang/src/rust/interpreter/io/agents.rs
@@ -17,9 +17,16 @@
 //!
 //! Native URNs each `.rho` file expects:
 //!
-//! - `file.rho` — `nRead`, `nSize`, `nClose` (extended as the
-//!   `File` surface grows in follow-up PRs).
+//! - `file.rho` — `nRead`, `nWrite`, `nSeek`, `nTell`, `nSize`,
+//!   `nTruncate`, `nFlush`, `nClose`.
+//! - `dir.rho`  — `nQuarantine`, `nEntries`, `nStat`, `nExists`.
 
 /// Source of the `File` agent block. Expects `File`, `nRead`,
-/// `nSize`, `nClose` to be bound in the enclosing scope.
+/// `nWrite`, `nSeek`, `nTell`, `nSize`, `nTruncate`, `nFlush`,
+/// `nClose` to be bound in the enclosing scope.
 pub const FILE_AGENT_SRC: &str = include_str!("agents/file.rho");
+
+/// Source of the `Dir` agent block. Expects `Dir`, `nQuarantine`,
+/// `nEntries`, `nStat`, `nExists` to be bound in the enclosing
+/// scope.
+pub const DIR_AGENT_SRC: &str = include_str!("agents/dir.rho");
--- a/rholang/src/rust/interpreter/io/agents.rs
+++ b/rholang/src/rust/interpreter/io/agents.rs
@@ -19,7 +19,8 @@
 //!
 //! - `file.rho` — `nRead`, `nWrite`, `nSeek`, `nTell`, `nSize`,
 //!   `nTruncate`, `nFlush`, `nClose`.
-//! - `dir.rho`  — `nQuarantine`, `nEntries`, `nStat`, `nExists`.
+//! - `dir.rho`  — `nQuarantine`, `nEntries`, `nStat`, `nExists`,
+//!   `nRemoveFile`, `nRemoveDir`, `nRename`, `nCopyFile`.
 
 /// Source of the `File` agent block. Expects `File`, `nRead`,
 /// `nWrite`, `nSeek`, `nTell`, `nSize`, `nTruncate`, `nFlush`,
@@ -27,6 +28,6 @@
 pub const FILE_AGENT_SRC: &str = include_str!("agents/file.rho");
 
 /// Source of the `Dir` agent block. Expects `Dir`, `nQuarantine`,
-/// `nEntries`, `nStat`, `nExists` to be bound in the enclosing
-/// scope.
+/// `nEntries`, `nStat`, `nExists`, `nRemoveFile`, `nRemoveDir`,
+/// `nRename`, `nCopyFile` to be bound in the enclosing scope.
 pub const DIR_AGENT_SRC: &str = include_str!("agents/dir.rho");
--- a/rholang/src/rust/interpreter/io/agents/dir.rho
+++ b/rholang/src/rust/interpreter/io/agents/dir.rho
@@ -1,17 +1,20 @@
 // Dir agent — wraps a directory root with the FIP `Dir` method surface.
 //
-// Currently ships: `entries`, `stat`, `exists`, `default`. Extended
-// in follow-up PRs with `openFile`, `openDir` (need File-agent
-// composition), `removeFile`, `removeDir`, `rename`, `copyFile`
-// (path-mutating natives), `chmod`, `chown` (need mode-string
-// parser).
+// Currently ships: `entries`, `stat`, `exists`, `removeFile`,
+// `removeDir`, `rename`, `copyFile`, `default`. Extended in
+// follow-up PRs with `openFile`, `openDir` (need File-agent
+// composition), `chmod`, `chown` (need mode-string parser).
 //
 // This file is an `agent` block that expects the enclosing scope
 // to bind:
 //   - `Dir`                — the agent's constructor channel
 //   - `nQuarantine`,
 //     `nEntries`, `nStat`,
-//     `nExists`            — fixed-channel Pars for the native
+//     `nExists`,
+//     `nRemoveFile`,
+//     `nRemoveDir`,
+//     `nRename`,
+//     `nCopyFile`          — fixed-channel Pars for the native
 //                            primitives, obtained via `NormalizerEnv`
 //                            injection at compile time (see
 //                            `interpreter::io::injections`).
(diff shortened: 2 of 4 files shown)


metaweta:
## Update — HIGH severity fix from security review

Security review of the mutating-methods commit surfaced a HIGH: `removeDir("", true)` (also `"."`, `"./"`, `"/"`) resolved through `canonicalize_and_quarantine` to the sandbox root itself, letting a caller who legitimately held only a `Dir` handle wipe the entire sandbox via `tokio::fs::remove_dir_all(root)`.

Fix landed as commit [`047338d6`](https://github.com/F1R3FLY-io/f1r3node-rust/commit/047338d6).

### The bug

`canonicalize_and_quarantine(root, "")`:
1. `trim_start_matches('/').trim_start_matches("./")` → `stripped = ""`.
2. `root.join("")` → `root` unchanged.
3. `canonicalize` succeeds (root exists).
4. `check_prefix` — `root.starts_with(root) == true` → returns `root` as the "canonical" path.

Then `Dir.removeDir("", true)` dispatches `nRemoveDir!?(root, true)` → `tokio::fs::remove_dir_all(root)`. Sandbox gone.

Not exploitable today (no `Fs.openDir` yet), but the pattern propagates to every future path-taking agent method, so the fix belongs at the quarantine layer.

### The fix

`io/path.rs::canonicalize_and_quarantine`: after stripping the leading `/`s and `./`s, if the remainder is empty or `Path::new(stripped).components()` yields only `Component::CurDir`, return `QuarantineError::BadArg("rel path resolves to the root itself; use root-scoped methods instead")`.

Chose `BAD_ARG` over `QUARANTINE`: the caller supplied a malformed argument, not an escape attempt. Callers who legitimately want to operate on the root use root-scoped methods (`entries()`, and future methods that take no relPath).

Chose to fix at the quarantine layer (not per-method in `dir.rho`): every current and future path-taking native (`stat`, `exists`, `removeFile`, `removeDir`, `rename`, `copyFile`, and eventually `chmod`, `chown`) now inherits the check without requiring per-method discipline.

### Tests

**5 new `io::path::tests`**:
- `empty_rel_is_bad_arg`
- `dot_only_rel_is_bad_arg` (covers `.`, `./`, `./.`, `././.`, `././`)
- `slash_only_rel_is_bad_arg` (covers `/`, `//`, `///`)
- `leading_slash_on_real_name_still_works` — regression guard that `/a` still resolves inside root.
- `leading_dot_slash_on_real_name_still_works` — same for `./a`.

**1 new integration test** `dir_agent_remove_dir_empty_relpath_is_rejected`: exercises all 4 spelling variants (`""`, `"."`, `"./"`, `"/"`), asserts each returns `[false, "FSERR_BAD_ARG", ...]` AND the sandbox root + canary file both survive on disk.

**Regression-verified in-workflow**: temporarily commented out the fix; the integration test failed with `[true]` (sandbox WAS wiped, exactly as the security review predicted). Restored fix; all 14 dir tests pass.

### Verified

- `cargo test -p rholang --lib io::` — 42 pass (37 prior + 5 new path tests)
- `cargo test -p rholang --test fileio_dir_agent_spec` — 14 pass (13 prior + 1 new regression guard)
- Other fileio suites (file, injection, lifecycle, replay) unchanged
- Full pre-push (11 slices) clean

Other security review findings (`entries()` trusts stashed root, `copyFile` byte-count "oracle") evaluated as accepted per FIP design or non-issues.

dylon (changes requested):
claude found 2 blockers and made a few recommendations:

  ## Review — PR #141 (Dir agent queries + mutations)

  **Verdict: blocked.**

  Quarantine-before-dispatch on every path-taking method is the right structure, and checking
  *both* `rename` arguments independently before touching the filesystem is exactly right. The
  empty-relpath fix shows the security thinking is active. It just doesn't go far enough.

  ### Blocking

  1. **The root-wipe fix is bypassable.** The new guard keys on the *spelling* of the relpath
     (empty, or all-`CurDir` components), not on the resolved path. `removeDir("sub/..", true)`
     has components `[Normal("sub"), ParentDir]` — not empty, not all-`CurDir` — so the guard
     doesn't fire; `joined = root/sub/..` canonicalizes (fast path, `sub` exists) to `root`;
     `check_prefix(root, root)` is true since `starts_with` is reflexive; and
     `tokio::fs::remove_dir_all(root)` deletes the entire sandbox. Same for `"a/../"`,
     `"./sub/.."`, `"x/y/../.."`, and for `rename`/`copyFile` arguments.
     `dir_agent_remove_dir_empty_relpath_is_rejected` covers only `["", ".", "./", "/"]`, so it
     passes. Please reject on the **resolved** path in `check_prefix`
     (`if candidate == root { return Err(BadArg(..)) }`), keeping the input guard as a cheap
     short-circuit, and add `"sub/.."` to the matrix.

     While you're in there: `rel.trim_start_matches('/').trim_start_matches("./")` applied in
     that fixed order turns `".//foo"` into `"/foo"`, and `root.join("/foo")` *replaces* the
     whole path with `/foo`. It fails closed at `check_prefix`, so it's not a hole — but looping
     the trim (or rejecting any `RootDir` component) would be less fragile.

  2. **The `Dir` constructor takes no mode, so FIP dir mode `"r"` is unrepresentable.** FIP lines
     249–256 make `openDir` mode the primary attenuation mechanism: `"r"` permits only
     `entries`/`exists`/`stat`/`openFile("r")`; `"rw"` additionally permits `removeFile`,
     `removeDir`, `rename`, `copyFile`, `chmod`, `chown`, and `openDir(path: true)`. Every `Dir`
     instance here carries the full mutation surface. #131 already parses and stores dir modes —
     nothing on the Rholang side receives them. Not exploitable yet (no `Fs` agent), but far
     cheaper to fix now than after `Fs.openDir` is written against this constructor:
     `constructor(@rootPath, @mode)` and gate the mutators with `FSERR_UNSUPPORTED`.

  ### Should fix

  3. **The check-to-use TOCTOU is now scheduler-sized.** `nQuarantine!?(root, rel)` and the
     subsequent `nRemoveFile!?(canonPath)` are separate messages, so arbitrarily many reductions
     can interleave between the check and the syscall — a microsecond window widened without
     bound. `path.rs` already disclaims TOCTOU against a concurrent external writer and that
     trade-off is reasonable, but this is a self-inflicted widening. Every path-taking native
     already receives a path; passing `(root, rel)` and quarantining *inside* the same Rust
     handler collapses the window to a single call. Keep `nativeQuarantine` for standalone checks.
  4. **`kind: "symlink"` is forwarded to Rholang unmapped**, contradicting FIP line 456.
  5. **The consensus-vs-oracular field filtering has no home.** `stat.rs`'s docstring states that
     stripping `mtime`/`ctime`/`atime`/`owner`/`group` in consensus mode "is the agent layer's
     job, not the native primitive's" — and the agent layer as shipped does no stripping. That
     cross-PR contract is currently unfulfilled by both sides.
  6. **Several test assertions can't distinguish success from failure.** They `contains`-match on
     `format!("{p:?}")` of the whole sink: `dir_agent_exists_false_for_absent_child` asserts only
     that both `GBool(true)` and `GBool(false)` appear somewhere, which an error tuple would also
     satisfy; `dir_agent_remove_dir_nonrecursive_on_nonempty_errors` asserts only `GBool(false)`.
     Given these are the tests establishing the quarantine and error-shape claims, asserting on
     tuple *structure* would be worth the effort.
  7. Argument type-checking: `removeDir("x", "yes")` aborts the deploy rather than returning
     `[false, FSERR_BAD_ARG, …]` — see #120 item 2. `Dir` forwards `recursive` unchecked.

metaweta:

PR #142, opened 2026-07-22
Issue 43 investigation and fix

### LINKED ISSUE #43: Per-deploy Rholang execution scales backwards with CPU
> Migrated from [F1R3FLY-io/f1r3node#493](https://github.com/F1R3FLY-io/f1r3node/issues/493) as part of the deprecation of that repository. Original author: @elmiringos. Comments on the legacy issue are not mirrored here — see the linked legacy issue for the full thread.

---

### What did you expect to see?
Per-deploy block creation and replay times that **stay flat or decrease** as the container's CPU limit grows. A `TransferTerm` deploy is a small, single-shard; doubling CPU should at worst leave it unchanged.

Concretely: a 1-deploy block replayed on a 2-CPU node in 0.8 s should not take longer to replay on an 8-CPU node of the same hardware family.

### What did you see instead?
Replay (and propose-core) time **grows with CPU limit**:

| k8s CPU limit (cgroup quota) | 1-deploy block replay | 10-deploy block replay |
|---|---|---|
| 2 CPU (`minimum`) | 0.75–0.82 s | not measured (test ramp stops earlier) |
| 8 CPU (`mid`) | 2.1–3.7 s (3–5× slower) | 14–26 s |

The downstream effect: with three validators concurrently proposing at the same height under heartbeat-off conditions, mid+ profiles can't accumulate.

FT returns zero finalizations even though the DAG is being populated (43 distinct blocks observed, 0 finalized). With a 15-minute window, the chain *can* converge — at `BestCap = 12, BestTPS = 0.20 deploys/sec` — which is still ~5× below the 1.0 deploys/sec/profile we'd extrapolate from the minimum-profile baseline.

### Steps to reproduce the bug
1. Deploy a 3-validator shard on Kubernetes with explicit cgroup CPU limits — e.g. one shard at 2 CPU/pod, another at 8 CPU/pod.
2. Disable heartbeat on both shards (`heartbeat-enabled = false` in `rnode.conf`).
3. From outside the validator pods, run a tight loop on each validator in parallel:
   - flood N transfer deploys via the public HTTP `/api/deploy` endpoint
   - call `POST /api/propose` on the admin port (40405)
   - wait for the validator's tip to advance via `/api/blocks/1`
   - repeat for ~3-15 minutes
4. Compare **per-deploy** block replay time across the two shards.

### What version of F1R3FLY Node are you using?
- **Source**: f1r3node Rust 0.4.13

### What is your environment?
- **OS / kernel**: Linux (k3s host)
- **Container runtime**: containerd via k3s
- **Orchestrator**: k3s single-node cluster
- **CPU limits**: tested across cgroup CFS quotas of 2, 4, 8, 16, 32 CPUs.

9Neechan:
<h2 data-heading="Summary">Summary</h2>
<p>Fixes #43 — per-deploy Rholang execution cost growing with block size (and, less directly, with CPU limit) during block replay.</p>
<ul>
<li><strong>Root cause</strong>: <code>replay_runtime.rs::run_user_deploy</code> calls <code>create_soft_checkpoint()</code> unconditionally before every user deploy (for failure rollback). <code>HotStore::snapshot()</code> deep-cloned all five <code>DashMap</code>s (data, continuations, installed<em>continuations, joins, installed_joins) into plain <code>HashMap</code>s on every call — <strong>O(total store size)</strong>, not O(deploy work). Since a block's store accumulates entries as deploys run, each subsequent deploy's checkpoint paid for a snapshot of the _whole accumulated state so far</em>, not just its own work. This is what made per-deploy <code>evaluate()</code> cost climb with block size.</li>
<li><strong>Fix</strong>: back <code>HotStore</code>'s five state maps with 256 independent <code>imbl::HashMap</code> shards behind per-shard <code>RwLock</code>s (<code>ShardedMap</code> in <code>hot_store.rs</code>) instead of <code>DashMap</code>. <code>snapshot()</code> now reads each shard's persistent-map value directly — O(1) refcount-bump clone per shard — so total snapshot cost is <strong>O(NUM_SHARDS)</strong>, independent of store size. Mirrors the striped-lock scheme #72 used in the propose path.</li>
<li>Also ports the same fix to <code>ReplayRSpace</code>'s <code>phase_a_locks</code>/<code>phase_b_locks</code>, which #72 fixed in the propose-path <code>RSpace</code> but never ported to replay — still a growing, unbounded <code>DashMap&#x3C;u64, Arc&#x3C;Mutex>></code> before this PR.</li>
</ul>
<h2 data-heading="Investigation">Investigation</h2>
<p>Found via <code>mntd</code>/<code>parallelism</code> cap-sweep diagnostics on a 3-validator testbed: a single deploy's <code>evaluate()</code> time grew from 57ms at block-size 1 to 372ms at block-size 100, while rspace++ mutex hold time stayed sub-millisecond throughout — ruling out ordinary lock contention and pointing at something scaling with accumulated store size instead of per-deploy work.</p>
<h2 data-heading="Validated on a real cluster">Validated on a real cluster</h2>

metric | before | after
-- | -- | --
mntd best cap | 1 | 10
evaluate() avg @ cap=100 | 372 ms | 49.7 ms (7.5×)
evaluate() avg @ cap=500 | n/a (never finalized) | 43.8 ms — flat
CPU peak @ N=250 | 74% | 100%
throughput @ N=250 | 4.10 d/s | 5.28 d/s (+29%)
1-deploy-block replay, mid (8 CPU) | 2.1–3.7 s | 0.39 s (~6-9×)

<p>Old super-linear (~n^1.76) growth with block size is gone on every CPU profile tested (2/8/16).</p>
<h2 data-heading="Regression test">Regression test</h2>
<p><code>concurrent_rspace_test.rs::soft_checkpoint_cost_does_not_scale_with_accumulated_store_size</code> — 100× more store state: ~31× cost before, ~1-3× after. Full <code>rspace_plus_plus</code> suite (257 tests) + workspace build clean.</p>
<h2 data-heading="Known, separately-tracked residuals (not blocking this PR)">Known, separately-tracked residuals (not blocking this PR)</h2>
<ul>
<li>A residual 2-4× replay-cost gap between low-CPU and high-CPU pods persists at every block size — likely Tokio's default worker-thread count reading cgroup CPU limit rather than workload size; untested before this fix since the O(store-size) cost masked it.</li>
<li>Finalization latency has a ~25-45s floor independent of replay cost — looks like a Casper fault-tolerance-threshold effect, unrelated to Rholang execution.</li>
</ul>

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/142"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787320722&installation_model_id=426883&pr_number=142&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F142&signature=e30936800ddb8523b5cf3ae4cf6d3d57dcc5d696472c13cd0183cf478567d3d0"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


9Neechan:
Should I reopen the PR from a non-fork repo to pass cicd?
@jeffrey-l-turner @spreston8

spreston8:
> Should I reopen the PR from a non-fork repo to pass cicd? @jeffrey-l-turner @spreston8

I think so. The CI/CD system is undergoing a lot of changes right now.

spreston8:
@jeffrey-l-turner — while digging into why this PR's integration tests never ran, I found a second blocker beyond the concurrency eviction #161 fixed, and it needs a one-line-per-site change that would fit naturally in one of your open CI/CD PRs (#162 or successor).

Both attempts of this PR's approved heavy run failed in seconds at `Clone Repository`:

> `##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. … To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.`

Recent `actions/checkout` releases added this guard, and the pipeline pins the moving `@v4` tag — so the fork path broke upstream with no change in this repo. Consequence: even with #161 merged and a maintainer approving the run, **every fork PR's heavy pipeline dies at clone**. This is why integration tests never ran here.

The fix is `allow-unsafe-pr-checkout: true` on the code-under-test checkout sites in `_integration-pipeline.yml` (the `inputs.checkout_repository`/`checkout_ref` checkouts — Build Docker Image plus the downstream jobs that re-fetch the same ref). The opt-in is sound for this pipeline's architecture, and stronger post-#161 than it ever was:

- the workflow file executes from the base branch — a fork cannot alter what runs;
- `await_approval` now gates the entire pipeline, so fork code is fetched only after a maintainer has reviewed the diff;
- the jobs that fetch fork code carry no secrets (the launch job is the only secret-bearing job and checks out only the pinned system-integration ref), and all these checkouts use `persist-credentials: false`.

Could you fold this into one of your open CI PRs? I can PR it separately on top if you'd rather keep scopes clean — but without it the fork lane stays broken for every future outside contributor, and @9Neechan's reopen-from-a-branch workaround becomes the permanent path instead of a stopgap.

jeffrey-l-turner:
> @jeffrey-l-turner — while digging into why this PR's integration tests never ran, I found a second blocker beyond the concurrency eviction #161 fixed, and it needs a one-line-per-site change that would fit naturally in one of your open CI/CD PRs (#162 or successor).
> 
> Both attempts of this PR's approved heavy run failed in seconds at `Clone Repository`:
> 
> > `##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow. This workflow runs with the base repository's GITHUB_TOKEN, secrets, default-branch cache scope, and runner access. … To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.`
> 
> Recent `actions/checkout` releases added this guard, and the pipeline pins the moving `@v4` tag — so the fork path broke upstream with no change in this repo. Consequence: even with #161 merged and a maintainer approving the run, **every fork PR's heavy pipeline dies at clone**. This is why integration tests never ran here.
> 
> The fix is `allow-unsafe-pr-checkout: true` on the code-under-test checkout sites in `_integration-pipeline.yml` (the `inputs.checkout_repository`/`checkout_ref` checkouts — Build Docker Image plus the downstream jobs that re-fetch the same ref). The opt-in is sound for this pipeline's architecture, and stronger post-#161 than it ever was:
> 
> * the workflow file executes from the base branch — a fork cannot alter what runs;
> * `await_approval` now gates the entire pipeline, so fork code is fetched only after a maintainer has reviewed the diff;
> * the jobs that fetch fork code carry no secrets (the launch job is the only secret-bearing job and checks out only the pinned system-integration ref), and all these checkouts use `persist-credentials: false`.
> 
> Could you fold this into one of your open CI PRs? I can PR it separately on top if you'd rather keep scopes clean — but without it the fork lane stays broken for every future outside contributor, and @9Neechan's reopen-from-a-branch workaround becomes the permanent path instead of a stopgap.
> 

see: https://github.com/F1R3FLY-io/f1r3node-rust/pull/171

spreston8:
Reviewed at `01981de1` against `dev`, reading the live files plus some measurements on the branch. No correctness defect found — the requested changes below.

**Medium**

1. **Every mutation deep-clones the channel's whole value vector.** All six `with_entry` sites read-clone-modify-reinsert (`hot_store.rs:667`): `map.get(channel).cloned()` → mutate → `map.insert(..)`, where `dev` did `occupied.get_mut().insert(0, d)` in place. `.cloned()` deep-clones the entire `Vec<Datum<A>>`, and in production `A = ListParWithRandom`, so every `put_datum`/`remove_datum` deep-clones a Rholang term per element — a channel holding *k* un-consumed datums makes produce cost O(k²) where `dev` was O(k). The TransferTerm benchmark can't see it (distinct vault channels, k≈1). `put_join`/`remove_join` same shape; `continuations` escapes via #72's `Arc`. Fix: `imbl::HashMap`'s `get_mut` (map.rs:723) or its DashMap-shaped `entry()` (map.rs:851, 1651) map ~1:1 onto the `dev` code and mutate in place unless the shard is shared with a live snapshot; or `Vec<Arc<Datum<A>>>` per #72's precedent.

2. **The shard-count invariant is guarded by an assertion that never runs.** `from_shards` (`:113`) and `restore_shards` (`:184`) use `debug_assert_eq!`, and with no `[profile.release]` in `Cargo.toml` plus release builds everywhere (`node/Dockerfile:80`, `Justfile:27`) both are compiled out of every build produced. Unreachable today so not blocking, but the failure modes are bad: short vector → bare index-out-of-bounds at some unrelated produce; long vector → shards ≥ `NUM_SHARDS` silently unaddressed; and `restore_shards`' `.zip()` silently partial-restores. Prefer `Box<[imbl::HashMap<K, V>; NUM_SHARDS]>` so length is a type guarantee and both assertions can go (`Box` because a bare array puts ~40 KB inline in every `SoftCheckpoint`), or make the fields private — nothing outside `hot_store.rs` touches them after this PR.

3. **Un-ignore the regression test and fix what it measures.** It has real teeth — at `c8c97569` it reports 246–285x against its own `MAX_RATIO = 20.0` — but `#[ignore]` keeps it out of CI, and it times 50 calls on a freshly-built store, so its ratio is heap first-touch rather than snapshot cost. Consequence: at block-realistic sizes it fails on correct code (10k/100k prefills → **20.8x** against the 20x bound). Timing three identical passes shows why: 10k = 50.2/17.9/17.9 µs, 100k = 1250.1/17.8/18.0 µs. Discarding one pass before measuring lets the bound tighten to ~3x *and* the `#[ignore]` be dropped; or drop timing for `imbl::HashMap::ptr_eq` (map.rs:216) asserting each snapshot shard shares a root with its live shard, which is the real invariant and is CI-safe.

4. **Extract the duplicated locking machinery.** `rspace.rs` and `replay_rspace.rs` now each carry verbatim copies of `channel_hash`, `acquire_locks`, `consume_lock`, `produce_lock`, `new_striped_locks`, `HeldLock`, `ChannelLockGuard` — ~56 lines. #72 fixed the play path and left replay behind, which is why this issue existed; shipping the fix by copying again preserves the precondition for a third divergence. One module, or onto the `SpaceMatcher` trait both types already implement. `ReportingRspace` delegates, so it's a two-caller change.

**Low**

5. Keys are hashed twice per operation: `shard_of` (`:61`) runs a full `DefaultHasher` pass, then `imbl` hashes the same key again with its own `RandomState`. Keys are `Par`/`Vec<Par>`, so both walk the term; `DashMap` computed one hash and reused it.

6. The hand-written `Default` is load-bearing — a derived one yields empty vectors and panics on first lookup — but nothing says so and no test pins `default().data.len() == NUM_SHARDS`.

7. Stripe count is a bare `(0..256)` in both files (`rspace.rs:130`, `replay_rspace.rs:134`) where the parallel shard count is a named `NUM_SHARDS` — companion to 4.

8. `#[allow(dead_code)]` on `impl ShardedMap` (`:98`) is covering two methods with no call sites: `new()` (construction all goes through `from_shards`) and `remove()` (removal happens inside `with_entry` closures). Please drop both.

9. The five proptests that actually protect this change — snapshot isolation, one per map — are named `remove_join_should_create_a_deep_copy_of_the_*` and none calls `remove_join`. They're what shows the `Arc`-sharing switch is safe, and the names hide it. This PR already touched all five lines.

10. Comment cleanups, one pass: the `ShardedMap` doc comment cites `soft_checkpoint_cost_scales_with_accumulated_store_size_not_deploy_size` (`:77`), the test's pre-rename name, which resolves to nothing at head; the adjacent "~31x" and the test's "~3x" measure 246–285x and 7.3–7.9x here, so as the stated evidence for the design they want re-measuring or making qualitative; and the pointers a reader can't resolve are worth dropping — `asi-chain-testbed issues/01-…, Update 2026-07-21c` is in another repo and "PR #72 fix #3/#4/#5" numbering exists nowhere in the tree. The mechanism explanations are good and worth keeping — it's the provenance that belongs in the PR description.

PR #144, opened 2026-07-22
feat(fileio): Dir.openFile + Dir.openDir (agent composition)

metaweta:
## Summary

Extends the `Dir` agent with the two remaining Phase-2 methods that compose one agent-class instance out of another:

- **`Dir.openFile(relPath, mode)`** — quarantines the relPath, opens a fd via `nativeOpen`, constructs a `File` agent around the fd, and returns `[true, fileBundle]`.
- **`Dir.openDir(relPath)`** — quarantines the relPath, verifies it names a directory via `nativeStat`, and constructs a nested `Dir` agent rooted at the canonical path. Returns `[true, dirBundle]`.

This is the first agent-composition point in the fileio stack. `Fs.openFile` in Phase 3 will follow the same shape but operate on absolute paths.

## The composition pattern

```rholang
try @[file] <- dir!openFile("notes.txt", "r") {
  try @[bytes] <- file!?("read", 100) { ... }
  catch @[code, msg] { ... }
}
catch @[code, msg] { ... }
```

`@[file] <- ...` binds `file` as a Process; `file!method(...)` uses it as a channel via Rholang's implicit Process→Name η-quoting.

## Test wrapper unification

`dir_agent_env()` and `wrap()` now bind BOTH the `Dir` and `File` constructor channels plus the union of their native URN sets, and inject `DIR_AGENT_SRC | FILE_AGENT_SRC` in parallel. The normalizer catches unbound `File` references in `Dir.openFile` at compile time, so both agent blocks need to be in scope regardless of whether `openFile` is actually invoked.

The extra bindings are inert for the 14 existing Dir tests — all still pass.

## Test plan

- [x] `cargo test -p rholang --test fileio_dir_agent_spec` — 20 pass (14 prior + 6 new)
- [x] `cargo test -p rholang --test fileio_file_agent_spec` — 8 pass
- [x] `cargo test -p rholang --lib io::` — 42 pass
- [x] `cargo fmt -p rholang --check` — clean
- [x] Pre-push (11 slices) — clean

## New tests (6)

Success cases exercise the returned bundle end-to-end (openFile→read, openDir→entries). Escape cases (`../secret.txt`, `../outside_dir`) return `FSERR_QUARANTINE`. Edge cases: `openFile` on a missing file returns `FSERR_NOT_FOUND` from the underlying `nativeOpen`; `openDir` on a regular file returns `FSERR_BAD_ARG`.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/144"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787334625&installation_model_id=426883&pr_number=144&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F144&signature=6b7633afa5a3671df5c008ab5220ce382eb8322627108907d8f8e7db03dc1e21"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/rholang/src/rust/interpreter/io/agents.rs
+++ b/rholang/src/rust/interpreter/io/agents.rs
@@ -20,14 +20,23 @@
 //! - `file.rho` — `nRead`, `nWrite`, `nSeek`, `nTell`, `nSize`,
 //!   `nTruncate`, `nFlush`, `nClose`.
 //! - `dir.rho`  — `nQuarantine`, `nEntries`, `nStat`, `nExists`,
-//!   `nRemoveFile`, `nRemoveDir`, `nRename`, `nCopyFile`.
+//!   `nRemoveFile`, `nRemoveDir`, `nRename`, `nCopyFile`, `nOpen`,
+//!   plus the `File` constructor channel (`openFile` composes
+//!   with the File agent).
 
 /// Source of the `File` agent block. Expects `File`, `nRead`,
 /// `nWrite`, `nSeek`, `nTell`, `nSize`, `nTruncate`, `nFlush`,
 /// `nClose` to be bound in the enclosing scope.
 pub const FILE_AGENT_SRC: &str = include_str!("agents/file.rho");
 
-/// Source of the `Dir` agent block. Expects `Dir`, `nQuarantine`,
-/// `nEntries`, `nStat`, `nExists`, `nRemoveFile`, `nRemoveDir`,
-/// `nRename`, `nCopyFile` to be bound in the enclosing scope.
+/// Source of the `Dir` agent block. Expects `Dir`, `File`,
+/// `nQuarantine`, `nEntries`, `nStat`, `nExists`, `nRemoveFile`,
+/// `nRemoveDir`, `nRename`, `nCopyFile`, `nOpen` to be bound in
+/// the enclosing scope.
+///
+/// The `File` binding is required because `Dir.openFile`
+/// constructs a `File` instance around the fd returned by
+/// `nativeOpen`. In practice this means the genesis-installed
+/// `Fs` agent's outer `new`-scope must bind BOTH agent
+/// classes and their combined native URN set.
 pub const DIR_AGENT_SRC: &str = include_str!("agents/dir.rho");
--- a/rholang/src/rust/interpreter/io/agents/dir.rho
+++ b/rholang/src/rust/interpreter/io/agents/dir.rho
@@ -273,7 +273,7 @@ agent Dir {
               }
             }
             _ => {
-              return!([false, "FSERR_BAD_ARG", ("path is not a directory:", relPath)])
+              return!([false, "FSERR_BAD_ARG", "path is not a directory: " ++ relPath])
             }
           }
         }
--- a/rholang/src/rust/interpreter/io/agents/dir.rho
+++ b/rholang/src/rust/interpreter/io/agents/dir.rho
@@ -1,20 +1,23 @@
 // Dir agent — wraps a directory root with the FIP `Dir` method surface.
 //
 // Currently ships: `entries`, `stat`, `exists`, `removeFile`,
-// `removeDir`, `rename`, `copyFile`, `default`. Extended in
-// follow-up PRs with `openFile`, `openDir` (need File-agent
-// composition), `chmod`, `chown` (need mode-string parser).
+// `removeDir`, `rename`, `copyFile`, `openFile`, `openDir`,
+// `default`. Extended in follow-up PRs with `chmod`, `chown`
+// (need mode-string parser).
 //
 // This file is an `agent` block that expects the enclosing scope
 // to bind:
 //   - `Dir`                — the agent's constructor channel
+//   - `File`               — the `File` agent's constructor
+//                            channel (`openFile` composes with it)
 //   - `nQuarantine`,
 //     `nEntries`, `nStat`,
 //     `nExists`,
 //     `nRemoveFile`,
 //     `nRemoveDir`,
 //     `nRename`,
-//     `nCopyFile`          — fixed-channel Pars for the native
+//     `nCopyFile`,
+//     `nOpen`              — fixed-channel Pars for the native
 //                            primitives, obtained via `NormalizerEnv`
 //                            injection at compile time (see
 //                            `interpreter::io::injections`).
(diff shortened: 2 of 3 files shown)


dylon (changes requested):
claude found 4 blockers and made a few recommendations:

  ## Review — PR #144 (Dir.openFile / Dir.openDir)

  **Verdict: blocked.**

  This is the first agent composition in the stack and the mechanics are right: a fresh reply
  channel per construction, `openDir` kind-checks the target before minting a `Dir`, the child
  `Dir`'s root is always the canonicalized path so scope attenuation is monotone (a child root
  is always under its parent's), and quarantine runs before any native fires on both paths. The
  tests are also a step up from the rest of the suite — `match openResult { [true, file] => … }`
  asserts on tuple *structure* rather than string-matching a debug dump, which is what I've been
  asking for elsewhere.

  It's blocked because it inherits #141's two blockers as its base and materially escalates one
  of them, and because it turns three previously-unreachable defects into user-reachable ones.

  ### Blocking

  1. **`openFile` escalates the missing `Dir` mode (#141, H3) from "mutators unavailable" to
     "arbitrary in-tree file destruction."** `Dir` still takes no mode, and `openFile` passes
     `mode` straight to `nativeOpen` with no gating. FIP line 253 is explicit that a `"r"`
     directory permits `openFile` *"restricted to file mode `"r"`"*. Here, any holder of any
     `Dir` can call `openFile("important.txt", "w")` — and `"w"` **truncates on open**. So a
     capability the FIP intends to be read-only silently becomes a delete primitive for every
     file in the tree. This is the concrete consequence I flagged abstractly on #141; it needs
     the mode-parameterized constructor before this merges.

  2. **`openFile` omits the kind check that `openDir` performs, so it inherits `native_open`'s
     missing file-type guard (#120).** `openDir` does the right thing —
     `match statRecord.get("kind") { "dir" => … _ => FSERR_BAD_ARG }` — but `openFile` goes
     straight from quarantine to `nOpen`. Combined with `native_open` having no file-type guard
     and no `O_NONBLOCK`, `dir.openFile("somefifo", "r")` **blocks forever and pins a tokio
     blocking-pool thread**, and `dir.openFile("<a subdir>", "r")` succeeds on Linux and hands
     back a `File` agent whose every read fails `EISDIR`. `entries()` reports both node types as
     `kind: "other"`, so they're discoverable. The symmetric `"file"` kind-check in `openFile`
     closes this at the agent layer; the native guard should land too.

  3. **Every agent instance is permanent tuplespace state, and `openFile`/`openDir` make the
     growth caller-driven.** I checked the desugaring against the parser corpus
     (`agent_minimal_desugared.rho`, `agent_with_methods_desugared.rho`): the dispatch loop is
     `for (...@args <= this)` — `<=`, persistent. So each constructed agent permanently leaves
     behind a persistent continuation on its fresh `this` *plus* the constructor's stored state
     datum (`@(*this,"root")!(rootPath)` / `@(*this,"fd")!(fd)`). Neither class has any teardown,
     and `File.close()` only releases the fd — the dispatch loop and the fd datum stay resident.

     Before this PR agents were only constructed by the test harness. Now a loop of
     `openDir("sub")` calls grows tuplespace state without bound. Node-local RSpace under the
     oracle model; on-chain state on any consensus path. And since there is no `charge` call
     anywhere in `dispatch.rs`, it is unpriced. The FIP's own wording — *"Creates a file agent
     **(or returns a cached one)**"* — points at the fix: cache per canonical path so repeated
     opens reuse one instance. Failing that, an explicit release path.

  4. **`openFile` makes the fd leak (#137, H4) user-reachable.** Every call allocates a fresh fd
     that is reclaimed only if the caller explicitly invokes `file!close()`. There is still no
     table bound, no per-deploy quota, and the rollback in #137 doesn't run on the casper deploy
     path. `openFile` is now the intended way to get fds, so this is the natural place for the
     lifecycle story — e.g. a per-`Dir` registry of issued `File`s with a bulk close.

  ### FIP alignment

  5. **The signature is a positional mode string, not the FIP's options map.** FIP §"Creating
     scoped agents" is `dir!openFile("notes/todo.txt", {"create": true})`, and §"The `openFile`
     and `openDir` methods" defines the map's `mode` field with default `"r"`. Two consequences
     beyond the shape: the FIP's own example — a call with **no** mode — falls through to
     `default` and returns `FSERR_UNSUPPORTED`, because `openFile` here requires exactly two
     arguments; and `"create": true` (FIP A3: `O_RDWR|O_CREAT` without truncation, the one
     combination fopen can't name) has no representation at all. The PR comment says the map
     form is deferred, which is fine — but please make the arity optional so the FIP's
     documented call shape doesn't hit the default arm.

  6. **`openDir` has no `"path": true`** (FIP A4, `mkdir -p`). Also deferred per the comment —
     worth noting there is no `mkdir` native anywhere in the stack yet, so this is blocked on
     #120 gaining one, not just on agent work.

  7. **Neither method caches** ("or returns a cached one"), which is FIP text and, per item 3, is
     also the cheapest fix for the state growth.

  8. **Child mode attenuation is undefined.** `openDir` constructs a fully-privileged child. Once
     the mode parameter lands (item 1), the rule needs to be explicit and **monotone** — a child
     must never be more privileged than its parent, and per FIP line 254 `openDir` with
     `path: true` requires the parent to be `"rw"`. Worth settling now, since `Fs.openDir` will
     copy whatever shape this establishes.

  ### Architecture

  9. **Embedding both agent blocks in one flat `new` scope gives every class direct access to
     every native.** The test harness now binds all thirteen native URNs plus both constructor
     channels in a single scope — and the comment correctly explains why ("the normalizer catches
     unbound names regardless of whether the referring method is invoked"). But that shape is
     what the genesis `Fs` deploy will inherit: one flat trust domain where `Dir` can call
     `nRead`/`nWrite`/`nClose` directly, bypassing `File` entirely, and vice versa. Worth
     considering per-class `new` scopes that bind only the natives that class needs, passing
     constructor channels between them — the attenuation the agent boundary is supposed to
     provide currently exists only by convention.

  10. **The write-only property of the returned handle is never tested.** `openFile` is the first
      place an agent handle crosses an agent boundary: `bundle+{*this}` is received as a Process,
      embedded in a reply tuple, and re-quoted by the caller as `@file!(…)`. Everything in the
      ocap story depends on the `bundle+` surviving that round trip — if it were ever flattened,
      the caller would gain *read* authority on the `File`'s dispatch channel and could consume
      the agent's own messages. The tests only ever send. A negative test that
      `for (x <- file) { … }` is rejected would lock in the property that makes every agent
      instance safe.

  11. Argument type-checking: `openFile("x", 42)` aborts the deploy via `IllegalArgumentError` →
      `NonDeterministicProcessFailure` rather than returning `[false, FSERR_BAD_ARG, …]` — the
      general issue raised on #120; `openFile`/`openDir` forward caller values unchecked.

  12. No replay test covers the composition, which introduces COMM events beyond what #137's
      native-level replay tests exercise.

  ### Verified clean (so nobody re-checks)

  - Child `Dir` roots are always under the parent's — `canonicalize_and_quarantine` guarantees
    the prefix, so path attenuation is monotone.
  - Symlink handling is consistent: quarantine canonicalizes first, so `nStat` on the returned
    path never sees a symlink, and an in-root symlinked subdirectory resolves and opens correctly
    while one pointing outside is rejected by `check_prefix`.
  - `Dir` referencing its own constructor channel for the nested construction resolves correctly
    through the enclosing `new`.
  - The reply-shape convention (`try @[file] <-` binding the single rest element) matches both
    the FIP's `try @[dir] <- fs!openDir(…)` example and the existing shipped `nQuarantine` call
    sites.
  - Client-side membranes still work: a client can wrap `dir` and wrap each child it hands out,
    so the FIP's §Membranes revocation pattern is supportable with this design.

metaweta:

PR #145, opened 2026-07-22
feat(fileio): line-based File methods (readLine, lines, appendLines)

metaweta:
## Summary

Adds the first slice of the FIP §"Per Line" surface: `readLine`, `lines`, and `appendLines` on the `File` agent, plus three matching native primitives.

## What ships

Native primitives:
- `nativeReadLine(fd) → [true, line] | [false, code, msg]` — reads until `\n` or EOF, advances cursor past the newline, strips `\n` (and preceding `\r` for CRLF), decodes as UTF-8. EOF returns `[true, ""]`.
- `nativeReadAllLines(fd) → [true, [line, ...]] | ...` — reads current position to EOF, splits on `\n`, absorbs trailing empty, strips per-line `\r`.
- `nativeAppendLines(fd, lines: List[String]) → [true, nBytes] | ...` — writes each String + `\n`, returns total bytes. Non-String elements rejected with `FSERR_BAD_ARG` before any write.

Agent methods (`file.rho`):
- `readLine()`, `lines()`, `appendLines(lineList)` — try/catch wrappers over the natives.

Per-call byte caps to bound single-call memory:
- `readLine`: 16 MiB per line
- `readAllLines`: 64 MiB per call (matches `native_read`)
- `appendLines`: 64 MiB per call (total bytes including newlines)

## Line-vs-byte mutex — deferred

The FIP §"Per Line" intro promises a coarse mutex between line and byte methods on the same file agent. This PR does NOT ship that mutex. Interleaving is safe from a data-integrity standpoint because the native handlers maintain fd-cursor consistency (readLine rewinds past the newline before returning), but honoring the FIP semantics is left to a follow-up.

## Registrations

Every new native adds: `FixedChannels::native_*()` byte (57-59), `BodyRefs::NATIVE_*` (56-58), entry in `non_deterministic_ops()`, row in `rho_runtime::std_system_processes()`, entry in `io::injections::fileio_native_urns()` (bundle count 19→22), plus lines in the two completeness assertions.

## Test plan

- [x] `cargo test -p rholang --test fileio_file_agent_spec` — 14 pass (8 prior + 6 new)
- [x] `cargo test -p rholang --test fileio_dir_agent_spec` — 20 pass (unchanged; extra URN bindings inert)
- [x] `cargo test -p rholang --lib io::` — 42 pass (bundle-completeness updated)
- [x] `cargo test -p rholang --lib non_deterministic` — 2 pass (nondet-completeness updated)
- [x] `cargo fmt -p rholang --check` — clean
- [x] `cargo clippy -p rholang --lib --tests` — clean
- [x] Pre-push (11 slices) — clean

## New tests (6)

- `file_agent_read_line_returns_first_line_and_advances_cursor` — verifies rewind logic; asserts cursor at byte 11 after reading a 10-char line + `\n`.
- `file_agent_read_line_strips_crlf` — both reads on a `\r\n` file return the bare line text.
- `file_agent_read_line_at_eof_returns_empty_string` — second readLine on a one-line file returns `[true, ""]`.
- `file_agent_lines_returns_all_lines_without_trailing_empty` — file ends in `\n`; asserts no trailing empty; per-line CRLF stripping verified.
- `file_agent_append_lines_writes_lines_then_readback_matches` — appends 3 lines, seeks to 0, reads back via `lines`; round-trip and 14-byte total verified.
- `file_agent_append_lines_rejects_non_string_element` — list contains an Int; asserts `FSERR_BAD_ARG` returned AND size still 0 (no partial write).

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/145"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787343531&installation_model_id=426883&pr_number=145&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F145&signature=a22362095e2917ab6081fad2b041d3a76c9cb3914f863c729fbdbb9140b43378"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/rholang/src/rust/interpreter/io/agents.rs
+++ b/rholang/src/rust/interpreter/io/agents.rs
@@ -18,7 +18,8 @@
 //! Native URNs each `.rho` file expects:
 //!
 //! - `file.rho` — `nRead`, `nWrite`, `nSeek`, `nTell`, `nSize`,
-//!   `nTruncate`, `nFlush`, `nClose`.
+//!   `nTruncate`, `nFlush`, `nClose`, `nReadLine`, `nReadAllLines`,
+//!   `nAppendLines`.
 //! - `dir.rho`  — `nQuarantine`, `nEntries`, `nStat`, `nExists`,
 //!   `nRemoveFile`, `nRemoveDir`, `nRename`, `nCopyFile`, `nOpen`,
 //!   plus the `File` constructor channel (`openFile` composes
@@ -26,7 +27,8 @@
 
 /// Source of the `File` agent block. Expects `File`, `nRead`,
 /// `nWrite`, `nSeek`, `nTell`, `nSize`, `nTruncate`, `nFlush`,
-/// `nClose` to be bound in the enclosing scope.
+/// `nClose`, `nReadLine`, `nReadAllLines`, `nAppendLines` to be
+/// bound in the enclosing scope.
 pub const FILE_AGENT_SRC: &str = include_str!("agents/file.rho");
 
 /// Source of the `Dir` agent block. Expects `Dir`, `File`,
--- a/rholang/src/rust/interpreter/io/agents/file.rho
+++ b/rholang/src/rust/interpreter/io/agents/file.rho
@@ -1,10 +1,11 @@
 // File agent — wraps a native fd with the FIP `File` method surface.
 //
 // Currently ships: `read`, `write`, `seek`, `tell`, `size`,
-// `truncate`, `flush`, `close`, `default`. Extended in follow-up
-// PRs with `readAt`, `writeAt`, `lockRange` (all need range-lock
-// infrastructure, Phase 4), `chmod`, `chown` (need mode-string
-// parser), plus line-based `text` / `lines` / `mapReduceLines*`.
+// `truncate`, `flush`, `close`, `readLine`, `lines`, `appendLines`,
+// `default`. Extended in follow-up PRs with `readAt`, `writeAt`,
+// `lockRange` (all need range-lock infrastructure, Phase 4),
+// `chmod`, `chown` (need mode-string parser), plus `text` /
+// `forEachLine` / `mapReduceLines*` (need callback-handler dispatch).
 //
 // This file is an `agent` block that expects the enclosing scope
 // to bind:
@@ -12,7 +13,10 @@
 //   - `nRead`, `nWrite`,
 //     `nSeek`, `nTell`,
 //     `nSize`, `nTruncate`,
-//     `nFlush`, `nClose`   — fixed-channel Pars for the native
+//     `nFlush`, `nClose`,
+//     `nReadLine`,
+//     `nReadAllLines`,
+//     `nAppendLines`       — fixed-channel Pars for the native
 //                            primitives, obtained via `NormalizerEnv`
 //                            injection at compile time (see
 //                            `interpreter::io::injections`).
@@ -156,6 +160,67 @@ agent File {
     for (@fd <<- @(*this, "fd")) {
       try <- nClose!?(fd) {
         return!([true])
+      }
+      catch @[code, msg] {
+        return!([false, code, msg])
+      }
+    }
+  } |
+
+  method readLine() {
+    // Reads bytes from the current position until the next `\n`
+    // (or EOF). Advances the cursor to just past the newline.
+    // Returns the line as a String with the trailing `\n` (and
+    // preceding `\r` for CRLF) stripped. EOF with no bytes read
+    // returns [true, ""] -- indistinguishable from a real empty
+    // line; callers who need to detect EOF unambiguously can
+    // compare `tell()` against `size()` afterward. Native caps
+    // any single line at 16 MiB and returns FSERR_BAD_ARG beyond
+    // that.
+    for (@fd <<- @(*this, "fd")) {
+      try @[line] <- nReadLine!?(fd) {
+        return!([true, line])
+      }
+      catch @[code, msg] {
+        return!([false, code, msg])
+      }
+    }
+  } |
+
+  method lines() {
+    // Reads from the current position to EOF, splits on `\n`,
+    // and returns the list of lines. A trailing newline is
+    // absorbed (so `"a\nb\n"` yields `["a", "b"]`, not
+    // `["a", "b", ""]`); each line has any trailing `\r`
+    // stripped. Advances the cursor to EOF. Native caps the
+    // total byte count at 64 MiB per call; larger files should
+    // be streamed via `readLine` in a loop.
+    for (@fd <<- @(*this, "fd")) {
+      try @[list] <- nReadAllLines!?(fd) {
+        return!([true, list])
+      }
+      catch @[code, msg] {
+        return!([false, code, msg])
+      }
+    }
+  } |
+
+  method appendLines(lineList) {
+    // Writes each String in `lineList` followed by a single
+    // `\n` at the current position, advancing the cursor.
+    // Returns `[true, nBytes]` where `nBytes` is total bytes
+    // written including the newlines. Requires a writeable fd
+    // (opened in "w", "a", "r+", "w+", or an "x"-suffixed
+    // variant); otherwise the native returns FSERR_PERM.
+    //
+    // The name "appendLines" refers to writing lines at the
+    // current cursor position -- it does NOT imply the fd was
+    // opened in append mode. For fds opened with "a"/"a+", the
+    // OS overrides the cursor and writes go to EOF, matching
+    // POSIX O_APPEND semantics.
+    for (@fd <<- @(*this, "fd")) {
+      try @[nBytes] <- nAppendLines!?(fd, lineList) {
+        return!([true, nBytes])
       }
       catch @[code, msg] {
         return!([false, code, msg])
--- a/rholang/src/rust/interpreter/io/injections.rs
+++ b/rholang/src/rust/interpreter/io/injections.rs
@@ -49,7 +49,7 @@ use crate::rust::interpreter::errors::InterpreterError;
 use crate::rust::interpreter::system_processes::FixedChannels;
 
 /// Return the `NormalizerEnv` bundle for the File I/O native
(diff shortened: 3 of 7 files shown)


dylon (changes requested):
claude found 8 blockers and made a few recommendations:

  ## Review — PR #145 (line-based File methods)

  **Verdict: blocked.**

  The plumbing here is the cleanest in the stack so far. Three new natives land across all
  seven integration points — `FixedChannels` (bytes 57–59), `BodyRefs` (56–58),
  `non_deterministic_ops`, `Definition` rows, the injections bundle (19 → 22 with the capacity
  hint and the completeness test both updated), and both test envs — with nothing forgotten and
  no channel or body-ref collision. The `readLine` rewind arithmetic is correct
  (`extras = n_read - nl_idx - 1`, seek back by that much lands the cursor exactly past the
  newline, and the test pins it at byte 11). Trailing-newline absorption in `readAllLines` is
  correct, including the `"a\n\n"` → `["a", ""]` case. `appendLines` extracts and validates every
  element *before* touching the fd, with a test asserting the file is still zero bytes after a
  rejection. And `readLine` is the first multi-syscall operation in the stack held under the
  per-handle mutex for its whole duration, which is the right instinct.

  It's blocked on semantics, not plumbing: three of the FIP's stated contracts are changed in
  ways that produce silently-wrong results rather than errors.

  ### Blocking — FIP alignment

  1. **`lines()` splits on `\n` only; the FIP requires `\R`.** FIP §`lines`: "reads the entire
     file, splits the string using the regular expression `\R` (Unicode line break)". `\R`
     matches `\r\n`, `\n`, `\r`, VT (U+000B), FF (U+000C), NEL (U+0085), LS (U+2028) and PS
     (U+2029). `native_read_all_lines` does `s.split('\n')` then strips a trailing `\r`. So a
     file with bare-`\r` terminators comes back as **one line with embedded `\r`s**, and
     LS/PS/NEL/VT/FF are not line breaks at all. No error is raised — the caller just gets the
     wrong list. (The FIP's own "we should deprecate this once we have regular expressions" note
     acknowledges `\R` is the contract until then.)

  2. **`lines()` reads from the cursor; the FIP says it reads the entire file.** After any prior
     `readLine()` or `read(n)`, `file!lines()` silently returns only the remainder. A client
     written against the FIP gets a wrong answer with no signal. Same for the `text()` model the
     FIP describes.

  3. **`appendLines` writes at the cursor; the FIP says it appends to the end.** FIP
     §`appendLines`: "takes a list of strings and **appends them to the end of the text file**",
     and §Per Line's intro frames the whole line surface as "reading from the beginning and
     appending to the end". The doc comment here is explicit that the name "does NOT imply the fd
     was opened in append mode" and that writes go to the current position. On an `"r+"` fd
     sitting at 0 — the default position after open — `appendLines` **overwrites the head of the
     file** where the FIP promises an append. That is a data-loss-shaped divergence from spec,
     and it is the one deviation a caller is most likely to hit by following the FIP literally.
     Either seek to end inside the method, or rename it (`writeLines`) and add a separate
     spec-conformant `appendLines`.

  4. **`appendLines` returns `[true, nBytes]`; the FIP's contract is `[true]`.** FIP §Error syntax
     names `close`, `flush`, `rename` as the `[true]`-only methods written `try <- …`, and the
     `appendLines` example uses exactly that bare form. The extra payload happens to remain
     compatible with the sugar (the success branch just drops the `let`), but it isn't the
     documented shape.

  5. **`readLine` is a `ReadOnlyStream` method, not a `File` method, in the FIP.** §Standard I/O
     assigns `readLine()` to the stdin class; §Per Line's `File` surface is `text`, `lines`,
     `mapReduceLines`, `mapReduceLinesOrdered`, `appendLines`. Adding it to `File` is a sensible
     extension — it's the only streaming primitive available given the 64 MiB `lines()` cap — but
     it's beyond the class's spec surface and should be called out as such, especially since the
     FIP is deliberate about which methods live on which class.

  ### Blocking — correctness

  6. **`readLine` cannot distinguish EOF from an empty line, so the loop this PR recommends
     cannot be written correctly.** `[true, ""]` is returned both at EOF and for a genuine empty
     line; the doc acknowledges this and suggests comparing `tell()` against `size()` afterward.
     That workaround is two extra agent round-trips per line *and* racy — there is no line-vs-byte
     mutex and nothing stops another process writing between the `tell` and the `size`. Meanwhile
     `lines()` caps at 64 MiB and the doc says "larger files should be streamed via `readLine` in
     a loop". Composing the two: **there is currently no correct way to read a text file larger
     than 64 MiB.** Compare `read(n)`, where the FIP's "at EOF the payload is an empty ByteArray"
     is unambiguous because a zero-length read can't be confused with data. Returning `[true, Nil]`
     at EOF versus `[true, ""]` for an empty line, or a third tuple element for "newline seen",
     would close this.

  7. **Error paths discard the bytes read but leave the cursor advanced.** In `native_read_line`,
     the oversized path (`line.len() > MAX_LINE_BYTES`) breaks out and returns `FSERR_BAD_ARG`
     after having consumed up to 16 MiB — with no rewind. Same for the `io_err` path, and same for
     `native_read_all_lines` on its 64 MiB cap. The caller gets an error, the data is gone, and
     the cursor is parked at an arbitrary offset mid-line, so a retry returns garbage. `read(n)`
     avoids this because it returns whatever it consumed. There's a narrower instance too: if the
     post-newline `seek` fails, `io_err` is set and a **successfully read line is thrown away**.
     Please capture the start position and restore it on every error path.

  8. **The byte cap doesn't bound the reply size.** `MAX_READ_BYTES` limits the input buffer, but
     the output is one `RhoString` `Par` per line and nothing caps the line *count*. 64 MiB of
     newline bytes sits exactly at the cap, splits into ~67M empty strings, and produces ~67M
     `Par`s — each a prost struct holding a `Vec<Expr>` holding a `GString`, so on the order of
     100+ bytes of heap apiece. That's multiple GB resident from a 64 MiB input, built in one
     handler call and then handed to RSpace as a produce payload. The amplification factor is
     roughly (Par overhead ÷ average line length), so short lines are the hazard generally, not
     just the all-newline extreme. And with no `charge` call anywhere in `dispatch.rs`, it's
     unmetered. Cap the line count as well, or bound the constructed payload rather than the input.

  ### Should fix

  9. **`appendLines` can't report a partial write.** If `write_all` fails midway, some bytes are
     on disk and the cursor has advanced, but the reply is a bare `from_io_error` with no byte
     count — the caller cannot determine the file's state. FIP `write` returns `nWritten` for
     exactly this reason.

  10. **Invalid UTF-8 in file *content* is reported as `FSERR_BAD_ARG`.** Per FIP §Standard error
      codes, `FSERR_BAD_ARG` is for "a mode string, offset, owner/group name, or other **argument**"
      — the argument here is fine, the file isn't. `FSERR_IO` fits better. (Same classification
      family as the NSS issue on #130.)

  11. **The unconditional trailing-`\r` strip also fires on a final line with no newline.** A file
      ending `"abc\r"` yields `"abc"`, silently dropping a byte that wasn't a line terminator.

  12. **`text()` is still missing, and the stated reason doesn't apply to it.** The header defers
      "`text` / `forEachLine` / `mapReduceLines*` (need callback-handler dispatch)" — but `text()`
      takes no callback. It's read-to-end plus one `String`, i.e. `readAllLines` without the split,
      and it's the FIP's simplest line method.

  13. **Line content isn't round-trip safe.** `appendLines(["a\nb"])` writes two physical lines;
      `lines()` returns `["a", "b"]`. No validation or escaping, and the FIP doesn't address it —
      worth at least documenting.

  14. **No replay coverage for the three new natives.** All three are in `non_deterministic_ops`
      and `fileio_replay_spec.rs` already has the play→mutate→replay harness; adding cases is
      cheap and these are the first natives whose captured output is a large structured payload.

  ### Interaction with the rest of the stack

  15. **This makes the missing line-vs-byte mutex concrete.** FIP F16 exists precisely because
      mixing line-based and positional access on one handle corrupts both. Until now only the byte
      surface shipped, so there was nothing to interleave. Now `read`/`seek` and
      `readLine`/`lines`/`appendLines` share one cursor with no mutex and no `FSERR_BUSY` — the
      hazard the FIP designed against is realizable today.

  16. **The Dir test scope had to grow again.** Adding three File natives forced
      `fileio_dir_agent_spec.rs` to bind them, because `FILE_AGENT_SRC` is embedded in the Dir
      scope. That's the flat-trust-domain issue raised on #144 showing up as maintenance cost —
      every File method now widens the set of natives directly reachable from the Dir agent's
      scope. Worth resolving before `Fs` freezes the shape.

  ### Verified clean

  - Channel bytes 57–59 and BodyRefs 56–58 are collision-free (prior max 56 / 55; test channels
    at 101+).
  - `RhoList::unapply` exists and returns `Option<Vec<Par>>` — `appendLines` compiles.
  - Replay guards on all three are correctly placed (after argument validation, before the
    syscall) and match the established `gpt4`/`dalle3` shape; ack-first destructuring is
    consistent with #139.
  - Empty file → `[true, []]`; `"\n"` → `[true, [""]]`; `"a\n\n"` → `["a", ""]`. All correct.
  - CRLF stripping is correct across chunk boundaries, since the `\r` is removed from the
    accumulated line rather than per-chunk.
  - `MAX_LINE_BYTES` is checked only on the no-newline branch, which is sound — the newline branch
    can overshoot by at most one 4 KiB chunk.

metaweta:

Issue #65, opened 2026-05-28
Cold-start Casper initialization hangs in `replay_blocks_for_mergeable_channels`


elmiringos:
## What did you expect to see?

A validator or observer that is cold-restarted (empty data directory) should download the last-finalized-state and historical blocks from peers, initialize Casper, and resume participation (finalization / proposals) within a reasonable window — order of minutes, proportional to chain depth.

With 14 DAG blocks of history and minimal load (100 deploys burst), cold-start should complete in under 5 minutes. With 49–71 finalized blocks it should complete in under 20 minutes.

## What did you see instead?

After cold restart the node enters a silent hang. It never reaches the Casper initialization loop and never logs `"Casper not initialized yet"` — meaning it is stuck *before* Casper even begins. The node reports no newly finalized blocks for the entire 20-minute observation window and beyond.

The hang is reproducible across all tested load levels:

| Run | Load | Deploy payload | Chain depth at restart | Node role | Recovery |
|---|---|---|---|---|---|
| Run 1 | 7 500 deploys burst | `GetNonceTerm` (read-only bridge query) | 7 finalized | observer | timeout 20 m |
| Run 2 | 7 500 deploys burst | `GetNonceTerm` (read-only bridge query) | 1 finalized | observer | timeout 20 m |
| Run 3 | ~2 500 deploys burst | `GetNonceTerm` (read-only bridge query) | 1 finalized | observer | timeout 20 m |
| Run 4 | 100 deploys burst | `BridgeTerm` (full bridge contract deploy) | **14 DAG blocks** (1 finalized) | validator | timeout 20 m |
| Run 4 | 1 deploy / 2 s | `GetNonceTerm` (read-only bridge query) | 49 finalized | observer | timeout 20 m |

Run 4 with the validator target is the clearest evidence: only 14 DAG blocks existed in the entire chain history. The node still could not complete cold-start in 20 minutes. This rules out load, deploy backlog, and resource pressure as causes.

The `"Casper not initialized yet"` log line count stays at **0** during the entire recovery window across all runs. The node is not reaching Casper initialization at all — it is stuck earlier, in the block replay path that precedes it.

## Deploy payloads used in testing

Two deploy types were used depending on the restart target:

**`GetNonceTerm`** — used for all observer-restart runs (Runs 1–3 and Run 4 slow). A read-only query that looks up the bridge contract in the registry and calls `getTotalLocked`. It never writes persistent state, making it the lightest possible flood payload.

[...]

PR #151, opened 2026-07-27
[agent] fix(soak): publish Pages dashboard, repair stakeholder alerting, protect published history

jeffrey-l-turner:
## What this fixes

The 60-hour weekend soak and the daily soaks had never produced a usable result, for three stacked reasons:

1. **Before PR #138** every scheduled run was a wall-clock gate no-op that reported success. The soak never executed.
2. **After #138** every real attempt died ~41s into `test_load.py` on a missing `validator4` TLS cert — `Failed to read the X.509 certificate: Is a directory (os error 21)`. Docker turns an absent bind-mount host path into a directory. That fix lives in `F1R3FLY-io/system-integration` PR #65.
3. **GitHub Pages was never enabled**, so even a passing soak could not publish its dashboard, and the stakeholder alert chain was broken end-to-end.

The OCI runners were never the problem — they launch, build, and run within two minutes on every attempt.

## Changes

**Soak operability**
- `soak-dashboard-pages.yml` (new): publishes the dashboard on push to `master` with an empty-state history, so the page exists without waiting for a weekend soak.
- Schedule-gate no-ops now emit a `::notice::` and a step summary, so a green check on a skipped cron is not mistaken for a soak result. The gate no-ops ~9× a week by design.
- `notify_failure` (new job): emails the ONS topic when a real soak fails **or is cancelled**. Runs GitHub-hosted with user-principal auth, so it still fires when the OCI runner never launched — a case the in-soak instance-principal step structurally cannot reach. Declares `permissions: {}` (no checkout, no GitHub API).

**Protecting published history** — Pages is the only durable store for the soak series, since artifacts expire at 14 days.
- Both the bootstrap deploy and `perf_report` now treat only HTTP 404 as "nothing published". Any other outcome — 5xx, timeout, connection failure — aborts instead of republishing an empty dashboard over real history. Previously a transient failure exited 0 and reset the series to the current run alone.
- Fetches are cache-busted (`?cb=<run_id>-<attempt>` plus `Cache-Control: no-cache`). Pages is CDN-backed, so a `200` could otherwise serve a copy older than the last publish and silently roll it back.
- Every fetched JSON artifact is validated before republishing — a `200` carrying a proxy error page is not data.

**Docs** (separately scoped commits, revertable independently)
- `staging` → `dev` across `CONTRIBUTING.md`, `SECURITY.md`, and the neutral-cloud benchmark guide. `staging` holds zero commits not already in `dev` and sits 158 behind.
- Documented the hotfix path — branch from `master`, PR to `master`, then merge `master` back into `dev` — which three merged PRs already follow but nothing described.
- Corrected `CLAUDE.md`'s branch strategy, which named a `main` branch that does not exist.

## Configuration applied out-of-band

Not visible in this diff, but required for any of it to work: GitHub Pages enabled (`build_type: workflow`); the `SOAK_ONS_TOPIC_OCID` repo variable set (it was unset, so every publish silently exited 0); and `use ons-topics` granted to the `ci-runner-ephemeral` dynamic group and `ci-runner-launchers` group — no IAM policy previously permitted ONS publish at all.

## Verification

The data-preservation logic was exercised against a local fixture server rather than reasoned about: a healthy `200` preserves and appends history, `404` bootstraps an empty series, and `503`, connection failure, and a `200` carrying HTML all abort without publishing. The pre-fix behaviour was reproduced first to confirm the bug was real — the old code exits **0** under HTTP 503 and publishes only the current week.

## Review

Two review passes are addressed in the comments below. A multi-agent review found the CDN staleness; @spreston8 found that the first fix was applied to the bootstrap path while the weekly production path — the one that actually appends history — kept the original hazard. Neither review caught the other's finding. Dismissed findings are documented with evidence so the calls can be challenged.

## Known remaining work

`SYSTEM_INTEGRATION_REF` is still pinned at `a50eeb19`, which predates the `validator4` certs. Until system-integration PR #65 merges and that pin is bumped, the soak still fails — but it now emails on failure instead of reporting a silent green.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/151"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787767347&installation_model_id=426883&pr_number=151&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F151&signature=6e9bd7e32faa2f60701478a27c7248187911f35ea750b0a3414f60ff4d4732ae"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -262,10 +262,21 @@ jobs:
           DURATION_SECONDS: ${{ needs.schedule_gate.outputs.duration_seconds }}
           DASHBOARD_URL: https://f1r3fly-io.github.io/f1r3node-rust/
         run: |
-          curl -fsS --max-time 30 "${DASHBOARD_URL}data/latest-summary.json" \
-            -o /tmp/soak-baseline.json 2>/dev/null \
-            && export BASELINE_JSON=/tmp/soak-baseline.json \
-            || echo "no published baseline yet (bootstrap run)"
+          # Cache-busted so the week-over-week deltas are not computed against
+          # a stale CDN copy. Unlike the history fetch this stays soft: a 60h
+          # soak's report is worth more than its baseline comparison, so a
+          # miss degrades to "no baseline" rather than discarding the run —
+          # but a non-404 miss is warned about, since a silently absent
+          # baseline makes the verdict look clean.
+          cb="${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          base_code="$(curl -sS --max-time 30 -H 'Cache-Control: no-cache' -w '%{http_code}' \
+            -o /tmp/soak-baseline.json \
+            "${DASHBOARD_URL}data/latest-summary.json?cb=${cb}")" || base_code=000
+          case "$base_code" in
+            200) export BASELINE_JSON=/tmp/soak-baseline.json ;;
+            404) echo "no published baseline yet (bootstrap run)" ;;
+            *)   echo "::warning::baseline fetch failed (HTTP ${base_code}); verdict computed without week-over-week comparison" ;;
+          esac
           ci-control/scripts/bench/aggregate-perf-report.sh
           cat "$OUT_DIR/perf-report.md" >> "$GITHUB_STEP_SUMMARY"
 
@@ -280,6 +291,16 @@ jobs:
         run: |
           # Instance-principal auth: works only on the OCI runner, no secrets.
           # Notification failure must never mask the soak result.
+          #
+          # This overlaps with the notify_failure job by design — do not
+          # "deduplicate" them. This step sends the weekly verdict with its
+          # metrics and week-over-week deltas, but only ever runs ON the soak
+          # runner and only once a report exists, so it cannot report a runner
+          # that never launched or a soak that died before writing one.
+          # notify_failure covers exactly those cases from a GitHub-hosted
+          # runner. A weekend soak that completes and then fails its verdict
+          # gate therefore sends two mails: the verdict (with numbers) and the
+          # job-level alert. That is the accepted cost of never going silent.
           ci-control/scripts/oci/publish-soak-alert.sh || \
             echo "::warning::ONS alert publish failed (non-fatal)"
 
@@ -329,9 +350,29 @@ jobs:
             echo "::error::soak artifact has no weekly-summary.json"; exit 1; }
           mkdir -p site/data
           cp -r ci-control/site/soak-dashboard/. site/
-          curl -fsS --max-time 30 "${DASHBOARD_URL}data/history.json" \
-            -o /tmp/history.json 2>/dev/null || echo '[]' > /tmp/history.json
-          jq -e 'type == "array"' /tmp/history.json > /dev/null || echo '[]' > /tmp/history.json
+          # Pages is the only durable store for the history series — soak
+          # artifacts expire after 14 days — and this is the job that appends
+          # to it. Treating a transient fetch failure as "nothing published"
+          # would permanently reset the series to this run alone, so only a
+          # 404 may start an empty history and anything else aborts. The
+          # cache-busting query string stops the CDN serving a copy older
+          # than the last publish, which would drop intervening weeks.
+          cb="${GITHUB_RUN_ID}-${GITHUB_RUN_ATTEMPT}"
+          code="$(curl -sS --max-time 30 -H 'Cache-Control: no-cache' -w '%{http_code}' \
+            -o /tmp/history.json "${DASHBOARD_URL}data/history.json?cb=${cb}")" || code=000
+          case "$code" in
+            200)
+              jq -e 'type == "array"' /tmp/history.json >/dev/null 2>&1 || {
+                echo "::error::published history.json is not a JSON array; refusing to reset the series"
+                exit 1
+              }
+              ;;
+            404) echo '[]' > /tmp/history.json ;;
+            *)
+              echo "::error::could not read published history.json (HTTP ${code}); aborting so the series is not reset to this run alone"
+              exit 1
+              ;;
+          esac
           jq -s '.[0] + [.[1]] | sort_by(.run.date)' \
             /tmp/history.json soak-results/report/weekly-summary.json \
             > site/data/history.json
@@ -426,12 +467,14 @@ jobs:
             exit 1
           fi
           if [ "$DURATION_SECONDS" = "216000" ]; then kind="weekend-60h"; else kind="daily-24h"; fi
-          # Name what actually broke. A soak that never ran and a dashboard
-          # that failed to publish after a clean soak are different pages.
+          # Name what actually broke. A soak that never ran and a report stage
+          # that failed after a clean soak are different pages. perf_report
+          # covers both the Pages deploy and the regression verdict gate, so
+          # its failure is named for the stage rather than the deploy alone.
           case "$LAUNCH_RESULT/$SOAK_RESULT" in
             failure/*|cancelled/*) what="runner launch $LAUNCH_RESULT" ;;
             */failure|*/cancelled) what="soak $SOAK_RESULT" ;;
-            *)                     what="dashboard publish $DASHBOARD_RESULT" ;;
+            *)                     what="report/verdict stage $DASHBOARD_RESULT" ;;
           esac
           title="F1R3FLY soak alert ($kind): $what — $(date -u +%F)"
           body="Merge Recovery Soak did not complete cleanly: $what.
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -364,13 +364,19 @@ jobs:
     # Runs on a GitHub-hosted runner with user-principal OCI auth so the
     # alert still goes out when the OCI runner itself failed to launch —
     # the in-soak ONS step (instance-principal) cannot cover that case.
+    # `cancelled` counts too: a 60h soak that hits its timeout, is cancelled
+    # by hand, or loses its runner to preemption produced no result either,
+    # and silence there is indistinguishable from success.
     if: >-
       always() &&
       needs.schedule_gate.outputs.should_run == 'true' &&
-      contains(needs.*.result, 'failure')
+      (contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled'))
     runs-on: ubuntu-latest
     environment: ephemeral-launch-internal
     timeout-minutes: 10
+    # Alerting needs OCI credentials, never the GitHub API — this job has no
+    # checkout and no reason to hold a token scope.
+    permissions: {}
     steps:
       - name: Install OCI CLI (pinned, checksum-verified)
         shell: bash -e {0}
@@ -403,7 +409,7 @@ jobs:
           EOF
           oci iam region list >/dev/null
 
-      - name: Publish failure alert
+      - name: Publish soak alert
         env:
           ONS_TOPIC_OCID: ${{ vars.SOAK_ONS_TOPIC_OCID }}
           LAUNCH_RESULT: ${{ needs.launch_runner.result }}
@@ -420,8 +426,15 @@ jobs:
             exit 1
           fi
           if [ "$DURATION_SECONDS" = "216000" ]; then kind="weekend-60h"; else kind="daily-24h"; fi
-          title="F1R3FLY soak FAILURE ($kind) $(date -u +%F)"
-          body="Merge Recovery Soak run failed.
+          # Name what actually broke. A soak that never ran and a dashboard
+          # that failed to publish after a clean soak are different pages.
+          case "$LAUNCH_RESULT/$SOAK_RESULT" in
+            failure/*|cancelled/*) what="runner launch $LAUNCH_RESULT" ;;
+            */failure|*/cancelled) what="soak $SOAK_RESULT" ;;
+            *)                     what="dashboard publish $DASHBOARD_RESULT" ;;
+          esac
+          title="F1R3FLY soak alert ($kind): $what — $(date -u +%F)"
+          body="Merge Recovery Soak did not complete cleanly: $what.
 
           Target:    $TARGET_REF
           Duration:  $kind
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -86,6 +86,15 @@ jobs:
           if [ "$duration_seconds" = "216000" ]; then
             run_benchmarks=true
           fi
+          if [ "$should_run" != "true" ]; then
+            echo "::notice::Schedule gate no-op: cron slot resolved outside the Pacific 22:00 launch window. No soak was attempted; this run's success is not a soak result."
+            {
+              echo "## Schedule gate: no-op"
+              echo ""
+              echo "This scheduled trigger resolved outside the Pacific 22:00 launch window."
+              echo "**No soak was attempted** — a green check on this run is not a soak result."
+            } >> "$GITHUB_STEP_SUMMARY"
+          fi
           {
             echo "should_run=$should_run"
             echo "duration_seconds=$duration_seconds"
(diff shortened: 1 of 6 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The notification design and schedule-gate annotation are generally sound, but the Pages bootstrap has a significant data-retention flaw: any transient failure while reading the currently published history can cause the next deployment to erase it. The alert condition also omits cancelled real runs. The history preservation behavior should be made fail-safe before merging.

Solid in-repo half of the soak operability fix: schedule-gate no-ops are clearly annotated, failure notification is correctly scoped with always() + should_run, and OCI user-principal auth on a GitHub-hosted runner is the right way to cover launch failures that instance-principal cannot. The Pages bootstrap workflow is thoughtfully designed to preserve published data with safe history/report fallbacks. Before merge, verify Pages deploys share the soak-pages concurrency lane with perf_report (to avoid stale overwrites) and that ephemeral-launch-internal will not gate alert jobs on manual approval; consider tightening alert wording for dashboard-only failures and adding empty defaults for the remaining JSON artifacts.

The code changes look good overall, addressing the soak dashboard deployment and failure notification. Minor improvements can be made in handling OCI credentials more securely and ensuring consistent workflow naming conventions.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Cancelled real soak runs do not trigger an alert</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 370)

The notification condition only matches a dependency result of `failure`. If runner launch, soak, or report generation is cancelled after a real soak was scheduled, all dependency results may be `cancelled` or `skipped`, and no stakeholder notification will be sent despite the soak not completing.

---

notify_failure runs in environment `ephemeral-launch-internal`. If that environment has required reviewers, wait timers, or restricted actors, soak failure emails will not send until approval — undermining the goal of covering runner-launch failures automatically.

**Suggestion:**
```
If incomplete real soak runs should also alert, explicitly test for both `failure` and `cancelled`, while retaining the schedule-gate check to suppress genuine no-op runs.
```

</details>

<details>
<summary><b>:red_circle: Critical: Transient download failures can erase existing soak history</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 49)

The workflow treats every failure to fetch history.json as an empty initial deployment and publishes `[]`. A temporary Pages outage, timeout, CDN error, or rate limit would therefore overwrite valid historical data. The other files are also removed independently when their downloads fail, producing a partially erased dashboard. This contradicts the stated goal of preserving previously published data.

---

history.json and latest-report.md get safe defaults when absent/invalid, but latest-summary.json and latest-verdict.json are left missing on first bootstrap or fetch failure. Dashboard JS that unconditionally fetches those files may 404 or error until the first successful soak publish.

**Suggestion:**
```
Add minimal valid defaults (e.g. {} or dashboard-expected shapes) when those files are missing after the curl loop, matching the history/report fallbacks.
```

</details>

<details>
<summary><b>:orange_circle: Major: Pages deploy race if perf_report lacks same concurrency group</b> (Reported by: xai)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 24)

This workflow fetches live data then redeploys the whole site under concurrency group `soak-pages`. If merge-recovery-soak's perf_report job deploys Pages without the exact same concurrency group (and cancel-in-progress: false), a master push that touches site/soak-dashboard/** can interleave with a weekend soak publish and overwrite fresher history/summary artifacts with stale or empty-state data.

**Suggestion:**
```
Confirm perf_report uses concurrency.group: soak-pages with cancel-in-progress: false. If it does not, add it in the same PR so bootstrap and soak publishes serialize on one Pages deploy lane.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Alert fires for dashboard-only failures as a full soak failure</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 368)

contains(needs.*.result, 'failure') treats perf_report failure the same as launch_runner/soak failure. The email title always says 'soak FAILURE', which can page stakeholders for a Pages publish problem after a successful soak.

**Suggestion:**
```
Derive severity/title from which dependency failed (e.g. distinguish launch/soak failure vs dashboard publish failure), or only notify when launch_runner or soak failed.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: OCI credentials handling</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 374)

The OCI credentials are stored in plain text in the runner's filesystem. Consider using GitHub's encrypted secrets for better security.

**Suggestion:**
```
Use GitHub's encrypted secrets to store OCI credentials and reference them securely in the workflow.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Workflow name consistency</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 43)

The workflow name 'Soak Dashboard Pages' could be more consistent with other workflow names in the repository.

**Suggestion:**
```
Consider renaming the workflow to match the naming conventions used in other workflows, e.g., 'Deploy Soak Dashboard'.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Verify OCI installer URL/checksum are in scope and pinned</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 376)

The install step trusts OCI_CLI_INSTALLER_URL and OCI_CLI_INSTALLER_SHA256 without defining them in this job. If they are not workflow-level env values (pinned URL + checksum), the step will fail open on unset vars or drift from the checksum-verified pattern used elsewhere.

**Suggestion:**
```
Ensure both values are defined at workflow env scope (or in this step) and remain pinned; fail closed if unset before curling.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Inconsistent bash strictness across notify_failure steps</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 375)

Install/configure/cleanup use `bash -e` while Publish failure alert uses `bash -euo pipefail`. The looser modes can miss unset vars or pipeline failures in credential setup.

**Suggestion:**
```
Use `bash -euo pipefail` consistently on all notify_failure run steps.
```

</details>
---

<details>

The code changes look good overall, addressing the soak dashboard deployment and failure notification. Minor improvements can be made in handling OCI credentials more securely and ensuring consistent workflow naming conventions.

#### openai (gpt-5.6-sol)

The notification design and schedule-gate annotation are generally sound, but the Pages bootstrap has a significant data-retention flaw: any transient failure while reading the currently published history can cause the next deployment to erase it. The alert condition also omits cancelled real runs. The history preservation behavior should be made fail-safe before merging.

#### xai (grok-4.5)

Solid in-repo half of the soak operability fix: schedule-gate no-ops are clearly annotated, failure notification is correctly scoped with always() + should_run, and OCI user-principal auth on a GitHub-hosted runner is the right way to cover launch failures that instance-principal cannot. The Pages bootstrap workflow is thoughtfully designed to preserve published data with safe history/report fallbacks. Before merge, verify Pages deploys share the soak-pages concurrency lane with perf_report (to avoid stale overwrites) and that ephemeral-launch-internal will not gate alert jobs on manual approval; consider tightening alert wording for dashboard-only failures and adding empty defaults for the remaining JSON artifacts.
</details>

---

spreston8:
Reviewed the full diff plus the surrounding workflow on master. This is a well-engineered change — checksum-pinned installer, env-indirection against expression injection, fail-closed on ambiguity, deliberate Pages-deploy serialization, credential cleanup. The notes below are one requested change and a few clarifying questions, not objections to the approach.

## What works

- The schedule-gate no-op annotation kills a real footgun: the gate no-ops most cron deliveries by design, and a green check on those runs reads as a soak result.
- `notify_failure` running GitHub-hosted with user-principal auth is the right call for the stated reason — instance-principal can't cover the runner-never-launched case. Failing loud (`exit 1`) when `SOAK_ONS_TOPIC_OCID` is unset is exactly right given that a silently-unset topic is what this PR is fixing.
- The bootstrap deploy's data-preservation logic is the best part of the PR: only a 404 counts as "nothing published", any other status aborts rather than risking an empty-dashboard overwrite, and `history.json` is validated as a JSON array before being accepted. Both this job and `perf_report` share the `soak-pages` concurrency group with `cancel-in-progress: false`, so the two workflows' Pages deploys serialize rather than race.

## Requested change

**Backport the defensive fetch to `perf_report`.** The new workflow's comment explains why treating a transient fetch failure as "no data" would republish an empty dashboard over real soak history — but `perf_report`'s own history fetch still does exactly that:

```
curl -fsS --max-time 30 "${DASHBOARD_URL}data/history.json" \
  -o /tmp/history.json 2>/dev/null || echo '[]' > /tmp/history.json
```

Any outage, timeout, or 5xx during the weekly run resets the published history to just the current run, and since the Pages site is the only durable history store (artifacts expire after 14 days), that loss is permanent. As written, the careful path protects only the rarely-run bootstrap deploy while the weekly production path keeps the hazard. Applying the same 200/404/abort pattern there would close it.

## Questions

1. **Double email on weekend failures — intended?** The in-soak step sends the weekly verdict email on every weekend run (`if: always() && run_benchmarks`), subject carrying the verdict ("F1R3FLY weekend soak: FAIL"). A failed weekend soak now produces that email plus the new failure alert. Defensible (different content, and `notify_failure` covers cases the in-soak step can't reach) — but if it's intended, a comment saying so would stop someone from "deduplicating" it later; if not, the in-soak send could be limited to pass verdicts.

2. **Queued-job-cancelled edge.** The alert condition matches only `'failure'`, which looks deliberate — `cancel-in-progress: true` cancels superseded runs and those shouldn't alert. But if the launched VM dies before registering, the `soak` job sits queued until GitHub kills it (~24h). If GitHub resolves that as `cancelled` rather than `failed`, no alert ever fires for a genuinely dead soak — the exact pathology this job exists to cover. Worth a comment in the workflow documenting which way that resolves (or a run-level guard if it resolves as cancelled).

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The notification workflow is appropriately scoped, avoids direct shell interpolation of GitHub context values, and handles OCI credentials carefully. The main concern is the dashboard preservation design: reconstructing canonical state from a cached public Pages endpoint can regress or erase newer soak results during cache-propagation or concurrent deployment scenarios. The deployment mechanism should use a durable authoritative data source or stronger cross-workflow serialization and version checks before merge.

Solid in-repo half of the soak operability fix: the Pages bootstrap workflow carefully refuses to clobber live history on non-404 fetch failures, and notify_failure correctly runs on GitHub-hosted runners with user-principal OCI auth so launch failures are covered. Main follow-ups before relying on this in production are verifying ephemeral-launch-internal will not gate the alert job on reviewers, deciding whether cancelled soaks should email stakeholders, and confirming perf_report shares the soak-pages concurrency group so dashboard deploys cannot race.

The code changes look good overall. There are minor improvements that can be made in terms of security, style, and documentation. The workflow changes are functional and safe to merge with the suggested improvements.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Restoring state from the public Pages URL can overwrite newer soak data</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 45)

On 404, latest-summary.json and latest-verdict.json are deleted and never replaced. history.json and latest-report.md get empty-state defaults, but the other two stay absent. If the dashboard JS unconditionally fetches/parses them, the bootstrap Pages deploy can look broken until the first weekend perf_report runs.

---

The workflow treats responses from the public GitHub Pages URL as the canonical dashboard state and then republishes them. Pages is CDN-backed, so these requests can return a cached version even after a newer perf_report deployment has completed. There is also a read-modify-deploy race if perf_report publishes between these downloads and this workflow's deployment unless that publisher uses the exact same repository-wide concurrency group. In either case, a merge-triggered bootstrap deployment can replace newer history and reports with stale or empty data, defeating the stated preservation guarantee.

**Suggestion:**
```
After the fetch loop, seed minimal valid empty-state JSON for latest-summary.json and latest-verdict.json when missing (matching whatever shape the dashboard expects), same as history.json and latest-report.md.
```

</details>

<details>
<summary><b>:orange_circle: Major: Failure alerts may block on environment protection rules</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 372)

The condition contains(needs.*.result, 'failure') ignores cancelled. A weekend 60h soak cancelled mid-run (timeout, manual cancel, or runner preemption) produces no ONS alert even though should_run was true and no successful soak completed.

---

notify_failure uses environment: ephemeral-launch-internal. If that environment has required reviewers, wait timers, or branch restrictions, the alert job will queue for approval instead of sending immediately — the exact window where you need unattended notification (including launch_runner failures). This would silently defeat the purpose of the job.

**Suggestion:**
```
Confirm ephemeral-launch-internal has no protection rules that require human approval. If it does, use a dedicated unprotected environment for alert-only secrets, or gate only the secret-bearing steps without deployment protection.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Secrets Management</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 112)

Ensure that secrets are properly managed and not exposed in logs or error messages. Consider using GitHub's secret scanning feature to detect any potential leaks.

**Suggestion:**
```
Review the usage of secrets and ensure they are not logged or exposed in any way.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Workflow Name Consistency</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 20)

Ensure that workflow names are consistent and descriptive. The name 'Soak Dashboard Pages' could be more specific.

**Suggestion:**
```
Consider renaming the workflow to something more descriptive, such as 'Deploy Soak Dashboard to GitHub Pages'.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Only the history file is validated before republishing</b> (Reported by: openai)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 58)

The workflow validates `history.json`, but any HTTP 200 response for `latest-summary.json` or `latest-verdict.json` is accepted without checking that it is valid JSON. A truncated response or an intermediary-generated HTML response with status 200 would then be published as dashboard data.

**Suggestion:**
```
Validate all downloaded JSON files with `jq -e` and reject empty or malformed responses before uploading the Pages artifact.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Branch Strategy Documentation</b> (Reported by: bedrock)</summary>

**File:** `CLAUDE.md` (line 180)

The branch strategy documentation could be more detailed and clear. Consider adding examples or a diagram to illustrate the workflow.

**Suggestion:**
```
Add a section with examples or a diagram to clarify the branch strategy.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Branching and Commits Documentation</b> (Reported by: bedrock)</summary>

**File:** `CONTRIBUTING.md` (line 15)

The branching and commits documentation could be more detailed. Consider adding examples or a diagram to illustrate the workflow.

**Suggestion:**
```
Add a section with examples or a diagram to clarify the branching and commit strategy.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: OCI key cleanup is good; consider restricting job permissions</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 396)

Checksum-pinned OCI CLI install, umask 077 key write, auth probe, and always() shred/rm of ~/.oci are solid. The job does not set permissions:, so it inherits the workflow default GITHUB_TOKEN scope (often contents: write on many repos). This job does not need the repo token at all.

**Suggestion:**
```
Set permissions: {} (or contents: read only if required) on notify_failure to least-privilege the GITHUB_TOKEN even though the sensitive material here is OCI user-principal.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Confirm perf_report shares concurrency group soak-pages</b> (Reported by: xai)</summary>

**File:** `.github/workflows/soak-dashboard-pages.yml` (line 26)

This workflow correctly uses concurrency.group: soak-pages with cancel-in-progress: false to avoid clobbering deploys. That only serializes against other jobs using the same group. If perf_report deploys Pages under a different group (or none), a master push during/after a weekend soak can still race and overwrite history.

**Suggestion:**
```
Verify the soak perf_report Pages deploy uses group: soak-pages and cancel-in-progress: false as well.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Branch-strategy docs bundled with soak infra fix</b> (Reported by: xai)</summary>

**File:** `CONTRIBUTING.md` (line 17)

CLAUDE.md / CONTRIBUTING.md / SECURITY.md branch-model updates are accurate and internally consistent, but unrelated to the soak Pages/ONS fix. Mixing them makes revert/blame noisier if either change needs to be rolled back alone.

**Suggestion:**
```
Prefer a separate docs PR next time; not a merge blocker.
```

</details>
---

<details>

The code changes look good overall. There are minor improvements that can be made in terms of security, style, and documentation. The workflow changes are functional and safe to merge with the suggested improvements.

#### openai (gpt-5.6-sol)

The notification workflow is appropriately scoped, avoids direct shell interpolation of GitHub context values, and handles OCI credentials carefully. The main concern is the dashboard preservation design: reconstructing canonical state from a cached public Pages endpoint can regress or erase newer soak results during cache-propagation or concurrent deployment scenarios. The deployment mechanism should use a durable authoritative data source or stronger cross-workflow serialization and version checks before merge.

#### xai (grok-4.5)

Solid in-repo half of the soak operability fix: the Pages bootstrap workflow carefully refuses to clobber live history on non-404 fetch failures, and notify_failure correctly runs on GitHub-hosted runners with user-principal OCI auth so launch failures are covered. Main follow-ups before relying on this in production are verifying ephemeral-launch-internal will not gate the alert job on reviewers, deciding whether cancelled soaks should email stakeholders, and confirming perf_report shares the soak-pages concurrency group so dashboard deploys cannot race.
</details>

---

jeffrey-l-turner:
## Review triage: 4 findings actioned, 6 dismissed

Each finding from the multi-agent review above was checked against the code rather than actioned on its own authority. Four were real and are fixed on this branch; six did not survive verification. Evidence for each dismissal is below so the calls can be challenged.

### Fixed

**1. Stale CDN reads could roll back published soak history** (critical — openai, xai)
GitHub Pages is CDN-backed, so a `200` can serve a copy cached from before the weekend soak's last publish, which this workflow would then republish over the newer data. Fetches now carry a per-run cache-busting query string (`?cb=<run_id>-<attempt>`) plus `Cache-Control: no-cache`, forcing a cache miss so current state is always read. Note this was **not** covered by the earlier fail-closed fix, which only distinguished HTTP status codes — a stale `200` and a fresh `200` are indistinguishable by status alone.

**2. Cancelled runs sent no alert** (major — xai)
`contains(needs.*.result, 'failure')` does not match `cancelled`, so a 60h soak that hit its timeout, was cancelled by hand, or lost its runner to preemption notified nobody, which is indistinguishable from success. The condition now matches `cancelled` as well.

**3. A `200` was trusted without validating its content** (minor — openai)
Only `history.json` was checked, and only for being an array. A truncated response or a proxy error page returned with `200` would have been republished as data. All three JSON artifacts are now parsed with `jq` before republishing, and the run aborts if any fails.

**4. Alert job held an unnecessary token scope** (suggestion — xai)
The reported risk did not apply here — the workflow already sets `permissions: contents: read` at top level, not `write`. But the observation pointed at something real: this job has no checkout and never calls the GitHub API, so it now declares `permissions: {}` and holds no token scope at all.

Alert titles now name what actually broke (`runner launch`, `soak`, or `dashboard publish`) instead of labelling every case a soak failure — a dashboard-only failure after a clean soak should not read as an incident.

Verified against a local fixture server across five cases: valid `200` preserves history; `404` bootstraps to empty; `503`, connection failure, and `200`-carrying-HTML all abort without publishing.

### Dismissed, with evidence

- **"Missing `latest-summary.json` / `latest-verdict.json` breaks the dashboard."** `index.html` never references `latest-summary.json` — its only consumers are the two workflows — and the `latest-verdict.json` fetch is inside a `try/catch`. A missing file leaves the verdict badge blank; nothing errors.
- **"Deploy race unless `perf_report` shares the concurrency group."** It does: `group: soak-pages`, `cancel-in-progress: false`, identical to the new workflow. Raised in both review rounds; the in-Actions race is genuinely closed. (The CDN staleness above is a separate problem, and is what got fixed.)
- **"`ephemeral-launch-internal` may gate alerts behind required reviewers."** That environment has `protection_rules: []`. Alerts fire unattended.
- **"OCI installer URL/checksum may be unset in this job."** Both are defined at workflow `env:` scope and inherit into every job.
- **"OCI credentials stored in plaintext; use encrypted secrets."** They *are* encrypted secrets. The OCI CLI requires a config file on disk; the job writes it under `umask 077` and shreds it in an `always()` step, matching the existing `launch_runner` pattern.
- **"Workflow name could be more specific" / "add diagrams to the branch-strategy docs."** Subjective; no defect identified.

### Fair call not actioned

One reviewer noted the branch-model doc changes (`CLAUDE.md`, `CONTRIBUTING.md`, `SECURITY.md`) are unrelated to the soak fix and make revert/blame noisier. That is correct as a scoping observation. They are kept here deliberately, as separately-scoped commits so either can be reverted alone.

### Caveat on this review

Anthropic abstained in both rounds on an API billing error, not a safety decline, so consensus rests on three of four providers and Claude never reviewed this change.

jeffrey-l-turner:
@spreston8 — thanks, the requested change was a real bug and a fair catch. All three points are addressed below.

### ✅ Requested change — backport the defensive fetch to `perf_report` — **fixed** in `af80fbc1`

You were right, and right about the severity: I fixed the bootstrap deploy and left the hazard in the path that actually matters. Reproduced against a fixture server before changing anything — under HTTP 503 the old code exits **0** and publishes `["newweek"]`, silently dropping every prior week:

```
OLD CODE under transient 503 => exit=0  published history = ["newweek"]
```

`perf_report` now uses the same 200/404/abort pattern, plus the cache-busting query string (it had the stale-CDN exposure too) and array validation. Verified across four cases:

| Case | Result |
|---|---|
| Existing history, HTTP 200 | exit 0, appends — `["deadbeefcafe","newweek"]` |
| Nothing published, HTTP 404 | exit 0, starts series — `["newweek"]` |
| Transient 503 | exit 1, aborts, nothing published |
| HTTP 200 carrying an HTML error page | exit 1, aborts, nothing published |

Your point about artifacts expiring at 14 days is what makes this worth failing closed over: Pages is the only durable copy, so the loss was permanent rather than recoverable from a re-run.

### ✅ Question 2 — queued-job-cancelled edge — **already fixed** in `7b19a124`

Your review landed about an hour before that commit, so it predates the fix. The condition now reads:

```yaml
(contains(needs.*.result, 'failure') || contains(needs.*.result, 'cancelled'))
```

That covers exactly the pathology you described — VM dies before registering, `soak` sits queued until GitHub kills it, and GitHub resolves it as `cancelled`. It also covers a 60h soak hitting its `timeout-minutes` and manual cancellation. Superseded runs still alert under this, which is the deliberate trade: a spurious alert is cheaper than silence on a genuinely dead soak.

### ✅ Question 1 — double email on weekend failures — **intended, now documented**

Intentional, and your instinct to pin it down was right — it reads like an oversight otherwise. Documented in-workflow so nobody deduplicates it later. The reasoning: the in-soak step runs *on the soak runner* and only once a report exists, so it structurally cannot report a runner that never launched or a soak that died before writing one. `notify_failure` exists for precisely those cases and runs GitHub-hosted. A weekend soak that completes and then fails its verdict gate does send both — the verdict with numbers, and the job-level alert — and that overlap is the accepted cost of never going silent.

### Two further fixes your review surfaced indirectly

- **Baseline fetch was also stale-prone.** `latest-summary.json` is fetched uncached for the week-over-week deltas, so a stale CDN copy meant comparing against the wrong week. Now cache-busted. Kept soft-failing — a 60h soak's report outweighs its comparison — but it now emits a `::warning::` on a non-404 miss, since a silently absent baseline makes the verdict look clean.
- **`perf_report` failures were mislabelled.** That job runs both the Pages deploy *and* the regression verdict gate, so a regression was being emailed as a "dashboard publish" failure. Alerts now say `report/verdict stage`.

Nothing here is left open from your review. Since these came in as a top-level comment rather than an inline review, there are no conversation threads to resolve — this comment is the resolution record.

spreston8 (approved):

PR #162, opened 2026-07-29
fix(ci): raise the link-checker and compartmentalize the OCI security with the CI

jeffrey-l-turner:
Not this process for the PR since it will properly compartmentalize OCI credentials and offers a more robustly secure CI implementation

 1. PR #162 merges into dev and post-merge CI is green.
 2. The migration is promoted to master.
 3. Post-merge master CI—including the release workflow—is green.
 4. Confirm no protected-branch workflow still references RELEASE_PAT or repository-level OCI_* secrets.

 Only then remove the old repository secrets. The oci-credentials environment secrets remain. Then merge dev into master and promote to production.


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :mag: Needs Review (66.6700% agreement)

**Reviewed by:**

| xai | `grok-4.5` | Needs Review | 0.84 |

---

### Summary

The migration from a long-lived administrator PAT to short-lived, repository-scoped GitHub App tokens is a meaningful security improvement, and the pinned token action and explicit permissions look appropriate. However, the release workflow now uses the `oci-credentials` environment solely to obtain GitHub App credentials, broadening its authorization boundary to unrelated OCI secrets and weakening the intended compartmentalization. The environment design should be separated before the old repository secrets are removed.

This PR meaningfully hardens CI by replacing long-lived RELEASE_PAT usage with short-lived GitHub App tokens, moving OCI/App secrets behind the oci-credentials environment, dropping secret plumbing from the reusable integration workflow, and tightening branch/tag triggers while fixing flaky link checks via a higher lychee timeout. The design matches the stated trust model (no fork code in credential-bearing jobs; approval gates before launch). I would not rubber-stamp merge solely from the diff: release.yml’s use of oci-credentials over-scopes release to cloud credential environment policy, token-in-remote-URL and github-script interpolation remain footguns, and success depends on App permissions, ruleset bypass, environment secret placement, and a full purge of RELEASE_PAT/repo-level OCI_* references outside the shown hunks. Approve the approach after those operational checks and a quick confirmation that the narrowed heavy-pipeline conditions still satisfy required checks for your branch topology.

The changes improve the security of the CI pipeline by replacing long-lived personal access tokens with short-lived GitHub App tokens. The increased timeout for the link checker is justified and should not impact performance significantly. Overall, the changes look good and are safe to merge.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Release job is granted access to the OCI credentials environment</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/release.yml` (line 75)

The release job only needs the GitHub App credentials, but assigning it to `oci-credentials` places it within the authorization boundary containing the OCI secrets as well. GitHub does not automatically inject every environment secret into each step, but the workflow is authorized to reference them, so a future compromised or accidentally modified release step could access unrelated cloud credentials. This undermines the stated goal of compartmentalizing OCI credentials and also unnecessarily couples release execution to the protection and deployment rules of the OCI environment.

---

The version bump/tag job now uses environment: oci-credentials even though it only needs the GitHub App signing key (contents:write), not OCI tenancy material. That couples release publishing to OCI environment protection rules and expands who/what must be trusted to touch release automation. It also makes least-privilege reviews harder: anyone approved for release effectively gets a path to the same environment that holds cloud credentials.

**Suggestion:**
```
Create a separate protected environment such as `release-credentials` or `github-app-credentials` containing only `CI_GITHUB_APP_ID` and `CI_GITHUB_APP_PRIVATE_KEY`, and use it for this job. Consider distinct environments for release, runner management, and OCI-only maintenance so each workflow receives only the credentials it requires.
```

</details>

<details>
<summary><b>:orange_circle: Major: Use of short-lived GitHub App token</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 229)

The change correctly replaces the use of `RELEASE_PAT` with a short-lived GitHub App token for runner registration. This is a security improvement as it reduces the risk associated with long-lived personal access tokens.

**Suggestion:**
```
Ensure that the GitHub App has the necessary permissions and that the private key is securely managed.
```

</details>

<details>
<summary><b>:orange_circle: Major: Migration completeness depends on unseen repo/environment state</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 280)

This PR removes workflow_call secrets (OCI_* and RELEASE_PAT) and switches privileged jobs to environment-scoped GitHub App credentials. That is directionally correct, but correctness is operational: CI_GITHUB_APP_ID/CI_GITHUB_APP_PRIVATE_KEY and OCI secrets must already exist on environment oci-credentials; the App must be allowed to mint runner registration tokens; and release rulesets must list the App as a bypass actor. Any remaining workflow still reading RELEASE_PAT or repository-level OCI_* will break or silently keep the old trust model. The diff does not prove reaper/other workflows fully dropped RELEASE_PAT usage.

**Suggestion:**
```
Before merge, grep the workflows for RELEASE_PAT and repository-level OCI_* references; confirm environment secrets/vars; confirm App permissions (administration for runners, contents for release); confirm ruleset bypass; and run the PR description’s staged promotion checklist before deleting old repo secrets.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Increased timeout for link checker</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/ci.yml` (line 206)

The timeout for the link checker has been increased from the default 20 seconds to 45 seconds. This is a reasonable change given the explanation provided in the comments.

**Suggestion:**
```
Consider adding a comment explaining the rationale for the increased timeout to improve clarity for future maintainers.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Heavy pipeline no longer runs on arbitrary branch pushes</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 319)

The pipeline/aggregator guards narrowed from “any push / non-promotion internal PR” to pushes on dev, master, or refs/tags/v*, and internal PRs into dev/master only (excluding heads named dev/master). Feature-branch pushes and removed branches (staging/trying) will no longer exercise the OCI-heavy path via ci.yml. That looks intentional with the branch-model cleanup, but it is a behavior change worth confirming against required checks and developer expectations.

**Suggestion:**
```
Double-check branch protection required checks still pass for the PR topologies you care about (feature → dev, dev → master, tag releases). Update contributor docs if staging/trying are fully retired.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: App token interpolated into git remote URL</b> (Reported by: xai)</summary>

**File:** `.github/workflows/release.yml` (line 163)

The short-lived App token is still embedded in the origin URL (https://x-access-token:${{ steps.app-token.outputs.token }}@...). This matches the previous PAT pattern, but any step that prints remotes, curl error URLs, or verbose git output can leak the token into logs for the remainder of its lifetime.

**Suggestion:**
```
Prefer git push https://x-access-token:${TOKEN}@github.com/... with the token only in the step env, or use a credential helper that does not persist the secret in remote config. Avoid set-url that leaves the token in local repo config for later steps.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: github-script interpolates TAG_NAME into JS source</b> (Reported by: xai)</summary>

**File:** `.github/workflows/release.yml` (line 171)

const tagName = '${{ steps.version.outputs.TAG_NAME }}' is expression-injected into JavaScript. If TAG_NAME is ever attacker-influenced or malformed, this is classic Actions script injection. Risk is low if the version step only emits strict semver you control, but the pattern is brittle.

**Suggestion:**
```
Pass the tag via env (e.g. env: { TAG_NAME: ${{ steps.version.outputs.TAG_NAME }} }) and read process.env.TAG_NAME inside the script.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: administration:write App token is powerful but short-lived</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 286)

Runner registration correctly requires admin-capable credentials; minting a short-lived GitHub App token with permission-administration: write is much better than a long-lived RELEASE_PAT. Residual risk is that any code execution in that job can use admin-equivalent API access until expiry.

**Suggestion:**
```
Keep the existing trust boundary strict: no PR/fork checkout in this job, minimal steps after token mint, and rely on environment approval gates. Consider minting the token immediately before use and avoiding writing it to durable files.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Link-checker timeout increase is well justified</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 226)

Raising lychee --timeout to 45 with a detailed rationale for doi.org latency is reasonable and unlikely to hide real 404/410 failures given --accept and retry settings. No functional concern.

**Suggestion:**
```
Optional: add a brief note in contributor docs so future timeout flakes are not “fixed” by excluding DOI links.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Approval then credential split is a good pattern</b> (Reported by: xai)</summary>

**File:** `.github/workflows/oci-validation.yml` (line 146)

Splitting await_approval (ephemeral-launch) from launch_ephemeral_runners (oci-credentials) improves compartmentalization versus a single privileged job. Same direction as moving reusable-pipeline launch onto oci-credentials and dropping workflow_call secret plumbing.

**Suggestion:**
```
Mirror this two-phase pattern anywhere else that still combines manual gate + cloud credential use in one environment.
```

</details>
---

<details>

The changes improve the security of the CI pipeline by replacing long-lived personal access tokens with short-lived GitHub App tokens. The increased timeout for the link checker is justified and should not impact performance significantly. Overall, the changes look good and are safe to merge.

#### openai (gpt-5.6-sol)

The migration from a long-lived administrator PAT to short-lived, repository-scoped GitHub App tokens is a meaningful security improvement, and the pinned token action and explicit permissions look appropriate. However, the release workflow now uses the `oci-credentials` environment solely to obtain GitHub App credentials, broadening its authorization boundary to unrelated OCI secrets and weakening the intended compartmentalization. The environment design should be separated before the old repository secrets are removed.

#### xai (grok-4.5)

This PR meaningfully hardens CI by replacing long-lived RELEASE_PAT usage with short-lived GitHub App tokens, moving OCI/App secrets behind the oci-credentials environment, dropping secret plumbing from the reusable integration workflow, and tightening branch/tag triggers while fixing flaky link checks via a higher lychee timeout. The design matches the stated trust model (no fork code in credential-bearing jobs; approval gates before launch). I would not rubber-stamp merge solely from the diff: release.yml’s use of oci-credentials over-scopes release to cloud credential environment policy, token-in-remote-URL and github-script interpolation remain footguns, and success depends on App permissions, ruleset bypass, environment secret placement, and a full purge of RELEASE_PAT/repo-level OCI_* references outside the shown hunks. Approve the approach after those operational checks and a quick confirmation that the narrowed heavy-pipeline conditions still satisfy required checks for your branch topology.
</details>

---

jeffrey-l-turner:
Responding to the multi-agent review. All three release-workflow findings are addressed on the branch; one operational item remains open and is called out at the bottom because it blocks the secret-deletion step, not the merge.

## Critical — release job over-scoped to the OCI credential environment

Fixed in `6186ae86`. The finding was correct: the release job needs only the App credentials, but naming `oci-credentials` put the five OCI secrets inside its authorization boundary — the same over-broad grant this PR exists to remove, just relocated.

- New `release-credentials` environment holding **only** `CI_GITHUB_APP_ID` and `CI_GITHUB_APP_PRIVATE_KEY`.
- Deployment branch policy: `master` only. Tighter than the credential environment it replaces — `release.yml` triggers exclusively on `push: master`, so nothing else can claim it.
- `release.yml:75` now names `release-credentials`.

Deliberately **not** in this PR, to keep it from turning into a credential-architecture redesign: a separate `runner-management-credentials`, a separate `oci-maintenance` environment, and distinct GitHub Apps for release versus runner administration. Those are follow-up work. This change resolves the concrete over-scoping introduced by this migration.

## Minor — App token persisted in the git remote URL

Fixed in `bd5cb0f8` (`release.yml:157`). `git remote set-url origin https://x-access-token:${TOKEN}@…` is gone. The token now lives only in the step's `env:` as `RELEASE_TOKEN` and reaches git through a `GIT_ASKPASS` helper written to `$RUNNER_TEMP` (mode 700, `trap … EXIT` removal), with `GIT_TERMINAL_PROMPT=0` and a credential-free push URL. Nothing durable in `.git/config`, so later steps that print remotes or emit verbose git output cannot leak it.

## Minor — `TAG_NAME` interpolated into `github-script` JS

Fixed in `bd5cb0f8` (`release.yml:179`). `const tagName = '${{ … }}'` became `process.env.TAG_NAME`, with the value passed via step `env:`. No expression reaches the JS source.

## Major — "migration completeness depends on unseen repo/environment state"

Fair, and not provable from the diff. Verified against live repository state:

| Check | Result |
|---|---|
| `RELEASE_PAT` references anywhere in `.github/` | zero |
| `secrets.OCI_*` in a job **not** naming a credential environment | none — every consumer in `_integration-pipeline.yml`, `ci-runner-reaper.yml`, `merge-recovery-soak.yml` (×2), `oci-validation.yml` is environment-scoped |
| App as ruleset bypass actor | Integration `4428544` present on both `devProtect` and `masterProtect` — the push to protected `master` will succeed |
| App can mint runner registration tokens | proven end-to-end: run `30495738655` completed `Create runner-management token` → `Launch 2 amd64 + 2 arm64 ephemeral runners`, and the full integration matrix ran on those runners |
| Approval gate intact | `await_approval` → `needs:` chain unchanged; `Await Launch Approval` passes ahead of every launch |
| `vars.CI_GITHUB_APP_ID` reachable from both environments | yes — present in `oci-credentials` and `release-credentials`; absent at repository level |

The narrowed heavy-pipeline trigger conditions were also exercised by that run: `feature → dev` produced `Integration Tests (amd64)` and `(arm64)` as required checks on the PR head.

## Still open — do not delete the repository-level `OCI_*` secrets yet

The environment-scoped OCI values are not yet correct, and `secrets: inherit` is currently masking that:

- Run `30495007422` (no `secrets: inherit`) **failed** at `Configure OCI auth from secrets` with `tenancy | malformed`, `user | malformed`, `fingerprint | malformed`.
- Run `30495738655` — same environment values, untouched since they were stored — **passed** that identical step. The only change between the two runs was `9e32b628` adding `secrets: inherit` to both callers.

If the environment copies were the values in use, making the repository-level copies reachable could not have changed the outcome. So the green pipeline is running on the repository-level values, and the `oci-credentials` copies are still malformed — most likely stored as whole `key=value` lines rather than bare values, which is why `region` (not OCID-validated) was the one scalar the CLI did not flag.

Consequences:

1. **Step 4 of the promotion checklist must wait.** Deleting the repository-level `OCI_*` secrets today breaks the heavy pipeline on the next run. `RELEASE_PAT` is a different matter — nothing references it any more, so it can go whenever.
2. **`secrets: inherit` should come back out once the environment values are fixed.** On `ci-fork-pr.yml:77` it hands the whole repository secret set to a `pull_request_target` pipeline. No fork-executed job references a secret — `_integration-pipeline.yml` touches secrets only in `launch_ephemeral_runners`, which checks out solely the pinned `system-integration` SHA — so there is no exposure today, but it widens the boundary and contradicts that file's own header comment. It is a workaround for the malformed values, not part of the design.

Verification sequence once the environment secrets are re-stored: revert `secrets: inherit` on both callers, confirm a green `Launch Ephemeral Runners`, and only then delete the repository-level `OCI_*` copies. That order proves the environment scoping is real exclusivity rather than a duplicate of the repository-level values.

jltatbeach (approved):

jeffrey-l-turner:
## Correction: `secrets: inherit` is required, not a workaround

My earlier comment on this PR said `secrets: inherit` was masking malformed environment values and should be removed in a follow-up. **That was wrong, and acting on it would break the heavy pipeline.** Superseding it here so nobody follows the old advice.

### What actually happens

A called workflow's jobs do not receive `secrets.*` merely by naming an `environment:`. The caller must pass secrets down. Without `secrets: inherit`, the launch job's `OCI_*` expressions resolve to empty strings — and an empty value produces the *identical* OCI CLI error as a genuinely malformed one:

```
| tenancy     | malformed | this must be an OCID
| user        | malformed | this must be an OCID
| fingerprint | malformed | ...
```

That ambiguity is what sent the earlier diagnosis down the wrong path. The evidence:

| Run | `secrets: inherit` | OCI auth step | Cause |
|-----|-----|-----|-----|
| [30495007422](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30495007422) | absent | failed | values arrived empty |
| [30495738655](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30495738655) | present | passed | values delivered |
| [30498890544](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30498890544) | absent | failed | empty again, *with correct values in the environment* |

The third run is decisive: the environment secrets had been re-stored and verified against the live OCI API immediately beforehand, and it still failed. Delivery, not value correctness, was the problem all along.

### Confirmation that the environment values are the ones in use

Run [30499363901](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30499363901) reported a byte length and truncated SHA-256 for each credential. All five matched the values stored in the `oci-credentials` environment, and the launch job completed end to end — auth, App token, four ephemeral runners, full integration matrix.

One caveat stated plainly: matching digests cannot *prove* the environment copies won, because the repository-level copies may hold the same values. What is established is that delivery works and the values are well-formed rather than empty.

### Consequence for the secret-deletion step

Deleting the repository-level `OCI_*` copies is still the right end state and is safe to attempt, because it is reversible — if a run then fails, re-uploading from a local `~/.oci/config` restores working credentials. By GitHub's documented precedence (environment secrets shadow repository secrets of the same name) it should be a no-op. `RELEASE_PAT` is unconditionally safe to delete; nothing references it any more.

### What changed on the branch

- `08a31a8b` restored `secrets: inherit` on both callers, with comments citing the run IDs above so this is not re-derived a third time.
- `0e42bd03` promoted the credential-delivery probe from a temporary diagnostic to a permanent step, `Report credential delivery`. It prints only lengths and truncated digests — GitHub masks secret values in logs but not digests, which is exactly what makes `len=0` diagnostic. It costs five `printf`s and removes the empty-vs-malformed ambiguity permanently.

There is no longer any step in this PR marked "remove before merging".

### Follow-up branches

`chore/ci-workflow-guardrails` adds a fail-closed invariant check wired into `Lint` (a required status check), asserting among other things that both callers pass `secrets: inherit` — so this specific regression cannot recur silently. It also adds `.github/CODEOWNERS` naming `@F1R3FLY-io/ci-maintainers`, which now exists with write access, with `require_code_owner_review` enabled on both rulesets.

`fix/split-ci-credential-envs` carries the remaining compartmentalization: an `oci-maintenance` environment (created and populated) for the reaper, which needs OCI credentials but not the App key, and a second GitHub App so runner administration no longer shares an identity with release pushes to protected `master`.

jeffrey-l-turner:
## Post-merge runbook

Ordered because several steps break CI if done early. Steps 1–3 concern this PR; 4–7 are the two follow-up branches. Tick these off as they land.

### 1. Merge this PR into `dev`

Nothing is marked "remove before merging" any more — the credential-delivery probe is a permanent step (`0e42bd03`).

### 2. Confirm post-merge `dev` CI is green

Specifically that `Launch Ephemeral Runners → Report credential delivery` prints non-zero lengths and `Configure OCI auth from secrets` passes. `len=0` on any value means delivery regressed.

### 3. Delete the superseded repository-level secrets

Only after step 2:

```bash
for s in OCI_FINGERPRINT OCI_PRIVATE_KEY OCI_REGION OCI_TENANCY_OCID OCI_USER_OCID RELEASE_PAT; do
  env -u GITHUB_TOKEN gh secret delete "$s" --repo F1R3FLY-io/f1r3node-rust
done
```

Keep `OCIR_*` and `DOCKERHUB_*` — `release_docker_image` still reads them at repository level.

`RELEASE_PAT` is unconditionally safe: nothing in `.github/` references it. The five `OCI_*` are the reversible experiment — if the next run reports `len=0`, re-upload them from a local `~/.oci/config` and the pipeline recovers. By GitHub's documented precedence (environment secrets shadow repository secrets of the same name) this should be a no-op, and the delivery report in step 2 is what proves it.

### 4. `chore/ci-workflow-guardrails` → PR

Already pushed. Contains `.github/CODEOWNERS` (owned by `@F1R3FLY-io/ci-maintainers`) and `.github/scripts/check-workflow-invariants.sh`, wired into `Lint` because `Lint` is a required status check — a guard living in a job that is merely a `needs:` ancestor can surface as `skipped`, which required-check evaluation may treat as satisfied.

Open it **after** this PR merges, otherwise its diff carries every commit from here.

Two things it will tell us:

- Whether `require_code_owner_review: true` has teeth at `required_approving_review_count: 0`. This PR touches owned paths, so if it merges without a code-owner approval, the count needs raising to 1.
- Whether the invariant script passes in CI, not just locally.

### 5. `fix/split-ci-credential-envs` → PR

The credential compartmentalization. Prerequisites already done out of band: `oci-maintenance` environment created and populated, and a second GitHub App registered (`administration: write` only, installed on this repo, **not** a ruleset bypass actor) with `RUNNER_ADMIN_APP_ID` / `RUNNER_ADMIN_APP_PRIVATE_KEY` stored in `oci-credentials`.

Changes four workflows:

| Workflow · job | Change |
|---|---|
| `ci-runner-reaper · reap` | → `oci-maintenance`; terminates instances via the OCI API, never registers a runner |
| `merge-recovery-soak · notify_failure` | → `oci-maintenance`; publishes to the ONS topic only |
| `merge-recovery-soak · launch_runner` | App refs → `RUNNER_ADMIN_APP_*` |
| `oci-validation · launch_ephemeral_runners` | App refs → `RUNNER_ADMIN_APP_*` |
| `_integration-pipeline · launch_ephemeral_runners` | App refs → `RUNNER_ADMIN_APP_*` |

`release.yml` is untouched and keeps `CI_GITHUB_APP_*` in `release-credentials` — it is the only consumer requesting `permission-contents: write`.

The reaper is the change that matters most: it runs unattended every 30 minutes with no approval gate, making it the widest ungated credential surface in the repository.

### 6. After step 5 merges green — shrink the release App

**Order matters.** Doing this before step 5 merges breaks runner registration everywhere.

- Remove `Administration` from App `4428544`, leaving `Contents: write`
- Consider dropping its `devProtect` bypass — `release.yml` only pushes `master`, so the `dev` bypass is unused
- Delete `CI_GITHUB_APP_ID` / `CI_GITHUB_APP_PRIVATE_KEY` from `oci-credentials`; after step 5 nothing there reads them

End state: three environments, each holding only what its consumers need, and two Apps that cannot substitute for one another — a compromise of the runner-registration path no longer buys a protected-branch push.

### 7. Update the invariant script for the App rename

A cross-branch dependency worth calling out rather than discovering later. `check-workflow-invariants.sh` (branch `chore/ci-workflow-guardrails`) detects App credentials with the literal pattern `secrets\.CI_GITHUB_APP_PRIVATE_KEY`. After step 5 renames the runner-admin references, that pattern stops matching and the credential-concentration check silently loses its App-key half — it would still pass on the `OCI_*` references alone, which is exactly the kind of quiet degradation these guards exist to prevent.

Generalize it to `secrets\.[A-Z_]*APP_PRIVATE_KEY` in whichever of the two branches merges second.

### Still open, unrelated to credentials

`test_validator_failure_recovery` fails intermittently on `arm64-docker` with `LFB spread 4 exceeds 3 after recovery`. Pre-existing, tracked separately, and unrelated to anything here — but it is the most likely reason a run on these branches goes red.

jeffrey-l-turner:
## Correction to runbook step 3: the secret deletion gates on `master`, not on `dev`

Step 3 said the repository-level `OCI_*` and `RELEASE_PAT` become deletable once this PR
merges into `dev`, on the grounds that nothing in `.github/` references `RELEASE_PAT` any
more. **That is true on this branch and false on `master`**, and `master` is what matters,
because scheduled workflows only ever run the default branch's copy.

On `master` at `a1d879ee`, `RELEASE_PAT` is still referenced by five workflows:
`_integration-pipeline.yml`, `ci-fork-pr.yml`, `merge-recovery-soak.yml`,
`oci-validation.yml`, `release.yml`. Concretely, the nightly soak's `launch_runner` job runs
under `environment: ephemeral-launch-internal` and reads repository-level `OCI_*` plus
`secrets.RELEASE_PAT` (`merge-recovery-soak.yml:303`) to mint a runner registration token.

Deleting either secret before the migration reaches `master` would break:

- the nightly and weekend soak (no runner can be registered), and with it the dashboard
- the release workflow's version bump, tag, and protected-branch push

So the correct gate is **after `dev → master`**, once master carries the environment-scoped
and GitHub App versions. The merge order already planned — `master → dev` first, then
`dev → master` — reaches that state; the deletion simply has to wait for the second leg
rather than the first.

Checkpoint publishing is unaffected either way: segments dispatch
`soak-checkpoint-publish.yml` with `github.token` plus job-level `actions: write`
(`merge-recovery-soak.yml:429` onward), not with a PAT.

### Related: the soak had been failing silently, and needed its own hotfix

While verifying the above, the nightly soak turned out to have been failing at shard
bring-up every night since ~2026-07-27, reporting green the whole time. Two cron slots fire
nightly; the 19:30 Pacific slot runs the real soak and the 20:30 slot no-ops through the
schedule gate and reports success — and the no-op is the later run, so it is the one people
see.

Root cause was a third `SYSTEM_INTEGRATION_REF` pin site. `.github/oci-validation.env` and
`_integration-pipeline.yml` are held identical by `build_base`'s drift check and had both
advanced to `06f2020c`; `merge-recovery-soak.yml` carries its own pin, sits outside that
check, and had rotted at `a50eeb19`. That predates `system-integration` `81284fc`, which
added `integration-tests/certs/validator4`. `compose.py` bind-mounts that path, so with no
host file present Docker created a **directory** and every node start died on
`Failed to read the X.509 certificate: IO error: Is a directory (os error 21)`.

Fixed on `hotfix/soak-system-integration-pin` (`4879a1f6`), which bumps that pin to
`29d7bd0`. Coverage for the two failure modes that let this hide — the unguarded third pin
and the indistinguishable no-op — is filed as `TASK-010-6` under EPIC-010 in
`docs/ToDos.md`.

PR #167, opened 2026-07-30
Feature/add commit version soak dashboard

jeffrey-l-turner:
---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/167"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787973456&installation_model_id=426883&pr_number=167&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F167&signature=5db5b0aef035b701b327d9ce942f668c78bff6840dcd342170f9fad3889f1b88"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #166, opened 2026-07-30
Hotfix/soak system integration pin

jeffrey-l-turner:
---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/166"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787964629&installation_model_id=426883&pr_number=166&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F166&signature=cf1c7f590013e58fe8b71a035019f0392cbca37f521635f6ae0a30f3b733e00c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

Issue #68, opened 2026-05-29
[Security] Hardcoded private keys and plaintext key logging expose validator credentials


Timekiller7:
**Severity: Critical**
---

## What did you expect to see?

- Private keys sourced exclusively from environment variables or secure secret stores, never embedded as literals in source files.
- Node startup logging the derived **public key** only — the private key must never appear in any log, stdout, or stderr output.
- Test-only key material gated behind `#[cfg(test)]` or a `test-utils` feature flag so it is excluded from production builds.
- Version-controlled Docker and CI files using environment variable references with no hardcoded fallback values for anything key-related.

---

### 1 — CRITICAL: Validator private key printed to stdout on every node startup

**File:** `casper/src/rust/validator_identity.rs`, line 57

```rust
pub fn from_private_key_with_logging(priv_key: Option<&str>) -> Option<Self> {
    println!("priv_key = {:?}", priv_key);   // raw private key emitted to stdout
```

Every validator node prints its private key in plaintext to stdout during boot. This function is called from five callsites in the boot path:

- `node/src/rust/runtime/setup.rs:322`
- `casper/src/rust/engine/casper_launch.rs:336`
- `casper/src/rust/engine/casper_launch.rs:470`
- `casper/src/rust/engine/casper_launch.rs:539`
- `casper/src/rust/engine/casper_launch.rs:688`

Any process that captures stdout — Docker log drivers, systemd journal, CI log aggregators, syslog forwarders — persistently stores the validator private key in plaintext.

---

### 2 — HIGH: Private keys hardcoded as public statics compiled into the production binary

**File:** `casper/src/rust/util/construct_deploy.rs`, lines 17–34

```rust
lazy_static! {
    pub static ref DEFAULT_SEC: PrivateKey = PrivateKey::from_bytes(
        &hex::decode("a68a6e6c...")...
    );
    pub static ref DEFAULT_SEC2: PrivateKey = PrivateKey::from_bytes(
        &hex::decode("5a0bde2f...")...
    );
}
```

These are `pub` statics with no `#[cfg(test)]` guard. They are compiled into every production build and exported from the crate's public API, making them accessible to any downstream code.

---

### 3 — HIGH: Validator private keys committed in version-controlled Docker and CI files

Private keys appear as hardcoded literal values in the following committed files:

| File | Variable(s) |
|---|---|
| `docker/standalone.yml` | `STANDALONE_PRIVATE_KEY` (default value) |
[...]

PR #169, opened 2026-07-30
fix(soak): survive failed iterations, size the RSS ceiling to the host, restart within the window

jeffrey-l-turner:
## Problem

Run [30516534214](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30516534214) — the first nightly soak ever to get **past bring-up** (post-#166) — died after 15 minutes and published `no passive soak summary was produced`, target `unknown @ unknown`. Three stacked causes, plus a fourth waiting behind them:

1. **The harness RSS watchdog killed the shard.** ~130s into every iteration: `Resource ceiling breached: total node RSS 9943MB exceeded 5000MB … killing nodes now`. The 5000MB ceiling is the harness **default** (`--rss-ceiling-mb` in system-integration's `conftest.py`; always active — `--monitor` only adds the CSV), sized for laptops. The soak VM is `VM.Standard.E6.Flex`, 16 OCPU / **32GB**, and a 6-node shard under `test_load` legitimately peaks ~10GB. The pytest `grpc UNAVAILABLE / connection refused` failures are downstream of that deliberate kill.
2. **A latent `set -e` bug turned one failed iteration into instant segment death.** The soak loop is designed to tolerate failures (count, sleep 30, continue), but `set -e` was re-enabled after each pytest, and the metric helpers' grep pipelines exit nonzero when a failed iteration leaves nothing to sample. The whole script died mid-loop — before `.soak-state` or the rollup were written — despite the comment above it promising metrics "must never fail the soak". Evidence: no `sleep 30`, no rollup, no `resuming soak` line, and each segment restarting at iteration 1.
3. **Consequence: nothing publishable.** No `summary.json` → verdict `regress`, "no data", `unknown @ unknown`.
4. **The reaper was next in line.** `ci-runner-reaper.yml` terminates `ci-eph-*` instances older than **2h**, every 30 minutes; its "no job runs longer than 60 minutes" premise predates the soak entirely. The first *healthy* 22h/60h run would have had its VM terminated at the 2h mark.

## Changes

**`scripts/run-merge-recovery-soak.sh`**
- Remove the erroneous `set -e` restores (the script's baseline is `set -uo pipefail`; errexit was never intended) and guard the metric pipelines with `|| true`. A failed iteration is now counted and the soak continues; the run still exits red at the end, with checkpoints, state and summary intact.
- Pass `--rss-ceiling-mb` sized to the host: `MemTotal − 8GB` (24,576MB on the soak VM), floor 5000, `SOAK_RSS_CEILING_MB` overrides, `0` disables. The watchdog stays on to catch a real leak before swap-thrash freezes the VM — a state that reports nothing and keeps billing — but never fires on normal load.

**Restart within the window** (`merge-recovery-soak.yml`, `scripts/restart-soak.sh`)
- `workflow_dispatch` gains `window_end_epoch`, `series`, `retry_attempt`. In restart mode the gate soaks only the *remainder* of the original window (2h floor, clean no-op below it) and derives the window's remaining Pacific checkpoints, so a restarted night still ends when the scheduled run would have and cannot spill into the next slot.
- Everything series-scoped now keys on `kind` rather than raw duration: job timeout, both concurrency groups, benchmark enablement, all six `merge_exit_min_seconds` sites.
- New `retry_within_window` job restarts **once**, only on infrastructure death: launch failure, or a soak job that ended without reaching its completion marker (VM preempted/reaped/frozen — a lost runner reports as *failure*, not cancellation). A soak that completed red produced a real verdict and is not retried; a manual cancel is an operator calling the night off and never restarts anything.
- `scripts/restart-soak.sh` drives the same mode by hand: `--last-failed`, `--until-next-slot`, `--hours N`, `--window-end`.

**Provenance** (`aggregate-perf-report.sh`, `soak-segment/action.yml`)
- The run object carries `restarted`, `retry_attempt` and `window_seconds`; the report header reads `restarted; covered 40h of the 60h window`. **Restarted runs are excluded from baseline duty** — a partial span would make the next full-window run look like a regression purely because the spans differ.
- Drive-by: the hardcoded "Weekend Soak Benchmark Report" title now follows the series, fixing daily reports titled "Weekend" (visible on the dashboard today).

**Reaper exemption** (`ci-runner-reaper.yml`, `merge-recovery-soak.yml`)
- The launch job tags its instance with `soak-deadline-epoch` (window end + 2h grace), plus `purpose`/`series`; the reaper skips it until expiry, then reaps it like any leak. An exemption **with an expiry**, not a bypass — untagged instances keep the 2h rule, and expired or unparseable tags are reaped.

**Pin bump**: `SYSTEM_INTEGRATION_REF` → `9ebdde01` (system-integration PR #68 merge): the `test_load.py` drain-snapshot convergence fix and registry pull retries, on top of the validator4 certs.

## Trade-off worth stating rather than having a reviewer discover it

**Reaper tagging fails closed.** If `oci compute instance update` hits a transient error, the launch job fails with a VM already running. That VM is untagged, so the reaper terminates it within 2h, and the retry launches a fresh one — a bounded double-spend of roughly one VM-hour. That is deliberate: the alternative is a soak that silently dies at hour two, which is the failure mode this PR exists to eliminate.

## Verification

- Extracted the metric helpers verbatim and ran them under `set -euo pipefail` (harsher than production) in both the zero-sample failed-iteration shape and a populated shape; the failed shape now emits `metrics.json` with nulls instead of killing the segment.
- Ceiling derivation: 32GB host → 24576, no `/proc/meminfo` → 5000, explicit `0`/`20000` honored, junk rejected with exit 2.
- Reaper jq filter against a 7-instance matrix: live soak skipped; expired, junk-tagged, untagged and stale-CI reaped; young CI and non-`ci-eph` untouched.
- Gate logic: the resolve script extracted verbatim and run through seven scenarios — partial weekend/daily restarts (correct remaining checkpoints, benchmarks, timeouts), sub-floor skip, plain dispatch, **scheduled cron path unchanged**, junk retry counter coerced to 0, junk window hard-fails. This caught a real bug: an empty `retry_attempt` input emitted an empty output, which would have silently disabled automatic retry for every scheduled run.
- Aggregator run against synthetic soaks: restarted weekend (stamp + fields), normal daily (no stamp, "Daily" title), restarted checkpoint (verdict suppressed, provenance carried), baseline selector picking the last non-restarted complete run.
- `bash -n`, `check-workflow-invariants.sh`, YAML parsing, and pre-commit fmt/clippy/deny all pass.

## Review rounds and resolutions

Two `/multi-review` rounds. Every finding was verified against the code before being accepted or rejected, rather than taken at face value — one "major" turned out to be false, and the second round caught a defect introduced while fixing the first.

### Round 1 — 0 critical, 3 major, 3 minor

| Finding | Verdict | Resolution |
|---|---|---|
| `end_epoch` never computed for non-restart runs, so automatic restart can never fire (xai, major) | **False** | `end_epoch=$((slot_epoch + duration_seconds))` is assigned unconditionally in the checkpoint block every path runs through. Confirmed by running the extracted gate on a simulated `schedule` trigger: it emits `end_epoch`. xai read only the restart branch. No change. |
| `window_end_epoch` unbounded → a leaked VM can stay reaper-exempt indefinitely (openai, major) | **Real** | Restart mode now rejects a remainder longer than the series' nominal window. A milliseconds-for-seconds typo is caught explicitly. Fixed in `150e2274`. |
| Restart duration measured before provisioning, so the soak overruns the window end (openai, major) | **Real, impact overstated** | Real drift of ~25 min; no actual slot overlap, since a daily window ends 2h before the next slot. Fixed properly anyway: the restarted final segment now takes `end_epoch` as an absolute deadline. `150e2274`. |
| `series` documented as required but not enforced (openai, minor) | **Real** | Now a hard failure. Also closed the hole underneath it — a window-bounded run is forced to `retry_attempt >= 1`, since such a run *is* a restart and must not publish a partial span as a full one. `150e2274`. |
| Reaper tag uses wall-clock now, not window end (xai, minor) | **Real** | Tag anchored to `end_epoch + 7200`. `150e2274`. |
| Unparseable deadline tags reaped silently (xai, minor) | Declined | Fail-open-to-reap is deliberate and commented: a typo must not buy permanent immunity. |
| Duplicated compartment OCID (xai, suggestion) | Deferred | Touches the reaper's blast-radius guarantee; tracked as **TASK-010-8**, own branch. |

Not from the review: the failure-alert email derived its label from raw duration, so a restarted weekend would have been emailed as `daily-24h`. Now keys on `kind` and names the restart.

### Round 2 — 0 critical, 1 major, 3 minor

| Finding | Verdict | Resolution |
|---|---|---|
| `--freeform-tags` replaces the whole tag map rather than merging (openai, major) | **Real, latent** | Correct about OCI semantics. Harmless today — `launch-runner.sh` sets no freeform tags at the pinned SHA, re-verified — but it collides with our own roadmap: TASK-010-7 asks system-integration to add cost-tracking tags at launch, which this step would then silently erase. Now a read-modify-write (`get` → `jq` merge → `update`), tested against populated / `null` / empty / non-JSON maps. `d62cbdd0`. |
| `restart-soak.sh` does not mirror the gate's upper bound (openai, minor) | **Real — introduced by the round-1 fix** | The bound was added to the gate only, so the layers disagreed. Simulated: `--until-next-slot` at 20:00 PT yields 23h against a 22h daily and the gate rejects it — precisely when an operator restarts after an evening failure. Explicit `--hours`/`--window-end` past nominal now fail locally; `--until-next-slot` is capped, since "end before the next launch" survives capping. `d62cbdd0`. |
| Two × "unnecessary newline" (bedrock, minor) | Declined | Same generic text pasted against two unrelated files, from the provider that voted approve. |

Both rounds had only 2 of 4 providers voting — Anthropic out of API credit, xai failing once — so the panel was thin and the verification step did real work.

## Cross-repo coordination (system-integration)

Conducted through tracked `docs/ToDos.md` files in both repos, **not** `docs/discoveries/`, whose contents are gitignored on both sides (`.gitignore:123` here, `:76-77` there) and therefore invisible to `git status`. An earlier message was lost exactly that way.

- **Their independent confirmation** of the RSS root cause from the same run logs, and confirmation that our pin `9ebdde01` is `main` HEAD needing no bump.
- **Answered their blocking question** on the tag contract: freeform tag, key `soak-deadline-epoch`, Unix epoch **seconds** + 7200s grace — and critically, the value is a JSON **string**, not a number, because OCI freeform tags are a string→string map. A numeric comparison would pass their tests and never exempt anything in production.
- **They corrected my suggested fix, rightly**: restricting their reaper to `ci-eph-*` prefixes alone would not have saved a soak, because soak runners *are* `ci-eph-*` (`launch-runner.sh:82`). The tag is the only discriminator; the prefix is blast-radius containment.
- **Their side is implemented** on `hotfix/provide-restart-resolve-soak-failure`: tag-aware reaper with 18 mutation-verified test cases, plus failure attribution in `test_load.py` covering both call sites.

## Follow-ups (durable — recorded in `docs/ToDos.md`, not only here)

- **TASK-010-7** — make system-integration's `reap-stale-runners.sh` soak-aware. At `9ebdde0` it filters only on `lifecycle-state` and age, with no name filter and no tag check, so it is blind to `soak-deadline-epoch` and would kill a 22h daily at hour 6. **Latent, not active**: nothing schedules it (only `smoke-test.yml` exists there). Also covers `ci-soak-*` naming, cloud-init self-destruct sized to a per-run budget (~$12 daily / ~$33 weekend), and a cost tag with a monthly OCI budget. *Their side is now implemented and awaiting review.*
- **TASK-010-7 (added)** — `conftest.py`'s `--rss-ceiling-mb` default of `5000` is a defect, not a tuning preference. `test_load.py:220` fixes the shard at 6 nodes regardless of host, and that shard peaks ~10GB, so the default sits at half the working set of the harness's own primary load test and kills it identically on a 64GB workstation. It went unnoticed because `_integration-pipeline.yml:482` `--deselect`s `test_load.py`, leaving the soak as its only automated caller — and the soak never got past bring-up until 2026-07-30, so the default had never once been exercised. Our host-derived override is a workaround; the shared default is the real fix.
- **TASK-010-8** — de-duplicate `CI_RUNNER_COMPARTMENT_OCID` without weakening the reaper. A repo-level Actions variable is likely wrong: it is admin-mutable, whereas the current hardcoding is what makes "the reaper can never touch other compartments" true. Preferred shape is an assertion in `check-workflow-invariants.sh` that both literals match.
- **Open, ours**: `oci-validation.env` and `_integration-pipeline.yml` agree at `06f2020`, satisfying the drift invariant — but `06f2020` predates the validator4 cert fix, so CI integration proves a ~3-week-old system-integration while the soak proves `main`. Deliberately not bumped today: one moving part at a time.
- Dashboard rendering of the restart stamp — deferred to `feature/add-commit-version-soak-dashboard`, which owns `index.html`; the data is already published for it to consume.
- Merge `master` back into `dev` after this lands (expect a small conflict in the aggregator's run object, where #167 adds `version` in the same block).
- **Owed after tonight's run**: report back to system-integration on whether the SELinux bind-mount labels are inert on the Ubuntu OCI host, and whether execution finally reaches `wait_for_lfb_converged` (`test_load.py:455`) — neither has ever been reached by a real soak.

### DIFF
--- a/.github/actions/soak-segment/action.yml
+++ b/.github/actions/soak-segment/action.yml
@@ -44,6 +44,15 @@ inputs:
   kind:
     required: true
     description: weekend | daily — selects the published series.
+  retry_attempt:
+    required: true
+    description: Restart counter from the gate; >0 marks a restarted run.
+  window_seconds:
+    required: true
+    description: >-
+      The series' nominal window span (79200 daily / 216000 weekend), as
+      opposed to this run's duration budget, which for a restarted run is
+      only the window's remainder.
   node_repo_dir:
     required: true
     description: Checkout of the node under test.
@@ -90,6 +99,8 @@ runs:
         RUN_ID: ${{ github.run_id }}
         RUN_ATTEMPT: ${{ github.run_attempt }}
         DURATION_SECONDS: ${{ inputs.duration_seconds }}
+        RETRY_ATTEMPT: ${{ inputs.retry_attempt }}
+        WINDOW_SECONDS: ${{ inputs.window_seconds }}
       shell: bash -euo pipefail {0}
       run: |
         # in_progress suppresses the verdict: a partial run has fewer
--- a/.github/workflows/ci-runner-reaper.yml
+++ b/.github/workflows/ci-runner-reaper.yml
@@ -6,7 +6,11 @@ name: CI Runner Reaper
 # June 2026 ~209 such zombies burned ~$1,084/day for 9 days (~$6k excess)
 # before anyone noticed. No pipeline job runs longer than 60 minutes
 # (_integration-pipeline.yml timeout-minutes), so any ci-eph instance older
-# than MAX_AGE_HOURS is a leak, busy or not.
+# than MAX_AGE_HOURS is a leak, busy or not — with one exception: soak
+# runners (merge-recovery-soak.yml) run 22h (daily) or 60h (weekend) and
+# carry a soak-deadline-epoch freeform tag naming the second after which
+# they may be reaped. The tag is an exemption with an expiry, not a bypass:
+# a leaked soak VM still dies, at most a grace period after its window.
 #
 # NOTE: the schedule trigger only fires from the default branch (master).
 # Until this file reaches master, run it manually via workflow_dispatch.
@@ -96,13 +100,21 @@ jobs:
           cutoff=$(date -u -d "-${MAX_AGE_HOURS} hours" +%Y-%m-%dT%H:%M:%S)
           echo "Reaping ci-eph-* instances created before ${cutoff}Z (max age ${MAX_AGE_HOURS}h)"
 
+          now_epoch=$(date -u +%s)
           oci compute instance list \
             --compartment-id "$CI_RUNNER_COMPARTMENT_OCID" --all --output json \
-            | jq -r --arg cutoff "$cutoff" '
+            | jq -r --arg cutoff "$cutoff" --argjson now "$now_epoch" '
                 .data[]
                 | select(."lifecycle-state" == "RUNNING" or ."lifecycle-state" == "STOPPED")
                 | select(."display-name" | startswith("ci-eph-"))
                 | select(."time-created" < $cutoff)
+                # Soak runners outlive any CI age bound (22h daily, 60h
+                # weekend). merge-recovery-soak.yml tags its instance with the
+                # epoch second after which reaping is allowed (window end +
+                # grace); skip until then. Untagged instances — every normal
+                # CI runner — keep the age rule, and an expired or unparseable
+                # tag is reaped like any other leak.
+                | select(((."freeform-tags"["soak-deadline-epoch"] // "0") | tonumber? // 0) < $now)
                 | "\(.id) \(."display-name") \(."time-created")"' \
             > /tmp/zombies.txt
 
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -453,20 +453,24 @@ jobs:
     needs: [schedule_gate, launch_runner, soak]
     # Restart only on infrastructure death: the launch never delivered a
     # runner, or the soak job ended without reaching its completion marker
-    # (VM preempted, reaped or frozen mid-run). A soak that completed red
-    # produced a real verdict — a second VM would only reproduce it. The
-    # restart conforms to the original window via window_end_epoch, so it can
-    # never spill into the next slot, and retry_attempt caps the chain at
-    # one: a second failure in the same window is a pattern to investigate,
-    # not a blip to retry, and OCI capacity shortages (the common launch
-    # failure) rarely clear within the night.
+    # (VM preempted, reaped or frozen mid-run — a lost runner surfaces as job
+    # FAILURE, not cancellation). A soak that completed red produced a real
+    # verdict — a second VM would only reproduce it. Cancellation never
+    # restarts anything: a manual cancel is an operator calling the night or
+    # weekend off, so !cancelled() skips this job for the whole run, and
+    # cancelled-shaped job results (timeout, next slot's concurrency cancel)
+    # stay dead too. The restart conforms to the original window via
+    # window_end_epoch, so it can never spill into the next slot, and
+    # retry_attempt caps the chain at one: a second failure in the same
+    # window is a pattern to investigate, not a blip to retry, and OCI
+    # capacity shortages (the common launch failure) rarely clear within the
+    # night.
     if: >-
-      always() &&
+      !cancelled() &&
       needs.schedule_gate.outputs.should_run == 'true' &&
       needs.schedule_gate.outputs.retry_attempt == '0' &&
       (needs.launch_runner.result == 'failure' ||
-       needs.launch_runner.result == 'cancelled' ||
-       ((needs.soak.result == 'failure' || needs.soak.result == 'cancelled') &&
+       (needs.soak.result == 'failure' &&
         needs.soak.outputs.completed != 'true'))
     runs-on: ubuntu-latest
     timeout-minutes: 5
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -472,9 +472,20 @@ jobs:
             --query 'data[0].id' --raw-output)"
           test -n "$iid" && [ "$iid" != "null" ]
           deadline="$((END_EPOCH + 7200))"
+          # --freeform-tags REPLACES the instance's whole tag map; it does not
+          # merge. The launcher sets no freeform tags today, so overwriting
+          # would be harmless right now — but system-integration TASK-010-7
+          # plans cost-tracking tags applied at launch, and this step would
+          # then silently erase them, breaking cost attribution in a way
+          # nobody notices until a bill arrives. Read-modify-write instead.
+          existing="$(oci compute instance get --instance-id "$iid" \
+            --query 'data."freeform-tags"' --raw-output 2>/dev/null || true)"
+          case "$existing" in ''|null) existing='{}' ;; esac
+          jq -e 'type == "object"' <<<"$existing" >/dev/null 2>&1 || existing='{}'
+          merged="$(jq -cn --argjson cur "$existing" --arg s "$KIND" --arg d "$deadline" \
+            '$cur + {purpose: "soak", series: $s, "soak-deadline-epoch": $d}')"
           oci compute instance update --instance-id "$iid" --force \
-            --freeform-tags "{\"purpose\":\"soak\",\"series\":\"${KIND}\",\"soak-deadline-epoch\":\"${deadline}\"}" \
-            >/dev/null
+            --freeform-tags "$merged" >/dev/null
           echo "tagged ${name} (${iid}): reapable after $(date -u -d "@${deadline}" +%FT%TZ)"
 
       - name: Remove OCI credentials
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -136,7 +136,28 @@ jobs:
               case "$INPUT_WINDOW_END" in
                 *[!0-9]*) echo "::error::window_end_epoch must be epoch seconds"; exit 1 ;;
               esac
+              # series is genuinely required here, not merely documented as
+              # such: a remainder span no longer identifies the series by its
+              # duration, so without it the guess below would publish the run
+              # into the wrong history with the wrong benchmark, timeout and
+              # concurrency behaviour. Fail rather than guess.
+              case "$INPUT_SERIES" in
+                daily) window_nominal=79200 ;;
+                weekend) window_nominal=216000 ;;
+                *)
+                  echo "::error::series must be daily or weekend when window_end_epoch is set"
+                  exit 1
+                  ;;
+              esac
               remaining="$((INPUT_WINDOW_END - slot_epoch))"
+              # A restart can only ever shorten the window it inherits. Without
+              # this bound an epoch typo — milliseconds for seconds, say — is
+              # accepted, and the reaper exemption derived from it would keep a
+              # leaked VM exempt for years. The exemption is meant to expire.
+              if [ "$remaining" -gt "$window_nominal" ]; then
+                echo "::error::window_end_epoch is ${remaining}s away, beyond the ${window_nominal}s $INPUT_SERIES window"
+                exit 1
+              fi
               # 2h floor: launch, image build and shard bring-up eat ~20min,
               # and a sub-2h remainder is not worth an OCI VM.
               if [ "$remaining" -lt 7200 ]; then
@@ -323,6 +344,15 @@ jobs:
           case "$retry_attempt" in
             ''|*[!0-9]*) retry_attempt=0 ;;
           esac
+          # A window-bounded run IS a restart, whatever the caller passed: it
+          # covers only part of its series' span. Forcing the counter keeps
+          # provenance, baseline exclusion and the one-retry cap consistent
+          # even for a hand-rolled dispatch that set window_end_epoch without
+          # retry_attempt — which would otherwise publish a partial run as a
+          # full one and still be eligible to retry.
+          if [ -n "${INPUT_WINDOW_END:-}" ] && [ "$retry_attempt" -eq 0 ]; then
+            retry_attempt=1
+          fi
           {
             echo "should_run=$should_run"
             echo "duration_seconds=$duration_seconds"
@@ -414,7 +444,7 @@ jobs:
         env:
           # Same pinned compartment as ci-runner-reaper.yml; not a secret.
           CI_RUNNER_COMPARTMENT_OCID: "ocid1.compartment.oc1..aaaaaaaalq6bh2a6dmq4h6i3nrripxlcevv7fa3goaf7wxve52qiuocmehia"
-          DURATION_SECONDS: ${{ needs.schedule_gate.outputs.duration_seconds }}
+          END_EPOCH: ${{ needs.schedule_gate.outputs.end_epoch }}
           KIND: ${{ needs.schedule_gate.outputs.kind }}
           SUPPRESS_LABEL_WARNING: "True"
         shell: bash -e {0}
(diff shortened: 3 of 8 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| bedrock | `us.amazon.nova-pro-v1:0` | Abstain | 0.5 |

---

### Summary

The retry and reporting changes are generally well structured, but the implementation does not yet enforce its central window-boundary guarantee: the remaining duration is measured before runner provisioning and subsequently runs past the original deadline. In addition, the unbounded manual deadline can grant a leaked OCI runner an extremely long reaper exemption. These resource-lifecycle issues should be addressed before merge; validating the documented series requirement would also prevent incorrectly classified manual restarts.

This PR thoughtfully adds in-window soak restart, reaper exemptions with expiry, restarted-run baseline exclusion, and an RSS ceiling fix — the design and docs are strong. The critical functional gap is that schedule_gate never assigns end_epoch for normal runs, so retry_within_window's END_EPOCH check will fail and automatic restarts will not fire. Fixing end_epoch emission (and ideally anchoring the reaper tag to it) should be done before relying on this hot fix in production; the remaining items are smaller robustness and maintainability nits.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Restart duration does not account for provisioning and setup time</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 139)

Restart mode calculates duration_seconds as window_end_epoch minus the time at which schedule_gate runs. That duration is later passed unchanged to the soak action, after runner provisioning, checkout, image building, and shard setup. Consequently, the actual soak can finish tens of minutes after the original window_end_epoch, contradicting the core guarantee that a restart cannot overlap the next slot. The reaper deadline has the same drift because it is calculated as the tagging time plus this already-stale duration.

**Suggestion:**
```
Treat window_end_epoch as the authoritative deadline throughout the workflow. Recalculate the remaining duration immediately before starting the soak, or pass the absolute deadline into the soak scripts and have their loops stop at that epoch. Set the reaper deadline directly to window_end_epoch + 7200 rather than current_time + duration_seconds + 7200.
```

</details>

<details>
<summary><b>:orange_circle: Major: Unbounded window input can exempt a runner from reaping indefinitely</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 137)

window_end_epoch is validated only as numeric and is not bounded against the selected series' nominal span. A mistaken or malicious manual dispatch can supply an epoch years in the future. The resulting duration is used to set soak-deadline-epoch, while the soak job itself still times out after 24 or 64 hours. If cleanup fails, the VM can therefore remain exempt from the reaper for years, undermining the intended expiry control and creating an avoidable resource-exhaustion and cost risk.

**Suggestion:**
```
Reject restart windows whose remaining duration exceeds the nominal series window (79,200 seconds for daily or 216,000 seconds for weekend), and consider applying a defensive maximum in ci-runner-reaper.yml as well, such as refusing exemptions implausibly far beyond the instance creation time.
```

</details>

<details>
<summary><b>:orange_circle: Major: end_epoch never computed for non-restart runs</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 439)

The schedule_gate job outputs end_epoch, and retry_within_window depends on it to re-dispatch with window_end_epoch. However, end_epoch is never assigned in the resolve step for a normal (non-restart) scheduled or dispatched run. Only remaining/duration_seconds are derived when INPUT_WINDOW_END is set. As a result, the first automatic infrastructure retry will fail the END_EPOCH validation (`no usable end_epoch from the gate; cannot restart`) and never re-dispatch — defeating the primary purpose of this PR.

**Suggestion:**
```
After resolving duration_seconds (and before writing outputs), set end_epoch for every run that should_run, e.g. `end_epoch=$(( slot_epoch + duration_seconds ))` for fresh runs, and `end_epoch=$INPUT_WINDOW_END` for restart-mode runs. Initialize end_epoch="" at the top so the output is always defined under set -u.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: tonumber? treats non-numeric deadline tags as immediately reapable</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci-runner-reaper.yml` (line 117)

The jq filter `((."freeform-tags"["soak-deadline-epoch"] // "0") | tonumber? // 0) < $now` collapses parse failures to 0, which is always < now. That is intentional fail-open-to-reap for garbage tags, but a partial write or OCI tag corruption would kill an in-window soak. Acceptable given the comments, but worth a reaper log line when the tag is present yet unparseable so the failure mode is visible.

**Suggestion:**
```
Consider emitting the display-name and raw tag value to the reaper step summary when tonumber? fails, so a bad tag is observable without changing the reap decision.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Restart series requirement is documented but not enforced</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 229)

The workflow input says series is required alongside window_end_epoch, but restart mode silently accepts an empty or invalid series and infers the kind from the otherwise ignored duration input. This can publish a manual restart into the wrong series and select the wrong benchmark behavior, timeout, nominal window size, and concurrency group.

**Suggestion:**
```
When window_end_epoch is non-empty, require INPUT_SERIES to be exactly daily or weekend and fail the gate otherwise. Optionally reject series when window_end_epoch is absent to avoid ambiguous dispatches.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Reaper exemption deadline uses wall-clock now, not window end</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 419)

The soak-deadline-epoch tag is computed as `now + DURATION_SECONDS + 7200`. For a promptly launched run this approximates window end + grace. If launch is delayed (queueing, image pull, prior step slowness), the tag extends past the intended window end, leaving the VM reap-exempt longer than documented. Conversely it is not anchored to the same end_epoch the restart logic uses, so the two clocks can drift.

**Suggestion:**
```
Prefer tagging with `end_epoch + 7200` (from schedule_gate.outputs.end_epoch) once end_epoch is correctly emitted, falling back to now+duration only if end_epoch is unavailable.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Hardcoded compartment OCID duplicated across workflows</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 406)

CI_RUNNER_COMPARTMENT_OCID is inlined in the new tagging step and noted as the same value used by ci-runner-reaper.yml. A compartment migration would need coordinated edits and could silently leave soak runners untagged (or un-reapable if only one side changes).

**Suggestion:**
```
Share the OCID via a repo variable or reusable workflow input so reaper and launcher cannot drift.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Solid operator tool; consider noting gh auth requirement in docs</b> (Reported by: xai)</summary>

**File:** `scripts/restart-soak.sh` (line 1)

The script correctly unsets ambient GITHUB_TOKEN, validates inputs, enforces the 2h floor, and always stamps retry_attempt=1 so manual runs cannot poison baselines. docs/soak-benchmarks.md already covers usage well.

**Suggestion:**
```
Optional: add a one-line note that the operator needs `gh auth` with actions:write on F1R3FLY-io/f1r3node-rust (the script header mentions it; a docs callout helps).
```

</details>
---

<details>

> **ABSTAIN**: {     "verdict": "provide_feedback",     "confidence": 0.9,     "issues": [         {             "severity": "minor",             "category": "style",             "file": ".github/actions/soak-segment/action.yml",             "line": 44,             "title": "Unnecessary comment",             "description": "The comment `# in_progress suppresses the verdict: a partial run has fewer` is incomplete and unclear. It should be removed or completed for clarity.",             "suggestion": "Remove or 

#### openai (gpt-5.6-sol)

The retry and reporting changes are generally well structured, but the implementation does not yet enforce its central window-boundary guarantee: the remaining duration is measured before runner provisioning and subsequently runs past the original deadline. In addition, the unbounded manual deadline can grant a leaked OCI runner an extremely long reaper exemption. These resource-lifecycle issues should be addressed before merge; validating the documented series requirement would also prevent incorrectly classified manual restarts.

#### xai (grok-4.5)

This PR thoughtfully adds in-window soak restart, reaper exemptions with expiry, restarted-run baseline exclusion, and an RSS ceiling fix — the design and docs are strong. The critical functional gap is that schedule_gate never assigns end_epoch for normal runs, so retry_within_window's END_EPOCH check will fail and automatic restarts will not fire. Fixing end_epoch emission (and ideally anchoring the reaper tag to it) should be done before relying on this hot fix in production; the remaining items are smaller robustness and maintainability nits.
</details>

---

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The failure-tolerant soak loop, host-sized RSS ceiling, bounded retry provenance, and reaper expiry logic are generally coherent. Before merging, the OCI tagging step should preserve existing instance metadata rather than replacing it. The manual restart helper should also validate the same maximum window bounds as the workflow so its documented invocations cannot dispatch runs that immediately fail their gate.

The code changes appear to address the identified issues effectively. The logic for handling soak runs, especially in cases of failures and restarts, seems well thought out. There are minor style issues regarding the formatting of new input parameters, but these do not affect the functionality. Overall, the changes look good and should be safe to merge.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Tagging the soak runner replaces all existing freeform tags</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 468)

`oci compute instance update --freeform-tags` replaces the instance's complete freeform-tag map; it does not merge the supplied keys. This step can therefore erase tags installed by `launch-runner.sh`, including ownership, cost-allocation, cleanup, or auditing metadata. The comment describes the new tags as being added alongside existing metadata, but the command is destructive.

**Suggestion:**
```
Fetch the current freeform tags, merge in `purpose`, `series`, and `soak-deadline-epoch` with `jq`, and submit the merged object. Alternatively, update `launch-runner.sh` to apply all required tags atomically when creating the instance.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary newline</b> (Reported by: bedrock)</summary>

**File:** `.github/actions/soak-segment/action.yml` (line 44)

The addition of new input parameters could be condensed to reduce the number of lines.

**Suggestion:**
```
Combine the descriptions and required flags into fewer lines.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary newline</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 17)

The addition of new input parameters could be condensed to reduce the number of lines.

**Suggestion:**
```
Combine the descriptions and required flags into fewer lines.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Manual restart options can generate windows rejected by the workflow</b> (Reported by: openai)</summary>

**File:** `scripts/restart-soak.sh` (line 194)

The script only enforces the 2-hour minimum, while the workflow rejects a remaining window greater than the nominal series span: 22 hours for daily and 60 hours for weekend. For example, `--series daily --hours 24` passes this script but fails in the workflow gate. `--series daily --until-next-slot` can likewise exceed 22 hours when invoked shortly after the previous slot, despite being presented as a supported command.

**Suggestion:**
```
Mirror the workflow's upper bound before dispatching: reject a daily remainder above 79200 seconds and a weekend remainder above 216000 seconds, with an actionable error. If `--until-next-slot` is intended to work at any time, cap or otherwise define its window so it remains within the selected series maximum.
```

</details>
---

<details>

The code changes appear to address the identified issues effectively. The logic for handling soak runs, especially in cases of failures and restarts, seems well thought out. There are minor style issues regarding the formatting of new input parameters, but these do not affect the functionality. Overall, the changes look good and should be safe to merge.

#### openai (gpt-5.6-sol)

The failure-tolerant soak loop, host-sized RSS ceiling, bounded retry provenance, and reaper expiry logic are generally coherent. Before merging, the OCI tagging step should preserve existing instance metadata rather than replacing it. The manual restart helper should also validate the same maximum window bounds as the workflow so its documented invocations cannot dispatch runs that immediately fail their gate.

#### xai (grok-4.5)

</details>

---

jltatbeach (approved):

PR #70, opened 2026-06-04
docs: align contributor CI policy

jeffrey-l-turner:

PR #170, opened 2026-07-30
fix(soak): retry the reaper tag through OCI's provisioning conflict; pin the compartment OCID under CI

jeffrey-l-turner:
## Problem

Validation run [30584775602](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30584775602) — the first launch after #169 merged — failed in the new reaper-tagging step **three seconds after the runner launched**:

```
"code": "Conflict",  "status": 409,
"message": "instance ocid1.instance...vongua is currently being modified, try again later",
"operation_name": "update_instance"
```

`launch-runner.sh` returns as soon as OCI accepts the launch call, but the instance keeps transitioning through `PROVISIONING` for another 30–60s, and `instance update` against a transitioning instance is refused. This is live on `master` now: **every** soak launch fails this way, including tonight's scheduled 19:30 run.

Nothing about the tag contract was wrong — `END_EPOCH: 1785463200` and the compartment both resolved correctly. Two things in #169 behaved exactly as designed and are worth recording:

- The step **failed closed**, so we got a failed launch rather than a soak that silently dies unexempt at hour two. That was the deliberate trade-off documented in #169.
- `retry_within_window` **skipped**, because the run already carried `retry_attempt=1`. The one-retry cap held instead of spawning a chain.

## Changes

**`merge-recovery-soak.yml` — retry the tag through the conflict.** HTTP 409 is explicitly retryable, so retry rather than fail: up to 12 attempts at 15s (~3 min), re-reading the tag map on each attempt so a write landing mid-retry is merged rather than clobbered by a stale snapshot. Fail-closed is preserved — after 12 attempts the launch still fails, because an untagged soak runner is reaped at the 2h mark. The OCI error is now printed before the annotation, so the next failure needs no log archaeology.

**`check-workflow-invariants.sh` — invariant 5, the compartment OCID (TASK-010-8).** `CI_RUNNER_COMPARTMENT_OCID` must be pinned identically in `ci-runner-reaper.yml` and `merge-recovery-soak.yml`. If they diverge, the soak writes its exemption tag into a compartment the reaper never queries, and the soak dies at 2h with a tag that exists but is invisible — the same symptom as no exemption at all, and much harder to diagnose.

This deliberately **asserts equality rather than de-duplicating**. TASK-010-8's original framing asked for the value to stop being hardcoded twice, but a repo-level Actions variable is admin-mutable, and the reaper's own comment claims it "can never touch other compartments" — a claim that depends on the value being immutable in-repo. Equality-under-CI keeps immutability and kills drift; that criterion is superseded rather than satisfied, and the reasoning is recorded in `docs/ToDos.md`.

**Cross-reference comments** at both OCID sites, naming the other file, the enforcing check, and the failure mode.

## Verification

- **Retry loop**, against a stubbed `oci` CLI: immediate success tags on attempt 1; three conflicts then success tags on attempt 4; permanent conflict fails closed after 12. Every success path preserved a pre-existing `owner: ci-team` tag, so #169's read-modify-write fix still holds under retry.
- **Invariant 5**, mutation-tested: a divergent OCID fails with both filenames; removing both literals fails with a message naming the cause.
- That mutation testing **caught a real defect in the guard itself** — under `set -e`, a no-match `grep` inside a command substitution killed the script before it could print why, making the "nobody pins it any more" branch unreachable and surfacing as a red job with no annotation. Fixed with `|| true` on both greps. A guard that cannot explain itself is the failure mode that file exists to prevent.
- YAML parses for both workflows; all six invariants pass.

## Third change: a live fail-open hole in the reaper (found via system-integration's review)

Their multi-agent review of PR #70 found that `float()` accepts `Infinity`/`inf`/`1e309`, granting a permanent exemption. **I tested whether the same hole existed here, and it does** — jq's `tonumber` accepts all of those, and no `< $now` comparison can ever be true for the result. A single bad tag value granted an instance **permanent immunity from the reaper**: the same unbounded-VM shape as the June 2026 leak (~209 zombies, ~$6k). Live on `master` since #169; neither of our own reviews of that PR caught it.

The filter now reaps on non-finite deadlines and on finite-but-absurd ones — the longest legitimate deadline is a 60h weekend plus 2h grace, so anything past a 7-day horizon is not a soak.

Verified across twelve value shapes:

| Tag value | Result |
|---|---|
| `Infinity`, `inf`, `-inf`, `1e309`, `9e99999` | reap (hole closed) |
| `1999999999` (year 2033, finite) | reap (past horizon) |
| `nan`, `junk`, `""`, tag absent | reap (unchanged) |
| live deadline, **including fractional** | exempt |

That last row is deliberate. The sibling agent declined their reviewer's suggestion to parse with `int()`, reasoning that a consumer stricter than its producer would discard a valid deadline and kill a live soak. Adopting that reasoning keeps producer and consumer aligned.

I also recorded **in the code** that the two components fail in deliberately opposite directions — the tagger fails **closed** (refuse to launch), this reaper fails **open toward cleanup** — with a note not to harmonise them. Side by side they read as an inconsistency someone would try to fix; composed, they give the property we want: an untagged soak runner never starts, and any untagged instance that does exist gets cleaned up.

## Review round 2 — resolutions

| Finding | Reporter | Verdict | Resolution |
|---|---|---|---|
| Invariant passes if **one** of the two declarations is removed | openai | **Real** | It asserted "at least one file pins it, and all literals agree", so deleting one copy left a single literal trivially agreeing with itself — the most likely drift, since it is the `${{ vars.X }}` migration the guard exists to discourage. My mutation tests missed it (divergent and both-removed, never one-removed). Both files now checked **by name**, each required to pin exactly one literal. `fd84d19d` |
| Failed tag read falls through to `{}` and erases all tags | openai | **Real** | `--freeform-tags` replaces the map, so a transient read failure followed by a successful update would wipe the tags the read-modify-write exists to protect. A failed read now retries **without writing**; only a successful read reporting empty/null becomes `{}`. `fd84d19d` |
| `grep` in the retry diagnostic aborts the loop under `pipefail` | xai | **False positive** | Pattern-matches to the `set -e`/`grep` bug this PR fixed in `check-workflow-invariants.sh`, but the syntactic context differs. Verified under `bash -eo pipefail`: `x="$(… \| grep nomatch \| head -1)"` exits **1** (a command substitution's status becomes the assignment's), while `printf '(%s)' "$(… \| grep nomatch \| head -1)"` exits **0** (discarded as an argument). The retry loop uses the argument form, and that step declares `shell: bash -e {0}` so `pipefail` is not enabled. No change. |
| Retry all errors rather than 409 only | openai, xai | Deferred | Fail-closed is preserved, so this is failure latency and diagnostics, not correctness. The code-matching cannot be tested against real OCI before tonight's run. |
| Assert jq `isinfinite` availability on the runner | xai | Deferred | GHA images ship jq 1.7; worth a follow-up assertion but not a blocker. |
| Unquoted `$ocid_sites`; cross-repo `COMP` note | xai, bedrock | N/A | The unquoted expansion is gone with the rewrite; the cross-repo blind spot is already logged OPEN in TASK-010-8. |

**Verified after the fixes:** all three invariant mutations now fail (soak-only removal, reaper-only removal, divergent values) with the baseline passing; and against a stubbed `oci` CLI, two consecutive read failures produced **zero** update calls, with the eventual write preserving pre-existing `owner` and `cost-center` tags.

## Cross-repo coordination (system-integration)

Conducted through tracked `docs/ToDos.md` in both repos — **not** `docs/discoveries/`, which is gitignored on both sides (`.gitignore:123` here, `:76-77` there) and swallowed an earlier message entirely.

**Their review found the bug in this PR's third change.** Their multi-agent review of system-integration PR #70 (now **merged**) surfaced a non-finite-deadline hole in *their* reaper; I tested whether the same weakness existed in ours and it did, live on `master`. Neither of my own two reviews of #169 caught it.

**Their three-item analysis, and what I did with each:**

1. **The two components fail in deliberately opposite directions** — our tagger fails **closed** (refuse to launch), their reaper fails **open toward cleanup** — and they asked that this not be "harmonised" later. Agreed, and now recorded *in the code* at the reaper's filter, since side by side the two read as an inconsistency someone would try to fix. Composed, they give the property both repos want: an untagged soak runner never starts, and any untagged instance that does exist gets cleaned up.
2. **A third site holds the compartment OCID** — system-integration's `ci/oci-runners/state.env` `COMP`, which decides both where instances are *created* and which compartment their reaper *scans*. Invariant 5 cannot see it. Logged **OPEN** in TASK-010-8 rather than guarded: cross-repo divergence fails **closed** (the launcher creates in one compartment, our tagging step lists the other, finds nothing, fails the launch) — loud and immediate, unlike the same-repo divergence the invariant catches, which would be silent until a soak died at 2h. Guarding it would need a network fetch of the pinned ref inside a required check.
3. **The orphaned VM from run 30584775602 is bounded**, covered by the cloud-init idle watchdog and the 2h reaper rule. Their observation worth keeping: **the missing tag is what makes it reapable** — had tagging succeeded and a later step failed, the exemption would have run to window end + 2h. That is an argument for keeping the tagger fail-closed even though it cost a launch.

**Tag contract** was confirmed to them verbatim from the diff rather than from prose, including a detail they could not have inferred: the value is a JSON **string**, not a number, because OCI freeform tags are a string→string map. A numeric comparison would pass their tests and never exempt anything in production.

**No hard dependency in either direction**, as they independently concluded: #169 landing first froze the tag contract before the consumer was written, which is the order both of us wanted.

## Follow-up (recorded in `docs/ToDos.md`, not only here)

Added to **TASK-010-7**: `launch-runner.sh` should tag **atomically at creation** (`oci compute instance launch --freeform-tags`) rather than leaving it to a follow-up update. This retry works, but it is a workaround — tagging at creation removes the race entirely and is the only way a tag is guaranteed present from the instance's first instant, closing the window in which a reaper could observe an untagged soak VM. Applies equally to the cost-tracking tags requested in the same handoff.

**TASK-010-8** is marked `review` in `docs/ToDos.md`.

## Note on scope

The branch is named for the invariant work, but the 409 retry is the headline change and the reason this needs to land promptly — without it no soak can launch at all.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/170"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788042540&installation_model_id=426883&pr_number=170&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F170&signature=d4737ccfef1c0fc87d575a6ba5c20439f9aae0846eece0062baefd451317c11d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Needs Review | 0.95 |

---

### Summary

The finite-deadline validation and bounded fail-closed retry are directionally sound, but two important guarantees are not fully implemented. The compartment invariant can pass with one of the two required declarations missing, and the tagger can erase the complete existing tag map when its read fails transiently. These should be corrected before merge; restricting retries to the documented transient conflict would also improve diagnostics and failure latency.

Solid, well-motivated fix: the 409 retry restores soak launches without weakening fail-closed tagging, invariant 5 correctly prefers equality-under-CI over a mutable Actions variable, and the reaper’s isinfinite/isnan/horizon select closes a real permanent-exemption billing hole (same class as system-integration PR #70). The one change that should land before merge is hardening the retry diagnostic grep with `|| true` — under Actions’ default `bash -eo pipefail`, a non-JSON update error aborts the step on the printf line and skips the remaining attempts, recreating the exact “guard cannot explain itself / set -e kills the loop” failure mode this PR already fixed in check-workflow-invariants.sh. Secondary nits: retry-all-errors vs 409-only, jq builtin availability on the reaper image, and the already-documented cross-repo COMP pin.

The code change addresses a critical issue with instance tagging during the provisioning phase and ensures that the compartment OCID is consistently pinned across workflows. The changes are well-documented, and the verification steps are thorough. The only minor issue is the documentation of a cross-repo invariant that is not currently guarded.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Invariant does not require both security-sensitive workflows to pin the OCID</b> (Reported by: openai)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 143)

The check verifies only that at least one workflow contains a matching literal and that all extracted literals have one unique value. If the declaration is removed from either ci-runner-reaper.yml or merge-recovery-soak.yml while the other remains, the invariant passes. It also passes if one declaration becomes malformed or unquoted and is no longer captured by the value regex. This leaves the stated compartment blast-radius and exemption-visibility contract unenforced.

**Suggestion:**
```
Extract and validate exactly one literal from each required workflow separately, fail if either is absent or malformed, and then compare the two values. Additional occurrences in other workflows can be checked independently if desired.
```

</details>

<details>
<summary><b>:orange_circle: Major: A failed tag read can cause all existing tags to be erased</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 497)

The `instance get` command suppresses errors and uses `|| true`, after which any empty, null, or non-object result is converted to `{}`. Because `--freeform-tags` replaces the complete map, a transient read failure followed by a successful update will overwrite all existing tags rather than preserving them. This contradicts the read-modify-write guarantee and can remove ownership or cost-attribution tags.

**Suggestion:**
```
Only use `{}` when a successful OCI response explicitly reports an empty tag map. Treat command failure or malformed output as a failed attempt and retry without issuing the update; after the retry budget, fail closed. Capture and report the read error rather than discarding it.
```

</details>

<details>
<summary><b>:orange_circle: Major: grep for OCI error code can abort retry loop under GHA pipefail</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 510)

GitHub Actions runs bash with `set -e -o pipefail` by default. On a failed update, the diagnostic line runs `grep -oE '"code": "[^"]*"' | head -1` inside a command substitution that is not a set -e exempt context. If the CLI error is non-JSON (network blip, TLS reset, plain-text rate limit, CLI crash), grep exits 1, the pipeline fails under pipefail, and the whole step aborts instead of sleeping and retrying. That is the same set -e/grep class of bug this PR just fixed in check-workflow-invariants.sh — and it undermines the retry for exactly the non-409 transient failures you most want to ride out. The happy path (OCI 409 JSON with a code field) works; other failures fail closed immediately with a worse signal than the intentional 12-attempt path.

**Suggestion:**
```
Add `|| true` (and optionally a fallback string):

```bash
"$(printf '%s' "$update_err" | grep -oE '"code": "[^"]*"' | head -1 || true)"
```

Or structure it as `code=$(... | grep ... || true)` before the printf.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unquoted $ocid_sites in error printf</b> (Reported by: xai)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 153)

`$(printf '%s ' $ocid_sites)` relies on word-splitting to turn newlines into spaces. Paths under `.github/workflows/` will not contain globs/spaces today, but it is inconsistent with the careful `set -e` hygiene elsewhere in this block.

**Suggestion:**
```
Use a quoted expansion with tr/paste, e.g. `$(printf '%s
' "$ocid_sites" | paste -sd' ' -)`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Retry loop retries every update failure rather than only provisioning conflicts</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 504)

Although the comments and PR description say HTTP 409 is explicitly retryable, the loop retries all nonzero exits, including authentication failures, invalid arguments, permission errors, and missing instances. Those failures cannot normally recover in three minutes, and delaying them obscures the real error because only an extracted code is logged during the loop.

**Suggestion:**
```
Inspect the OCI error status/code and retry only the expected 409 provisioning conflict (and any intentionally supported transient statuses). For other errors, print the complete error and fail immediately. Also avoid saying `retrying` after attempt 12 when no further retry occurs.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Cross-repo invariant documentation</b> (Reported by: bedrock)</summary>

**File:** `docs/ToDos.md` (line 51)

The documentation mentions a cross-repo invariant that is not currently guarded. Consider adding a note or a future task to address this.

**Suggestion:**
```
Add a note or a future task to address the cross-repo invariant mentioned in the documentation.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Confirm jq on the reaper runner provides isinfinite/isnan</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci-runner-reaper.yml` (line 145)

The new filter depends on `isinfinite` and `isnan` (jq ≥ 1.5/1.6). If this job ever runs on an image with an older jq, the filter fails as a whole and that reaper pass does nothing — fail-open toward leaving zombies, the opposite of the intent of this change. Likely fine on current GHA images, but it is now load-bearing for the billing-leak fix.

**Suggestion:**
```
Pin/assert jq version in the reaper setup step (e.g. `jq --version` gate), or use a portable predicate such as `tostring | test("^-?(inf|Infinity|nan)"; "i")` alongside the numeric comparisons so an ancient jq still fails toward cleanup.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Retry loop treats every update failure as retryable</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 505)

Any non-zero exit from `oci compute instance update` burns a full 12×15s window before fail-closed. Permanent errors (404 NotAuthorizedOrNotFound, 400 InvalidParameter, auth/config problems) will delay a doomed launch by ~3 minutes and add noise. Fail-closed is preserved, so this is latency/UX rather than correctness.

**Suggestion:**
```
Retry only on Conflict/409 (and maybe 429). On other codes, break immediately and fail closed with the error body. Example: extract code once per attempt; continue only if it matches Conflict; otherwise exit the loop and hit the existing failure path.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Cross-repo COMP OCID blind spot left open (acknowledged)</b> (Reported by: xai)</summary>

**File:** `docs/ToDos.md` (line 867)

Invariant 5 only covers in-repo workflow literals. system-integration's `state.env` `COMP` is a third site that decides create/list compartment; divergence fails loud on launch (list finds nothing) rather than silent 2h death, so risk is lower, but the blast-radius story is still multi-repo. The OPEN note is good; no code change required for this PR.

**Suggestion:**
```
Optional follow-up: when Lint already checks out the pinned SYSTEM_INTEGRATION_REF, assert `COMP=` equals the in-repo literal so the third site cannot drift unnoticed.
```

</details>
---

<details>

The code change addresses a critical issue with instance tagging during the provisioning phase and ensures that the compartment OCID is consistently pinned across workflows. The changes are well-documented, and the verification steps are thorough. The only minor issue is the documentation of a cross-repo invariant that is not currently guarded.

#### openai (gpt-5.6-sol)

The finite-deadline validation and bounded fail-closed retry are directionally sound, but two important guarantees are not fully implemented. The compartment invariant can pass with one of the two required declarations missing, and the tagger can erase the complete existing tag map when its read fails transiently. These should be corrected before merge; restricting retries to the documented transient conflict would also improve diagnostics and failure latency.

#### xai (grok-4.5)

Solid, well-motivated fix: the 409 retry restores soak launches without weakening fail-closed tagging, invariant 5 correctly prefers equality-under-CI over a mutable Actions variable, and the reaper’s isinfinite/isnan/horizon select closes a real permanent-exemption billing hole (same class as system-integration PR #70). The one change that should land before merge is hardening the retry diagnostic grep with `|| true` — under Actions’ default `bash -eo pipefail`, a non-JSON update error aborts the step on the printf line and skips the remaining attempts, recreating the exact “guard cannot explain itself / set -e kills the loop” failure mode this PR already fixed in check-workflow-invariants.sh. Secondary nits: retry-all-errors vs 409-only, jq builtin availability on the reaper image, and the already-documented cross-repo COMP pin.
</details>

---

jltatbeach (approved):

PR #161, opened 2026-07-29
Fix CI Runners, canceling queues, approval, and better parallelism

jeffrey-l-turner:
## What this fixes

CI runs were being silently cancelled, and PRs were losing checks with nothing to
point at. On 2026-07-29 this happened twice, in two different workflows, from the
same underlying GitHub behaviour: **only one run may be *pending* per concurrency
group, so each new arrival cancels the previously pending one** — and a cancelled
run reports no status at all.

- PR #143's `CI (fork PRs)` run (`30457308520`) was cancelled with **zero jobs
  started**. Its integration tests simply never happened; the PR showed no failure.
- PR #158's `CI` run for head `1d3d79b1` was cancelled while queued. All **12
  required checks never reported**, so the PR sat `BLOCKED` showing
  *"Expected — waiting for status to be reported"* forever.

Four fixes here.

### 1. Every fork run shared one concurrency lane

`ci-fork-pr.yml` used a single `ci-fork-pr` group for all PRs. `pull_request_target`
also fires for same-repo PRs — their jobs skip via the `fork == true` guards, but the
*run* still joined the lane. Six internal pushes between 13:41 and 15:07 each evicted
whatever was queued, including #143.

Now every run gets `ci-fork-pr-<PR number>`.

### 2. Approval waits held the serialization lock

The OCI serialization group and the `ephemeral-launch` reviewer gate were on the same
job. A job sitting in `waiting` holds any concurrency group it names, and GitHub lets
approvals pend for up to 30 days. Run `30422042122` (PR #142) waited from 04:21 and
held the lane for twelve hours.

Split into a lock-free `await_approval` job that holds only the wait, and a
`launch_ephemeral_runners` job that holds `oci-ephemeral-launch` for the ~1 minute of
actual VM creation. `await_approval` carries **no** `timeout-minutes` — a bound on a
job whose purpose is waiting on a human would cancel the pipeline whenever a
maintainer didn't click Approve promptly.

### 3. The approval gate did not cover most of the pipeline

`build_docker_image` had no dependencies, so an **unapproved** fork PR already compiled
fork-authored code on the permanent self-hosted runners, and the three `smoke_*` jobs
spent GitHub-hosted minutes. The gate only ever covered OCI VM creation.

`build_docker_image` now depends on `await_approval`, so every job is transitively
behind the gate. An unapproved fork PR consumes nothing. Approval — not the
concurrency grouping — is what holds untrusted work back, which matters now that fork
runs no longer share a single lane.

[...]

PR #171, opened 2026-07-30
fix(ci): restore the fork PR lane broken by actions/checkout's new guard

jeffrey-l-turner:
## Problem

Reported by @spreston8 on [#142](https://github.com/F1R3FLY-io/f1r3node-rust/pull/142#issuecomment-5124651243). Every fork PR dies within seconds at `Clone Repository`:

> `##[error]Refusing to check out fork pull request code from a 'pull_request_target' workflow. … To opt in, review the risks at https://gh.io/securely-using-pull_request_target and set 'allow-unsafe-pr-checkout: true' on the actions/checkout step.`

Recent `actions/checkout` releases added this guard, and the pipeline pins the **moving `@v4` tag** — so the fork lane broke *upstream*, with no change in this repo and no signal that anything had happened. This is why #142's integration tests never ran even after #161 fixed the concurrency eviction, and why the reopen-from-a-branch workaround has become the de facto path for outside contributors.

I verified every load-bearing claim in the report before acting:

| Claim | Verified |
|---|---|
| Pipeline pins the moving `@v4` tag | Yes — **all seven** checkout sites |
| Five checkouts of the code under test via `inputs.checkout_repository` | Yes — lines 101, 419, 516, 625, 725 |
| Those five set `persist-credentials: false` | Yes, all five |
| The system-integration checkouts are separate | Yes — they pin a trusted ref and need no opt-in |

## Changes

**`allow-unsafe-pr-checkout: true` on the five code-under-test checkouts.** The two system-integration checkouts fetch a pinned trusted ref and are deliberately untouched.

**A new invariant (#5) requiring both `persist-credentials: false` and `allow-unsafe-pr-checkout: true`** on every checkout of the code under test.

## Why the opt-in is safe here

Its premises are **machine-enforced, not conventional** — `check-workflow-invariants.sh` already asserts three of the four:

- the job **gates on `await_approval`**, so fork code is fetched only after a maintainer has reviewed the diff;
- **credentials are confined to `launch_ephemeral_runners`**, which checks out only the pinned system-integration ref and never the code under test;
- every credential-reading job **names an environment**;
- and the workflow file itself **executes from the base branch**, so fork code cannot alter what runs.

The fourth premise — `persist-credentials: false` — was convention only. That is the control keeping a writable token out of a workspace where untrusted code gets built, and opting past a security guard is precisely when convention stops being sufficient. Hence invariant 5. It also asserts the opt-in itself, so a future removal fails CI loudly rather than silently breaking outside contributors again — which is exactly how this went unnoticed for days.

## Mutation testing caught a defect in the new guard before it landed

The `persist-credentials` half was **unfalsifiable**. The explanatory comment beside these settings contains the literal string `persist-credentials: false`, and the unanchored regex matched the *comment* rather than the setting — so that half reported `ok` no matter what. Both patterns are now anchored to the start of the line, requiring a YAML key rather than a mention of one.

A guard defeated by its own documentation is worse than no guard: it manufactures confidence.

**Verified:** baseline passes; removing the opt-in from one site fails naming that site; removing `persist-credentials` from one site fails naming that site; YAML parses.

## Open question for reviewers

**Should `actions/checkout` be pinned by SHA across all seven sites?** The moving `@v4` tag is the *root cause* — an upstream behaviour change silently broke CI with no local edit, and the next one will do the same. Pinning would fix the class rather than this instance, the way `OCI_CLI_INSTALLER_URL` is already pinned with a checksum in this repo. Left out here because it touches two sites this bug does not require; happy to add it in this PR or a follow-up.

## Credit

Diagnosis and the proposed fix are @spreston8's; this PR implements it with the invariant added. They offered to submit it separately — taken up here to keep it moving, not to duplicate their work.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/171"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788045385&installation_model_id=426883&pr_number=171&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F171&signature=de9b82ef69923bb56c3bea2348e3c564cb76509efe165a698877594c5e03b11e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/scripts/check-workflow-invariants.sh
+++ b/.github/scripts/check-workflow-invariants.sh
@@ -124,6 +124,61 @@ EOF
 done
 ok "both pipeline callers pass secrets down"
 
+# 5. Fork-checkout hygiene. Every checkout of the code under test must set
+#    BOTH `persist-credentials: false` and `allow-unsafe-pr-checkout: true`.
+#
+#    The second is what makes the fork lane work at all: actions/checkout
+#    refuses fork code under pull_request_target without it, and because the
+#    action is pinned to the moving `@v4` tag, that guard arrived upstream and
+#    silently broke every fork PR at Clone Repository with no change in this
+#    repo. Asserting it means a future removal fails CI loudly instead of
+#    breaking outside contributors invisibly again.
+#
+#    The first is the control that opt-in leans on. Once we tell checkout to
+#    fetch untrusted code in a base-repo context, `persist-credentials: false`
+#    is what keeps a writable token out of a workspace that fork code will be
+#    built in. It was previously convention; opting past a security guard is
+#    exactly when convention stops being good enough.
+fork_bad_persist=""
+fork_bad_optin=""
+while read -r line_no flags; do
+    case "$flags" in
+        *P*) ;;
+        *) fork_bad_persist="$fork_bad_persist $PIPELINE:$line_no" ;;
+    esac
+    case "$flags" in
+        *A*) ;;
+        *) fork_bad_optin="$fork_bad_optin $PIPELINE:$line_no" ;;
+    esac
+done <<EOF
+$(awk '
+    function flush() {
+        if (is_fork) printf "%d %s%s\n", start, (has_persist ? "P" : "-"), (has_optin ? "A" : "-")
+        is_fork = 0; has_persist = 0; has_optin = 0
+    }
+    # Anchored to the start of the line so a YAML KEY is required, not just a
+    # mention of one. Unanchored, the comment above these settings — which
+    # names `persist-credentials: false` while explaining why it matters —
+    # satisfied the match by itself, and that half of the check could never
+    # fail. Caught by mutation testing; a guard defeated by its own
+    # documentation is worse than no guard, because it reports ok.
+    /^      - (name|uses):/ { flush(); start = NR }
+    /inputs\.checkout_repository/                       { is_fork = 1 }
+    /^[[:space:]]*persist-credentials:[[:space:]]*false/     { has_persist = 1 }
+    /^[[:space:]]*allow-unsafe-pr-checkout:[[:space:]]*true/ { has_optin = 1 }
+    END { flush() }
+' "$PIPELINE")
+EOF
+if [ -n "$fork_bad_persist" ]; then
+    err "checkout of fork-authored code without 'persist-credentials: false' at:${fork_bad_persist}; a writable token must never reach a workspace holding untrusted code"
+fi
+if [ -n "$fork_bad_optin" ]; then
+    err "checkout of fork-authored code without 'allow-unsafe-pr-checkout: true' at:${fork_bad_optin}; actions/checkout refuses fork code under pull_request_target without it, which breaks every fork PR at clone"
+fi
+if [ -z "$fork_bad_persist$fork_bad_optin" ]; then
+    ok "fork-code checkouts set persist-credentials:false and allow-unsafe-pr-checkout:true"
+fi
+
 if [ "$fail" -ne 0 ]; then
     printf '::error::%s\n' "workflow security invariants violated; see errors above"
     exit 1
--- a/.github/workflows/_integration-pipeline.yml
+++ b/.github/workflows/_integration-pipeline.yml
@@ -104,6 +104,21 @@ jobs:
           repository: ${{ inputs.checkout_repository != '' && inputs.checkout_repository || github.repository }}
           ref: ${{ inputs.checkout_ref != '' && inputs.checkout_ref || github.sha }}
           persist-credentials: false
+          # Fork PRs reach this pipeline through ci-fork-pr.yml's
+          # pull_request_target, and recent actions/checkout releases refuse to
+          # check out fork code from that trigger without this opt-in. Nothing
+          # changed here — the pin is the moving `@v4` tag, so the fork lane
+          # broke upstream and every fork PR has died at Clone Repository since.
+          #
+          # The opt-in is sound only because its premises are machine-enforced,
+          # not merely conventional. check-workflow-invariants.sh asserts that
+          # this job gates on await_approval (fork code is fetched only after a
+          # maintainer reviews the diff), that credentials are confined to
+          # launch_ephemeral_runners (which never checks out the code under
+          # test), and that `persist-credentials: false` is set right here. The
+          # workflow file itself executes from the base branch, so fork code
+          # cannot alter what runs. Do not add secrets to this job.
+          allow-unsafe-pr-checkout: true
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v3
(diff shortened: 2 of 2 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The five checkout sites are configured consistently and the immediate fork-PR failure should be resolved without persisting checkout credentials. The main concern is that the new security invariant is still syntactic and can be satisfied by keys outside the checkout action's `with` mapping, so it does not fully provide the machine-enforced guarantee described by the change. Pinning checkout to a reviewed SHA would also address the moving-tag failure class highlighted by this PR.

Solid, well-motivated fix: opting in with allow-unsafe-pr-checkout: true on the five untrusted checkouts restores the fork lane, persist-credentials: false remains in place, and invariant #5 makes both controls machine-enforced — including a real fix for the comment-matching false negative caught by mutation testing. The security story is credible given base-branch workflow execution, approval gating, and credential confinement (assuming those existing invariants still cover the same jobs). No critical vulnerabilities in the diff itself; main follow-ups are pinning actions/checkout by SHA to fix the class of breakages, softening the awk indentation assumption, and reducing the five-way duplicated comment block.

The code change effectively addresses the issue by adding the necessary opt-in for fork PRs and enforcing invariants to maintain security. The changes are well-documented and the security implications are thoroughly considered. A minor suggestion for further improvement would be to pin actions/checkout by SHA to prevent future breakages.

---

### Issues Found

<details>
<summary><b>:yellow_circle: Minor: Step boundary regex is hard-coded to exactly six spaces</b> (Reported by: xai)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 160)

Patterns require bare false/true. Values like persist-credentials: "false" or allow-unsafe-pr-checkout: "true" are valid YAML and accepted by actions/checkout, but would be reported as missing and fail the invariant. Unlikely given current style, but the failure mode would be a false CI red rather than a security gap.

---

The awk rule /^      - (name|uses):/ only recognizes steps indented with exactly six spaces. That matches the current workflow style, but a future edit that re-indents, uses a different step list depth, or writes '- uses:' under a differently nested job will silently stop classifying checkouts. Missed sites would not be checked for P/A flags.

**Suggestion:**
```
Optionally allow optional quotes: /^[[:space:]]*persist-credentials:[[:space:]]*["']?false["']?/ and the analogous true pattern.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Invariant does not verify that controls are checkout inputs</b> (Reported by: openai)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 160)

The regexes accept the required keys anywhere between step boundaries, including under `env:`, inside a `run: |` script, or in another nested mapping. In those cases the invariant would pass even though `actions/checkout` would not receive the setting. The patterns also accept prefixes such as `false-invalid` and `true-invalid` because they are not anchored at the end. Since this check is intended to machine-enforce a security premise, a false positive could allow a future workflow edit to restore credential persistence while the invariant still reports success.

**Suggestion:**
```
Parse the workflow as YAML and inspect each matching checkout step's `with` mapping directly. If retaining awk, track entry into the checkout step's `with:` block, require the expected indentation, and anchor scalar values, for example allowing only `false` or `true` followed by whitespace and an optional YAML comment.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Pin security-sensitive checkout actions by commit SHA</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 104)

The change explicitly relies on the behavior of a moving `actions/checkout@v4` tag, which already caused an unreviewed upstream behavior change to break this lane. A future tag update could again change the semantics of the unsafe-checkout opt-in or credential handling, and a compromised moving tag would create a supply-chain risk.

---

The root cause of this outage was the moving @v4 tag picking up an upstream behavior change with no local signal. Adding allow-unsafe-pr-checkout: true fixes this instance, but leaves the class open: the next checkout major/minor guard or default change will break the fork lane the same way. The PR already notes OCI_CLI_INSTALLER_URL is pinned with a checksum; the seven checkout sites should get the same treatment (pin to full SHA, with an optional floating version comment for humans).

**Suggestion:**
```
Pin all seven `actions/checkout` uses to a reviewed full commit SHA, retaining a comment indicating the corresponding release version, and update them through a controlled dependency-update process.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Invariant #5 is a strong addition; document the approval premise link</b> (Reported by: xai)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 127)

The new guard correctly enforces both controls and fixes the unanchored-regex self-match via mutation testing — that is the right response to opting past an upstream security guard. The comment states that await_approval gating and credential confinement are already asserted elsewhere; worth confirming those earlier invariants name the same five jobs that use inputs.checkout_repository so the 'premises are machine-enforced' claim stays end-to-end.

**Suggestion:**
```
In the invariant #5 header comment, cite the invariant numbers that enforce await_approval and credential confinement (e.g. 'see invariants #1 and #3') so future readers can verify the full chain without re-deriving it.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Consider pinning actions/checkout by SHA</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 104)

The PR mentions an open question about whether actions/checkout should be pinned by SHA across all seven sites to avoid future breakages due to upstream changes. This could be considered as a follow-up improvement.

**Suggestion:**
```
Pin actions/checkout to a specific SHA in a follow-up PR to ensure stability against upstream changes.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Identical 15-line security comment repeated five times</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 107)

The same block explaining why allow-unsafe-pr-checkout is safe is duplicated at every code-under-test checkout. That helps local readability, but drift is likely (one site gets an edited rationale, others do not), and it inflates review noise. The load-bearing enforcement already lives in check-workflow-invariants.sh.

**Suggestion:**
```
Keep a short one- or two-line pointer at each site (e.g. 'Required for fork PRs via pull_request_target; see invariant #5 — do not add secrets') and a single full rationale next to invariant #5 in the shell script, or above the first checkout only.
```

</details>
---

<details>

The code change effectively addresses the issue by adding the necessary opt-in for fork PRs and enforcing invariants to maintain security. The changes are well-documented and the security implications are thoroughly considered. A minor suggestion for further improvement would be to pin actions/checkout by SHA to prevent future breakages.

#### openai (gpt-5.6-sol)

The five checkout sites are configured consistently and the immediate fork-PR failure should be resolved without persisting checkout credentials. The main concern is that the new security invariant is still syntactic and can be satisfied by keys outside the checkout action's `with` mapping, so it does not fully provide the machine-enforced guarantee described by the change. Pinning checkout to a reviewed SHA would also address the moving-tag failure class highlighted by this PR.

#### xai (grok-4.5)

Solid, well-motivated fix: opting in with allow-unsafe-pr-checkout: true on the five untrusted checkouts restores the fork lane, persist-credentials: false remains in place, and invariant #5 makes both controls machine-enforced — including a real fix for the comment-matching false negative caught by mutation testing. The security story is credible given base-branch workflow execution, approval gating, and credential confinement (assuming those existing invariants still cover the same jobs). No critical vulnerabilities in the diff itself; main follow-ups are pinning actions/checkout by SHA to fix the class of breakages, softening the awk indentation assumption, and reducing the five-way duplicated comment block.
</details>

---

jltatbeach (approved):

PR #175, opened 2026-07-31
Fix/soak runner isolation and postmortem

jeffrey-l-turner:
Eight fixes to the merge-recovery soak, each diagnosed from a named failed run. The last one closes the root cause; the rest make failures survivable and diagnosable.

## Root cause, and the commit that closes it

`launch-runner.sh` registered every VM with one shared label set, so GitHub routed each queued job to whichever runner claimed it first — regardless of which workflow paid to launch it. The soak repeatedly ran on VMs launched by PR CI builds and died when the CI side finished with the runner it believed it owned.

Confirmed twice, same ~4 minute lag:

| CI integration finished | Soak died |
|---|---|
| 05:27:35Z | ~05:31Z |
| 07:08:14Z | 07:12:37Z |

**The quieter half:** a soak running on another workflow's VM boots *that* workflow's cloud-init, so `SYSTEM_INTEGRATION_REF` silently stopped describing the machine the soak ran on. Run `30606130771` executed the pre-fix idle watchdog while the pin named the commit that fixed it — 2410 lines of console history with no `pgrep` and no `Runner.Worker` in them. A pin that does not bind is worse than none, because it reads as a guarantee and stops people re-checking.

`d2b1d69c` bumps the pin to system-integration `main` `5b3144e4` and sets `RUNNER_LABELS`, so the soak registers under a label nothing else requests and runs on the VM it launched.

## Commits

| | |
|---|---|
| `c39f89c1` | Post-mortem was dead code — gated on a job output that is empty precisely when the runner is lost. Now finds the instance by a `github-run-id` tag that lives on the VM. |
| `8b9b6cc2` | Exclusive label via post-registration swap; `oci compute instance-console-history` → `console-history` (the former is not a real subcommand, and `\|\| true` hid it). |
| `e2ee9ae7` | Lookup piped through `2>/dev/null \|\| true` made auth failure, outage and genuine no-match indistinguishable, then exited green. |
| `e88d5a6a` | `Soak Checkpoint Publish` had **zero runs ever** — no `always()`, so a composite abort took the checkpoint down with the segment it was meant to salvage. |
| `43017d87` | Correct labels ≠ available runner. A poached VM passed every check, then hung against `timeout-minutes: 1440`. |
| `ac76e581` | Bounded reclaim: relaunch and retry, three launches max. |
| `7a14636c` | Abandon rather than terminate the poached VM (see below). Also made the shared-label removal best-effort. |
| `d2b1d69c` | Pin bump + `RUNNER_LABELS` on both launch sites. Closes the race at registration. |

## Two decisions worth recording

**Abandon, not terminate** (`7a14636c`, reversing `ac76e581`). Terminating a poached VM mid-job produces a runner-loss failure in the workflow that stole it — the exact signature this PR exists to diagnose. Three mechanisms already present identically from outside; a deliberate fourth means the next person debugging one must first rule out that we did it on purpose. Same money either way, since the runner is `--ephemeral` and self-terminates. system-integration's call, not ours — the original framing weighed only cost and soak windows, where it is genuinely close.

**No patching the pinned launcher.** Forcing the override in by rewriting `launch-runner.sh` between checkout and execution would have worked, and would have meant the pinned SHA no longer described what ran — the exact defect being fixed. Working around a pinning problem by breaking the pin is not a fix.

## Verified, not assumed

- Override on `main` by **ancestry** (`git merge-base --is-ancestor`), not grep — `main` already contained `__RUNNER_LABELS__`, the cloud-init placeholder, so grep would have reported success against a launcher that ignores the variable
- `RUNNER_LABELS` set on **both** steps invoking `launch-runner.sh`; missing it on the relaunch would make the retry re-enter the race it exists to escape, silently and only on the retry path
- Every `runs-on` label appears in the registered set — checked mechanically, since a mismatch is a silent overnight hang
- Value satisfies their guard: `REQUIRED_LABELS=(self-hosted linux $LABEL_ARCH oracle-cloud)`, both sites pass `amd64`
- Reclaim loop tested across idle / poached-once / poached-twice / poached-thrice, with the `oci` stub rigged to fail loudly if called — never invoked; the give-up path leaves no fourth VM
- Checkpoint readiness tested across no-output / empty-dir / aggregation-failure / success
- Lookup tested across auth-failure / unparseable / clean-no-match / match
- Compartment OCID pinning mutation-tested — one character breaks invariant 6
- 404-on-`DELETE` forward-compatibility tested both before and after `RUNNER_LABELS` takes effect

## Still open after this merges

**One overnight failure is unexplained.** `30611992344` died at 09:47Z after 2h33m in segment 2 with **no CI running**. The collision does not account for it. It is now diagnosable — post-mortem and checkpoint both work — but a similar failure tonight will have a different cause.

**Deferred by agreement with system-integration:** dropping `linux`/`x64`/`self-hosted` from their `DEFAULT_LABELS` and narrowing the guard to `oracle-cloud`, the only label whose omission actually breaks routing. GitHub auto-assigns the other three as read-only. Waits for one green soak — there are currently zero successes to regress against.

**Known gap:** invariant 8 (`bash -n` over every `run:` block) sweeps `.github/workflows/*` only. Composite actions under `.github/actions/*` have `run:` blocks too and were verified by hand. Same blind spot that let the reaper syntax error through.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/175"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788071070&installation_model_id=426883&pr_number=175&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F175&signature=576b20d0e45bda4143b81077b5b40741db3e3ad1a6b34a6a9518445855d708da"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/actions/soak-segment/action.yml
+++ b/.github/actions/soak-segment/action.yml
@@ -89,8 +89,25 @@ runs:
     # Everything below is checkpoint-only. The final segment publishes through
     # the normal perf_report job, which computes a real verdict and appends to
     # the history series.
+    #
+    # always(), because these steps exist for the runs that do NOT finish. A
+    # composite action aborts on the first failed step, so without it a segment
+    # that dies takes its own checkpoint down with it -- and the checkpoint is
+    # the only artifact a partial night can produce. That is how Soak Checkpoint
+    # Publish reached 2026-07-31 with zero runs in its entire history while
+    # three consecutive nights failed mid-segment. The salvage path was wired to
+    # the success path, where it is redundant: a segment that completes cleanly
+    # is one whose run goes on to publish a real verdict anyway.
+    #
+    # Note what this does NOT fix. When the RUNNER is lost rather than the step
+    # failing, no further step executes at all -- there is no machine left to
+    # run them on -- and always() cannot help. That was the actual failure mode
+    # on 30606130771, 30607922155 and 30611992344. This change matters for the
+    # segment that fails while its runner is still alive; the dead-runner case
+    # needs the soak to stop dying, not a better guard.
     - name: Aggregate checkpoint (${{ inputs.label }})
-      if: inputs.deadline_epoch != ''
+      id: aggregate
+      if: always() && inputs.deadline_epoch != ''
       env:
         SOAK_DIR: /tmp/merge-recovery-soak
         OUT_DIR: /tmp/merge-recovery-soak/checkpoint
@@ -101,16 +118,42 @@ runs:
         DURATION_SECONDS: ${{ inputs.duration_seconds }}
         RETRY_ATTEMPT: ${{ inputs.retry_attempt }}
         WINDOW_SECONDS: ${{ inputs.window_seconds }}
+        # Environment, not interpolation -- same rule the publish step below
+        # states explicitly: `${{ }}` inside `run:` is substituted before bash
+        # sees it, so the value is parsed as shell rather than quoted as data.
+        LABEL: ${{ inputs.label }}
       shell: bash -euo pipefail {0}
       run: |
         # in_progress suppresses the verdict: a partial run has fewer
         # iterations, a lower peak RSS and a throughput figure over a shorter
         # window than the baseline it would be compared against, so a verdict
         # here would measure the clock rather than the code.
-        ci-control/scripts/bench/aggregate-perf-report.sh
+        #
+        # Reachable now via always() with the soak step failed, so "the soak
+        # produced nothing at all" is a real case rather than an impossible
+        # one. That is not an error worth reporting -- a segment that died
+        # before writing a single result has nothing to checkpoint, and saying
+        # so once is more useful than failing a second time on top of the
+        # failure that caused it. Report readiness and let the caller decide.
+        if [ ! -d "$SOAK_DIR" ] || [ -z "$(ls -A "$SOAK_DIR" 2>/dev/null)" ]; then
+          echo "no soak output at $SOAK_DIR; nothing to checkpoint"
+          echo "ready=false" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+        if ! ci-control/scripts/bench/aggregate-perf-report.sh; then
+          echo "::warning::checkpoint aggregation failed for ${LABEL}; the segment result stands on its own"
+          echo "ready=false" >> "$GITHUB_OUTPUT"
+          exit 0
+        fi
+        echo "ready=true" >> "$GITHUB_OUTPUT"
 
+    # Gated on the aggregate having actually produced something, which keeps
+    # if-no-files-found: error meaningful. Dropping it to warn would have been
+    # the easy way to survive always(), at the cost of silencing a genuinely
+    # empty checkpoint on the success path -- the same trade that made the
+    # console-history typo invisible for as long as it was.
     - name: Upload checkpoint (${{ inputs.label }})
-      if: inputs.deadline_epoch != ''
+      if: always() && inputs.deadline_epoch != '' && steps.aggregate.outputs.ready == 'true'
       uses: actions/upload-artifact@v4
       with:
         name: soak-checkpoint-${{ github.run_id }}-${{ github.run_attempt }}-${{ inputs.label }}
@@ -119,7 +162,7 @@ runs:
         if-no-files-found: error
 
     - name: Publish checkpoint (${{ inputs.label }})
-      if: inputs.deadline_epoch != ''
+      if: always() && inputs.deadline_epoch != '' && steps.aggregate.outputs.ready == 'true'
       env:
         GH_TOKEN: ${{ inputs.gh_token }}
         # Passed as environment, never interpolated into the script body:
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -85,8 +85,25 @@ env:
   #     the RSS/floor guardian names the ceiling that killed it instead of
   #     surfacing a raw gRPC traceback — without which the tightened limits
   #     below would be self-obscuring. Plus their reaper's soak guards.
-  # On their `dev` and heading to `main`; a SHA is immutable either way.
-  SYSTEM_INTEGRATION_REF: 7b6c9dd9674b19732923ed49feee758829b2c27a
+  # 2026-07-31: bumped to system-integration main (PR #76 merge). This is the
+  # bump that makes every bump above it mean anything.
+  #
+  # Until now this pin did not bind. launch-runner.sh registered every VM with
+  # one shared label set, so GitHub routed the soak job to whichever runner
+  # claimed it first -- frequently a VM some CI run had launched, carrying the
+  # cloud-init THAT workflow pinned. Run 30606130771 executed the pre-fix idle
+  # watchdog while this variable named the commit that fixed it; the console
+  # history has no pgrep and no Runner.Worker anywhere in 2410 lines. A pin that
+  # does not bind is worse than no pin, because it reads as a guarantee and
+  # stops people re-checking.
+  #
+  # This ref carries the RUNNER_LABELS override, so the soak now registers its
+  # runner under a label nothing else requests and therefore runs on the VM it
+  # launched. Verified on main by ancestry (git merge-base --is-ancestor), not
+  # by grep: main already contained the string __RUNNER_LABELS__, the cloud-init
+  # template placeholder, long before the override existed. Grepping would have
+  # said yes against a launcher that ignores the variable.
+  SYSTEM_INTEGRATION_REF: 5b3144e4e567e985fa36cc8c69ee70ee63d599ac
   # Pinned OCI CLI installer (release tag + sha256). The checksum fails closed
   # if the tag is ever moved. Bump both together when upgrading.
   OCI_CLI_INSTALLER_URL: "https://raw.githubusercontent.com/oracle/oci-cli/v3.89.1/scripts/install/install.sh"
@@ -473,6 +490,18 @@ jobs:
           # The next step needs it to find this exact runner among all the
           # ci-eph-* runners registered at any moment.
           RUNNER_NAMES_FILE: /tmp/soak-runner-name
+          # Register under the exclusive label from the start rather than
+          # relabelling afterwards. Relabelling cannot revoke a job already
+          # assigned, so it always left a window; this closes it.
+          #
+          # The whole set, not just the pool label: RUNNER_LABELS replaces
+          # DEFAULT_LABELS outright. launch-runner.sh requires self-hosted,
+          # linux, the arch and oracle-cloud, and refuses to launch without
+          # them -- an omission here is a hard failure at launch, not a VM that
+          # boots healthy and never matches a runs-on. oracle-cloud in
+          # particular is the one whose absence actually breaks routing, and it
+          # must stay in step with the soak job's runs-on below.
+          RUNNER_LABELS: self-hosted,linux,x64,f1r3fly-rust-soak,oracle-cloud
         shell: bash -e {0}
         run: system-integration/ci/oci-runners/launch-runner.sh amd64
 
@@ -505,6 +534,12 @@ jobs:
           # Same file the launch step wrote to. launch-runner.sh APPENDS, so on
           # a relaunch below the newest name is the last line.
           RUNNER_NAMES_FILE: /tmp/soak-runner-name
+          # MUST match the launch step. This step relaunches on a lost race, and
+          # without it the replacement VM would register under the shared label
+          # set -- so the retry that exists to escape the race would re-enter it,
+          # and the soak would then wait on an exclusive label its own
+          # replacement does not carry. Silent, and only on the retry path.
+          RUNNER_LABELS: self-hosted,linux,x64,f1r3fly-rust-soak,oracle-cloud
         shell: bash -e {0}
         run: |
           # Bounded reclaim loop. Losing the registration race is recoverable:
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -404,7 +404,7 @@ jobs:
     environment: oci-credentials
     # 60, not 25: this job waits for the runner to register before relabelling
     # it (cloud-init 3-5 min, bounded at 15), and on a lost registration race it
-    # terminates the poached VM and launches another, up to three times. Worst
+    # abandons the poached VM and launches another, up to three times. Worst
     # case is three full 15-minute waits plus boots. The soak window absorbs
     # this; a hung launch job would not.
     timeout-minutes: 60
@@ -508,19 +508,14 @@ jobs:
         shell: bash -e {0}
         run: |
           # Bounded reclaim loop. Losing the registration race is recoverable:
-          # the poached VM is ours, so terminate it, launch another and try
-          # again. Three launches, because a pool busy enough to take three in a
-          # row is not a race any longer -- it is CI saturation, and burning
-          # more 16-OCPU VMs against it is waste, not persistence.
+          # abandon the poached VM to the job that took it, launch another and
+          # try again. Three launches, because a pool busy enough to take three
+          # in a row is not a race any longer -- it is CI saturation, and
+          # burning more 16-OCPU VMs against it is waste, not persistence.
           #
-          # Terminating a poached VM FAILS THE JOB THAT STOLE IT, with the same
-          # runner-loss error this branch exists to diagnose. That is a real
-          # cost paid by another workflow, accepted deliberately: the alternative
-          # is abandoning the VM to finish its stolen job, which spends the same
-          # money and leaves the soak waiting on a runner that will deregister
-          # the moment it is done. A re-runnable CI job is cheaper than a lost
-          # soak window. Once launch-runner.sh takes RUNNER_LABELS this loop
-          # stops firing, because there is no window left to lose.
+          # See the abandon-rather-than-terminate note below for why we do not
+          # simply take the VM back. Once launch-runner.sh takes RUNNER_LABELS
+          # this loop stops firing, because there is no window left to lose.
           MAX_LAUNCHES=3
           bound=""
           for launch in $(seq 1 "$MAX_LAUNCHES"); do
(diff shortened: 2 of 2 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The diagnostic command correction and tag-based instance resolution are useful improvements, but the runner isolation is not complete because the VM still registers in the shared pool before being relabeled. Additionally, the new fallback lookup suppresses operational errors and can misleadingly report that no tagged instance existed. Both issues affect the primary behavior this PR is intended to fix and should be addressed before merge.

Strong, well-motivated fix for a real cross-workflow runner theft bug and a dead postmortem path. Exclusive label bind/remove (fails closed), github-run-id freeform tagging, capture_diagnostics resolution by tag, and the console-history CLI name correction are all correct and carefully documented. No merge-blocking security or logic defects; remaining notes are hardening (safe jq arg passing), scaling of compartment-wide instance list, and closing the residual register-to-relabel race upstream in launch-runner.sh.

The PR addresses important isolation and postmortem issues for the soak runner. The changes look solid and well-documented. A minor improvement would be to avoid hardcoding sensitive information and to refactor the shell script for better readability.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Runner remains claimable by shared CI before relabeling</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 511)

The runner initially registers with the shared f1r3fly-rust-ci-ephemeral label and is only relabeled after polling detects it. A queued CI job can be assigned immediately upon registration, before the DELETE request removes the shared label. Changing labels afterward does not revoke an already assigned job, so the soak can still lose its VM or later run on a VM whose lifecycle is controlled by another workflow. This preserves the original isolation failure under concurrent CI load, despite narrowing the race window.

**Suggestion:**
```
Modify launch-runner.sh to accept a RUNNER_LABELS override and register this runner with f1r3fly-rust-soak instead of the shared label from the outset. Pass that override from this workflow and retain post-registration verification as a safety check.
```

</details>

<details>
<summary><b>:orange_circle: Major: OCI lookup failures are silently treated as an absent instance</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1347)

The instance search redirects OCI and jq errors to /dev/null and ends with `|| true`. Authentication failures, missing permissions, API outages, malformed responses, or jq errors therefore all produce an empty result and a successful diagnostic job claiming the runner likely died before self-tagging. This can silently discard the postmortem in exactly the failure path this change is intended to improve.

**Suggestion:**
```
Capture the OCI response in a separate command and fail or emit a distinct actionable error if listing instances fails. Only treat a successful list operation with no matching tag as the benign 'no instance found' case. Likewise, allow jq parsing failures to surface rather than suppressing them.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded OCID</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 91)

The OCID for the CI runner compartment is hardcoded. While this is mentioned as a security precaution, it would be better to use a secret or environment variable to store such sensitive information.

**Suggestion:**
```
Store the OCID in a GitHub secret or environment variable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Shell script readability</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 455)

The shell script could be broken down into smaller functions for better readability and maintainability.

**Suggestion:**
```
Refactor the shell script into smaller, reusable functions.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Runner name interpolated into jq expression</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 512)

The runner name is shell-interpolated into the --jq filter string (select(.name == "$name")). If the name ever contained jq metacharacters (quotes, pipes), the filter would break or behave unexpectedly. Input is currently controlled by launch-runner.sh, so this is low risk today, but the pattern is fragile.

**Suggestion:**
```
Pipe the API JSON to jq and pass the name via --arg so it is never parsed as code:
  rid="$(gh api \"repos/$GH_REPO/actions/runners\" --paginate \
    | jq -r --arg name \"$name\" '.runners[] | select(.name == $name) | .id' \
    | head -1)"
Or equivalent safe construction.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Wait loop could surface gh api errors</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 511)

Registration polling swallows all gh api errors (2>/dev/null || true) and only distinguishes 'not registered yet' from 'API broken' by eventually timing out after 15 minutes. That makes auth/permission regressions look like slow cloud-init.

**Suggestion:**
```
Optionally track consecutive API failures and fail fast with the gh error body after N failures, while still tolerating empty runner lists during cloud-init.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Documented residual race still worth closing upstream</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 548)

The comment correctly notes a window between registration (with the shared f1r3fly-rust-ci-ephemeral label) and the DELETE where a queued CI job can still claim the soak VM. The dual-label bind is a solid mitigation and fails closed, but the race remains real under load.

**Suggestion:**
```
Follow through on the noted RUNNER_LABELS (or equivalent) override in launch-runner.sh so the soak VM registers with f1r3fly-rust-soak from the start and never advertises the shared label. Until then this step is the right fail-closed guard.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Full compartment instance list filtered client-side</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1352)

Resolve uses `oci compute instance list --compartment-id ... --all` and then filters every instance in jq by freeform tag. As terminated ci-eph-* instances accumulate, this grows unbounded (pages, latency, memory) for a job that only needs one match.

**Suggestion:**
```
Prefer server-side filtering if the tenancy/CLI version supports it, e.g. --freeform-tags "{\"github-run-id\":\"$GITHUB_RUN_ID\"}", and/or narrow by display-name prefix / lifecycle-state when possible. Keep --all only as a fallback.
```

</details>
---

<details>

The PR addresses important isolation and postmortem issues for the soak runner. The changes look solid and well-documented. A minor improvement would be to avoid hardcoding sensitive information and to refactor the shell script for better readability.

#### openai (gpt-5.6-sol)

The diagnostic command correction and tag-based instance resolution are useful improvements, but the runner isolation is not complete because the VM still registers in the shared pool before being relabeled. Additionally, the new fallback lookup suppresses operational errors and can misleadingly report that no tagged instance existed. Both issues affect the primary behavior this PR is intended to fix and should be addressed before merge.

#### xai (grok-4.5)

Strong, well-motivated fix for a real cross-workflow runner theft bug and a dead postmortem path. Exclusive label bind/remove (fails closed), github-run-id freeform tagging, capture_diagnostics resolution by tag, and the console-history CLI name correction are all correct and carefully documented. No merge-blocking security or logic defects; remaining notes are hardening (safe jq arg passing), scaling of compartment-wide instance list, and closing the residual register-to-relabel race upstream in launch-runner.sh.
</details>

---

jeffrey-l-turner:
## Review resolutions (`e2ee9ae7`)

Thanks — two of these were real, and one of them is the same class of bug this PR exists to fix.

### Fixed

**`major` — OCI lookup failures silently treated as an absent instance** (openai, `merge-recovery-soak.yml:1347`)

Correct, and the more serious of the two. The lookup piped the CLI into jq under `2>/dev/null || true`, so expired credentials, a missing IAM policy, an API outage, a malformed response, and a genuine no-match all produced the same empty string — after which the job emitted the benign "died before self-tagging" warning and exited green.

That is precisely the defect this PR already documents in the `instance-console-history` typo: an error swallowed into a plausible-looking negative result. A post-mortem that cannot distinguish *nothing to find* from *could not look* is not a post-mortem.

The list and the filter are now separate commands, each failing loudly. Behaviour-tested against a stubbed CLI across all four paths:

| Scenario | Exit | Signal |
|---|---|---|
| OCI auth failure | 1 | `could not list instances…` |
| Unparseable listing | 1 | `could not parse the instance listing` |
| Clean listing, no match | 0 | `listing read cleanly` |
| Clean listing, match | 0 | `resolved by tag: …` |

Before the change, all four were indistinguishable.

**`minor` — runner name interpolated into jq expression** (xai, `:512`)

Fixed via `--arg`. Worth doing precisely because `launch-runner.sh` controls that name and is *outside this repository* — which is the entire premise of the step it appears in.

**`suggestion` — wait loop swallows `gh api` errors** (xai, `:511`)

Fixed. Polling now counts consecutive failures, prints the response body on the first and every tenth, and hard-fails after 20 — rather than timing out at 15 minutes with a message blaming slow cloud-init for an auth or permissions fault.

### Not fixable in this repository

**`major` — runner remains claimable by shared CI before relabeling** (openai, `:511`)

Correct, and your framing is sharper than my code comment: relabelling after assignment does not revoke an assignment already made, so the race survives under concurrent CI load. Narrowing is not closing.

`launch-runner.sh` lives in `F1R3FLY-io/system-integration` and hardcodes `LABELS`, so the VM cannot register correctly from this side. **Update:** that repo is implementing a `RUNNER_LABELS` override now, and their version is better than the one-liner I was going to propose — `${VAR:-default}` does not substitute for whitespace-only values, so my draft would have allowed an empty label set and a VM no `runs-on` could ever match.

Ordering matters and is agreed: their change lands, then the pin bump, then this repo sets the override and demotes the post-registration swap to a verification-only check. Setting it earlier would be worse than the status quo — the variable would be ignored and the soak would wait forever on a label nothing carries. Until then this step is the fail-closed guard.

### Declined, with reasons

**`minor` — hardcoded OCID** (bedrock, `:91`) — deliberate, and the alternative is the hazard. An Actions variable is mutable by anyone with repo admin, which is exactly what the comment above the literal argues against. `check-workflow-invariants.sh` enforces that this file and `ci-runner-reaper.yml` name the same compartment; mutation-verified that changing one character fails CI.

**`minor` — refactor shell into functions** (bedrock, `:455`) — these are GitHub Actions `run:` blocks, not a shell library. Each runs in its own process with no shared scope, so there is nothing for functions to be reused across.

**`suggestion` — server-side tag filtering** (xai, `:1352`) — reasonable in principle, but `--freeform-tags` is not a filter parameter on `oci compute instance list`; it would need `oci search`. Not worth the rewrite for a job that only runs when a soak has already failed. Revisit if compartment listing latency becomes real.

### Note on the abstention

`anthropic` abstained on `credit balance is too low` — a dead API key, not a review signal. Three voters produced three different verdicts, which is why consensus reads 33%; that is provider disagreement about severity, not an unresolved defect.

jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The exclusive runner-labeling and checkpoint changes appear well considered, but the postmortem path still has two material reliability gaps. Re-run attempts can resolve an older attempt's VM because the durable tag lacks run_attempt, and console-history API errors remain broadly swallowed and can be mistaken for an expected terminated-instance result. Both issues undermine the diagnostic guarantees central to this pull request and should be addressed before merge.

Strong, well-motivated infra fix: exclusive RUNNER_LABELS on both launch paths, runs-on aligned to f1r3fly-rust-soak, bounded reclaim without terminating poached VMs, checkpoint steps on always() with a ready gate, postmortem resolution via github-run-id tag, and the console-history subcommand correction all directly address the named failure modes. The main follow-up is confirming that every step after Resolve the soak instance is safe when no VM is found—otherwise the job-condition fix can recreate noisy diagnostic failures. Secondary nits are busy=unknown abandon behavior, substring label checks, and --all compartment listing cost; none look like merge-blocking security holes.

The code changes appear to be well-implemented and address the identified issues effectively. The logic for handling runner labels and ensuring exclusive runner usage is sound. Minor style improvements could enhance readability.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Instance tag does not distinguish workflow re-run attempts</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 881)

The VM is tagged only with github.run_id, but GitHub keeps the same run ID when jobs or an entire workflow are re-run and increments github.run_attempt instead. The resolver searches all instances, including terminated ones, and selects the newest instance carrying that run ID. If a later attempt loses its runner before self-tagging, the lookup can therefore resolve a VM from an earlier attempt and publish stale console history as the current failure's postmortem.

**Suggestion:**
```
Tag the instance with GITHUB_RUN_ATTEMPT as well as GITHUB_RUN_ID, and require both values in the resolver filter. A single composite tag such as github-run-key="$GITHUB_RUN_ID-$GITHUB_RUN_ATTEMPT" would also work. Consider including the repository identity too if the OCI compartment is shared across repositories.
```

</details>

<details>
<summary><b>:orange_circle: Major: Post-resolve steps may be ungated when instance is absent</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1470)

capture_diagnostics previously short-circuited on needs.soak.outputs.instance_id != ''. That guard is removed (correctly) and only 'Capture console history' is gated on steps.resolve.outputs.instance_id. Any later steps in this job (artifact upload, packing /tmp/postmortem, etc.) that still assume console output or a non-empty IID will now run on the 'died before self-tag' path and can fail the diagnostic job—or worse, upload an empty/partial postmortem with if-no-files-found: error—reintroducing a second red X on top of the real soak failure. Confirm every downstream step in this job is either gated on steps.resolve.outputs.instance_id != '' or tolerates a missing capture.

**Suggestion:**
```
Gate the remainder of the postmortem body on steps.resolve.outputs.instance_id != '', or make upload steps if-no-files-found: warn only when resolve produced no IID, and keep credentials cleanup on always().
```

</details>

<details>
<summary><b>:orange_circle: Major: Console-history API failures are still misreported as terminated instances</b> (Reported by: openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1574)

The capture, polling, and content-download commands suppress stderr and convert failures into empty or fallback values. Consequently, expired credentials, missing console-history permissions, OCI outages, rate limits, or another CLI regression are reported as the VM likely being terminated or as empty console content. This retains the same error/no-result ambiguity that the new instance-resolution code explicitly fixes and can silently defeat the postmortem path.

**Suggestion:**
```
Capture stderr and exit status separately. Treat failure as expected only when the previously queried lifecycle state establishes that capture is unavailable for a terminated instance; otherwise emit the OCI error and fail the diagnostic step. Also report a polling timeout explicitly and validate that get-content succeeded before treating the capture as complete.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Code formatting</b> (Reported by: bedrock)</summary>

**File:** `.github/actions/soak-segment/action.yml` (line 101)

Consider breaking long lines to improve readability.

**Suggestion:**
```
Break the long lines into multiple lines for better readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Code formatting</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 455)

Consider breaking long lines to improve readability.

**Suggestion:**
```
Break the long lines into multiple lines for better readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Failed gh api bodies may land in logs via 2>&1</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 560)

Registration wait captures gh api … 2>&1 and prints tails on repeated failure. That is the right trade for diagnosing auth vs slow boot, but app-token or gateway error pages can include sensitive headers/tokens in rare failure modes.

**Suggestion:**
```
Keep the consecutive-failure counter and message; redact or head/tail more aggressively (e.g. strip Authorization-like lines) before printf to the log.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Label presence checks use substring case patterns</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 620)

Verification is `case "$have" in *f1r3fly-rust-soak*)` / `*f1r3fly-rust-ci-ephemeral*)`. Because $have is a comma-joined sorted list this is probably fine with current names, but it is not a set membership test—future labels that are prefixes/extensions of either string would false-pass or false-fail.

**Suggestion:**
```
Test membership with jq on the labels array (e.g. `any(. == $l)`) or pad the joined string as `,${have},` and match `,f1r3fly-rust-soak,` exactly.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: busy=unknown is treated as claimed and triggers abandon</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 650)

busy is fetched with `2>/dev/null || echo unknown`, and only the exact string false binds the runner. A transient gh/API blip yields unknown, which falls through to abandon-and-relaunch. That is fail-safe against the 1440-minute hang, but three consecutive listing successes followed by three busy-endpoint blips could burn MAX_LAUNCHES without a real poach. The runner-list path already distinguishes API failure from slow boot; the busy probe does not.

**Suggestion:**
```
Retry the busy read a few times on non-true/false values before abandoning, or count unknown separately from busy=true and only abandon after repeated unknown responses.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Comments are excellent postmortems but very long in-line</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 486)

The narrative comments (pin non-binding, abandon-vs-terminate, always() vs lost runner) are the right knowledge to keep and match the PR body. In-workflow they also make the executable surface harder to scan and will drift if upstream launch-runner.sh behavior changes.

**Suggestion:**
```
Keep the fail-closed invariants and the 'do not terminate poached VMs' rule next to the code; move the multi-run chronology and decision history to docs/ or the PR description linked from a short comment.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Bind-loop body indentation makes control flow harder to audit</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 530)

The for launch in $(seq 1 "$MAX_LAUNCHES") body is effectively at column 0 inside the run: block. Given nested registration wait, label mutate, busy check, and relaunch, denser indentation (or a small helper script in-repo) would make matching done/fi and the relaunch guard easier to review next time.

**Suggestion:**
```
Indent the loop body consistently, or extract bind-and-reclaim into ci-control/scripts so the workflow only orchestrates.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unbounded compartment instance list on every postmortem</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1510)

oci compute instance list --compartment-id … --all pulls every instance page in ci-runner, including long-retained terminated VMs, then filters client-side. As the compartment accumulates history this grows latency and memory for a 15-minute diagnostic job.

**Suggestion:**
```
If the tenancy/IAM allows, narrow with display-name prefix, lifecycle-state, or freeform-tag filters server-side; at minimum document reaper retention so --all stays bounded.
```

</details>
---

<details>

The code changes appear to be well-implemented and address the identified issues effectively. The logic for handling runner labels and ensuring exclusive runner usage is sound. Minor style improvements could enhance readability.

#### openai (gpt-5.6-sol)

The exclusive runner-labeling and checkpoint changes appear well considered, but the postmortem path still has two material reliability gaps. Re-run attempts can resolve an older attempt's VM because the durable tag lacks run_attempt, and console-history API errors remain broadly swallowed and can be mistaken for an expected terminated-instance result. Both issues undermine the diagnostic guarantees central to this pull request and should be addressed before merge.

#### xai (grok-4.5)

Strong, well-motivated infra fix: exclusive RUNNER_LABELS on both launch paths, runs-on aligned to f1r3fly-rust-soak, bounded reclaim without terminating poached VMs, checkpoint steps on always() with a ready gate, postmortem resolution via github-run-id tag, and the console-history subcommand correction all directly address the named failure modes. The main follow-up is confirming that every step after Resolve the soak instance is safe when no VM is found—otherwise the job-condition fix can recreate noisy diagnostic failures. Secondary nits are busy=unknown abandon behavior, substring label checks, and --all compartment listing cost; none look like merge-blocking security holes.
</details>

---

PR #178, opened 2026-07-31
fix(ci): move smoke-test host ports out of the ephemeral range

jeffrey-l-turner:
The heavy pipeline's recurring 'just re-run it and it passes' failures are not flaky. They are deterministic, and the reason they looked random is that the condition is decided before the job starts.

Host ports 40400-40455 sit inside the kernel ephemeral range (32768-60999). The GitHub Actions agent opens its control connections to GitHub before the job exists and can source them from that range:

  ESTAB 10.1.0.54:40418 -> 20.85.130.105:443  users:(("Runner.Listener",pid=2060))
  ESTAB 10.1.0.32:40422 -> 140.82.112.24:443  users:(("hosted-compute-",pid=2034))

Those sockets are ESTAB for the life of the job. They never drain, so the step's 90-second wait could not have helped -- it only delayed the failure by a minute and a half. sysctl ip_local_reserved_ports cannot help either: it blocks new allocations and cannot reclaim a socket that already exists, which the step's own comment says. Re-running passed because a fresh runner's agent picked different source ports.

CI now binds 14400-14455, below the ephemeral floor, where nothing is ever auto-assigned. Container ports are unchanged, so nothing inside the node or on the compose network moves; docker/ci-ports.*.yml remap only the host side and are used only by this pipeline, so local development keeps 40400-40405. smoke_local_standalone binds on the host directly rather than through compose, so it moves via the node's own --*-port flags.

ports: !override is load-bearing and is called out in both files. Compose MERGES list-valued keys, so the first version of this change ADDED 14400-1445x while leaving 40400-4045x bound -- the collision would have survived, silently, and the smoke tests would have kept failing at the same rate. Caught by running docker compose config rather than by reading the spec.

The reserve step is now a precondition check rather than a drain wait, and it separates the two cases it previously conflated: a non-TIME_WAIT holder fails immediately and names it, because waiting cannot clear it; TIME_WAIT still gets its drain window, sampled through t=60 inclusive so the last sample is the decision. Behaviour-tested against both real signatures plus a clear range and a never-clearing TIME_WAIT.

Evidence, for anyone auditing this later: three re-run events across the last 60 CI runs, two of them this step. The six 'Check matrix slots' failures in the same window are the aggregator gate reporting those, not independent faults. Note that a naive search finds nothing -- a re-run flips the run conclusion to success and the jobs API serves only the latest attempt, so prior failures are visible only via /runs/{id}/attempts/{n}/jobs.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/178"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788121164&installation_model_id=426883&pr_number=178&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F178&signature=64d34f0d4ade18e3aa0ea33754f912b3b6ef1d1d6bf4da5041b71c68ea9c7054"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/scripts/check-node-ports.sh
+++ b/.github/scripts/check-node-ports.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+# Verify a host port range is free before starting anything that binds it.
+#
+# Usage: check-node-ports.sh <first-port> <last-port>
+#
+# CI binds 14400-14455 (see docker/ci-ports.*.yml), which is BELOW the kernel
+# ephemeral floor of 32768, so nothing can be auto-assigned into it and no
+# reservation is needed. This is a precondition check, not a drain wait.
+#
+# It used to reserve and wait on 40400-40455, inside the ephemeral range, and
+# that is the whole reason the heavy pipeline had to be re-run so often. The
+# Actions agent opens its control connections to GitHub before the job exists
+# and can source them from that range:
+#
+#   ESTAB 10.1.0.54:40418 -> 20.85.130.105:443  (("Runner.Listener",pid=2060))
+#   ESTAB 10.1.0.32:40422 -> 140.82.112.24:443  (("hosted-compute-",pid=2034))
+#
+# Those are ESTAB for the life of the job. They never drain, so the old 90s
+# wait could not have helped -- it only delayed the failure. A re-run passed
+# because a fresh runner's agent picked different source ports, which is
+# exactly why this read as an intermittent flake for months.
+#
+# TIME_WAIT is still worth waiting out: it genuinely clears in ~60s, and a
+# previous job on the same runner can leave some behind. So the two cases are
+# separated -- wait for what drains, fail immediately and name the holder for
+# what does not.
+set -euo pipefail
+
+if [ "$#" -ne 2 ]; then
+  echo "usage: $0 <first-port> <last-port>" >&2
+  exit 2
+fi
+first="$1"
+last="$2"
+for p in "$first" "$last"; do
+  case "$p" in
+    ''|*[!0-9]*)
+      echo "::error::port arguments must be numeric; got '$first' '$last'" >&2
+      exit 2
+      ;;
+  esac
+done
+if [ "$first" -gt "$last" ]; then
+  echo "::error::first port ($first) is above last port ($last)" >&2
+  exit 2
+fi
+# Refuse a range the kernel can auto-assign into. The whole point of the move
+# was to get below the ephemeral floor; a caller that passes an overlapping
+# range has reintroduced the original bug, and silently checking it would let
+# that pass as a green step.
+floor="$(sysctl -n net.ipv4.ip_local_port_range 2>/dev/null | awk '{print $1}')"
+case "$floor" in
+  ''|*[!0-9]*) floor=32768 ;;
+esac
+if [ "$last" -ge "$floor" ]; then
+  echo "::error::range ${first}-${last} reaches into the ephemeral range (starts at ${floor}); the Actions agent can already hold ports there before this job exists"
+  exit 1
+fi
+
+range="sport >= :${first} and sport <= :${last}"
+# 13 samples at t=0,5,...,60 -- the last sample IS the decision, so the loop
+# never sleeps past a check it does not then act on.
+for i in $(seq 1 13); do
+  snapshot="$(sudo ss -Htanp "$range" 2>/dev/null || true)"
+  busy="$(printf '%s' "$snapshot" | grep -c . || true)"
+  [ "$busy" -eq 0 ] && exit 0
+  stuck="$(printf '%s\n' "$snapshot" | grep -v '^TIME-WAIT' || true)"
+  if [ -n "$stuck" ]; then
+    echo "FAIL: port(s) in ${first}-${last} held by a socket that will not drain:"
+    printf '%s\n' "$stuck"
+    echo "Not TIME_WAIT, so waiting cannot clear these. If the holder is the"
+    echo "Actions agent, the CI range now overlaps the ephemeral range again --"
+    echo "check net.ipv4.ip_local_port_range on this runner."
+    exit 1
+  fi
+  [ "$i" -eq 13 ] && break
+  echo "  [${i}] ${busy} TIME_WAIT socket(s) in ${first}-${last}; waiting for drain..."
+  sleep 5
+done
+echo "FAIL: ${busy} TIME_WAIT socket(s) still holding ${first}-${last} after 60s:"
+printf '%s\n' "$snapshot"
+exit 1
--- a/.github/workflows/_integration-pipeline.yml
+++ b/.github/workflows/_integration-pipeline.yml
@@ -589,43 +589,78 @@ jobs:
         # the reservation already in place). Wait for the range to drain
         # (TIME_WAIT <= 60s) before starting anything that binds.
         run: |
-          sudo sysctl -w net.ipv4.ip_local_reserved_ports=40400-40455
-          for i in $(seq 1 18); do
-            busy=$(ss -Htan 'sport >= :40400 and sport <= :40455' | wc -l)
+          # CI binds 14400-14455 (see docker/ci-ports.*.yml), which is BELOW the
+          # kernel ephemeral floor of 32768, so nothing can be auto-assigned
+          # into it and no reservation is needed. This step is now a fast
+          # precondition check, not a drain wait.
+          #
+          # It used to reserve and wait on 40400-40455, inside the ephemeral
+          # range, and that is the whole reason the heavy pipeline had to be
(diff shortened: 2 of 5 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.93 |
| xai | `grok-4.5` | Feedback Provided | 0.9 |

---

### Summary

The functional change looks sound: CI-only Compose overrides correctly replace rather than merge host port mappings, all affected workflow endpoints and cleanup commands use the overrides, and the direct-host smoke test receives matching CLI port flags. No security, logic, or performance defects are evident. The only actionable issue is stale workflow documentation that still describes the removed sysctl reservation and old port range.

Solid, well-diagnosed CI fix: moving host binds from 40400-40455 to 14400-14455 (below the ephemeral floor), using ports: !override so compose does not silently merge both ranges, and splitting ESTAB/non-drain holders from TIME_WAIT wait logic are all correct and match the failure mode described. No security concerns. Remaining feedback is maintainability and polish—deduplicate the thrice-pasted precondition script, refresh stale step comments that still mention reservation/40400, and keep shard compose -f flags consistent with standalone.

The code change effectively addresses the issue of port conflicts by moving the CI host ports out of the ephemeral range. The logic for checking and waiting for port availability is well-implemented. Minor improvements to comment clarity could enhance readability, but overall the change is solid and ready for approval.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Existing comments still describe the removed reservation behavior</b> (Reported by: bedrock, openai, xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 589)

The comments explaining the port range change could be slightly more concise to improve readability.

---

The comments immediately above each run block still talk about sysctl reservation and draining 40400-40455 (e.g. 'the reservation already in place', 'binds 40400-40405 directly on the host. Reservation + drain'). The script bodies were updated, but the surrounding step comments were not, which will confuse the next reader.

---

The comments immediately preceding the three port checks still say that the range is reserved with sysctl and then drained, including references to ports 40400-40455. The new implementation explicitly removes the reservation, checks 14400-14455, and only waits for TIME_WAIT sockets. These stale comments contradict the new detailed comments and could mislead future maintainers. Similar stale text appears above the smoke_docker_shard and smoke_local_standalone checks.

**Suggestion:**
```
Consider rephrasing the comments to be more direct and concise.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Shard job compose invocations may be inconsistent</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 810)

standalone updates every docker compose invocation (up, logs, down) to include -f docker/ci-ports.standalone.yml. In the shard job, the diff only shows Start shard and Cleanup gaining -f docker/ci-ports.shard.yml; intermediate steps such as 'Wait for bootstrap Running state' appear unchanged. Logs usually still resolve via project name, so this is likely benign, but it is inconsistent and could bite if compose config is re-read without the override.

**Suggestion:**
```
Pass the same -f docker/shard.yml -f docker/ci-ports.shard.yml pair on every compose call in the job for symmetry with the standalone job.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Port-precondition script duplicated three times</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 592)

The same ~40-line reserve/drain script (and the long rationale comment) is copy-pasted into smoke_docker_standalone, smoke_docker_shard, and smoke_local_standalone. Any future fix to the ss parsing, sample count, or messaging has to be applied in three places and can drift (the local job already carries a slightly wrong 'see docker/ci-ports.*.yml' framing even though it binds via CLI flags).

**Suggestion:**
```
Extract the check into a small shared script (e.g. scripts/ci/wait-for-host-ports.sh 14400 14455) and invoke it from each job. Keep the long rationale once in that script or in the ci-ports yml files.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: TIME-WAIT detection is slightly brittle on ss output format</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 622)

stuck sockets are detected with grep -v '^TIME-WAIT'. That matches current iproute2 ss -H output (state is the first column, hyphenated), and the PR notes it was behaviour-tested. It can still mis-classify if ss ever pads the state column, localizes output, or if a line is a banner/error that does not start with a state token—those would be treated as non-draining holders and fail the job immediately.

**Suggestion:**
```
Prefer matching a leading state token explicitly, e.g. awk '$1 != "TIME-WAIT"' or grep -vE '^(TIME-WAIT)', and/or pin the ss flags further (-4 if you only bind IPv4 host ports) so the format stays predictable.
```

</details>
---

<details>

The code change effectively addresses the issue of port conflicts by moving the CI host ports out of the ephemeral range. The logic for checking and waiting for port availability is well-implemented. Minor improvements to comment clarity could enhance readability, but overall the change is solid and ready for approval.

#### openai (gpt-5.6-sol)

The functional change looks sound: CI-only Compose overrides correctly replace rather than merge host port mappings, all affected workflow endpoints and cleanup commands use the overrides, and the direct-host smoke test receives matching CLI port flags. No security, logic, or performance defects are evident. The only actionable issue is stale workflow documentation that still describes the removed sysctl reservation and old port range.

#### xai (grok-4.5)

Solid, well-diagnosed CI fix: moving host binds from 40400-40455 to 14400-14455 (below the ephemeral floor), using ports: !override so compose does not silently merge both ranges, and splitting ESTAB/non-drain holders from TIME_WAIT wait logic are all correct and match the failure mode described. No security concerns. Remaining feedback is maintainability and polish—deduplicate the thrice-pasted precondition script, refresh stale step comments that still mention reservation/40400, and keep shard compose -f flags consistent with standalone.
</details>

---

spreston8 (approved):
Non-blocking: the 25-line port-check script is duplicated verbatim across all three smoke jobs — executable logic ×3 that must evolve in lockstep. A `check-node-ports.sh` in `.github/scripts/` (beside `check-workflow-invariants.sh`) taking the range as an argument collapses the triplication.

jeffrey-l-turner:
Good catch, and taken — `d756eee2`.

The check is now `.github/scripts/check-node-ports.sh <first> <last>`, beside `check-workflow-invariants.sh` as you suggested. **145 lines out of the workflow, 12 in**, and the three smoke jobs are each a single call.

Worth adding that the copies **had already started to diverge**: a docs pass reworded the comments per job, so the three blocks were no longer byte-identical even though the logic still was. Your "must evolve in lockstep" concern was already one commit from being real rather than hypothetical.

### Parameterising created a hole, so the script closes it

Taking the range as an argument means a caller can now pass a range back **inside** the ephemeral window (32768–60999) — silently reintroducing the exact bug this branch exists to fix, since the Actions agent can already hold ports there before the job starts. The inline version could not have had this problem because the range was hardcoded.

So the script reads `net.ipv4.ip_local_port_range` and refuses any range reaching into it:

```
::error::range 40400-40455 reaches into the ephemeral range (starts at 32768);
the Actions agent can already hold ports there before this job exists
```

A mistaken caller now fails loudly at the check rather than passing green and flaking weeks later.

### Re-verified after extraction

Six paths against a stubbed `ss`/`sysctl`:

| Case | Exit |
|---|---|
| range clear | 0 |
| agent `ESTAB` — fails fast, names the holder | 1 |
| `TIME_WAIT` never clears — fails after 60s | 1 |
| **range overlaps ephemeral floor** | 1 |
| non-numeric args | 2 |
| inverted range | 2 |

The first three match the inline version exactly; the fourth is new.

One thing worth flagging from doing it: my first extraction pass was wrong. It matched only two of the three blocks — the reworded comments meant text matching no longer found the third — and its end marker hit the first `exit 1` inside the loop rather than the last, leaving corrupted YAML. Reverted and redone with a step-boundary pass. Mentioning it because "three identical blocks" was exactly the assumption that made the naive approach look safe, which is the same reason the duplication was worth removing.

PR #82, opened 2026-06-17
chore: clean up clippy and deny warnings

jeffrey-l-turner:

PR #190, opened 2026-08-01
fix(soak): fail closed on host protection; repair checkpoint, bench, …

jeffrey-l-turner:
…and dashboard paths

Post-mortem of weekend soak run 30713818751 (VM lost to kernel OOM after repeated guardian breaches):

- write-soak-summary.sh: rename jq variable $def — a jq keyword that jq 1.6 on the Ubuntu 22.04 runner rejects; add a jq-keyword lint to the summary test harness
- run-merge-recovery-soak.sh: end the whole soak on any host-protection breach (no retry, later segments no-op); add an orchestrator-level host guardian that survives pytest death and kills node processes/containers on memory-floor breach; make the failure-evidence copy selective and timed; emit a minimal fallback summary.json so checkpoints keep valid metadata; surface bench segment logs on failure
- run-bench-segment.sh: default DEPLOYER_KEY to the funded bootstrap fixture key from docker/.env (bench segments failed 100% silently)
- soak-checkpoint-publish.yml: pass --repo to gh run download (ran in a non-git cwd and misreported every failure as 'artifact not ready')
- merge-recovery-soak.yml: dashboard falls back to the newest checkpoint when the runner is lost (latest-* only, no history append); post-mortem job releases the reaper exemption so a dead VM is collected in ~30min
- docs: work log for this sweep plus the randomized-exercise planning log

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/190"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788220591&installation_model_id=426883&pr_number=190&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F190&signature=d57145bd97d0aad88c7ffc22cfab9f89c610e288cfd6aac23a19058bb192a486"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,4 +1,5 @@
 self-hosted-runner:
   labels:
     - f1r3fly-rust-ci-ephemeral
+    - f1r3fly-rust-soak
     - oracle-cloud
--- a/.github/oci-validation.env
+++ b/.github/oci-validation.env
@@ -14,4 +14,4 @@
 # .github/workflows/_integration-pipeline.yml — a reusable workflow cannot
 # source this file at env-resolution time, so the pin is duplicated there.
 # Bump both in the same commit.
-SYSTEM_INTEGRATION_REF=369d49df2f97e65b3d0ad869aa668a7383b11179
+SYSTEM_INTEGRATION_REF=66de4f95141cbef9c1308d58fbb872b66cf75197
--- a/.github/scripts/check-workflow-invariants.sh
+++ b/.github/scripts/check-workflow-invariants.sh
@@ -18,12 +18,12 @@ PIPELINE=.github/workflows/_integration-pipeline.yml
 fail=0
 
 err() {
-    printf '::error::%s\n' "$1"
-    fail=1
+	printf '::error::%s\n' "$1"
+	fail=1
 }
 
 ok() {
-    printf 'ok: %s\n' "$1"
+	printf 'ok: %s\n' "$1"
 }
 
 # Emit one line of facts per job in a workflow file:
@@ -35,7 +35,7 @@ ok() {
 # and counting those would let the guard pass on a workflow whose real `needs:`
 # had been deleted.
 scan_jobs() {
-    awk '
+	awk '
         /^[[:space:]]*#/ { next }
         /^  [A-Za-z_][A-Za-z0-9_-]*:[[:space:]]*$/ {
             job = $1
@@ -71,11 +71,11 @@ scan_jobs() {
 #    Removing either `needs:` is the abuse path the ephemeral-launch approval
 #    exists to close.
 for job in launch_ephemeral_runners build_docker_image; do
-    if scan_jobs "$PIPELINE" | awk -v j="$job" '$1 == j && $5 == 1 { found = 1 } END { exit !found }'; then
-        ok "$job gates on await_approval"
-    else
-        err "$job in $PIPELINE no longer depends on await_approval; unapproved fork PRs could spend CI resources"
-    fi
+	if scan_jobs "$PIPELINE" | awk -v j="$job" '$1 == j && $5 == 1 { found = 1 } END { exit !found }'; then
+		ok "$job gates on await_approval"
+	else
+		err "$job in $PIPELINE no longer depends on await_approval; unapproved fork PRs could spend CI resources"
+	fi
 done
 
 # 2. Credential concentration. Exactly one job in the reusable pipeline may hold
@@ -84,24 +84,24 @@ done
 cred_jobs="$(scan_jobs "$PIPELINE" | awk '$3 == 1 || $4 == 1 { print $1 }')"
 cred_count="$(printf '%s' "$cred_jobs" | grep -c . || true)"
 if [ "$cred_count" = "1" ] && [ "$cred_jobs" = "launch_ephemeral_runners" ]; then
-    ok "credentials confined to launch_ephemeral_runners"
+	ok "credentials confined to launch_ephemeral_runners"
 else
-    # Flattened to one line: a GitHub error annotation captures only its first
-    # line, so a multi-line job list would hide every name after the first.
-    cred_list="$(printf '%s' "$cred_jobs" | tr '\n' ' ')"
-    err "expected exactly one credential-bearing job (launch_ephemeral_runners) in $PIPELINE, found: ${cred_list:-none}"
+	# Flattened to one line: a GitHub error annotation captures only its first
+	# line, so a multi-line job list would hide every name after the first.
+	cred_list="$(printf '%s' "$cred_jobs" | tr '\n' ' ')"
+	err "expected exactly one credential-bearing job (launch_ephemeral_runners) in $PIPELINE, found: ${cred_list:-none}"
 fi
 
 # 3. Environment scoping. Any job anywhere that reads OCI or GitHub App
 #    credentials must name an environment, so access is a property of the
 #    environment rather than of the repository secret store.
 for workflow in .github/workflows/*.yml; do
-    while read -r job has_env reads_oci reads_app _gated _calls _inherits; do
-        [ -n "${job:-}" ] || continue
-        if [ "$reads_oci$reads_app" != "00" ] && [ "$has_env" = "0" ]; then
-            err "$workflow job '$job' reads privileged credentials without naming an environment"
-        fi
-    done <<EOF
+	while read -r job has_env reads_oci reads_app _gated _calls _inherits; do
+		[ -n "${job:-}" ] || continue
+		if [ "$reads_oci$reads_app" != "00" ] && [ "$has_env" = "0" ]; then
+			err "$workflow job '$job' reads privileged credentials without naming an environment"
+		fi
+	done <<EOF
 $(scan_jobs "$workflow")
 EOF
 done
@@ -113,12 +113,12 @@ ok "every credential-reading job names an environment"
 #    config as malformed (runs 30495007422 and 30498890544). Guarded because the
 #    line reads like redundant over-sharing and invites deletion.
 for caller in .github/workflows/ci.yml .github/workflows/ci-fork-pr.yml; do
-    while read -r job _has_env _reads_oci _reads_app _gated calls inherits; do
-        [ -n "${job:-}" ] || continue
-        if [ "$calls" = "1" ] && [ "$inherits" = "0" ]; then
-            err "$caller job '$job' calls the integration pipeline without 'secrets: inherit'; the launch job would see empty credentials"
-        fi
-    done <<EOF
+	while read -r job _has_env _reads_oci _reads_app _gated calls inherits; do
+		[ -n "${job:-}" ] || continue
+		if [ "$calls" = "1" ] && [ "$inherits" = "0" ]; then
+			err "$caller job '$job' calls the integration pipeline without 'secrets: inherit'; the launch job would see empty credentials"
+		fi
+	done <<EOF
 $(scan_jobs "$caller")
 EOF
 done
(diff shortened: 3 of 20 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The checkpoint, dashboard, benchmark-key, and summary fixes are generally well targeted, but the new host guardian has a fail-closed correctness defect in its cleanup path: a normal no-match result from `pkill` can prevent Docker containers from being killed while still recording a breach. Since this mechanism is intended to prevent another kernel OOM and directly covers the prior Docker-only failure mode, it should be corrected before merge. The jq keyword lint is also incomplete, though that is a lower-severity test-hardening concern.

This is a strong, well-motivated post-mortem fix set: jq 1.6 `$def` breakage, silent bench key misconfig, checkpoint download `--repo`, dashboard fallback for lost runners, selective evidence copy, fail-closed host protection, and reaper-exemption release all map cleanly to the incident. The implementation is mostly careful (fail-soft reaper release, in_progress history exclusion, minimal summary fallback). Before merge, fix or verify the EXIT-trap composition and the `$ITERATION_DIR/.started` reference used by harness-breach detection—those are the two places the new safety logic can silently weaken. Minor hardening around `.env` quote parsing and post-iteration guardian checks would make the fail-closed path more reliable.

The changes address several critical issues that led to silent failures and incorrect behavior in the soak tests. The improvements in failure logging, variable name changes to avoid jq keyword conflicts, and the addition of the --repo argument in the workflow file are all positive steps. However, ensure that all changes are thoroughly tested and documented to prevent future issues.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Guardian can exit before killing Docker containers</b> (Reported by: openai, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 442)

The guardian subshell inherits the script's errexit behavior. `pkill -9 -f '/tmp/rnode'` returns status 1 when no matching host process exists, causing the subshell to exit before running `docker kill`. This is especially likely in the exact failure mode being addressed, where the remaining nodes exist only as Docker containers. The breach marker is written first, so the main loop may then report a protected shutdown even though the memory-consuming containers were never stopped. There is also a race where the main process sees the marker and its EXIT trap kills the guardian before mitigation completes.

---

The new host-guardian setup installs `trap '...' EXIT`, which replaces any previous EXIT trap rather than composing with it. If this script already (or later) registers EXIT cleanup (compose down, sampler stop, temp dir removal, marker writes), that cleanup will silently stop running once the guardian is enabled. That is a common bash footgun and is especially risky in a soak orchestrator whose post-failure evidence path matters.

**Suggestion:**
```
Make every cleanup command explicitly fail-soft and distinguish breach detection from mitigation completion, for example `pkill ... || true` and `{ docker ps ... | xargs -r docker kill; } || true`, then atomically publish the breach marker after cleanup attempts complete. Alternatively have the main process wait for the guardian before finalizing when it observes the marker.
```

</details>

<details>
<summary><b>:orange_circle: Major: Missing DEPLOYER_KEY handling</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/run-bench-segment.sh` (line 35)

The script now defaults DEPLOYER_KEY to the bootstrap key from the environment file, which is a good fix for the silent failures. Ensure that this default is well-documented and that users are aware of the requirement for a funded deployer key.

**Suggestion:**
```
Add a comment explaining the requirement for a funded deployer key and the default behavior.
```

</details>

<details>
<summary><b>:orange_circle: Major: Improved failure logging</b> (Reported by: bedrock)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 345)

The addition of logging the tail of the segment and bench logs on failure is a good improvement for debugging. Ensure that these logs are easily accessible and that the output is clear and concise.

**Suggestion:**
```
Consider adding more context to the log messages to make them more informative.
```

</details>

<details>
<summary><b>:orange_circle: Major: Harness breach probe depends on possibly missing .started file</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 562)

Breach detection uses `find ... -newer "$ITERATION_DIR/.started"`, but this diff never creates `$ITERATION_DIR/.started`. On GNU find, a missing reference file makes `-newer` error (stderr discarded) and the time filter ineffective, so the harness-marker channel degrades to the pytest.log grep fallback. That fallback is weaker and version-dependent, undercutting the 'three channels' design.

**Suggestion:**
```
At iteration start, `touch "$ITERATION_DIR/.started"` (or compare against the iteration directory mtime / pytest invocation timestamp you already record). If `.started` is created elsewhere, add a guard that skips `-newer` when the file is absent.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Added --repo argument</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/soak-checkpoint-publish.yml` (line 95)

The addition of the --repo argument to gh run download is a good fix for the issue where the command failed due to not being in a git repository. Ensure that this change is well-documented.

**Suggestion:**
```
Add a comment explaining the necessity of the --repo argument in this context.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: .env value may retain quotes when loading DEPLOYER_KEY</b> (Reported by: xai)</summary>

**File:** `scripts/bench/run-bench-segment.sh` (line 45)

BOOTSTRAP_PRIVATE_KEY is parsed with a simple awk `key=value` split. Checked-in `.env` files often quote values (`BOOTSTRAP_PRIVATE_KEY="..."` or `'...'`). If quotes are present, DEPLOYER_KEY will include them and latency-benchmark.sh will still fail, only now with a more confusing key format error.

**Suggestion:**
```
Strip optional surrounding single/double quotes after the read, e.g. `DEPLOYER_KEY="${DEPLOYER_KEY%\"}"; DEPLOYER_KEY="${DEPLOYER_KEY#\"}"` (and the same for `'`), or reuse whatever env loader the repo already uses for docker/.env.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: jq compatibility lint does not cover all variable declarations</b> (Reported by: openai)</summary>

**File:** `scripts/bench/test-write-soak-summary.sh` (line 73)

The grep only recognizes variables introduced by `as`, `--arg`, `--argjson`, and `--slurpfile`. jq variables can also be introduced in constructs such as function parameters or `label`, so a keyword variable can still reach jq 1.6 while this lint passes. This makes the new regression protection less comprehensive than its comment suggests.

**Suggestion:**
```
Prefer running a jq 1.6 parser in CI, such as through a pinned container or downloaded binary. If the heuristic is retained, document its limitations and extend it to all jq variable-binding forms used by the scripts.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Variable name change</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/write-soak-summary.sh` (line 48)

The change from $def to $mdef to avoid conflicts with jq keywords is appropriate. Ensure that all instances of this variable are updated consistently throughout the script.

**Suggestion:**
```
Perform a thorough search to ensure all instances of $def are replaced with $mdef.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Guardian breach is only polled at loop head during success path</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 455)

The top-of-loop check correctly fail-closes between units of work, and the failure path also inspects `$HOST_GUARDIAN_BREACH`. However, if the guardian fires during a long in-flight iteration/bench that still exits 0 (or hangs until an external timeout), the soak can continue longer than intended after the host was already defended. Given the post-mortem (OOM of pytest while containers lived), this is mostly mitigated but not fully closed.

**Suggestion:**
```
After each iteration/bench returns, unconditionally test `-s "$HOST_GUARDIAN_BREACH"` before scheduling more work; optionally have the guardian also signal the main PID (e.g. kill -USR1) for faster bailout.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Open item on RSS ceiling is the real remaining soak risk</b> (Reported by: xai)</summary>

**File:** `docs/work-logs/soak-oci-image-defects-20260801.md` (line 88)

The work log correctly notes that fail-closed stops the burn but the load profile still peaks above the 14336MB ceiling on the 32GB VM. After this PR, weekend soaks may now fail fast every time until ceiling or load is retuned—by design, but operators should expect red runs rather than a 'fixed soak'.

**Suggestion:**
```
Track a follow-up issue/PR for ceiling vs profile tuning before the next gated weekend run; the canary plan already listed is the right gate.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: jq-keyword lint is a good regression net; tighten spacing variants</b> (Reported by: xai)</summary>

**File:** `scripts/bench/test-write-soak-summary.sh` (line 72)

The lint would have caught the `$def` outage and is well justified. The `as \$` alternative only allows a single space, so variants like `as  $def` or `as	$def` would slip through.

**Suggestion:**
```
Use `as[[:space:]]+\$($JQ_KEYWORDS)` in addition to the `--arg/--argjson/--slurpfile` patterns.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Broad pkill pattern is acceptable only because this is a dedicated soak VM</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 436)

`pkill -9 -f '/tmp/rnode'` can match any process whose cmdline contains that string. In the locked-down soak image this is likely fine and matches the fail-closed goal, but it is worth keeping scoped if this script is ever reused on a shared runner.

**Suggestion:**
```
Prefer tighter matchers (pid files written at node start, `pgrep -f '^/tmp/rnode(\s|$)'`, or docker-only kill when provider=docker) and document that this path assumes a dedicated VM.
```

</details>
---

<details>

The changes address several critical issues that led to silent failures and incorrect behavior in the soak tests. The improvements in failure logging, variable name changes to avoid jq keyword conflicts, and the addition of the --repo argument in the workflow file are all positive steps. However, ensure that all changes are thoroughly tested and documented to prevent future issues.

#### openai (gpt-5.6-sol)

The checkpoint, dashboard, benchmark-key, and summary fixes are generally well targeted, but the new host guardian has a fail-closed correctness defect in its cleanup path: a normal no-match result from `pkill` can prevent Docker containers from being killed while still recording a breach. Since this mechanism is intended to prevent another kernel OOM and directly covers the prior Docker-only failure mode, it should be corrected before merge. The jq keyword lint is also incomplete, though that is a lower-severity test-hardening concern.

#### xai (grok-4.5)

This is a strong, well-motivated post-mortem fix set: jq 1.6 `$def` breakage, silent bench key misconfig, checkpoint download `--repo`, dashboard fallback for lost runners, selective evidence copy, fail-closed host protection, and reaper-exemption release all map cleanly to the incident. The implementation is mostly careful (fail-soft reaper release, in_progress history exclusion, minimal summary fallback). Before merge, fix or verify the EXIT-trap composition and the `$ITERATION_DIR/.started` reference used by harness-breach detection—those are the two places the new safety logic can silently weaken. Minor hardening around `.env` quote parsing and post-iteration guardian checks would make the fail-closed path more reliable.
</details>

---

jltatbeach (approved):

jeffrey-l-turner:
## Cross-repo merge sequence (paired with system-integration#82)

This PR is the consumer half of the host-protection breach-marker contract; F1R3FLY-io/system-integration#82 is the producer (`ResourceMonitor` writes `host-protection-breach.txt` in the monitor output dir).

**Sequence:**

1. system-integration#82 → `main` — all checks green now.
2. This PR → `master` when its pipeline is green. **Merge order between 1 and 2 does not matter:** if this lands first, the pinned harness never writes the marker and the driver's other two channels (orchestrator guardian marker, pytest.log grep) carry the fail-closed behavior; if #82 lands first, the marker is written and unread. Both partial states are safe.
3. **`SYSTEM_INTEGRATION_REF` pin bump** to the post-#82 `main` SHA — neither PR contains it, and it is the step that activates the harness-marker channel. Recommended before the next weekend soak window.

Producer-side note: #82 now also clears any inherited marker at monitor start and writes it via temp-file rename, so your `-newer "$ITERATION_DIR/.started"` filter is defense in depth rather than the only guard against stale markers. Absence remains inconclusive by design (a monitor OOM-killed with pytest writes nothing) — exactly why your guardian marker stays the primary channel.

PR #193, opened 2026-08-02
fix: RSS Runaway Fix

jeffrey-l-turner:
Fixes the node RSS runaway that killed soak runs between 1.5 and 4.5 hours in (kernel OOM on the 32GB soak VM after repeated host-protection breaches).

## Root cause

The replay cache was capped by entry count only — each entry retains a full event log, so 192 entries could still hold gigabytes. Block processing ran up to 100 blocks concurrently, each holding a live replay runtime and hot store, multiplying the pressure.

## Fixes

- **Replay cache is byte-capped**: 32MB alongside the 192-entry cap, with per-entry retained-bytes accounting that includes key bytes; oversized entries are rejected, LRU eviction enforces both caps, and `get` preserves the stored key.
- **Bounded parallel block processing**: default drops from a hardcoded 100 to 2, overridable via `F1R3_MAX_PARALLEL_BLOCKS` (clamped to tokio's semaphore limit so bad values cannot panic startup).
- **Observability for both sides of the change**: new gauges for `replay-cache.entries` / `replay-cache.retained-bytes` and `block-processing.active` / `parallel-limit` / `queue.pending` — the last one instruments the inbound queue where pressure moves when drain concurrency shrinks. The soak driver now scrapes these per node each iteration into `node-memory-timeseries.tsv`, so a future breach names the responsible cache instead of just total RSS.
- **Benchmark harness hardening**: Casper readiness preflight before flooding, live finalization harvesting via the deploy-status API (previously every deploy showed 0% finalized), and deduplicated polling.

## Review resolutions

All critical/major findings from the multi-agent review and reviewer comments are addressed on the branch:

| Finding | Resolution |
|---------|------------|
| Byte cap excluded key allocations; `get` could swap the stored key for a larger-backed clone (critical) | `c3a6e931` — admission/eviction charge `key heap bytes + entry bytes` with length-based (reproducible) accounting; `get` uses `move_index` so the stored key is never replaced; proptest invariants extended to the charged-bytes contract |
| Finalization harvester re-polled every deploy each cycle (major) | `c3a6e931` — in-memory finalized-sig set; only pending deploys are re-polled |
| Unvalidated `F1R3_MAX_PARALLEL_BLOCKS` could exceed tokio's permit max and panic startup (major) | `c3a6e931` — clamped to `Semaphore::MAX_PERMITS`, with test |
| 100→2 shifts memory into the uninstrumented inbound queue; queue-depth gauge requested | `39c3200f` — `block-processing.queue.pending` gauge sampled from the channel's permit accounting via `WeakSender`; scraped per node by the soak driver |
| 100→2 throughput cost unquantified | Deliberate default (documented in the work log); empirical 2-vs-100 comparison comes from the next dev daily soak using the new queue/latency instrumentation — see Follow-ups |
| `Bytes::len()` accounting blind to shared backing allocations — a small cached slice could pin a whole decode buffer (round-2 major, GPT + Claude) | `f6fccb31` — every `Bytes` is deep-copied into a private, exactly-`len` buffer at cache insert (entries and keys); accounting is now exact and cached events no longer pin decode buffers, with pointer-inequality tests proving the copy |
| `event_heap_bytes` complexity (style) | Acknowledged; left as-is — the per-variant walk is the accounting itself |

## Verification

- Property-based invariant suite on the cache (entry/byte caps hold, accounting equals the sum over live entries, admission contract, LRU ordering) plus updated unit tests; runs in normal CI.
- Soak driver behavior pinned by the fail-closed driver test, including the per-node metrics scrape and the queue-pending column.
- Full 11-crate release suite green locally and in CI.

## Follow-ups

- Throughput evidence for parallelism 2 vs 100 (block-processing rate, catch-up drain, finalization latency) comes free from the next dev daily soak once this merges; the new queue-depth gauge is the instrument for it.
- The soak RSS ceiling (14336MB) vs load profile tuning remains an open item tracked in the work log.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/193"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788265740&installation_model_id=426883&pr_number=193&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F193&signature=a222fedda27a56854e1a55b2dd8fc595b32e399044f0d81060b6a50858e70076"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Needs Review | 0.89 |

---

### Summary

This PR addresses RSS runaway via two product changes — a byte-capped (32MB) replay cache with per-entry retained-bytes accounting and LRU eviction, and a reduction of block-processing parallelism from a hardcoded 100 to a configurable default of 2 — plus observability (new gauges), soak-driver metric scraping, and CI test coverage including a solid proptest invariant suite for the cache. The Rust code is careful (saturating arithmetic, poisoning-aware locks, guard-based gauge tracking) and well-tested. The main concern requiring human judgment is the 50x parallelism reduction as a product-wide default, which trades sync/catch-up throughput for memory headroom; the work log confirms it is intentional, but the throughput impact should be validated. Remaining items are minor: silent fallback on invalid env values, lock-scope nit in put(), an approximate byte-accounting model worth documenting (especially shared prost::Bytes pinning), and a benign but racy concurrent file read plus substring-grep dedup in the reworked latency benchmark script. Overall functional and safe to merge with the noted improvements.

The change adds useful byte-based replay-cache eviction, lower block-processing concurrency, observability, and targeted tests. However, the new cache accounting does not cover all memory actually retained and can therefore provide a false sense of a strict 32 MB bound, which is significant for a PR specifically addressing RSS runaway. Additionally, an oversized but valid parallelism environment value can panic the node at startup. These issues should be resolved or explicitly bounded and documented before merge.

Solid RSS-focused fix: the replay cache gains a real byte cap with careful admission/eviction accounting and a strong proptest invariant suite, block processing concurrency becomes explicitly bounded (default 2) with active/limit gauges, and soak/latency tooling is improved to attribute memory to those caches. The implementation looks correct and well-tested; the main concern is product impact of dropping parallel block processing from 100→2, plus the byte accounting being a payload estimate rather than true RSS. No security issues stood out. Safe to merge after consciously accepting the parallelism default and ideally tightening defaults/docs around the byte cap.

The code changes address performance issues by introducing a byte-capped replay cache and reducing block processing parallelism. The changes are generally well-implemented, but there are opportunities for improving readability and maintainability through refactoring and adding comments.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Byte cap does not account for all memory retained by cache entries</b> (Reported by: openai, xai)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 47)

The retained-byte calculation counts the event vector, selected nested buffers, and post-state length, but excludes the ReplayCacheKey allocations, IndexMap storage, Arc allocation overhead, and potentially backing allocations retained by prost::bytes::Bytes slices. Bytes::len() reports the visible slice length rather than necessarily the size of the shared backing allocation. The get path also discards the stored key and inserts key.clone() from the lookup argument, which can replace a compact key with one backed by a larger shared allocation. Consequently, retained_bytes <= max_bytes is not a reliable bound on memory actually retained by the cache, undermining the primary RSS-runaway mitigation.

---

retained_bytes accounts for event payload heap data and post_state length, but not ReplayCacheKey storage, IndexMap/hash-table overhead, Arc allocation overhead, or allocator slop. With MAX_REPLAY_CACHE_BYTES at 32MiB, actual process RSS attributable to the cache can exceed the gauge/cap, so the metric may under-explain future RSS breaches.

**Suggestion:**
```
Either document that retained_bytes is a payload estimate (not RSS), or add a conservative per-entry/key overhead constant so the cap is a closer upper bound. Including key fingerprint/state-hash lengths in accounting would be a cheap improvement.
```

</details>

<details>
<summary><b>:orange_circle: Major: Complexity in event_heap_bytes calculation</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 11)

The function event_heap_bytes has a complex calculation that could be simplified or broken down into smaller functions for better readability and maintainability.

**Suggestion:**
```
Consider refactoring the event_heap_bytes function to use helper functions for each event type to reduce complexity.
```

</details>

<details>
<summary><b>:orange_circle: Major: Parallel block processing default cut from 100 to 2</b> (Reported by: xai)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 24)

max_parallel_blocks changes from a hardcoded 100 to a product-wide default of 2 (overridable via F1R3_MAX_PARALLEL_BLOCKS). This is a large behavioral change that will materially reduce concurrent block processing and can increase backlog latency under bursty catch-up or high inbound block rates. The work log states this is intentional for RSS control, but it is still a production throughput/latency tradeoff that should be explicitly validated outside soak (steady-state TPS, catch-up time, propose latency) before merge.

**Suggestion:**
```
Call out the default change in release/operator notes, and consider documenting recommended values by machine class. If soak is the only evidence, add a short justification (RSS delta vs finalization lag) near the constant or in the PR body so future tuners know why 2 was chosen.
```

</details>

<details>
<summary><b>:orange_circle: Major: Default parallel block limit dropped from 100 to 2 — significant throughput regression risk</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 24)

MAX_PARALLEL_BLOCKS_DEFAULT of 2 replaces the previous hardcoded value of 100. While this is intentional per the work log (to bound RSS), a 50x reduction in block-processing parallelism can severely slow initial sync / catch-up for nodes joining an established network, and could cause the bounded queue (F1R3_MAX_BLOCKS_IN_PROCESSING=512) to back up under bursty block arrival. The trade-off between memory ceiling and sync throughput deserves explicit benchmarking beyond the soak test.

**Suggestion:**
```
Document expected sync-time impact and consider a higher default (e.g., 4-8) or an adaptive limit based on available memory. At minimum, prominently document F1R3_MAX_PARALLEL_BLOCKS in operator-facing docs since it's now a critical tuning knob.
```

</details>

<details>
<summary><b>:orange_circle: Major: Unbounded environment value can panic Semaphore construction</b> (Reported by: openai)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 60)

configured_max_parallel_blocks accepts any positive usize. Tokio Semaphore::new panics when the requested permit count exceeds Semaphore::MAX_PERMITS, so a syntactically valid value such as usize::MAX in F1R3_MAX_PARALLEL_BLOCKS can crash the node during startup. Large values below that limit can also defeat the memory-safety goal of this change.

**Suggestion:**
```
Validate the configured value against tokio::sync::Semaphore::MAX_PERMITS and preferably a practical product maximum. Reject invalid values with a clear startup error or clamp them while emitting a warning. Add tests for oversized numeric values.
```

</details>

<details>
<summary><b>:orange_circle: Major: Harvester re-polls finalization status for every unfinalized deploy each cycle</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `scripts/bench/latency-benchmark.sh` (line 218)

Each poll interval rewrites over the entire submits.tsv and issues one HTTP finalization request per not-yet-finalized deploy. For longer/higher-rate runs this becomes O(pending) HTTP per tick and can perturb the system under test.

---

Every POLL_INTERVAL the harvester iterates the full submits list and issues one curl per not-yet-finalized sig. For long durations at higher rates this becomes O(n) HTTP requests per poll cycle against the node under test, which can itself perturb the latency being measured.

---

For every submitted deploy on every polling cycle, the harvester runs grep over finals.tsv. This makes finalization tracking grow quadratically with the number of submissions and creates many subprocesses. Long-duration or high-rate benchmark runs may spend substantial time scanning files rather than polling status.

**Suggestion:**
```
Acceptable at the default 2 deploys/sec, but consider batching or capping the per-cycle poll count if this script is used at higher rates, and note the observer effect in the script header.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: InMemoryReplayCache::new still means unlimited bytes</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 148)

new(max_entries) delegates to with_limits(max_entries, usize::MAX). RuntimeManager correctly uses with_limits(..., 32MiB), but any other caller of new()/default_capacity() silently gets no byte cap—the original RSS failure mode. This is an easy regression footgun.

**Suggestion:**
```
Give default_capacity()/new() a sane byte default (e.g. the same 32MiB), or #[deprecated] new() in favor of with_limits so unlimited-byte caches are explicit.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Oversized-entry rejection check performed while holding the mutex</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 186)

In put(), the `self.max_entries == 0 || entry_bytes > self.max_bytes` check only depends on immutable fields and the entry, but it is evaluated after acquiring the state mutex. Rejected oversized entries needlessly serialize on the lock.

**Suggestion:**
```
Move the early-return check before `self.state.lock()` so oversized/disabled puts never contend on the mutex.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Invalid F1R3_MAX_PARALLEL_BLOCKS values are silently ignored</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 59)

configured_max_parallel_blocks silently falls back to the default of 2 when the env var is empty, '0', or non-numeric. An operator who sets F1R3_MAX_PARALLEL_BLOCKS=0 or a typo'd value gets no feedback that their configuration was rejected.

**Suggestion:**
```
Emit a tracing::warn! when a set-but-invalid value is discarded, so misconfiguration is visible in logs (the 'Starting bounded block processing' info line helps, but doesn't explain why the operator's value was ignored).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent indentation</b> (Reported by: bedrock)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 162)

The indentation in the block processing loop is inconsistent, which can make the code harder to read.

**Suggestion:**
```
Ensure consistent indentation throughout the file.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Env-var read moved into BlockProcessorInstance::new reduces testability</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 165)

The max_parallel_blocks parameter was removed from the constructor and replaced with an internal call to max_parallel_blocks() which reads the process environment. This makes the instance harder to construct deterministically in tests (env-var-dependent) and couples construction to global process state, whereas the previous design allowed dependency injection.

**Suggestion:**
```
Keep a parameterized constructor (or a with_parallel_limit variant) and have the call site in node_runtime.rs pass max_parallel_blocks(); this preserves both the new default behavior and injectability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Harvester reads SUBMITS_FILE concurrently with the deploy-flood writer; dedup uses substring grep</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/latency-benchmark.sh` (line 217)

harvest_finalizations runs in the background reading $SUBMITS_FILE while the main loop appends to it, so a poll iteration can read a partially written last line (truncated sig). Additionally, `grep -q -F "$sig" "$FINALS_FILE"` is a substring match — a deploy sig that is a prefix of another already-finalized sig would be wrongly skipped, and a truncated sig read mid-write could match spuriously.

**Suggestion:**
```
Anchor the dedup match to the tab-delimited field (e.g., `grep -q -F $'	'"$sig"$'	'` or use awk field comparison), and consider having the writer append via a temp-line+mv or having the harvester skip malformed lines defensively.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: retained_bytes is an approximation — document what it excludes</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 44)

The retained-bytes accounting covers Vec capacity and per-event heap payloads but excludes ReplayCacheKey heap size, IndexMap bucket overhead, Arc allocation headers, and Bytes refcount/shared-buffer semantics (a Bytes slice into a large shared buffer is counted at slice length while potentially pinning the whole buffer). The 32MB cap is therefore a lower bound on actual retention in some scenarios.

**Suggestion:**
```
Add a doc comment on ReplayCacheEntry::new / retained_bytes stating the estimate's scope and known exclusions, particularly the shared-Bytes pinning caveat, so future readers don't treat the gauge as exact RSS attribution.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Magic 32MiB replay-cache byte limit lacks inline rationale</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 114)

MAX_REPLAY_CACHE_BYTES = 32 * 1024 * 1024 is central to the RSS fix but has no comment tying it to observed soak RSS, entry-size distribution, or relationship to MAX_REPLAY_CACHE_ENTRIES / MAX_REPLAY_CACHE_EVENT_LOG_ENTRIES.

**Suggestion:**
```
Add a short comment referencing the soak ceiling and that admission is fail-open (oversized entries are skipped, not fatal).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Replay cache gauges only refreshed on the put path</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 230)

record_replay_cache_metrics is invoked only when an entry is stored. Cache hits (get) and clear() do not refresh the gauges, so replay-cache.entries / retained-bytes can be stale after a clear until the next admission, which could confuse soak-time attribution the work log describes.

**Suggestion:**
```
Also refresh the gauges in (or after) clear(), or emit them from a periodic metrics tick alongside the other cache-size gauges.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Env knobs added without operator-facing docs in-tree</b> (Reported by: xai)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 59)

F1R3_MAX_PARALLEL_BLOCKS is now a supported product default override, but the diff only documents it in a work-log discovery note. Operators/debuggers may not find it.

**Suggestion:**
```
Mention F1R3_MAX_PARALLEL_BLOCKS (and the new block-processing/replay-cache metrics) in README or runbook config section alongside other F1R3_* env vars.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Whole-file reformat (spaces→tabs) mixed with functional changes</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/latency-benchmark.sh` (line 30)

The switch to tab indentation and comment realignment touches nearly every line of latency-benchmark.sh, obscuring the actual functional changes (preflight readiness loop, background harvester, deploy-finalization-status API polling) in the diff.

**Suggestion:**
```
In future PRs, land mechanical reformatting as a separate commit/PR from behavioral changes to keep reviews tractable.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Finalization harvest uses substring grep for deploy IDs</b> (Reported by: xai)</summary>

**File:** `scripts/bench/latency-benchmark.sh` (line 220)

harvest_finalizations uses grep -q -F "$sig" against FINALS_FILE/SUBMITS_FILE. Fixed-string substring matches can false-positive if one deploy id is a prefix/substring of another, skipping work or mis-detecting finalization. Unlikely with full hex hashes, but brittle.

**Suggestion:**
```
Match on the full TSV field (e.g. awk -F'	' -v s="$sig" '$2==s{found=1} END{exit !found}') or keep an associative set of seen sigs in the harvester process.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add comments for script functionality</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/test-latency-benchmark.sh` (line 1)

The script lacks comments explaining its purpose and functionality, which could be helpful for future maintainers.

**Suggestion:**
```
Add comments at the top of the script to describe its purpose and key functionality.
```

</details>
---

<details>

This PR addresses RSS runaway via two product changes — a byte-capped (32MB) replay cache with per-entry retained-bytes accounting and LRU eviction, and a reduction of block-processing parallelism from a hardcoded 100 to a configurable default of 2 — plus observability (new gauges), soak-driver metric scraping, and CI test coverage including a solid proptest invariant suite for the cache. The Rust code is careful (saturating arithmetic, poisoning-aware locks, guard-based gauge tracking) and well-tested. The main concern requiring human judgment is the 50x parallelism reduction as a product-wide default, which trades sync/catch-up throughput for memory headroom; the work log confirms it is intentional, but the throughput impact should be validated. Remaining items are minor: silent fallback on invalid env values, lock-scope nit in put(), an approximate byte-accounting model worth documenting (especially shared prost::Bytes pinning), and a benign but racy concurrent file read plus substring-grep dedup in the reworked latency benchmark script. Overall functional and safe to merge with the noted improvements.

The code changes address performance issues by introducing a byte-capped replay cache and reducing block processing parallelism. The changes are generally well-implemented, but there are opportunities for improving readability and maintainability through refactoring and adding comments.

#### openai (gpt-5.6-sol)

The change adds useful byte-based replay-cache eviction, lower block-processing concurrency, observability, and targeted tests. However, the new cache accounting does not cover all memory actually retained and can therefore provide a false sense of a strict 32 MB bound, which is significant for a PR specifically addressing RSS runaway. Additionally, an oversized but valid parallelism environment value can panic the node at startup. These issues should be resolved or explicitly bounded and documented before merge.

#### xai (grok-4.5)

Solid RSS-focused fix: the replay cache gains a real byte cap with careful admission/eviction accounting and a strong proptest invariant suite, block processing concurrency becomes explicitly bounded (default 2) with active/limit gauges, and soak/latency tooling is improved to attribute memory to those caches. The implementation looks correct and well-tested; the main concern is product impact of dropping parallel block processing from 100→2, plus the byte accounting being a payload estimate rather than true RSS. No security issues stood out. Safe to merge after consciously accepting the parallelism default and ideally tightening defaults/docs around the byte cap.
</details>

---

spreston8:
Reviewed at `1aeb7903` against `dev`, reading the live files. No correctness defect found; the new tests pass here (`casper replay_cache` 12 passed incl. the proptest, `node parallel_block_limit` 2 passed). Requested changes below.

Worth noting up front that the two mechanisms are independent — the 32 MiB replay-cache cap and the parallelism change — and the risky half is a one-line default, since the semaphore already existed. Both asks are about the parallelism side.

**Medium**

1. **What does 100 → 2 cost in throughput?** This is the substantive change and the description doesn't cover it. It may well be free — a single block's replay already fans out across cores through the reducer's par branches, so 2 concurrent blocks can still saturate CPU — but the argument isn't made, and the failure mode matters: a node behind on blocks now drains a 2,048-deep queue two at a time, so catch-up and LFS sync are where a regression would show. You have soak evidence already (the 1.5–4.5 hour runs); block-processing throughput and catch-up at 2 vs 100 would settle it.

2. **Add a queue-depth gauge — the memory moves into the inbound queue, which isn't instrumented.** `block_processor_queue` is bounded at 2,048 `BlockMessage`s (`setup.rs:37`), and with 50x less drain concurrency it will sit far deeper under load. That's very likely still a large net win, since a queued `BlockMessage` is much cheaper than a live replay runtime plus hot store — but this PR instruments what got smaller (`block-processing.active`, `replay-cache.retained-bytes`) and not what got bigger. As it stands a soak can show `active = 2` and flat cache bytes while RSS still climbs through queue depth, which would read as "fixed" when it's "moved". `PROPOSER_QUEUE_PENDING_METRIC` is the precedent to copy.

**Low**

3. `retained_bytes` counts the event log and post-state but not the cache key (`parent_state` + `sender_pk` + `payload_hash`, ~100 bytes per entry) or `IndexMap`'s per-slot overhead. Immaterial for KB-scale event logs, but it makes 32 MiB a floor on retention rather than a ceiling — worth knowing when reading the new gauge against RSS.

4. `MAX_REPLAY_CACHE_BYTES` is hardcoded while `max_parallel_blocks` got an env override. It's exactly the value you'd want to sweep across soak runs — could it take the same treatment?

5. `F1R3_MAX_PARALLEL_BLOCKS` is env-only and undocumented. Consistent with `F1R3_MAX_BLOCKS_IN_PROCESSING` in the same file, so no objection to the mechanism, but a knob with this much leverage on memory and throughput is worth a line in the config docs.

6. The soak-harness rewrite (~820 of the 1303 added lines, `run-merge-recovery-soak.sh` alone at 943 changed) is bundled with a 5-file node fix. Not worth unpicking now, but splitting them would make both easier to review and let either be reverted independently.

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a well-engineered memory-pressure fix: the replay cache gains byte-cap accounting with a strong property-based invariant suite, block-processing parallelism becomes bounded and configurable with safe clamping, and both sides get observability gauges wired into the soak driver. The main substantive concerns are (1) the byte accounting charges `Bytes::len()` and thus can undercount retention when slices pin large shared allocations — the exact hazard the code's own comments acknowledge for keys — meaning the 32MB cap is a soft bound; (2) replay-cache gauges are only refreshed on `put`, so they go stale after `clear()`; and (3) the 100→2 parallelism default is a 50x behavioral change merged before throughput evidence exists, with real risk to catch-up sync (acknowledged as a follow-up but worth a less aggressive default or pre-merge measurement). Minor items include the `local -A` bash-4 requirement in the benchmark script, silent normalization of `F1R3_MAX_PARALLEL_BLOCKS=0`, and lock-scope/duplication nits. Note the diff was truncated (15 files, partial coverage), so the soak driver script changes themselves were not fully reviewed. Nothing here is a security issue or merge blocker; feedback-level fixes recommended.

The bounded block-processing change and LRU admission logic appear coherent, but the replay cache's central byte-bound guarantee is not reliable because `Bytes::len()` does not measure the backing allocation retained by a shared slice. That should be resolved or explicitly bounded through compact copies before relying on the 32MB cap as an RSS safeguard. The new observability also needs initialization and clear-path updates to avoid stale data, while the benchmark poller would benefit from bounded concurrency.

The code changes look generally good with some minor improvements possible for performance and style consistency. The new features and fixes seem well-implemented, and the additional tests and metrics provide good coverage. A few minor adjustments and additional comments could further enhance the code quality.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Byte accounting can substantially underestimate memory retained by Bytes</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 62)

The accounting uses `prost::bytes::Bytes::len()` for keys, hashes, output values, and post-state data. A `Bytes` value may be a small slice backed by a much larger shared allocation, so its length is not necessarily the amount of memory kept alive by the cache. Prost decoding from a shared input buffer can produce exactly this ownership pattern. Consequently, a cache reporting at most 32MB may retain substantially more backing memory, undermining the RSS fix's core byte-cap guarantee. Preserving the stored key during `get` does not solve this because two equal keys with identical lengths can retain differently sized backing allocations.

**Suggestion:**
```
Normalize cached byte fields into compact owned allocations at admission, for example with `Bytes::copy_from_slice`, and account those compact allocations. Alternatively, serialize/cache a compact representation whose allocation size is known. Add a test using a small `Bytes::slice` of a large backing buffer to ensure admission cannot retain unaccounted large buffers.
```

</details>

<details>
<summary><b>:orange_circle: Major: Byte accounting undercounts `Bytes` shared-allocation retention</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 95)

`event_heap_bytes` and `produce_heap_bytes` charge `prost::bytes::Bytes::len()` per buffer, but a `Bytes` value is a refcounted slice: a short slice can keep an arbitrarily large parent allocation alive. The `get` implementation's own comment acknowledges this hazard ('could swap a compact key for one backed by a larger shared allocation'), yet the cap accounting itself has the same blind spot. In the worst case, entries charged well under 32MB collectively pin far more real memory, weakening the RSS guarantee this PR is meant to provide.

**Suggestion:**
```
Where the event logs originate from decoded protobuf messages backed by a large shared buffer, consider copying into freshly-allocated (compact) buffers before insertion, or document that inputs must be compact. Alternatively, track and charge unique underlying allocations where feasible.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Replay-cache gauges can remain absent or stale</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/runtime_manager.rs` (line 228)

Replay-cache metrics are only recorded after a `put` attempt. They are not initialized when the cache is created and are not updated when `ReplayCache::clear` empties the cache. A node with no admissions may expose no replay-cache series, while a cleared cache may continue reporting its previous entry and byte counts. This weakens the intended soak diagnostics.

**Suggestion:**
```
Initialize both gauges to zero when constructing the runtime/cache and update them after every state-changing operation, including `clear`. A cache-level metrics callback or wrapper would make it harder for callers to miss updates.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Potential performance improvement</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 100)

The `charged_bytes` function calculates the size of the key and entry each time it is called. This could be optimized by caching the result if the key and entry do not change frequently.

**Suggestion:**
```
Consider caching the result of `charged_bytes` if the key and entry are stable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Cheap rejection checks performed under the lock</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 176)

In `put`, `charged` is computed before the lock (good), but the `max_entries == 0 || charged > max_bytes` rejection check happens after acquiring the mutex. Both operands are lock-independent, so rejected puts take the lock unnecessarily, adding contention on the hot replay path.

**Suggestion:**
```
Move the `if self.max_entries == 0 || charged > self.max_bytes { return false; }` check above the `self.state.lock()` call.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Replay-cache gauges go stale after `clear()` and on cache misses</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 224)

`record_replay_cache_metrics` is only invoked on the `put` path. If the cache is cleared (e.g., `RuntimeManager` calls `clear()` elsewhere) or entries are only being evicted via reads/replacements outside the instrumented sites, the `replay-cache.entries` / `replay-cache.retained-bytes` gauges will report the last put-time snapshot indefinitely. Since the whole point is diagnosing a future breach, a stale gauge could name the wrong suspect.

**Suggestion:**
```
Update the gauges inside `clear()` (or in the cache implementation itself), or sample them on the same periodic loop used for the block-processing queue gauge so they reflect current state regardless of activity.
```

</details>

<details>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 50)

The function `max_blocks_in_processing` uses snake_case, while `configured_max_parallel_blocks` and `max_parallel_blocks` use camelCase. Consistency in naming conventions should be maintained.

**Suggestion:**
```
Rename `max_blocks_in_processing` to `configured_max_blocks_in_processing` to maintain consistency.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: F1R3_MAX_PARALLEL_BLOCKS=0 silently falls back to default 2</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 59)

`configured_max_parallel_blocks` filters out `0` and falls back to the default of 2, the same treatment as an unparseable value. An operator setting `0` (intending to pause processing or expecting an error) gets silent behavior divergence with no warning log distinguishing 'invalid' from 'valid but out of range'.

**Suggestion:**
```
Emit a `tracing::warn!` when the env var is present but rejected (non-numeric, zero, or clamped), so misconfiguration is visible in logs rather than silently normalized.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: `block-processing.active` may count permit-waiting tasks depending on acquire placement</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/block_processor_instance.rs` (line 222)

`ActiveBlockProcessingGuard::new()` increments the gauge at the top of the spawned task. If the semaphore permit is acquired inside the task (rather than in the recv loop before spawn), the gauge counts tasks queued waiting on a permit, not tasks actively processing — which would make `active` exceed `parallel-limit` and confuse the exact analysis this PR wants to enable. The acquire site isn't visible in the diff; please verify the guard is created only after the permit is held.

**Suggestion:**
```
Ensure the guard is constructed after `semaphore.acquire_owned().await` succeeds, or add a debug assertion / test that `active <= parallel_limit`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Default parallelism drop from 100 to 2 has no throughput evidence yet</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/runtime/node_runtime.rs` (line 639)

The default drops 50x with verification deferred to 'the next dev daily soak'. Catch-up sync of a node that is far behind is the highest-risk scenario: with only 2 concurrent blocks, drain rate may fall below block-production rate on a busy shard, causing the inbound queue (now bounded at 512 via F1R3_MAX_BLOCKS_IN_PROCESSING) to saturate and blocks to be dropped/re-requested. The new queue gauge helps observe this, but merging a large behavioral change before measuring is a risk the PR itself acknowledges.

**Suggestion:**
```
Consider a less aggressive default (e.g., 4–8) or gating the change behind the env var for one soak cycle, then locking in the default with measured evidence. At minimum, verify catch-up sync of a fresh node against an existing shard before releasing.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: `local -A` requires bash >= 4; breaks on macOS default bash</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/latency-benchmark.sh` (line 216)

`harvest_finalizations` uses `local -A finalized_sigs=()`, an associative array requiring bash 4+. macOS ships bash 3.2 by default; the script uses `#!/usr/bin/env bash` so a developer running it locally on macOS without a newer bash will get a hard failure ('declare: -A: invalid option') only at runtime, after the flood has started.

**Suggestion:**
```
Either add an early bash-version check (`(( BASH_VERSINFO[0] >= 4 )) || die ...`) or track finalized sigs in a temp file/grep -F set to stay bash-3 compatible.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Finalization polling is serial and can stall under API degradation</b> (Reported by: openai)</summary>

**File:** `scripts/bench/latency-benchmark.sh` (line 225)

Each polling cycle issues one synchronous curl request per pending deploy, with a five-second timeout for each request. With many outstanding deploys, an unhealthy API can make one cycle take `pending_count * 5s`, preventing timely polling and making the nominal 90-second finalization wait ineffective because the background worker may still be stuck in a long serial pass.

**Suggestion:**
```
Use bounded parallel polling, batch status queries if the API supports them, or impose a deadline for an entire polling pass rather than only per-request timeouts.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Trait `put` return-value contract undocumented</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/replay_cache.rs` (line 130)

The `ReplayCache::put` signature changed to return `bool` (admission result), which is a semantic change for all implementors and callers, but the trait carries no doc comment explaining what `true`/`false` means or when an entry may be rejected (oversized, zero-capacity cache).

**Suggestion:**
```
Add a doc comment on the trait method documenting the admission contract: returns `false` when the entry cannot be admitted (charged bytes exceed max_bytes, or max_entries == 0), `true` otherwise.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Duplicated replay-cache admission blocks</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 378)

The two cache-admission + metrics + debug-log blocks (around lines 378 and 488) are byte-for-byte identical. The change amplified pre-existing duplication; a helper taking (cache, key, entry, seq_num) would keep the admission/logging/metrics contract in one place.

**Suggestion:**
```
Extract a private helper, e.g. `fn store_replay_entry(cache: &InMemoryReplayCache, key: ReplayCacheKey, entry: ReplayCacheEntry, seq_num: i64)`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add comments for clarity</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/test-latency-benchmark.sh` (line 70)

The script could benefit from additional comments explaining the purpose of each section and the expected behavior of the mock functions.

**Suggestion:**
```
Add comments to describe the purpose and expected behavior of each section and mock function.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Brittle grep pattern in soak driver test assertion</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/test-run-merge-recovery-soak.sh` (line 76)

`grep -q 'test.validator1.*12.*1048576.*2.*2.*7'` uses unescaped dots and greedy `.*` between all values; it can match unintended lines (e.g., any line containing those digit substrings in order) and will silently keep passing if column order changes. This weakens the fail-closed guarantee the test is meant to pin.

**Suggestion:**
```
Match with anchored field separators, e.g. a tab-delimited awk check on exact columns, or escape the pattern (`test\.validator1	...`).
```

</details>
---

<details>

This is a well-engineered memory-pressure fix: the replay cache gains byte-cap accounting with a strong property-based invariant suite, block-processing parallelism becomes bounded and configurable with safe clamping, and both sides get observability gauges wired into the soak driver. The main substantive concerns are (1) the byte accounting charges `Bytes::len()` and thus can undercount retention when slices pin large shared allocations — the exact hazard the code's own comments acknowledge for keys — meaning the 32MB cap is a soft bound; (2) replay-cache gauges are only refreshed on `put`, so they go stale after `clear()`; and (3) the 100→2 parallelism default is a 50x behavioral change merged before throughput evidence exists, with real risk to catch-up sync (acknowledged as a follow-up but worth a less aggressive default or pre-merge measurement). Minor items include the `local -A` bash-4 requirement in the benchmark script, silent normalization of `F1R3_MAX_PARALLEL_BLOCKS=0`, and lock-scope/duplication nits. Note the diff was truncated (15 files, partial coverage), so the soak driver script changes themselves were not fully reviewed. Nothing here is a security issue or merge blocker; feedback-level fixes recommended.

The code changes look generally good with some minor improvements possible for performance and style consistency. The new features and fixes seem well-implemented, and the additional tests and metrics provide good coverage. A few minor adjustments and additional comments could further enhance the code quality.

#### openai (gpt-5.6-sol)

The bounded block-processing change and LRU admission logic appear coherent, but the replay cache's central byte-bound guarantee is not reliable because `Bytes::len()` does not measure the backing allocation retained by a shared slice. That should be resolved or explicitly bounded through compact copies before relying on the 32MB cap as an RSS safeguard. The new observability also needs initialization and clear-path updates to avoid stale data, while the benchmark poller would benefit from bounded concurrency.

#### xai (grok-4.5)

</details>

---

jeffrey-l-turner:
Resolutions for the review rounds above (@spreston8's numbered points and the multi-agent findings), now on the branch:

**1. 100 → 2 throughput cost** — deliberate product default, documented in the work log; the empirical 2-vs-100 comparison (block-processing rate, catch-up drain of the 2,048-deep queue, finalization latency) comes from the next dev daily soak after merge, using the instrumentation added below. The PR description's Follow-ups section tracks it.

**2. Queue-depth gauge** — done in `39c3200f`. `block-processing.queue.pending`, sampled every 5s from the channel's own permit accounting (`max_capacity − capacity`) via a `WeakSender` so the sampler can't hold the queue open past the last real producer; initialized to 0 per the `PROPOSER_QUEUE_PENDING_METRIC` precedent. The soak driver scrapes it per node into `node-memory-timeseries.tsv` alongside active/limit and the replay-cache gauges, so "moved, not fixed" would be visible directly.

**3. Key bytes / accounting floor-vs-ceiling** — done in two steps. `c3a6e931` charges key heap bytes (`parent_state` + `sender_pk` + `payload_hash`) alongside entry bytes, with length-based accounting so equal keys always charge equally, and `get` now preserves the stored key via `move_index`. `f6fccb31` goes further: every `prost::bytes::Bytes` is deep-copied into a private, exactly-`len` buffer at cache insert (entries and keys), so a small cached slice can no longer pin a large decode buffer — the accounting is exact per buffer and 32 MiB is a true ceiling on buffer retention. Still uncounted: `IndexMap` per-slot overhead, bounded by the 192-entry cap (~tens of KB worst case). Pointer-inequality tests pin the copies; the proptest invariants cover the charged-bytes contract.

**4. `MAX_REPLAY_CACHE_BYTES` env override** and **5. config-docs line for `F1R3_MAX_PARALLEL_BLOCKS`** — open; happy to add both in a small follow-up commit on this branch if wanted (the byte cap is indeed the value worth sweeping across soak runs).

**6. Bundling** — acknowledged; keeping as-is per your note, and the soak-harness half is independently revertable at the script level.

Multi-agent round 2 confirms the trajectory: critical count 1 → 0, majors 5 → 2, and both remaining majors were the shared-backing-allocation finding that `f6fccb31` closes. Full mapping lives in the PR description's "Review resolutions" table.

PR #187, opened 2026-08-01
feat(ci): dispatch merge recovery soaks from OCI

jeffrey-l-turner:
- replace delayed GitHub cron events with DST-aware OCI scheduling
- add a Bash OCI Function and idempotent deployment automation
- route daily and weekend series while preserving restart behavior

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/187"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788178633&installation_model_id=426883&pr_number=187&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F187&signature=1a2ed4750083c46e03c3d3487339bc342858dea16f15d243718dcc41a3ffbff7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #196, opened 2026-08-04
fix(ci): make soak cron a self-suppressing fallback for the OCI sche…

jeffrey-l-turner:
Completes the OCI soak-scheduler cutover (PR #187's design, never deployed) and makes the GitHub cron a permanent, self-suppressing fallback instead of removing it.

## Why

The scheduler infrastructure from #187 was never stood up — zero Resource Scheduler schedules, no Function, zero `[scheduled:…]` runs ever — so soaks have been triggered by GitHub cron, delivered 2.5–3.5 hours late every night. Worse, the original cutover plan (remove the crons) hid a live-fire hazard: during any overlap the soak concurrency group (`cancel-in-progress: true`) means the hours-late cron would cancel the OCI-dispatched soak mid-run, every night.

## Changes

- **Cron slot dedup**: an eligible cron delivery yields when a live-or-successful dispatch titled `Merge Recovery Soak [scheduled:<slot>]` already claimed the slot. Dead claims (gate failures, cancellations) do not suppress the fallback; a failed claim query fails open with a visible `::warning::`. The crons stay enabled permanently — if the Function ever dies silently, the soak still happens, hours late instead of not at all.
- **Routing tests**: eight scenarios including live/dead/done claims, fail-open on query failure, and a fake `gh` that pins the exact query shape.
- **Scheduler image build fix**: `dnf` (Oracle Linux 8.10 base), not `microdnf`.
- **README**: cutover step 5 rewritten to the keep-cron design.

## Deployment status (already live)

Function `f1r3node-soak-scheduler`, both UTC schedules (02:30/03:30), Vault secret (GitHub App PEM), dynamic groups and IAM policies deployed to the `ci-runner` compartment; GitHub App validated (actions:write, metadata:read, this repo only); handler unit tests pass and the late-invocation guard verified live. First automatic dispatch: **02:30 UTC tomorrow** — this PR's dedup must be on `master` (scheduled runs execute master's workflow file) before then, or the late cron will cancel the Function's soak.

## Review resolutions

All multi-agent review findings are addressed — see the resolution table in the PR comments (critical claim-status filter in `47581f2b`, observability and query-shape hardening in `8c3c5c8a`).

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/196"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788405917&installation_model_id=426883&pr_number=196&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F196&signature=b97a19e680af171eeec0256e7947273547000f76f81f869e2452cbd98f727fbc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/scripts/test-soak-schedule-routing.sh
+++ b/.github/scripts/test-soak-schedule-routing.sh
@@ -36,7 +36,14 @@ cat >"$TMP/bin/gh" <<'SH'
 #!/usr/bin/env bash
 case "$*" in
   *"/commits?"*) printf '1\n' ;;
-  *) printf '%s\n' '{"workflow_runs":[]}' ;;
+  *)
+    if [ -n "${FAKE_CLAIMED_SLOT:-}" ]; then
+      printf '{"workflow_runs":[{"id":123,"display_title":"Merge Recovery Soak [scheduled:%s]"}]}\n' \
+        "$FAKE_CLAIMED_SLOT"
+    else
+      printf '%s\n' '{"workflow_runs":[]}'
+    fi
+    ;;
 esac
 SH
 chmod +x "$TMP/bin/date" "$TMP/bin/gh"
@@ -48,6 +55,7 @@ run_gate() {
 	PATH="$TMP/bin:$PATH" \
 		FAKE_NOW_EPOCH="$((slot + 60))" \
 		FAKE_SLOT_EPOCH="$slot" \
+		FAKE_CLAIMED_SLOT="${FAKE_CLAIMED_SLOT:-}" \
 		FAKE_PACIFIC_WEEKDAY="$weekday" \
 		EVENT_NAME="$([ "$mode" = oci ] && printf workflow_dispatch || printf schedule)" \
 		EVENT_SCHEDULE="30 2 * * *" \
@@ -81,6 +89,14 @@ for mode in oci cron; do
 	grep -qx 'kind=daily' "$monday"
 	grep -qx 'run_benchmarks=false' "$monday"
 done
+
+# A cron delivery whose slot was already claimed by an OCI-dispatched run
+# must yield: the soak concurrency group cancels in-progress, so a late cron
+# would otherwise kill a soak hours into its run.
+claimed="$(FAKE_CLAIMED_SLOT=1785810600 run_gate cron 1 1785810600)"
+grep -qx 'should_run=false' "$claimed"
+# The unclaimed cron slot from the loop above must still have run.
+grep -qx 'should_run=true' "$TMP/output-oci-1"
 
 if PATH="$TMP/bin:$PATH" \
 	FAKE_NOW_EPOCH=1785551460 \
--- a/.github/scripts/test-soak-schedule-routing.sh
+++ b/.github/scripts/test-soak-schedule-routing.sh
@@ -36,13 +36,20 @@ cat >"$TMP/bin/gh" <<'SH'
 #!/usr/bin/env bash
 case "$*" in
   *"/commits?"*) printf '1\n' ;;
-  *)
+  *"runs?event=workflow_dispatch"*)
+    if [ -n "${FAKE_CLAIM_QUERY_FAIL:-}" ]; then
+      exit 1
+    fi
     if [ -n "${FAKE_CLAIMED_SLOT:-}" ]; then
       printf '{"workflow_runs":[{"id":123,"display_title":"Merge Recovery Soak [scheduled:%s]","status":"%s","conclusion":%s}]}\n' \
         "$FAKE_CLAIMED_SLOT" "${FAKE_CLAIMED_STATUS:-in_progress}" "${FAKE_CLAIMED_CONCLUSION:-null}"
     else
       printf '%s\n' '{"workflow_runs":[]}'
     fi
+    ;;
+  *)
+    printf 'unexpected gh invocation: %s\n' "$*" >&2
+    exit 1
     ;;
 esac
 SH
@@ -58,6 +65,7 @@ run_gate() {
 		FAKE_CLAIMED_SLOT="${FAKE_CLAIMED_SLOT:-}" \
 		FAKE_CLAIMED_STATUS="${FAKE_CLAIMED_STATUS:-}" \
 		FAKE_CLAIMED_CONCLUSION="${FAKE_CLAIMED_CONCLUSION:-}" \
+		FAKE_CLAIM_QUERY_FAIL="${FAKE_CLAIM_QUERY_FAIL:-}" \
 		FAKE_PACIFIC_WEEKDAY="$weekday" \
 		EVENT_NAME="$([ "$mode" = oci ] && printf workflow_dispatch || printf schedule)" \
 		EVENT_SCHEDULE="30 2 * * *" \
@@ -108,6 +116,9 @@ done_claim="$(FAKE_CLAIMED_SLOT=1785810600 FAKE_CLAIMED_STATUS=completed \
 grep -qx 'should_run=false' "$done_claim"
 # The unclaimed cron delivery from the loop above must still have run.
 grep -qx 'should_run=true' "$TMP/output-cron-1"
+# A failed claim query must fail OPEN: soak anyway rather than silently skip.
+open_fail="$(FAKE_CLAIM_QUERY_FAIL=1 run_gate cron 1 1785810600 query-fail)"
+grep -qx 'should_run=true' "$open_fail"
 
 if PATH="$TMP/bin:$PATH" \
 	FAKE_NOW_EPOCH=1785551460 \
--- a/.github/scripts/test-soak-schedule-routing.sh
+++ b/.github/scripts/test-soak-schedule-routing.sh
@@ -38,8 +38,8 @@ case "$*" in
   *"/commits?"*) printf '1\n' ;;
   *)
     if [ -n "${FAKE_CLAIMED_SLOT:-}" ]; then
-      printf '{"workflow_runs":[{"id":123,"display_title":"Merge Recovery Soak [scheduled:%s]"}]}\n' \
-        "$FAKE_CLAIMED_SLOT"
+      printf '{"workflow_runs":[{"id":123,"display_title":"Merge Recovery Soak [scheduled:%s]","status":"%s","conclusion":%s}]}\n' \
+        "$FAKE_CLAIMED_SLOT" "${FAKE_CLAIMED_STATUS:-in_progress}" "${FAKE_CLAIMED_CONCLUSION:-null}"
     else
       printf '%s\n' '{"workflow_runs":[]}'
     fi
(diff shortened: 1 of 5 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.89 |

---

### Summary

This is a well-reasoned CI safety change: the cron schedule is retained as a self-suppressing fallback for the OCI Function scheduler, correctly failing open so an API error can never silently skip a soak, and the routing test covers both the claimed-slot suppression and unclaimed pass-through paths. The core logic is sound — the event=workflow_dispatch filter prevents the cron from matching its own schedule-event run, and the slot-epoch-embedded title makes stale runs from prior days unmatchable. Remaining concerns are minor: the claim check ignores run status (a dead-on-arrival OCI dispatch still suppresses the fallback), API failures are silenced so fail-open events leave no log trace, the suppression trusts a title string that any dispatcher can produce, and the test stub doesn't pin the event filter in the query. None of these block merge, but the status-blind claim check and silent fail-open are worth addressing to fully honor the 'a silently skipped soak is not recoverable' design goal.

The self-suppressing cron fallback addresses the concurrency cancellation problem and correctly fails open on API or parsing failures. However, matching solely by display title means even a failed or cancelled OCI-dispatched workflow permanently claims the slot and prevents the fallback from recovering it. The added test also contains a cron-versus-OCI output-file typo.

Solid, well-motivated fix: keeping GitHub crons as a permanent late fallback and suppressing them when an OCI dispatch already claimed the slot directly addresses the concurrency-group cancel hazard, and fail-open matches the existing idle-check philosophy. The main gap is that any matching workflow_dispatch title counts as a claim — including failed or cancelled OCI runs — so the fallback can still skip a night when the Function dispatched but the soak did not succeed. Tighten the status/conclusion filter, fix the test assertion to actually cover unclaimed cron pass-through, and consider bounding or observing claim-query failures.

The code changes look good and address the issue of having a fallback mechanism for the soak tests. The logic for checking if a slot is already claimed and suppressing the cron job if necessary is well-implemented. Minor documentation improvements could enhance clarity.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Failed or cancelled dispatches suppress the fallback soak</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 405)

The lookup treats every workflow-dispatch run with the expected title as a successful claim, regardless of its status or conclusion. If the OCI scheduler dispatches the workflow but that run fails or is cancelled before performing the soak, the later cron will still suppress itself. This leaves a gap in the fallback behavior and can silently skip the night's soak.

---

The runs listing uses event=workflow_dispatch&per_page=100 with client-side title filter and no created/branch window. For a nightly soak this is likely fine, but a busy workflow_dispatch history could push tonight's run off the first page and cause a false miss (cron runs despite an active OCI soak) — reintroducing the cancel-in-progress failure mode this PR aims to prevent.

---

The slot-claim check treats any workflow_dispatch run with the expected display_title as a claim, regardless of status/conclusion. That correctly protects in-progress OCI soaks from being cancelled by a late cron, and correctly skips after a successful run. But if the OCI Function dispatches and the run fails or is cancelled early (bad ref, infra blip, manual cancel), the cron fallback will still suppress itself and the night produces no soak — weakening the 'Function dies silently' reliability goal. A duplicate start is described as recoverable; a failed claim should probably not count as claimed.

**Suggestion:**
```
If the API usage allows, narrow the query (e.g. created date from slot time, or branch) or paginate until a match or a run older than the slot is seen. At minimum, document the per_page assumption next to the fail-open comment.
```

</details>

<details>
<summary><b>:orange_circle: Major: Unclaimed cron assertion checks the OCI output file</b> (Reported by: openai, xai)</summary>

**File:** `.github/scripts/test-soak-schedule-routing.sh` (line 99)

Comment says "The unclaimed cron slot from the loop above must still have run" but the assertion checks $TMP/output-oci-1. That does not guard the cron pass-through path the new suppression logic could regress. If the loop already asserts cron should_run=true, this line is misleading; if not, the unclaimed-cron case is under-tested.

---

The comment says this assertion verifies that the unclaimed cron slot passed through, but it reads "$TMP/output-oci-1". Consequently, this specific assertion does not validate the cron path and could miss a mode-specific regression.

**Suggestion:**
```
Assert against the cron loop output (e.g. $TMP/output-cron-1, or whatever path run_gate writes for mode=cron) and align the comment. Optionally also keep an explicit unclaimed cron run_gate call next to the claimed case for symmetry.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Claimed-slot check ignores run status/conclusion</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 403)

Both stderr (2>/dev/null) and the pipeline failure (|| true) are swallowed. The fail-open behavior is correct by design, but when the gh API call errors there is no log line at all — the run proceeds and an operator cannot tell whether the dedup check actually ran and found no claim, or whether it silently failed. This makes it hard to notice a broken token/permission that would have masked a genuine duplicate.

---

The gh api query matches any workflow_dispatch run whose display_title equals the slot title, regardless of its status. If the OCI-dispatched run was cancelled seconds after starting, or failed immediately at setup (before doing any actual soaking), the late cron will still yield and no soak happens that night. This partially defeats the stated goal that 'a silently skipped soak is not recoverable' — the fallback only covers the Function never dispatching, not the dispatched run dying at birth.

**Suggestion:**
```
Capture the gh exit status separately and emit an explicit log line like 'slot-claim query failed; failing open and soaking anyway' so fail-open events are observable in the run log.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Slot suppression is triggered by display_title string match alone</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 405)

Any actor with workflow_dispatch permission (or a run whose run-name resolves to 'Merge Recovery Soak [scheduled:<epoch>]') can suppress that night's cron soak by creating a matching-title run that does nothing. Exposure is limited to repo collaborators, but it means the fallback can be defeated by a stray/mistaken manual dispatch that passes scheduled_slot_epoch, even if the resulting run soaks nothing.

**Suggestion:**
```
If feasible, additionally check the triggering actor (e.g. the OCI Function's machine identity) or an input marker, rather than title alone; otherwise document this as an accepted trust assumption.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Clarify documentation</b> (Reported by: bedrock)</summary>

**File:** `oci/soak-scheduler/README.md` (line 51)

The documentation update could be more concise and clear. Consider rephrasing for better readability.

**Suggestion:**
```
Rephrase the documentation to clearly state the purpose and behavior of the cron fallback.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Fake gh does not discriminate on the event=workflow_dispatch query parameter</b> (Reported by: anthropic)</summary>

**File:** `.github/scripts/test-soak-schedule-routing.sh` (line 39)

The stub returns the claimed run for every non-commits endpoint, so the test would still pass if the production query dropped the event=workflow_dispatch filter. That filter matters: without it, other run listings (or a future refactor that queries all events) could self-match, and the test would not catch the regression.

**Suggestion:**
```
Make the fake gh match on the actual runs URL including 'event=workflow_dispatch' (e.g. a case pattern like *'runs?event=workflow_dispatch'*) so the test pins the query shape.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: per_page=100 without pagination relies on default sort order</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 404)

The query fetches only the first 100 workflow_dispatch runs. This is fine today because GitHub returns runs newest-first and the claiming dispatch is at most hours old, but the assumption is implicit. A busy repo dispatching this workflow frequently could theoretically push the claim past page 1.

**Suggestion:**
```
Add a brief comment noting the reliance on newest-first ordering, or add a created=>=<date> filter to bound the query to the current day.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: API failures are fully silenced</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 404)

2>/dev/null combined with || true implements the intended fail-open behavior, but operators get no signal when the claim check is degraded and every cron night "soaks anyway". That can mask token/permission/API outages until duplicates or concurrency cancels show up.

**Suggestion:**
```
On empty/failed query paths, emit a single notice (echo or ::warning::) that claim detection failed open, without changing should_run.
```

</details>
---

<details>

This is a well-reasoned CI safety change: the cron schedule is retained as a self-suppressing fallback for the OCI Function scheduler, correctly failing open so an API error can never silently skip a soak, and the routing test covers both the claimed-slot suppression and unclaimed pass-through paths. The core logic is sound — the event=workflow_dispatch filter prevents the cron from matching its own schedule-event run, and the slot-epoch-embedded title makes stale runs from prior days unmatchable. Remaining concerns are minor: the claim check ignores run status (a dead-on-arrival OCI dispatch still suppresses the fallback), API failures are silenced so fail-open events leave no log trace, the suppression trusts a title string that any dispatcher can produce, and the test stub doesn't pin the event filter in the query. None of these block merge, but the status-blind claim check and silent fail-open are worth addressing to fully honor the 'a silently skipped soak is not recoverable' design goal.

The code changes look good and address the issue of having a fallback mechanism for the soak tests. The logic for checking if a slot is already claimed and suppressing the cron job if necessary is well-implemented. Minor documentation improvements could enhance clarity.

#### openai (gpt-5.6-sol)

The self-suppressing cron fallback addresses the concurrency cancellation problem and correctly fails open on API or parsing failures. However, matching solely by display title means even a failed or cancelled OCI-dispatched workflow permanently claims the slot and prevents the fallback from recovering it. The added test also contains a cron-versus-OCI output-file typo.

#### xai (grok-4.5)

Solid, well-motivated fix: keeping GitHub crons as a permanent late fallback and suppressing them when an OCI dispatch already claimed the slot directly addresses the concurrency-group cancel hazard, and fail-open matches the existing idle-check philosophy. The main gap is that any matching workflow_dispatch title counts as a claim — including failed or cancelled OCI runs — so the fallback can still skip a night when the Function dispatched but the soak did not succeed. Tighten the status/conclusion filter, fix the test assertion to actually cover unclaimed cron pass-through, and consider bounding or observing claim-query failures.
</details>

---

jltatbeach (approved):

jeffrey-l-turner:
Resolutions for the multi-agent review findings, all landed on this branch:

| Finding | Resolution |
|---------|------------|
| **Critical** — failed or cancelled dispatches still claimed the slot, so the cron would yield to a dead claim and silently skip the night | `47581f2b` — the claim filter now requires the dispatch be live (`status != completed`) or have concluded `success`; dead claims no longer suppress the fallback. Test cases: live-claim yields, dead-claim soaks, done-claim yields. |
| **Major** — the "unclaimed cron still ran" assertion checked the OCI-mode output file (and the claimed-run invocation overwrote the file it should have checked) | `47581f2b` — `run_gate` takes an output tag so every scenario writes a distinct file; the unclaimed assertion now reads the cron output it was meant to. |
| **Minor** — fail-open was silent: a broken token looked identical to "no claim found" (also raised as a suggestion by a second reviewer) | `8c3c5c8a` — a failed query emits `::warning::slot-claim query failed; failing open and soaking anyway`; an unparseable response warns separately. Degraded claim detection is now visible in the run log. |
| **Minor** — title-based claims can be held by any collaborator's dispatch | `8c3c5c8a` — documented as the accepted trust assumption: such a dispatch passes through this same gate and either soaks or, on failure, stops claiming under the status filter. |
| **Suggestion** — fake `gh` didn't pin the `event=workflow_dispatch` query shape | `8c3c5c8a` — the stub answers only the exact runs query and errors on anything else; dropping the event filter now fails the test. New fail-open case asserts a failed query still soaks. |
| **Suggestion** — `per_page=100` relies on newest-first ordering implicitly | `8c3c5c8a` — reliance documented in a comment (claiming dispatch is at most hours old on a workflow that runs a handful of times daily). |

Also on the branch since the review ran: `6720bfb1` fixes the scheduler image build (`dnf`, not `microdnf`, on the Oracle Linux 8.10 base — verified by the successful live deployment), and `eeabdc78` merges dev (resolving the `.gitignore` conflict from #193's merge).

Deployment status: the Function, both Resource Scheduler schedules (02:30/03:30 UTC), Vault secret, and IAM policies are live in the `ci-runner` compartment; the handler's late-invocation guard verified against both of tonight's already-passed slots. First automatic end-to-end dispatch: tomorrow 02:30 UTC — which is why the cron dedup here must be on `master` before then.

PR #189, opened 2026-08-01
feat(soak): randomized exercise soak on the verified harness (parked until the release candidate)

jeffrey-l-turner:
## Status: parked (2026-09-18)

This draft keeps the randomized soak intent and drops its original implementation. The randomized soak will be rebuilt on the formally verified soak harness and its profiles. Do not rebase or update this branch until the chain below completes.

## Dependency chain

1. Issue #24 is resolved, so the sustained phase of the soak holds. The nightly evidence of 2026-09-16 shows the residual sustained-phase failures continue after PR #382.
2. The pull request stack merges into `dev` in order: #430, #431, #432, #433, then #436, then #441.
3. A `weekend-60h` soak in exact-candidate mode passes and publishes the stability-soak gate document on the candidate prerelease. That is the release candidate.
4. This branch is then recreated from `dev` and rewritten.

## What this branch carries today, and what survives

- `.github/scripts/resolve-ci-pins.mjs`, its test, `.github/ci-pins.jsonc`, its schema, and `docs/ci-pins.md` are superseded. `dev` pins the system-integration harness through `scripts/repin-system-integration.sh` and `.github/oci-validation.env`, in bash. The repository ships Rust and bash only, so the JavaScript resolver is dropped.
- The ToDos and UserStories additions date from 2026-08-01 and are overtaken by EPIC-017 and EPIC-018.
- `docs/randomized-exercise-soak.md`, the value-stream design, survives and is rewritten for the verified harness.

## Shape of the rewrite

- In the verified harness a randomized soak is not a new profile. It is a seeded, coverage-constrained scheduler that composes the seven profile scenarios and takes each verdict from the existing classifier.
- Verification follows the profile pattern: a scheduler model with seed determinism and coverage bounds plus unsafe controls under `formal/tlaplus/casper_soak/`, a Rust module with tests in `scripts/casper-soak`, a check script, and a workflow.
- The claim audit fixes the registry at CLAIM-CASPER-SOAK-001 through 008 and refuses any other set. The scheduler either lives under the harness claim 001 as a scheduling mode, or the registry extends to 009 with an audit change. This decision is recorded when the rewrite starts.
- Only the authority-finality and publication profiles exist so far. The scheduler also depends on TASK-017-7 through TASK-017-11, which build the other five.

## Procedure when the chain completes

1. Recreate the branch from `dev` with only the rewritten value-stream document.
2. Add the scheduler model, controls, Rust module, tests, check script, and workflow.
[...]

PR #84, opened 2026-06-25
test: validate gated fork-PR CI [do-not-merge]

spreston8:
Throwaway PR to validate the gated fork-PR pipeline on `ci/fork-runner-fix` (startup fix + ephemeral-launch approval gate). Will be closed after the run.

Co-Authored-By: Claude <noreply@anthropic.com>

PR #83, opened 2026-06-22
ci: gated fork-PR ephemeral runners via pull_request_target

jeffrey-l-turner:
Adds a privileged `CI (fork PRs)` workflow (`ci-fork-pr.yml`) that runs the secret-bearing ephemeral-runner + integration matrix for fork PRs via `pull_request_target`, gated behind the `ephemeral-launch` environment (maintainer/owner approval releases the OCI secrets). Extracts the heavy pipeline into a reusable `_integration-pipeline.yml` and slims `ci.yml` accordingly. Cheap checks still run unprivileged on the normal `pull_request` path.

Fork `pull_request` runs get no secrets, so the prior in-line `launch_ephemeral_runners` job failed for fork PRs (e.g. #69). This routes the privileged path through `pull_request_target` (base-branch workflow, fork cannot alter it) so it can target a fork's head after approval.

PR #199, opened 2026-08-04
fix(soak): restore docker-iteration telemetry (dual harness roots)

jeffrey-l-turner:
Restores per-node soak telemetry on docker iterations: every docker iteration to date silently lost its monitor time-series, RSS attribution, finalization-latency samples, and the harness breach-marker channel.

## Root cause (cross-repo)

The soak driver searches `integration-tests/data/` for harness telemetry, but docker sessions write monitor artifacts and archived node logs under `integration-tests/log-archive/` — and until the paired system-integration fix, the docker provider's `monitor_output_dir` returned `None`, so the monitor CSVs and `host-protection-breach.txt` were never written at all on docker runs. Observable damage: `rss_peak_mb` and `finalization_latency` null on runs 30880995655 (the breached daily) and 30906818259 (the attribution canary); the 2026-08-04 observer-node diagnosis had to be made from the teardown peak/avg table alone.

## Changes

- **Driver** (`run-merge-recovery-soak.sh`): `HARNESS_TELEMETRY_DIRS` covers both roots across all six search sites — monitor snapshot loop, RSS fallback, finalization-latency and too-far-ahead greps, harness breach marker, and `SOAK_METRIC` collection. Failure-evidence copy preserves both trees. Roots are created up front so multi-root `find` never errors.
- **`collect-soak-metrics.sh`**: the log-root argument accepts a colon-separated list, skipping missing dirs; single-path callers unchanged.
- **Driver test**: plants the monitor CSV *only* under `log-archive/` and the metrics CSV *only* under `data/` — a driver searching a single root fails the test.

## Pairing and sequencing (updated 2026-08-07)

Pairs with system-integration `fix/docker-monitor-output-dir` (docker provider exposes its per-session archive dir as `monitor_output_dir`; reviewed and approved by the system-integration session — see that repo's `docs/discoveries/2026-08-04-docker-monitor-output-dir.md`).

Agreed merge order — supersedes the earlier "before tonight's daily" target:

1. **Now, before the weekend soak**: merge system-integration #83 (→ `main`) and #84 (→ `dev`), the shared `fix/docker-monitor-output-dir` head. This is safe ahead of the soak because every f1r3node-rust workflow resolves system-integration through the pinned `SYSTEM_INTEGRATION_REF` SHA (`66de4f95`) — nothing tracks a branch, so those merges cannot change soak behavior until the pin moves.
2. **Weekend soak runs on the current pin**, unchanged baseline. This PR and #189 are held until after it completes.
3. **After the soak**: bump `SYSTEM_INTEGRATION_REF` at the 3 legacy sites *on this branch* to the post-merge system-integration SHA (picking up the docker-monitor fix, the `352d479e` torn-read test fix, and the bridge-admin flake hardening if merged by then), update the branch against dev, then merge this PR. This satisfies the review's item 1: the bump lands before the merge, not after, so the telemetry restoration is real on the first post-merge soak.
4. **#189 lands after this PR** and re-expresses the 3 legacy pin sites in its `runnerRef`/`catalogRef` registry format — the explicit ordering the review's item 2 asked for.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/199"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788446550&installation_model_id=426883&pr_number=199&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F199&signature=8f1e1337ecdb636e86bad4ff03946f0e39f7b962125d4068bfc80702bcd0919a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :bulb: Feedback Provided (50.0000% agreement)

**Reviewed by:**

| openai | `gpt-5.6-sol` | Approve | 0.94 |

---

### Summary

This is a well-reasoned, well-documented fix to a real telemetry-loss bug: docker soak iterations searched only integration-tests/data/ while docker sessions write under log-archive/. The dual-root array is threaded correctly through all six search sites, the colon-list extension to collect-soak-metrics.sh is backward compatible for single-path callers, and the test is thoughtfully constructed to fail if either root is dropped (monitor CSV only in archive, metrics CSV only in data). Remaining concerns are moderate: possible double-counting if any log ever appears under both roots (worth an enforced or documented invariant), the breach-marker lookup uses find-order `head -1` instead of the newest-file pattern used everywhere else, the unconditional up-front mkdir can mask a misconfigured SYSTEM_INTEGRATION_DIR by creating an empty skeleton, and the colon delimiter cannot represent paths containing ':' (relevant for timestamped archive dirs). None of these block merge; the change is functional and safe, and the identified items are hardening follow-ups.

The change consistently searches both harness telemetry roots across monitor snapshots, RSS fallback, latency/error extraction, breach detection, metric collection, and failure-evidence preservation. The collector remains compatible with ordinary single-path callers, safely skips unavailable roots, and preserves degraded-input behavior. The updated driver test appropriately distributes artifacts across both roots, providing useful regression coverage. No actionable security, logic, performance, or style issues were identified.

Solid, well-motivated fix: centralizing dual harness roots in HARNESS_TELEMETRY_DIRS and threading them through all six search sites (plus collect-soak-metrics multi-root support and a split-placement driver test) correctly addresses the silent docker telemetry loss. No security concerns; array expansion and join/split handling are correct for the controlled paths. Main actionable nit is failure-evidence copy over empty pre-created roots potentially tripping GNU tar’s empty-archive refusal and logging a false 'incomplete' warning—filter empty finds before tar. Safe to merge after that small hardening (or consciously accepting the noise).

The code changes look good and address the identified issues with telemetry collection. The introduction of handling multiple log roots is well-implemented. Minor improvements could be made by adding comments for better readability.

---

### Issues Found

<details>
<summary><b>:yellow_circle: Minor: Code readability</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/collect-soak-metrics.sh` (line 28)

The script could benefit from adding comments to explain the purpose of each section, especially around the new logic for handling multiple log roots.

**Suggestion:**
```
Add comments to clarify the purpose and functionality of each section of the script.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: mkdir -p on telemetry roots runs unconditionally, before any SYSTEM_INTEGRATION_DIR sanity check</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 83)

The up-front `mkdir -p "${HARNESS_TELEMETRY_DIRS[@]}"` executes regardless of whether SYSTEM_INTEGRATION_DIR points at a real system-integration checkout. If SYSTEM_INTEGRATION_DIR is misconfigured (e.g., a typo'd path), this silently creates an integration-tests/data and integration-tests/log-archive skeleton in the wrong location, and every subsequent find succeeds against empty directories — the exact null-telemetry failure mode this PR fixes, but now harder to spot because the dirs exist. Also note mkdir failure is unguarded; under a read-only SYSTEM_INTEGRATION_DIR the behavior depends on whether the script runs with `set -e`.

**Suggestion:**
```
Guard the mkdir behind a check that "$SYSTEM_INTEGRATION_DIR/integration-tests" already exists (fail loudly otherwise), so a misconfigured path aborts rather than degrades silently.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Potential double-counting when logs exist under both telemetry roots</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 293)

iteration_finalization_latency, iteration_too_far_ahead_errors, and the SOAK_METRIC collection now grep across both data/ and log-archive/. If a session ever writes a node log under data/ during the run and then archives a copy into log-archive/ at teardown (which 'log-archive' naming suggests is plausible for subprocess sessions), the same 'Propose timing:' samples, 'too far ahead' occurrences, and SOAK_METRIC lines will be counted twice, skewing latency percentiles and error counts. The PR description asserts the roots are disjoint per provider, but nothing in the code enforces or documents that invariant defensively.

**Suggestion:**
```
Deduplicate matched files (e.g., resolve to canonical paths and `sort -u` the find output before xargs), or add a comment documenting the disjointness invariant and where it is enforced in system-integration so a future archiving change doesn't silently double-count.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Empty evidence root can trigger non-fatal tar failure noise</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 664)

HARNESS_TELEMETRY_DIRS are mkdir -p'd up front, so both data/ and log-archive/ always exist. On failure, the evidence loop iterates both roots. For the mode that did not write artifacts (subprocess → empty log-archive/, docker → possibly empty data/), find emits no files and GNU tar typically exits non-zero with 'Cowardly refusing to create an empty archive'. That is handled by '|| printf ... incomplete', but it will emit a misleading 'failure-evidence copy incomplete' warning on every failed iteration even when the other root copied fine.

**Suggestion:**
```
Only invoke tar when find produced inputs, e.g. build a file list first and skip when empty; or short-circuit when find ... -print -quit finds nothing before entering the tar pipeline. Keep the non-fatal guard for real copy errors.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: HARNESS_MARKER picks first breach marker in find traversal order, not the newest</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 702)

The host-protection-breach.txt lookup uses `find ... | head -1`, which returns whichever match the multi-root traversal encounters first, not the most recent. With two roots this is now order-dependent: if a stale-but-newer-than-.started marker exists under data/ while the authoritative docker marker is under log-archive/ (or vice versa), the wrong breach line could be reported. The other lookups in this file (iteration_rss_peak_mb, snapshot_iteration_monitor_outputs) already use `-print0 | xargs -0 -r ls -t | head -1` to select the newest; this one does not.

**Suggestion:**
```
Use the same newest-file selection pattern here: `find "${HARNESS_TELEMETRY_DIRS[@]}" -name host-protection-breach.txt -newer "$ITERATION_DIR/.started" -print0 2>/dev/null | xargs -0 -r ls -t 2>/dev/null | head -1`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Colon-separated root list cannot represent paths containing ':'</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/collect-soak-metrics.sh` (line 35)

Using ':' as the list separator (PATH-style) means any log root whose path contains a colon (uncommon but legal, e.g. timestamped dirs like '2026-08-04T02:30') would be split incorrectly and silently skipped by the -d check, degrading to '{}' with no diagnostic. Given log-archive dirs are often timestamp-named, this is not purely theoretical.

**Suggestion:**
```
Either document the constraint prominently at the call site, or accept multiple positional root arguments instead of a delimited single argument (e.g., shift MARKER and REGISTRY handling so roots are "$@"), which sidesteps delimiter collisions entirely.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Colon-delimited roots are a known fragile multi-path convention</b> (Reported by: xai)</summary>

**File:** `scripts/bench/collect-soak-metrics.sh` (line 35)

Splitting LOG_ROOT on ':' works for the controlled SYSTEM_INTEGRATION_DIR paths used here and preserves single-path backward compatibility, but it inherits the classic PATH problem if a root ever contains a colon. Acceptable for this harness, worth a brief comment that roots must not contain ':'.

**Suggestion:**
```
Optional hardening later: accept repeated --log-root flags or NUL-separated input. For now, a one-line comment near the IFS split is enough.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test comment claims metrics CSV under data/ but diff only moves monitor CSV</b> (Reported by: xai)</summary>

**File:** `scripts/bench/test-run-merge-recovery-soak.sh` (line 17)

The new comment says the monitor CSV is planted only under log-archive/ and the metrics CSV only under data/ so a single-root driver fails. The visible diff only relocates resource-timeseries.csv to FAKE_ARCHIVE_DIR; if the metrics CSV dual-placement is not actually asserted elsewhere in the test, the comment overstates coverage.

**Suggestion:**
```
Confirm the test also plants a SOAK_METRIC-bearing log (or metrics CSV) exclusively under data/ and asserts both rss/monitor and registry metrics are collected; otherwise adjust the comment to match what is actually enforced.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Newest-file selection via `xargs -0 ls -t | head -1` breaks on filenames with whitespace</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 200)

The pattern `find ... -print0 | xargs -0 -r ls -t | head -1` parses ls output, so a path containing a newline (or, for downstream consumers, spaces) is mangled. This is pre-existing, but the change widens its blast radius to log-archive/, where docker-provider session dirs may embed container names or timestamps that are less controlled than data/ layouts.

**Suggestion:**
```
If GNU find is guaranteed, prefer `find ... -printf '%T@ %p
' | sort -rn | head -1 | cut -d' ' -f2-` or a small stat-based loop; otherwise document the no-whitespace assumption for session dir names.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Inline subshell IFS-join for the colon-separated argument is hard to read</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 353)

The `"$( IFS=:; printf '%s' "${HARNESS_TELEMETRY_DIRS[*]}" )"` construct inside the collect-soak-metrics.sh invocation is correct (IFS mutation is confined to the subshell) but dense, and it's the only place the array is serialized. If another caller ever needs the joined form, this will get copy-pasted.

**Suggestion:**
```
Compute the joined string once near the array definition (e.g., `HARNESS_TELEMETRY_DIRS_JOINED="$(IFS=:; printf '%s' "${HARNESS_TELEMETRY_DIRS[*]}")"`) or add a small join helper, then pass the variable.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: SYSTEM_INTEGRATION_REF pin bump is deferred — merge-order hazard worth a tracking note</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 366)

The PR states the fix only fully works paired with system-integration `fix/docker-monitor-output-dir`, and the 3-site SYSTEM_INTEGRATION_REF pin bump lands later on this branch. Between this merge and the pin bump, docker iterations will still produce no monitor CSVs (provider returns None), so log-archive/ will be searched but empty — behavior is safe (multi-root find with created dirs), but the daily run tonight will only capture full telemetry if the pin bump also lands before 02:30 UTC.

**Suggestion:**
```
Add a TODO comment near HARNESS_TELEMETRY_DIRS referencing the pending pin bump and the paired system-integration fix, so the intermediate no-op-on-docker state is discoverable from the code rather than only from PR history.
```

</details>
---

<details>

This is a well-reasoned, well-documented fix to a real telemetry-loss bug: docker soak iterations searched only integration-tests/data/ while docker sessions write under log-archive/. The dual-root array is threaded correctly through all six search sites, the colon-list extension to collect-soak-metrics.sh is backward compatible for single-path callers, and the test is thoughtfully constructed to fail if either root is dropped (monitor CSV only in archive, metrics CSV only in data). Remaining concerns are moderate: possible double-counting if any log ever appears under both roots (worth an enforced or documented invariant), the breach-marker lookup uses find-order `head -1` instead of the newest-file pattern used everywhere else, the unconditional up-front mkdir can mask a misconfigured SYSTEM_INTEGRATION_DIR by creating an empty skeleton, and the colon delimiter cannot represent paths containing ':' (relevant for timestamped archive dirs). None of these block merge; the change is functional and safe, and the identified items are hardening follow-ups.

The code changes look good and address the identified issues with telemetry collection. The introduction of handling multiple log roots is well-implemented. Minor improvements could be made by adding comments for better readability.

#### openai (gpt-5.6-sol)

The change consistently searches both harness telemetry roots across monitor snapshots, RSS fallback, latency/error extraction, breach detection, metric collection, and failure-evidence preservation. The collector remains compatible with ordinary single-path callers, safely skips unavailable roots, and preserves degraded-input behavior. The updated driver test appropriately distributes artifacts across both roots, providing useful regression coverage. No actionable security, logic, performance, or style issues were identified.

#### xai (grok-4.5)

Solid, well-motivated fix: centralizing dual harness roots in HARNESS_TELEMETRY_DIRS and threading them through all six search sites (plus collect-soak-metrics multi-root support and a split-placement driver test) correctly addresses the silent docker telemetry loss. No security concerns; array expansion and join/split handling are correct for the controlled paths. Main actionable nit is failure-evidence copy over empty pre-created roots potentially tripping GNU tar’s empty-archive refusal and logging a false 'incomplete' warning—filter empty finds before tar. Safe to merge after that small hardening (or consciously accepting the noise).
</details>

---

jeffrey-l-turner:
Follow-up for the review's double-counting finding (the one substantive minor): when teardown archives a copy of a log that also lives under `data/`, the counting metrics saw it twice — `too_far_ahead_errors` and `SOAK_METRIC` sample counts inflate ~2×; percentile metrics were unaffected (duplicated samples don't move percentiles).

**Fix** (lands with the next commit on this branch):

- `dedup_files_by_content` — drops content-duplicate files from the candidate list before the counting greps, first occurrence wins, hashed by content rather than path (the two roots' layouts differ, so the same file carries unrelated relative paths). Fail-soft: an unhashable file passes through rather than being dropped.
- Wired into `iteration_finalization_latency`, `iteration_too_far_ahead_errors`, and `collect-soak-metrics.sh`'s SOAK_METRIC pass.
- Driver test now plants an **identical** node log in both roots with one `too far ahead` line and asserts `too_far_ahead_errors == 1` — the assertion fails without the dedup.

The remaining minors are acknowledged as cosmetic (mkdir ordering behind the `:?` guard, tar noise on an empty root, first-vs-newest marker selection which is pre-existing single-marker-per-session behavior, readability note).

spreston8:
Reviewed at `da663425` against `dev`. Requested changes below.

**Medium**

1. **Land the `SYSTEM_INTEGRATION_REF` pin bump before merging, not after.** The driver now searches both roots, but the docker provider still returns `None` for `monitor_output_dir` until system-integration `fix/docker-monitor-output-dir` merges and the 3-site bump lands here — so merging as-is restores the search without the files, and tonight's daily would produce the same null `rss_peak_mb` and `finalization_latency` this PR exists to fix. The body already plans it that way; flagging it because the stated 02:30 UTC target creates pressure to merge ahead of the bump, and the failure mode is silent — the change looks landed and nothing improves.

2. **Decide the merge order against #189 explicitly.** The 3 `SYSTEM_INTEGRATION_REF` sites this PR is about to bump are the same legacy sites #189 replaces with the `runnerRef`/`catalogRef` registry, and #189's drift check asserts the legacy value equals *both* registry refs. Whichever lands second has to redo the bump in the other's format, so it is worth choosing the order now rather than resolving it as a conflict later.

**Low**

3. Extract `dedup_files_by_content` into a shared helper. It is currently 16 identical lines in both `run-merge-recovery-soak.sh` and `collect-soak-metrics.sh`, and divergence would not break loudly — it would make the driver's own counting metrics disagree with the registry metrics, which reads as a telemetry inconsistency rather than a bug.

4. Consider a size prefilter before hashing. Every fresh `.log`/`.txt` under both roots is content-hashed three times per iteration — twice in the driver (finalization latency, too-far-ahead) and once in `collect-soak-metrics.sh`. Hashing only files whose size collides removes nearly all of that work, which matters on the long soaks this telemetry serves.

jltatbeach (approved):

PR #200, opened 2026-08-04
Title: Block API silently drops user transfers when the precharge is not the first transfer in report[0]

Timekiller7:
**Problem** 

`extract_transfers_from_report` inferred the deployer from the first transfer in `report[0]`, assuming that batch always holds the precharge. When it didn't, `deployer_addr` was `None` and every transfer of the deploy was dropped. `.skip(1)` also lost user transfers landing in batch 0.

**Fix:**

- Deployer address now comes from `deploy_info.deployer` (`PublicKey` → `VaultAddress` → base58), not from report contents.
- The precharge is skipped only when there is one: `phlo_limit * phlo_price > 0` (genesis runs without cost accounting, so nothing is skipped there).
- Before skipping, `report[0]` is checked against the precharge shape — one transfer, deployer-sent, amount == `total_phlo_charge()`. On a mismatch it is not skipped: all transfers are returned and a `warn!` is logged. Over-reporting plus a log line, never a silent drop.

Batch order (precharge → user → refund) is a contract of `process_deploy_with_cost_accounting`, so it's now pinned by a test in `casper` instead of worked around here.

Identifying the precharge by recipient doesn't work: it goes to `posVaultAddr`, an unforgeable name inside `PoS.rhox`, not derivable from any public key. A phase marker on `SingleReport` would solve this properly, but touches the proto, `rspace++` and `casper` — proposing it as a follow-up.

**Consumer note:** the map still gives every deploy with a non-empty report an entry (possibly empty). Deploys with no transfers are filtered in `runtime/setup.rs`, so they no longer appear in the `transfers_available` event


spreston8:
Reviewed at `e2a5105c` against `dev`. Deriving the deployer from `deploy_info.deployer` is the right fix — it breaks the circular dependency where identifying the deployer required already having found one of their transfers. The assumption the skip rests on also checks out: `reporting_casper.rs:138` passes `with_cost_accounting = !is_genesis`, and `replay_deploy_e` routes through `process_deploy_with_cost_accounting` (`replay_runtime.rs:217`), so for non-genesis blocks the precharge really is inside each user deploy's report and comes first. The `.skip(1)` is also correctly placed after `.filter()`, and the refund is excluded for free since it flows PoS vault → deployer. Tests pass locally (3/3).

Requested changes below.

**Medium**

1. **"Batch ordering no longer matters" isn't quite right — `.skip(1)` is still order-dependent.** The assumption moved rather than disappeared: from *which batch holds the precharge* to *the precharge is the first deployer-sent transfer in flattened order*. This matters because if that assumption breaks the way the batch-position one did, the new failure mode is worse than the old one. Before, a violated assumption dropped every transfer for the deploy — wrong, but visibly absent. Now it would drop the first genuine user transfer and report the precharge in its place, i.e. silently wrong data in the API response, which is harder to notice than missing data. Identifying the precharge structurally would remove the positional dependency altogether — its recipient is the validator/PoS vault, so filtering on that (or on the known precharge amount) is order-free. Failing that, a comment stating the assumption would at least make the next reader aware there is one.

2. **Genesis blocks have no precharge to skip.** Since `with_cost_accounting` is `!is_genesis`, a report for the genesis block contains no precharge, so the first deployer-sent transfer there is genuine and `.skip(1)` discards it. The practical impact is likely nil — genesis pre-mine creates vaults rather than transferring out of the genesis deployer's vault — but there is no guard and no test, so it is worth either confirming that no matching transfer can arise or skipping conditionally.

**Low**

3. Both new failure paths `continue` with no log. Note `deployer_pk` comes from `unwrap_or_default()`, so a deploy with no deployer info yields `""` → `hex::decode("")` succeeds as an empty vec → `PublicKey::from_bytes(&[])` → `from_public_key()` returns `None` → the whole deploy is skipped silently. That is the same shape as the bug being fixed here, so a `tracing::warn!` on both branches would make the next occurrence visible instead of invisible.

4. The tests still encode precharge-first — `make_deploy_reports` returns `vec![precharge_report, user_report]` — so nothing covers the case the PR is about (user transfer in batch 0, precharge later). Reversing the batch order in one test would pin the claim directly.

jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR fixes a real bug where user transfers were silently dropped when the precharge was not the first transfer in report batch 0, by deriving the deployer address deterministically from deploy_info.deployer (PublicKey -> VaultAddress -> base58) and filtering all flattened report batches by sender. The approach is a clear improvement and the added tests (including a two-deployer scenario and the empty-entry behavior change) are good. However, the claim that 'batch ordering no longer matters' is not fully accurate: the .skip(1) still positionally assumes the precharge is the first deployer-originated transfer in flattened iteration order, so a user transfer preceding the precharge, or a missing precharge, would still cause silent misattribution or loss. Additionally, deploys with undecodable deployer keys are skipped silently without logging. The code is functional and safe to merge, but these edge cases and the doc/description accuracy deserve attention.

Deriving the deployer address from `deploy_info.deployer` fixes the original dependence on the first transfer's sender, and omitting empty map entries matches the PR description. However, the new implementation still uses positional logic to identify the precharge, so reordered or missing precharge events can silently drop a real user transfer and expose the precharge instead. This ordering-sensitive case should be resolved and tested before merge.

Solid fix for a real silent-data-loss bug: deriving the deployer from deploy_info and flattening all batches removes the brittle report[0]/skip(1) assumptions and is clearly an improvement. The implementation is readable and the new tests cover basic multi-deploy and empty cases. Main residual risk is the new assumption that the first deployer-sourced transfer is always the precharge—.skip(1) will mis-drop a user transfer if precharge is missing—and the suite does not yet lock in the original failure layout (user/precharge ordering inside batch 0). Also note the intentional map-entry behavior change for deploys with no user transfers; verify callers. Safe to merge after adding a regression test and tightening the precharge documentation/invariant if possible.

The code change addresses a logic error where user transfers were silently dropped if the precharge was not the first transfer. The new approach correctly derives the deployer address and processes all batches, ensuring no transfers are lost. The tests added cover the new functionality adequately. The code looks good and is safe to merge.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Behavior change: deploys with no user transfers no longer get a map entry</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 52)

After flattening, `.skip(1)` assumes the precharge is the first transfer whose sender is the deployer. If a user-transfer batch appears before the precharge batch, the user transfer is discarded and the precharge is returned as a user transfer. The same problem occurs when the precharge event is absent or cannot be parsed: the first legitimate user transfer is silently dropped. This contradicts the stated guarantee that batch ordering no longer matters. The added tests always construct reports as `[precharge_report, user_report]`, so they do not cover this case.

---

Previously every deploy (with a non-empty report) received an entry in the map, possibly empty. Now empty entries are omitted. This is documented in the PR description and covered by the updated test, but any downstream consumers using presence of a key to distinguish 'deploy processed, no transfers' from 'deploy not processed' will see different behavior.

---

Previously every deploy got a map entry (possibly empty Vec). Now entries are omitted when there are no user transfers. Callers that did contains_key / iterated keys expecting all deploy sigs will change behavior. The test update shows this is intentional, but it is an API contract change for consumers of extract_transfers_from_report.

**Suggestion:**
```
Confirm all call sites treat a missing key the same as an empty vec. If any UI/API enumerates deploys from this map alone, document or adjust those call sites in the same PR.
```

</details>

<details>
<summary><b>:red_circle: Critical: skip(1) still depends on iteration order — precharge must be the first matching transfer</b> (Reported by: anthropic, xai)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 47)

After filtering to transfers where from_addr == deployer_addr, .skip(1) unconditionally drops the first match as the precharge. This is correct when precharge always runs and appears first in report order, but it will silently drop a real user transfer if: (1) there is no precharge transfer in the report (e.g. failed/absent precharge, or a deploy type that does not emit one), or (2) report event ordering ever places a user transfer before the precharge. The old code had a related batch-order assumption; this replaces it with a temporal-order assumption that is better but still unvalidated for edge cases.

---

The PR description claims 'Batch ordering no longer matters', but the new code flattens all report batches in order and uses .skip(1) to drop the precharge. This still assumes the precharge (sender == deployer) is the *first* deployer-originated transfer in flattened iteration order. If a user transfer lands in a batch that precedes the precharge's batch, the user transfer will be dropped and the precharge will be reported as a user transfer. Similarly, if a deploy has deployer-originated transfers but no precharge appears in the report (e.g., a truncated/partial report), the first legitimate user transfer is silently discarded.

**Suggestion:**
```
Consider making precharge identification more explicit if the report model allows it (e.g. batch role, system vs user marker). At minimum, add tests for: only-precharge (no user transfer), precharge absent with one user transfer, and multiple user transfers — and document the ordering invariant in the doc comment.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Comment update</b> (Reported by: bedrock)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 10)

The comment should be updated to reflect the new logic where the deployer address is derived from `deploy_info.deployer` and batch ordering no longer matters.

**Suggestion:**
```
Update the comment to explain the new logic.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Deploys with malformed deployer keys are silently skipped without logging</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 36)

When hex::decode fails or VaultAddress::from_public_key returns None, the deploy is skipped with `continue` and no diagnostic. If a deploy actually contained user transfers, they would silently vanish from the API response — the exact class of bug this PR is fixing — with no way to detect it operationally.

**Suggestion:**
```
Emit a warn/debug log with the deploy signature when the deployer public key cannot be decoded or converted to a vault address, so silent data loss is observable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Missing regression test for the bug described in the PR</b> (Reported by: xai)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 330)

The PR fixes silent drops when the precharge is not the first transfer in report[0], and when user transfers land in batch 0. The new tests cover the happy path, empty reports, and two deployers — but none reconstruct the failing layout (user transfer in batch 0, or non-precharge first in batch 0). Without that, the exact failure mode can regress unnoticed.

**Suggestion:**
```
Add a test where batch 0 contains a non-deployer or non-precharge transfer before/with the precharge, and/or a user transfer appears in report[0], asserting those user transfers are still returned.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unused `_from` parameter in test helper</b> (Reported by: anthropic, xai)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 252)

make_block_event_info_with_transfer now takes an unused `_from` parameter (the from address is derived from deployer_hex internally). Keeping a dead parameter makes the helper's signature misleading.

---

make_block_event_info_with_transfer still takes _from but ignores it and re-derives the vault address from deployer_hex. Call sites pass deployer_addr into _from, which is misleading.

**Suggestion:**
```
Drop the from parameter and only take deployer_hex (plus to/amount/sig), deriving the address once inside the helper or via make_deploy_reports as you already do.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Doc comment omits precharge skip invariant</b> (Reported by: xai)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 15)

The updated docs say transfers are those where sender == deployer, but not that the first such transfer is treated as precharge and discarded. That invariant is load-bearing for correctness.

**Suggestion:**
```
Extend the doc comment to state that among transfers with from_addr == deployer, the first in report order is assumed to be the precharge and is skipped.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unnecessary clones in extract_transfers_two_users test</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/web/block_info_enricher.rs` (line 350)

deployer1_hex.clone() and deployer2_hex.clone() are cloned when the originals are not used afterwards; the strings could be moved directly.

**Suggestion:**
```
Drop the .clone() calls and move the hex strings into the DeployInfo structs.
```

</details>
---

<details>

This PR fixes a real bug where user transfers were silently dropped when the precharge was not the first transfer in report batch 0, by deriving the deployer address deterministically from deploy_info.deployer (PublicKey -> VaultAddress -> base58) and filtering all flattened report batches by sender. The approach is a clear improvement and the added tests (including a two-deployer scenario and the empty-entry behavior change) are good. However, the claim that 'batch ordering no longer matters' is not fully accurate: the .skip(1) still positionally assumes the precharge is the first deployer-originated transfer in flattened iteration order, so a user transfer preceding the precharge, or a missing precharge, would still cause silent misattribution or loss. Additionally, deploys with undecodable deployer keys are skipped silently without logging. The code is functional and safe to merge, but these edge cases and the doc/description accuracy deserve attention.

The code change addresses a logic error where user transfers were silently dropped if the precharge was not the first transfer. The new approach correctly derives the deployer address and processes all batches, ensuring no transfers are lost. The tests added cover the new functionality adequately. The code looks good and is safe to merge.

#### openai (gpt-5.6-sol)

Deriving the deployer address from `deploy_info.deployer` fixes the original dependence on the first transfer's sender, and omitting empty map entries matches the PR description. However, the new implementation still uses positional logic to identify the precharge, so reordered or missing precharge events can silently drop a real user transfer and expose the precharge instead. This ordering-sensitive case should be resolved and tested before merge.

#### xai (grok-4.5)

Solid fix for a real silent-data-loss bug: deriving the deployer from deploy_info and flattening all batches removes the brittle report[0]/skip(1) assumptions and is clearly an improvement. The implementation is readable and the new tests cover basic multi-deploy and empty cases. Main residual risk is the new assumption that the first deployer-sourced transfer is always the precharge—.skip(1) will mis-drop a user transfer if precharge is missing—and the suite does not yet lock in the original failure layout (user/precharge ordering inside batch 0). Also note the intentional map-entry behavior change for deploys with no user transfers; verify callers. Safe to merge after adding a regression test and tightening the precharge documentation/invariant if possible.
</details>

---

Note: @Timekiller7 ^^

jeffrey-l-turner (changes requested):

Timekiller7:
hey, @jeffrey-l-turner @spreston8 
I've been reviewing feedback (from Jeff: `Consider making precharge identification more explicit if the report model allows it (e.g. batch role, system vs user marker`) and want to check the approach before writing it.

An explicit phase marker on SingleReport (like a clean fix) touches the proto plus rspace++, casper, and the web layer, and it changes a contract the external services may consume.

So I'd suggest splitting it. In this PR: keep the positional skip but make it conditional (only when phlo_limit * phlo_price > 0, which excludes genesis - the actual cause of the problem) and verify it, if the skipped batch doesn't look like a precharge, don't skip and log a warning, so a broken invariant is noisy instead of silently swapping a user transfer for the precharge. Plus a test in casper pinning the report shape. Then the phase marker as a follow-up PR.

Does that split work for you, or would you rather the marker land here?


PS: still working on providing other changes to PR

jeffrey-l-turner:
> So I'd suggest splitting it. In this PR: keep the positional skip but make it conditional (only when phlo_limit * phlo_price > 0, which excludes genesis - the actual cause of the problem) and verify it, if the skipped batch doesn't look like a precharge, don't skip and log a warning, so a broken invariant is noisy instead of silently swapping a user transfer for the precharge. Plus a test in casper pinning the report shape. Then the phase marker as a follow-up PR.

Yes -- this makes sense to me....

spreston8:
> > So I'd suggest splitting it. In this PR: keep the positional skip but make it conditional (only when phlo_limit * phlo_price > 0, which excludes genesis - the actual cause of the problem) and verify it, if the skipped batch doesn't look like a precharge, don't skip and log a warning, so a broken invariant is noisy instead of silently swapping a user transfer for the precharge. Plus a test in casper pinning the report shape. Then the phase marker as a follow-up PR.
> 
> Yes -- this makes sense to me....

Oh I didn't see this when I brought it up in standup. Whatever works then

Timekiller7:
> > > So I'd suggest splitting it. In this PR: keep the positional skip but make it conditional (only when phlo_limit * phlo_price > 0, which excludes genesis - the actual cause of the problem) and verify it, if the skipped batch doesn't look like a precharge, don't skip and log a warning, so a broken invariant is noisy instead of silently swapping a user transfer for the precharge. Plus a test in casper pinning the report shape. Then the phase marker as a follow-up PR.
> > 
> > 
> > Yes -- this makes sense to me....
> 
> Oh I didn't see this when I brought it up in standup. Whatever works then

okey, thanks!

Timekiller7:
@jeffrey-l-turner, PR is ready for merge

spreston8 (approved):

PR #198, opened 2026-08-04
feat(formal): TLA+ model of byte-bounded block admission; FV process doc

jeffrey-l-turner:
Design proof for the 2026-08-04 daily-soak breach (run 30880995655): per-node attribution showed the readonly observer at 6,492MB against a 947-3,371MB validator baseline — the block-processor queue bounds message count (2048), not bytes, on exactly the node whose whole workload is inbound blocks.

- formal/tlaplus/block_admission: BlockAdmission.tla models the mpsc queue, semaphore drain (in-flight residency included in the budget), and retriever-backed deferral. TLC-validated: the fix config is clean (byte bound + no shedding + liveness under weak fairness); the pre-fix config violates Inv_RetainedBytesBounded (count cap is not a byte cap); the drop config violates liveness (shedding wedges the shard)
- scripts/ci/check-tla-invariants.sh: generalized from slashing-only to <area>/<config> entries; block_admission/MC_BlockAdmission joins the nightly TLC frontier
- docs/formal-verification.md: umbrella treatment of the verification stack, tier ladder, area conventions, verified-areas index, and the block-admission implementation obligations; linked from the top-level README, with the per-area README slimmed to model/config tables

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/198"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788445446&installation_model_id=426883&pr_number=198&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F198&signature=dd47a48a0909e122776529172040436b479a50004f1cd17c32c1d71ef078cd2f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
[...]

PR #201, opened 2026-08-04
Hotfix/fix tla+ ci and soak verdict status for TLA+ gate

jeffrey-l-turner:
Expedited hotfix for two independent defects that keep the repository's top-level status red even when every build/test check passes. Soak state belongs to the README badges and the dashboard; the commit status above the file list should reflect build/test checks only.

## 1. Release soak gate painted a red ✗ on green builds

On every push to `master`, `release.yml`'s `soak_gate` job read the published weekend soak verdict and **failed the job** on `regress` — turning the commit status red even though all CI checks passed (observed on `3e119ee0`, where the verdict gate was the only failing check).

Now the gate **holds without failing**:

- On a `regress` verdict the job succeeds, emits `proceed=false`, and the release job skips. The held state is surfaced on the commit as a **neutral `release-held` check run** with the verdict, failure list, and dashboard link.
- Blocking semantics are unchanged: a regress still blocks the release; `[soak-override]` and the pre-bootstrap path still allow it.
- Hardened failure modes (from the multi-agent review): only a true HTTP 404 (dashboard bootstrap) allows the release through — network errors, 5xx, and malformed JSON **hold** the release rather than failing open or crashing the step; the `GITHUB_OUTPUT` heredoc uses a randomized delimiter so fetched JSON cannot inject outputs; check-run publication failures degrade to a warning instead of re-redding the job.

Validated with a stubbed-curl simulation of all six gate paths (pass, regress with delimiter-injection attempt, 404, network failure, malformed JSON, override during outage) — every path exits 0 with correct proceed/held outputs.

## 2. Nightly TLA+ job has never been green

All eleven nightly `slashing-tests` runs since the schedule began (2026-07-25) failed the same way: `MC_EquivocationDetector` and `MC_EquivocationDetectorEager_3v` hit the 45-minute per-config cap (their interleaved liveness passes go superlinear past ~100M states) while every other config passes in seconds to minutes.

`check-tla-invariants.sh` moves the two cap-busting configs to the opt-in `RUN_EXHAUSTIVE_TLA=1` tier alongside `MC_EquivocationDetector_safety`. The nightly gate now runs the 8 fast configs.

Verified via `workflow_dispatch` run [30931698777](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/30931698777): **all 8 configs clean in ~5½ minutes** — the first passing TLA+ invariant check since the schedule began. Follow-up (tracked): a liveness/safety split for the two heavy models (the pattern that lets `MC_EquivocationDetector_liveness` pass in 3 s) so they return to the nightly tier. Note the nightly workflow also has an independent pre-existing failure in the `cargo-mutants` job (canceled at its deadline every night); that is out of scope here and tracked separately.

## Merge notes

- Expedite path: dev → promote to master before the next nightly.
- PR #198 also touches `check-tla-invariants.sh` (area-namespaced entries, fail-on-missing). Small conflict expected when both reach dev; both sides' semantics should be kept — the reconciliation is planned on the #198 branch.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/201"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788454572&installation_model_id=426883&pr_number=201&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F201&signature=1493a9e585d32b5cc41cbb5025f25f85f49e9db182e7b730539e1a410b097d07"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -39,8 +39,13 @@ jobs:
     timeout-minutes: 5
     if: >-
       !startsWith(github.event.head_commit.message, 'chore(release)')
+    permissions:
+      checks: write
+    outputs:
+      proceed: ${{ steps.gate.outputs.proceed }}
     steps:
-      - name: Require latest weekend soak verdict to be pass
+      - name: Read latest weekend soak verdict
+        id: gate
         shell: bash -euo pipefail {0}
         env:
           OVERRIDE: ${{ contains(github.event.head_commit.message, '[soak-override]') }}
@@ -48,8 +53,16 @@ jobs:
           # Releases must not promote past a regressing weekend soak
           # (EPOCH-010 TASK-010-3). Maintainer override: include
           # [soak-override] in the release commit message.
+          #
+          # A regress HOLDS the release (the release job skips on
+          # proceed=false) without failing this job: soak state belongs to
+          # the README badges and the dashboard, not to the commit's build
+          # status, which must reflect build/test checks only. The hold is
+          # surfaced on the commit as a neutral `release-held` check run
+          # (next step) instead of a red X.
           if [ "$OVERRIDE" = "true" ]; then
             echo "::warning::Soak gate overridden via [soak-override] in commit message"
+            echo "proceed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
           verdict_json="$(curl -fsS --max-time 30 \
@@ -57,16 +70,48 @@ jobs:
             2>/dev/null || true)"
           if [ -z "$verdict_json" ]; then
             echo "::warning::No published soak verdict yet (dashboard bootstrap); allowing release"
+            echo "proceed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
           verdict="$(printf '%s' "$verdict_json" | jq -r '.verdict // "unknown"')"
           date="$(printf '%s' "$verdict_json" | jq -r '.run.date // "unknown"')"
           echo "Latest weekend soak verdict: $verdict ($date)"
           if [ "$verdict" != "pass" ]; then
-            printf '%s' "$verdict_json" | jq -r '.failures[] | "FAIL: " + .'
-            echo "::error::Latest weekend soak regressed — release blocked. Fix the regression or use [soak-override] in the release commit message."
-            exit 1
+            printf '%s' "$verdict_json" | jq -r '.failures[]? | "FAIL: " + .'
+            echo "::warning::Latest weekend soak verdict is '$verdict' ($date) — release held. Fix the regression or use [soak-override] in the release commit message."
+            {
+              echo "held_summary<<HELD_EOF"
+              echo "Latest weekend soak verdict: \`$verdict\` ($date)."
+              echo ""
+              printf '%s' "$verdict_json" | jq -r '.failures[]? | "- " + .'
+              echo ""
+              echo "The release job was skipped; the build itself is unaffected. Fix the regression or include \`[soak-override]\` in the release commit message. Soak state and history: https://f1r3fly-io.github.io/f1r3node-rust/"
+              echo "HELD_EOF"
+            } >> "$GITHUB_OUTPUT"
+            echo "proceed=false" >> "$GITHUB_OUTPUT"
+            exit 0
           fi
+          echo "proceed=true" >> "$GITHUB_OUTPUT"
+
+      - name: Publish neutral release-held check
+        if: steps.gate.outputs.proceed == 'false'
+        uses: actions/github-script@v7
+        env:
+          HELD_SUMMARY: ${{ steps.gate.outputs.held_summary }}
+        with:
+          script: |
+            await github.rest.checks.create({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              name: 'release-held',
+              head_sha: context.sha,
+              status: 'completed',
+              conclusion: 'neutral',
+              output: {
+                title: 'Release held — weekend soak verdict is not pass',
+                summary: process.env.HELD_SUMMARY,
+              },
+            });
 
   release:
     name: Bump Version & Tag
@@ -75,7 +120,8 @@ jobs:
     environment: release-credentials
     timeout-minutes: 10
     if: >-
-      !startsWith(github.event.head_commit.message, 'chore(release)')
+      !startsWith(github.event.head_commit.message, 'chore(release)') &&
+      needs.soak_gate.outputs.proceed == 'true'
     permissions:
       contents: write
     steps:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,30 +65,53 @@ jobs:
             echo "proceed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
-          verdict_json="$(curl -fsS --max-time 30 \
+          # Randomized heredoc delimiter: the summary embeds strings from
+          # the fetched JSON, so a static delimiter would let a crafted
+          # line terminate the block early and inject further outputs.
+          delim="HELD_EOF_${RANDOM}${RANDOM}_$$"
+          hold() {
+            {
+              echo "held_summary<<$delim"
+              cat
+              echo ""
+              echo "The release job was skipped; the build itself is unaffected. Fix the regression or include \`[soak-override]\` in the release commit message. Soak state and history: https://f1r3fly-io.github.io/f1r3node-rust/"
+              echo "$delim"
+            } >> "$GITHUB_OUTPUT"
+            echo "proceed=false" >> "$GITHUB_OUTPUT"
+          }
+          body="$RUNNER_TEMP/latest-verdict.json"
+          http_code="$(curl -sS --max-time 30 --retry 3 --retry-delay 5 \
+            -o "$body" -w '%{http_code}' \
             "https://f1r3fly-io.github.io/f1r3node-rust/data/latest-verdict.json" \
-            2>/dev/null || true)"
-          if [ -z "$verdict_json" ]; then
+            2>/dev/null)" || http_code=000
+          if [ "$http_code" = "404" ]; then
             echo "::warning::No published soak verdict yet (dashboard bootstrap); allowing release"
             echo "proceed=true" >> "$GITHUB_OUTPUT"
             exit 0
           fi
-          verdict="$(printf '%s' "$verdict_json" | jq -r '.verdict // "unknown"')"
-          date="$(printf '%s' "$verdict_json" | jq -r '.run.date // "unknown"')"
+          if [ "$http_code" != "200" ]; then
+            # Fail closed on an unverifiable verdict (network error, 5xx):
+            # holding a release on a transient fetch failure is recoverable;
+            # releasing past an unseen regress is not. Only a true 404
+            # (pre-bootstrap dashboard) allows the release through.
+            echo "::warning::Could not fetch soak verdict (HTTP $http_code) — release held rather than promoting past an unverifiable verdict. Re-run the workflow to retry."
+            hold <<EOF
+          Could not fetch the weekend soak verdict (HTTP $http_code); the release is held rather than promoted past an unverifiable verdict. Re-run the workflow to retry.
+          EOF
+            exit 0
+          fi
+          verdict_json="$(cat "$body")"
+          verdict="$(printf '%s' "$verdict_json" | jq -r '.verdict // "unknown"' 2>/dev/null || echo unknown)"
+          date="$(printf '%s' "$verdict_json" | jq -r '.run.date // "unknown"' 2>/dev/null || echo unknown)"
           echo "Latest weekend soak verdict: $verdict ($date)"
           if [ "$verdict" != "pass" ]; then
-            printf '%s' "$verdict_json" | jq -r '.failures[]? | "FAIL: " + .'
+            printf '%s' "$verdict_json" | jq -r '.failures[]? | "FAIL: " + .' 2>/dev/null || true
             echo "::warning::Latest weekend soak verdict is '$verdict' ($date) — release held. Fix the regression or use [soak-override] in the release commit message."
             {
-              echo "held_summary<<HELD_EOF"
               echo "Latest weekend soak verdict: \`$verdict\` ($date)."
               echo ""
-              printf '%s' "$verdict_json" | jq -r '.failures[]? | "- " + .'
-              echo ""
-              echo "The release job was skipped; the build itself is unaffected. Fix the regression or include \`[soak-override]\` in the release commit message. Soak state and history: https://f1r3fly-io.github.io/f1r3node-rust/"
-              echo "HELD_EOF"
-            } >> "$GITHUB_OUTPUT"
-            echo "proceed=false" >> "$GITHUB_OUTPUT"
+              printf '%s' "$verdict_json" | jq -r '.failures[]? | "- " + .' 2>/dev/null || true
+            } | hold
             exit 0
           fi
           echo "proceed=true" >> "$GITHUB_OUTPUT"
@@ -100,18 +123,26 @@ jobs:
           HELD_SUMMARY: ${{ steps.gate.outputs.held_summary }}
         with:
           script: |
-            await github.rest.checks.create({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
-              name: 'release-held',
-              head_sha: context.sha,
-              status: 'completed',
-              conclusion: 'neutral',
-              output: {
-                title: 'Release held — weekend soak verdict is not pass',
-                summary: process.env.HELD_SUMMARY,
-              },
-            });
+            // Publication is best-effort: a failed check-run API call must
+            // not fail this job, or the gate re-creates the red commit
+            // status the hold mechanism exists to avoid. The held state is
+            // still visible in the job log and the ::warning annotation.
+            try {
+              await github.rest.checks.create({
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                name: 'release-held',
+                head_sha: context.sha,
+                status: 'completed',
+                conclusion: 'neutral',
+                output: {
+                  title: 'Release held by the weekend soak verdict gate',
+                  summary: process.env.HELD_SUMMARY,
+                },
+              });
+            } catch (error) {
+              core.warning(`Could not publish the release-held check run: ${error.message}`);
+            }
 
   release:
     name: Bump Version & Tag
--- a/docs/soak-benchmarks.md
+++ b/docs/soak-benchmarks.md
@@ -282,10 +282,18 @@ Thresholds live in one file: `scripts/bench/soak-gate-thresholds.json`
 the soak run**: failure rate +5pts, peak RSS +20%, finalization p95 +20%,
 throughput −20%. Active-segment metrics warn.
 
-**Releases are gated**: the `release.yml` workflow refuses to bump/tag on
-`master` while the latest published soak verdict is `regress`. Maintainer
-override: include `[soak-override]` in the release commit message (the gate
-logs a warning and proceeds).
+**Releases are gated**: while the latest published soak verdict is `regress`,
+the `release.yml` workflow holds the bump/tag without failing anything — the
+gate job stays green (soak state belongs to the badges and dashboard, not to
+the commit's build status), the release job is skipped, and the hold is
+surfaced as a neutral `release-held` check run on the commit plus a
+`::warning::` annotation on the gate job. A held release is quiet by design:
+no red ✗ appears; the release simply does not happen until the regression is
+fixed or overridden. The gate also holds when the verdict cannot be fetched
+(network error, 5xx, malformed JSON) — only a true 404 (pre-bootstrap
+dashboard) lets the release proceed. Maintainer override: include
+`[soak-override]` in the release commit message (the gate logs a warning and
+proceeds).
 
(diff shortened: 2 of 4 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR makes two reasonable, well-documented changes: (1) the release workflow's weekend-soak gate now 'holds' releases (release job skipped via a `proceed` output, surfaced as a neutral `release-held` check run) instead of failing the commit's build status, and (2) two TLA+ configs that consistently exceed the 45-minute nightly wall-clock cap are moved to the opt-in exhaustive tier. The gate logic is sound and fails closed if the gate step errors, and the added `checks: write` permission is appropriately scoped. Main feedback: harden the GITHUB_OUTPUT heredoc against delimiter injection from the remote verdict JSON, consider distinguishing endpoint outage from bootstrap in the fail-open path, make the check-run publication non-fatal, fix an apparent date typo in the TLA+ script comment, and ensure the demoted heavy TLA+ configs retain some scheduled automated coverage. The code is functional and safe to merge with these improvements considered.

The release-hold output wiring and exhaustive TLA+ tier changes are internally consistent, but the soak verdict retrieval remains fail-open for all fetch failures, allowing releases when the gate cannot actually verify the latest verdict. The neutral check output should also be bounded so an unusually large verdict does not turn the intended neutral hold into a failed workflow.

The release soak gate redesign (fail-the-job → proceed output + neutral release-held check + release job skip) matches the stated intent that soak regressions should hold promotion without redding build status, and the TLA script change correctly demotes known 45m-timeout models to the exhaustive tier with updated comments. Before merge, confirm the release job already depends on soak_gate via needs; without that dependency the new proceed output cannot work and releases will either never run or no longer be held. Also consider continue-on-error on the checks.create step so an API failure cannot reintroduce a red soak_gate status.

The changes appear to be improvements in handling the soak verdict and updating the TLA+ configurations. However, ensure that the new soak verdict handling aligns with the release strategy and update the comments to reflect the current configurations.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Soak gate fails open on network and server errors</b> (Reported by: openai)</summary>

**File:** `.github/workflows/release.yml` (line 65)

The `curl` command suppresses every failure with `2>/dev/null || true`, and an empty result is interpreted as the dashboard bootstrap case. Consequently, DNS failures, timeouts, TLS errors, GitHub Pages outages, HTTP errors, and genuinely missing verdicts all set `proceed=true`. A release can therefore bypass the soak gate precisely when the verdict cannot be verified.

**Suggestion:**
```
Handle the expected bootstrap condition separately, such as allowing only an explicit HTTP 404 when no verdict has ever been published. Treat timeouts, transport failures, 5xx responses, and invalid or empty responses as a held release or a failed gate. Consider validating the JSON schema before setting `proceed=true`.
```

</details>

<details>
<summary><b>:orange_circle: Major: Untrusted remote JSON written into GITHUB_OUTPUT via static heredoc delimiter</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/release.yml` (line 83)

The `held_summary` output is built from the fetched soak verdict JSON (`.verdict`, `.run.date`, `.failures[]`) and written to $GITHUB_OUTPUT using a fixed heredoc delimiter `HELD_EOF`. If any value in the fetched JSON contains a line equal to `HELD_EOF` (or newline-embedded `key=value` pairs), it can prematurely terminate the heredoc and inject arbitrary step outputs. The data comes from a GitHub Pages endpoint that is nominally under the repo's control, but anyone who can influence the published verdict artifact (e.g., via the soak pipeline) could tamper with workflow outputs. Note that `proceed=false` is written after the heredoc so the release gate itself is hard to bypass, but injected outputs are still a hardening gap.

**Suggestion:**
```
Use a randomized delimiter, e.g. `delim="HELD_EOF_$(openssl rand -hex 8)"` (or `$RANDOM$RANDOM`), and/or sanitize the JSON-derived strings (strip newlines) before embedding them.
```

</details>

<details>
<summary><b>:orange_circle: Major: Verify release job declares needs: soak_gate</b> (Reported by: xai)</summary>

**File:** `.github/workflows/release.yml` (line 120)

The release job condition now gates on `needs.soak_gate.outputs.proceed == 'true'`, and soak_gate newly exposes a `proceed` output. The diff does not show a `needs: soak_gate` (or equivalent) entry on the release job. In GitHub Actions, without an explicit `needs` dependency, `needs.soak_gate.outputs.proceed` is always empty, so the comparison to `'true'` fails and the release job will never run. Conversely, if the previous block relied only on soak_gate exiting non-zero plus branch protection, switching to `exit 0` without a `needs` edge would also fail to hold releases. Confirm `needs` already includes `soak_gate`; if not, add it.

**Suggestion:**
```
Ensure the release job includes something like:

```yaml
release:
  needs: soak_gate
  if: >-
    !startsWith(github.event.head_commit.message, 'chore(release)') &&
    needs.soak_gate.outputs.proceed == 'true'
```

If other dependencies already exist, append `soak_gate` to that list.
```

</details>

<details>
<summary><b>:orange_circle: Major: Neutral check-run publication can fail silently on forks / restricted tokens</b> (Reported by: anthropic, openai)</summary>

**File:** `.github/workflows/release.yml` (line 92)

The `release-held` check is created with the default GITHUB_TOKEN and the newly added `checks: write` permission. If the workflow ever runs in a context where check creation fails (token restrictions, API error), the step will fail the job even though the gate step deliberately exited 0 to keep the commit status green — undermining the stated goal that soak state should not redden the commit's build status.

---

The complete remote `failures` array is copied into `HELD_SUMMARY` and sent to the Checks API. Check output summaries have size limits; a large verdict can make `checks.create` return an error, causing the soak-gate job to fail with a red status instead of producing the intended neutral `release-held` result.

**Suggestion:**
```
Add `continue-on-error: true` to the 'Publish neutral release-held check' step so a check-API failure cannot turn the hold into a red X on the commit.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improvement in handling soak verdict</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/release.yml` (line 48)

The change in handling the soak verdict seems to be an improvement as it now holds the release instead of blocking it outright. However, ensure that this behavior aligns with the team's release strategy and that all stakeholders are aware of this change.

**Suggestion:**
```
Consider adding a comment to explain the rationale behind this change and how it fits into the overall release process.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Soak gate fails open when the verdict endpoint is unreachable</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/release.yml` (line 71)

`curl -fsS ... || true` combined with the empty-response branch means any network failure, 404, or endpoint outage silently allows the release (`proceed=true` with only a warning). This is described as 'dashboard bootstrap' behavior but it cannot distinguish bootstrap from an outage or a mis-deployed dashboard, so a real regression could be released past simply because the JSON fetch failed. This is pre-existing behavior carried forward, but the refactor was an opportunity to tighten it.

**Suggestion:**
```
Distinguish HTTP 404 (bootstrap, allow) from other failures (fail closed or retry), e.g. capture the HTTP status with `curl -w '%{http_code}'` and only allow release on an explicit 404.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Check-run publish failure would re-red the gate job</b> (Reported by: xai)</summary>

**File:** `.github/workflows/release.yml` (line 97)

The stated goal is that soak hold state must not fail the commit's build status (neutral `release-held` check instead of a red X). The new "Publish neutral release-held check" step runs without `continue-on-error`. If `checks.create` fails (token/permission glitch, API outage, etc.), soak_gate itself goes red and undermines that design.

**Suggestion:**
```
Add `continue-on-error: true` to the check-publish step, and optionally emit a `::warning::` if you care to surface API failures without failing the job:

```yaml
- name: Publish neutral release-held check
  if: steps.gate.outputs.proceed == 'false'
  continue-on-error: true
  uses: actions/github-script@v7
  ...
```
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Update comment to reflect current TLA+ configurations</b> (Reported by: bedrock)</summary>

**File:** `scripts/ci/check-tla-invariants.sh` (line 4)

The comment in the script does not match the current configurations being checked. Update the comment to accurately reflect the TLA+ configurations that are being checked.

**Suggestion:**
```
Update the comment to include the correct configurations: MC_EquivocationDetector_liveness, MC_EquivocationDetectorEager, MC_ConcurrentTracker, MC_SlashFlow, and MC_TwoLevelSlashing.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Likely incorrect date in comment ('2026-07-25')</b> (Reported by: anthropic)</summary>

**File:** `scripts/ci/check-tla-invariants.sh` (line 23)

The comment states the nightly schedule began on '2026-07-25', which is a future date and is almost certainly a typo (probably 2025-07-25). Incorrect dates in rationale comments undermine their value as historical record.

**Suggestion:**
```
Correct the date to the actual schedule start (e.g., 2025-07-25).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Nightly TLA+ coverage regression: full detector and 3-validator eager configs no longer run by default</b> (Reported by: anthropic)</summary>

**File:** `scripts/ci/check-tla-invariants.sh` (line 74)

Moving MC_EquivocationDetector and MC_EquivocationDetectorEager_3v to the opt-in RUN_EXHAUSTIVE_TLA tier means the default nightly gate no longer exercises the interleaved safety+liveness detector model or the 3-validator eager variant at all. The rationale (45-minute wall-clock cap consistently hit) is documented, but until the promised liveness/safety split lands, regressions in those state spaces will only be caught if someone remembers to run the exhaustive tier.

**Suggestion:**
```
Track the liveness/safety split for these two configs in an issue referenced from the comment, and/or schedule a periodic (e.g., weekly) job that runs with RUN_EXHAUSTIVE_TLA=1 so the heavy configs still get automated coverage.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: No audit check run on [soak-override]</b> (Reported by: xai)</summary>

**File:** `.github/workflows/release.yml` (line 64)

Override and "no verdict yet" paths set `proceed=true` and exit successfully with only a workflow warning. A neutral/notice check run (e.g. `release-soak-overridden`) would make overrides visible on the commit UI and dashboard history the same way holds are now surfaced.

**Suggestion:**
```
Consider publishing a neutral check when `OVERRIDE=true`, reusing the same checks:write permission, so overrides are reviewable without digging through workflow logs.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Markdown injection into check-run summary from remote data</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/release.yml` (line 88)

Failure strings from the verdict JSON are embedded verbatim into the check-run's Markdown summary. Malformed or malicious content in the failures array would render as arbitrary Markdown (links, images) on the commit's checks page. Low impact, but easy to sanitize.

**Suggestion:**
```
Escape or code-fence the failure lines (e.g., wrap each in backticks after stripping backticks from the input) before embedding them in the summary.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Default TLA gate coverage intentionally reduced</b> (Reported by: xai)</summary>

**File:** `scripts/ci/check-tla-invariants.sh` (line 7)

MC_EquivocationDetector and MC_EquivocationDetectorEager_3v move from the always-on POST_FIX_CONFIGS list into the RUN_EXHAUSTIVE_TLA=1 tier because they hit the 45m cap. That fixes flaky/timeout CI signal, but the nightly/default gate no longer model-checks those two instances unless exhaustive mode is enabled somewhere on a schedule.

**Suggestion:**
```
Confirm a scheduled or manual workflow actually sets RUN_EXHAUSTIVE_TLA=1 periodically so these configs do not silently rot. If not, add a non-gating nightly job or track the liveness/safety split called out in the comment.
```

</details>
---

<details>

This PR makes two reasonable, well-documented changes: (1) the release workflow's weekend-soak gate now 'holds' releases (release job skipped via a `proceed` output, surfaced as a neutral `release-held` check run) instead of failing the commit's build status, and (2) two TLA+ configs that consistently exceed the 45-minute nightly wall-clock cap are moved to the opt-in exhaustive tier. The gate logic is sound and fails closed if the gate step errors, and the added `checks: write` permission is appropriately scoped. Main feedback: harden the GITHUB_OUTPUT heredoc against delimiter injection from the remote verdict JSON, consider distinguishing endpoint outage from bootstrap in the fail-open path, make the check-run publication non-fatal, fix an apparent date typo in the TLA+ script comment, and ensure the demoted heavy TLA+ configs retain some scheduled automated coverage. The code is functional and safe to merge with these improvements considered.

The changes appear to be improvements in handling the soak verdict and updating the TLA+ configurations. However, ensure that the new soak verdict handling aligns with the release strategy and update the comments to reflect the current configurations.

#### openai (gpt-5.6-sol)

The release-hold output wiring and exhaustive TLA+ tier changes are internally consistent, but the soak verdict retrieval remains fail-open for all fetch failures, allowing releases when the gate cannot actually verify the latest verdict. The neutral check output should also be bounded so an unusually large verdict does not turn the intended neutral hold into a failed workflow.

#### xai (grok-4.5)

The release soak gate redesign (fail-the-job → proceed output + neutral release-held check + release job skip) matches the stated intent that soak regressions should hold promotion without redding build status, and the TLA script change correctly demotes known 45m-timeout models to the exhaustive tier with updated comments. Before merge, confirm the release job already depends on soak_gate via needs; without that dependency the new proceed output cannot work and releases will either never run or no longer be held. Also consider continue-on-error on the checks.create step so an API failure cannot reintroduce a red soak_gate status.
</details>

---

jltatbeach (approved):

jeffrey-l-turner:
## Review resolutions (multi-agent review → `aff2428b`)

All four majors dispositioned; three fixed in `aff2428b`, one confirmed false positive.

| # | Finding | Disposition |
|---|---|---|
| 1 | **Soak gate fails open on network/server errors** (OpenAI) | **Fixed.** The gate now inspects the HTTP code explicitly: only a true 404 (pre-bootstrap dashboard) allows the release through. Network errors (curl failure/HTTP 000), 5xx, and malformed JSON all **hold** the release (`proceed=false` + neutral `release-held` check) — fail closed, but still green. `curl` also gained `--retry 3`. Note the fail-open on an empty body predates this PR; it is now removed entirely. |
| 2 | **Static `HELD_EOF` heredoc delimiter → output injection** (Anthropic) | **Fixed.** The `held_summary` block uses a randomized per-run delimiter (`HELD_EOF_${RANDOM}${RANDOM}_$$`), so a crafted line in the fetched JSON cannot terminate the block early and inject further outputs. Verified in simulation with a literal `HELD_EOF` line in the failures array — it renders inert inside the block. |
| 3 | **Verify release job declares `needs: soak_gate`** (xAI) | **False positive.** The release job has declared `needs: soak_gate` since before this PR (unchanged by the diff), and additionally now gates on `needs.soak_gate.outputs.proceed == 'true'`. |
| 4 | **Neutral check-run publication can fail silently / re-red the job** (Anthropic, OpenAI) | **Fixed.** `checks.create` is wrapped in try/catch; an API failure degrades to `core.warning` instead of failing the gate job (which would have re-created exactly the red X this mechanism removes). The held state remains visible in the job log and `::warning` annotation. This workflow runs only on push to `master`, so the fork-token variant of the concern does not apply, but the guard covers restricted-token failures too. |

**Validation:** stubbed-curl simulation of all six gate paths — `pass`, `regress` (including the delimiter-injection attempt), `404`, network failure, malformed JSON, and `[soak-override]` during an outage — every path exits 0 with the correct `proceed`/`held_summary` outputs. `.github/scripts/check-workflow-invariants.sh` passes (all `run:` blocks parse; untrusted inputs stay in expression/env contexts).

**Minors, dispositioned without change:**
- *"Likely incorrect date 2026-07-25"* — false positive: that is the actual first scheduled nightly run of `slashing-tests` in this repo.
- *Nightly TLA+ coverage reduction* — the deliberate, documented tradeoff: the two configs have never completed inside the cap, so nightly "coverage" was nominal; they remain runnable via `RUN_EXHAUSTIVE_TLA=1` and the liveness/safety split to restore them is tracked as follow-up.
- *No audit trail on `[soak-override]`* — the override already emits a `::warning` annotation on the run and requires the marker in the release commit message itself (permanent in git history); left as is for this expedited hotfix.
- *Markdown injection into check-run summary from remote data* — content comes from the repo's own published dashboard endpoint over HTTPS; with output-injection closed (#2), residual markdown styling in a neutral check summary is accepted risk.

spreston8:
Reviewed at `aff2428b` against `dev`. Both fixes look correct, and the gate rework is fail-closed in every direction constructed against it: `needs: soak_gate` is declared on the release job (referencing `needs.soak_gate` without it would have silently skipped every release), an unset or empty `proceed` skips the release, a gate-job failure skips it via `needs`, and only a literal 404 proceeds — network failure, non-200, and jq failure all hold. Worth noting an accidental extra layer: `hold()` writes the summary heredoc before `proceed=false`, and later assignments to the same output key win, so even a successful delimiter injection setting `proceed=true` inside the summary would be overridden.

A few things below, none blocking.

**Medium**

1. **The doc the script cites as its reference is now stale.** `check-tla-invariants.sh` still names `docs/theory/slashing/design/14-test-plan.md` §14.6/§14.9 as its reference, and that document (lines 907–910) says `MC_EquivocationDetector_safety.cfg` *is* the exhaustive detector run, "intentionally excluded from the default PR-gate script … enabled by `RUN_EXHAUSTIVE_TLA=1`". After this change the exhaustive tier holds three configs. The script's own header comment was updated carefully; the doc it points at now contradicts it, which is the version a reader chasing the reference will land on.

**Low**

2. The `GITHUB_OUTPUT` round-trip is what forces the hardening. Passing a multi-line summary through an output is the reason the randomized delimiter and the dedented heredoc both exist. Writing the summary to `$RUNNER_TEMP/held-summary.md` and reading it with `fs.readFileSync` in the `github-script` step would remove the multi-line output entirely — no delimiter to randomize, no injection surface to harden, no heredoc indentation to preserve. The hardening as written is sound; the design that requires it is avoidable.

3. Related: the heredoc is correct today only because `hold <<EOF` sits at 12 spaces while its body and `EOF` sit at 10 — exactly the `run: |` block-scalar base indent, so the terminator lands at column 0 after YAML dedenting. It reads as a mis-indentation, and any reindentation to match the enclosing `if` breaks the terminator silently.

4. Nightly equivocation coverage is now at most two validators. `MC_EquivocationDetectorEager` is `{v1, v2}` and `_3v` is `{v1, v2, v3}` per their `.tla` modules, and `MC_EquivocationDetector_liveness` is a 1-validator model by its own comment — so moving `_3v` out leaves the nightly tier checking an inherently multi-validator property at two validators at most. Accepted and tracked per the description; worth scoping the liveness/safety-split follow-up to restore `_3v` specifically, not only the combined model.

5. The hold is quieter than a maintainer may expect. A regress used to be an unmissable red ✗; now the gate is green, the release job skips, and the signal is a neutral `release-held` check plus a `::warning::`. That is the intent, but "the release silently did not happen" is a new failure mode, and `docs/soak-benchmarks.md` — which carries a "Gates releases: yes" row and is where someone would look for expected behaviour — does not mention the new signal.

jeffrey-l-turner:
Thanks @spreston8 — addressed in `638d919f` (docs only; no workflow/script changes):

| # | Item | Disposition |
|---|---|---|
| 1 (Medium) | Stale reference doc | **Fixed.** All three stale passages in `docs/theory/slashing/design/14-test-plan.md` (§14.6 intro, the §14.6 tier description at the cited lines, §14.9 runbook) now describe the three-config exhaustive tier, the 45-minute-cap history behind the move, and the tracked liveness/safety-split follow-up. |
| 5 (Low) | Quiet hold undocumented | **Fixed.** `docs/soak-benchmarks.md`'s "Releases are gated" paragraph now documents the hold semantics: green gate job, skipped release job, neutral `release-held` check + `::warning::`, fail-closed on unfetchable verdicts with 404-only bootstrap allow — and calls out explicitly that a held release is quiet by design. |
| 2–3 (Low) | File-based summary instead of the `GITHUB_OUTPUT` heredoc | **Deferred.** Agreed it's the cleaner design — it removes the injection surface and the heredoc-dedent fragility outright rather than hardening around them. Declining for this expedited hotfix only because the current mechanism is simulation-validated and approved; recorded as an adoptable follow-up for the next touch of `release.yml`. |
| 4 (Low) | Nightly detector coverage ≤2 validators | **Accepted; follow-up rescoped.** The liveness/safety-split follow-up is now scoped (in the test-plan doc and tracking notes) to restoring `MC_EquivocationDetectorEager_3v` specifically, as the only three-validator detector model. |

Also appreciated the note on `hold()`'s write order making a hypothetical injected `proceed=true` lose to the later `proceed=false` — happy accident, now load-bearing defense-in-depth.

PR #209, opened 2026-08-07
fix(network): preserve transient peer liveness

a-k-l-sdao:
## Summary

- Retain a peer connection across transient heartbeat failures.
- Reset the consecutive-failure streak after successful communication.
- Remove and rediscover a peer only after the configured failure threshold is reached.

## Regression coverage

`integration-tests/test/tests/custom/test_transient_peer_liveness.py::test_transient_peer_failure_does_not_disconnect`

The regression pauses a peer repeatedly and verifies retention after a transient failure, streak reset after recovery, removal after three consecutive failures, and subsequent rediscovery.

## Validation

- Baseline `dev`: the targeted system regression failed.
- This branch: all ten `clear_connections` Rust tests passed.
- Strict Clippy for `comm` and `node` passed.

## Dependency


a-k-l-sdao:
Updated in `904e832c` with the strict-Clippy generic-bound cleanup. All ten `clear_connections` tests, formatting, and strict Clippy pass after rebasing onto current `dev`.

spreston8 (approved):
Reviewed at `6d75f8b7f` against `dev`. Approving.

Pushed to this branch rather than requested: the `dev` merge, a `PeerLivenessTracker` refactor, `--discovery-heartbeat-failure-threshold`, a revert of the two timing defaults, both regression gates, and the pin.

The regression could not run: `test_transient_peer_liveness.py` was deselected and gated on a capability no manifest advertised. Both gates are now open.

**Worth reconsidering** — `network-timeout` is back to 5s and `cleanup-interval` to 10s. Three consecutive failures already covers a transient blip; widening these defends against the same thing twice and compounds, taking a dead peer from ~45s to ~135s. `network-timeout` is also the deadline for every protocol send and for the Kademlia RPC, so tripling it slows block fetch and discovery too. Nothing in the branch measures the false-strike rate at 5s. If there is load data behind 15 and 30, the revert should be undone.

Issue #48, opened 2026-05-20
Reporter (block_report API) replay panics on LFS-synced READONLY observers


spreston8:
> Migrated from [F1R3FLY-io/f1r3node#509](https://github.com/F1R3FLY-io/f1r3node/issues/509) as part of the deprecation of that repository. Original author: @spreston8. Comments on the legacy issue are not mirrored here — see the linked legacy issue for the full thread.

---

## Symptom

A fresh readonly observer attached mid-shard via LFS sync emits panic clusters in `tokio-rt-worker` whenever the gRPC API server's transfer-enrichment path triggers `block_report.trace(block)`:

```
WARN Deploy replay failed, returning empty events
  deploy_index=0  error="System runtime error: Unable to consume results of system deploy"
... (one per deploy in the block)
WARN Discarded 0 replay events during precharge error path

thread 'tokio-rt-worker' panicked at rholang/src/rust/interpreter/rho_runtime.rs:386:14:
called `Result::unwrap()` on an `Err` value:
  BugFoundError("Unused COMM event: replayData multimap has 471 elements left")
```

The reporter creates a fresh `RhoReportingRspace` over the same store, resets to the block's `pre_state_hash`, then for each deploy replays the precharge system deploy. `consume_system_result` returns `None` (no result on the precharge return channel), each deploy's replay aborts, the rspace's `replayData` multimap stays full of unconsumed COMM events, and `create_checkpoint()` panics on the dirty trace.

## Bisect

Reproducible across PR branches and `rust/staging` baseline (panic counts vary slightly with timing; pattern is identical):

| Binary | Test | observer1 panics | observer1 deploy failures |
|---|---|---|---|
| `rust/staging` (`bf6a03d2`) | test_observer_lfs_sync | 2 | 12 |
| `feat/lfs-coverage` (PR #507) | test_observer_lfs_sync | 2-4 | 12-24 |
| `feat/bonding-additiveset` (this branch) | test_observer_lfs_sync | 2 | 12 |
| `feat/bonding-additiveset` | test_bonding_validators (B5) | 4 | 18 |

**The bug pre-exists on `rust/staging`**, not introduced by any of the bonding/LFS PRs in flight. Was hidden until today by a transient-observer scanner gap in the integration-test framework (now fixed on system-integration `refactor/integration-test-framework`).

## Bug class (narrowed)

[...]

Issue #37, opened 2026-05-18
LFS-synced observer cannot validate post-LFS blocks: parent walk fails on hashes below min_height, observer stuck at approved-state height indefinite


Timekiller7:
### What did you expect to see?
A freshly-started observer with a clean data directory should be able to join a long-running shard and catch up to the chain tip by validating and replaying new blocks broadcast by validators after the initial LFS sync completes.

### What did you see instead?
The observer completes the initial LFS sync successfully but then never processes a single subsequent block, getting stuck at the approved-state height indefinitely. CLI confirms the observer at block=5639, finalized=5639 while validators advance past 5879+.
The observer logs are flooded with the following errors — ~53,700 occurrences in 1h 43min, all on the same 2 incoming blocks from validators:
```
ERROR] Error processing block c568440f63...:
        KvStore error: Invalid argument: DAG storage is missing hash 3ae31924a0...

[ERROR] Error processing block f20c8fe076...:
        KvStore error: Invalid argument: DAG storage is missing hash 3ae31924a0...

[ERROR] [RootRepository] validateAndSetCurrentRoot FAILED:
        Blake2b256Hash(875d3b30e7...) not in roots store
```
The missing hash 3ae31924a0 is block #5588 — exactly one block below min_height=5589 of the LFS-restore range. The missing roots (one of them is 875d3b30e7) are post-states of blocks #5581 and #5582 — even deeper.
The regular Casper block validation pipeline tries to walk parents from incoming #5640+ blocks back to the approved state, and the walk goes below the LFS window's min_height. There it hits hashes that LFS never imported, and the validation cycle fails indefinitely — the observer retries the same blocks ~26,600 times each without ever advancing.

This appears related to [#509](https://github.com/F1R3FLY-io/f1r3node/issues/509), but manifests on a different code path. 

Key differences:
- Not the reporter — enable-reporting = false would not help. The error happens on the normal block_processor path during incoming-block validation
- No panic — observer keeps running, just retries the same blocks indefinitely, never advancing
- Triggered without any gRPC API calls, purely from validator broadcasts

Likely the same underlying class of bug

### Steps to reproduce the bug

1. Start a network of 3 validators from genesis and let it run until it reaches a height around 5000–6000 blocks (or any height where typical multi-parent depth exceeds the LFS min_height window — roughly 50 blocks below the approved block).
2. Start a fresh observer node from scratch (no existing approved block, new node identity)
[...]

PR #208, opened 2026-08-07
fix(casper): reject expired deploys at admission

a-k-l-sdao:
## Summary

- Reject deploys whose valid-after block has fallen outside the configured deploy lifespan.
- Preserve the inclusive admission boundary for deploys still inside the valid window.
- Prevent expired deploys from entering deploy storage.

Refs #197

This addresses expired deploy admission, one input to the stale-backlog finalization stall described in #197. It does not address the rejected-deploy re-inclusion window tracked by #42.

## Regression coverage

Corresponds to F1R3FLY-io/system-integration#85:

`integration-tests/test/tests/standalone/test_expired_deploy_admission.py::test_expired_deploy_rejected_at_admission`

The regression checks rejection at the expired boundary, acceptance immediately inside the valid window, finalization of the valid deploy, and absence of the expired deploy from storage.

## Validation

- Baseline `dev`: the expired deploy was accepted.
- This branch: the targeted expiration regression passed.
- Casper strict Clippy passed.
- Formatting and whitespace checks passed.

## Dependency

None. Rebased directly onto current `dev`.

PR #210, opened 2026-08-07
fix(observer): reject invalid replay reports

a-k-l-sdao:
## Summary

- Propagate replay and report-collection failures instead of caching empty report events.
- Reject replay reports when their computed post-state hash differs from the block's recorded post-state hash.
- Distinguish post-state divergence from deploy replay failure.
- Allow observer catch-up to retry missing blocks without publishing invalid reports.

Refs #48

This addresses invalid reporter replay and post-state recovery on LFS-synced observers. It does not address the separate normal block-processing parent-walk failure tracked by #37.

## Regression coverage

`integration-tests/test/tests/custom/test_observer_missing_block_retry.py::test_observer_retries_missing_block_after_peer_returns`

The regression removes every block source during observer synchronization, restores them, and verifies retry, LFB catch-up, target-block visibility, and exact post-state-hash recovery.

## Validation

- Baseline `dev`: functional assertions completed, but teardown detected invalid empty replay events.
- This branch: observer recovery regression passed with no forbidden replay warning.
- Reporting Casper behavior regression passed.
- Strict Clippy for `casper` and `node` passed.

## Dependency


jeffrey-l-turner (approved):
## Multi-Agent Code Review

**Verdict:** :white_check_mark: Approved (75.0000% agreement)

**Reviewed by:**

| openai | `gpt-5.6-sol` | Approve | 0.93 |

---

### Summary

This PR correctly fixes a real bug: replay/report-collection failures were previously swallowed and cached as empty report events, causing observers to serve invalid reports and preventing retry. The new fail-fast propagation plus post-state hash validation is the right direction and is backed by an integration regression test. The main concern is the all-or-nothing replay semantics: if `replay_deploy_e` can return an error for deploys that legitimately failed during original execution, valid blocks containing errored deploys would now have permanently unavailable reports — this should be verified before merge. Secondary suggestions: ensure validation strictly precedes any report caching, introduce a dedicated error variant for post-state mismatch, and prefer typed errors over strings so retryable vs. non-retryable failures can be distinguished. Overall, the change is functional and safe to merge once the errored-deploy replay behavior is confirmed.

The change correctly propagates deploy replay and report-collection failures instead of converting them into successful empty reports, and it prevents invalid replay reports from being cached by verifying the computed post-state hash against the block's recorded value. Error messages include useful replay context, and the added hash encoding occurs only on failure paths. No actionable security, logic, performance, or style issues are evident in the provided diff.

Solid correctness fix: replay and report-collection failures are now propagated instead of being swallowed as empty event lists, and the API rejects reports whose computed post-state hash does not match the block header. That directly addresses invalid/empty observer reports and enables safe retry on catch-up. The changes are focused, the error messages are actionable, and no security or logic regressions stand out; only a minor API-clarity suggestion around error typing.

The code changes look good and address the intended fixes. Minor style improvements can be made for consistency and readability.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: All-or-nothing replay may block reports for blocks containing legitimately failed deploys</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/reporting_casper.rs` (line 177)

Previously, a failed deploy replay produced empty events for that deploy while the rest of the block's report was still generated. Now, any single deploy replay error aborts the entire block report. If `replay_deploy_e` returns `Err` for deploys that legitimately errored during original execution (e.g., user-level Rholang errors, out-of-phlo deploys that were recorded as errored in the block), then blocks containing any such deploy will permanently fail report generation, even though the block itself is valid. This is a behavioral regression risk that depends on whether `replay_deploy_e` distinguishes 'expected' deploy failures (returns Ok) from genuine replay divergence (returns Err).

**Suggestion:**
```
Verify that `replay_deploy_e` returns `Ok` for deploys whose failure matches the recorded execution outcome (errored=true in the ProcessedDeploy). If not, only propagate errors when the replay outcome diverges from the recorded outcome; otherwise collect the (possibly empty) events and continue.
```

</details>

<details>
<summary><b>:orange_circle: Major: Reusing ReplayFailed for post-state mismatch conflates distinct failure modes</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 103)

A post-state hash mismatch after a successful replay is semantically different from a replay execution failure. Reusing `BlockReportError::ReplayFailed` makes it harder for callers (and observability/alerting) to distinguish 'replay engine failed' from 'replay succeeded but produced divergent state', which usually indicates a more serious consensus/state issue.

---

The post-state hash mismatch is reported via BlockReportError::ReplayFailed. That works, but a distinct variant (e.g. PostStateHashMismatch) would make it easier for callers and metrics to distinguish replay execution failures from integrity mismatches without parsing the error string.

**Suggestion:**
```
Add a BlockReportError::PostStateHashMismatch { computed, expected, block_hash } variant (or similar) and return that from the new check.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improve error message formatting</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 98)

Consider using a more concise format for the error message to improve readability.

**Suggestion:**
```
Use a single format! macro for the error message.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Post-state validation may occur after report caching in lower layers</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 101)

The post-state hash validation happens in `BlockReportAPI` after `trace`/replay returns. If the replay result (or any intermediate report) is cached inside `ReportingCasper` or the `ReportStore` before this validation runs, an invalid report could still be persisted and served on subsequent requests, defeating the purpose of the check. The PR description says failures should 'allow observer catch-up to retry missing blocks without publishing invalid reports', so cache writes must strictly follow this validation.

**Suggestion:**
```
Confirm that no caching (ReportStore write or in-memory memoization) occurs before the post-state hash check, or move the validation into `RhoReporterCasper::trace` so it runs before any persistence.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Consistent error handling</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/reporting_casper.rs` (line 175)

Ensure consistent error handling by using the same pattern for both deploy and system deploy replays.

**Suggestion:**
```
Use map_err for both replay_deploy_e and replay_block_system_deploy.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Missing comment explaining the post-state validation invariant</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 101)

The new hash comparison is a correctness-critical guard but has no comment explaining why the computed post-state hash must equal the block's recorded post-state hash, or what conditions (missing blocks during observer catch-up) can cause a mismatch.

**Suggestion:**
```
Add a brief comment referencing the observer missing-block scenario (system-integration#85) so future maintainers understand why the guard exists and why failure here should not be cached.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Stringly-typed errors in replay paths</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/reporting_casper.rs` (line 180)

The new error propagation formats everything into `String` errors (e.g., `format!("Deploy replay failed at index {} ...")`). This loses the underlying error type and makes programmatic handling (retry vs. permanent failure classification) at higher layers impossible — relevant since the observer is supposed to retry on transient failures.

**Suggestion:**
```
Consider a typed error enum (thiserror) for the reporting replay path so callers can distinguish transient failures (retryable) from deterministic replay divergence (non-retryable).
```

</details>
---

<details>

This PR correctly fixes a real bug: replay/report-collection failures were previously swallowed and cached as empty report events, causing observers to serve invalid reports and preventing retry. The new fail-fast propagation plus post-state hash validation is the right direction and is backed by an integration regression test. The main concern is the all-or-nothing replay semantics: if `replay_deploy_e` can return an error for deploys that legitimately failed during original execution, valid blocks containing errored deploys would now have permanently unavailable reports — this should be verified before merge. Secondary suggestions: ensure validation strictly precedes any report caching, introduce a dedicated error variant for post-state mismatch, and prefer typed errors over strings so retryable vs. non-retryable failures can be distinguished. Overall, the change is functional and safe to merge once the errored-deploy replay behavior is confirmed.

The code changes look good and address the intended fixes. Minor style improvements can be made for consistency and readability.

#### openai (gpt-5.6-sol)

The change correctly propagates deploy replay and report-collection failures instead of converting them into successful empty reports, and it prevents invalid replay reports from being cached by verifying the computed post-state hash against the block's recorded value. Error messages include useful replay context, and the added hash encoding occurs only on failure paths. No actionable security, logic, performance, or style issues are evident in the provided diff.

#### xai (grok-4.5)

Solid correctness fix: replay and report-collection failures are now propagated instead of being swallowed as empty event lists, and the API rejects reports whose computed post-state hash does not match the block header. That directly addresses invalid/empty observer reports and enables safe retry on catch-up. The changes are focused, the error messages are actionable, and no security or logic regressions stand out; only a minor API-clarity suggestion around error typing.
</details>

---

jeffrey-l-turner:
## Multi-review response

**Major 1 — "All-or-nothing replay may block reports for blocks containing legitimately failed deploys" (anthropic): disposition — does not occur; no change.**
`replay_deploy_e` errors only on a **status mismatch**, not on recorded failure: `replay_runtime.rs:383` compares `processed_deploy.is_failed != !eval_successful`, so a deploy recorded as failed that replays as failed returns `Ok` and contributes its events to the report. The hard-fail this PR introduces therefore triggers only on genuine replay divergence (recorded-success replaying failed, recorded-failure replaying clean, or infra errors) — which is exactly the class of invalid report the PR exists to reject.

**Major 2 — "Reusing `ReplayFailed` for post-state mismatch conflates distinct failure modes" (anthropic, xai): fixed.**
`BlockReportError` gains a dedicated `PostStateMismatch { block, computed, recorded }` variant, used by the post-state check in `block_report_api.rs`; `reporting_routes.rs` maps it in its own match arm (same 500 status, distinguishable error body and log line). "Every deploy replayed cleanly yet the aggregate state diverged" — a consensus-integrity signal — is now structurally distinct from "the trace itself failed."

a-k-l-sdao:
Fixed: post-state divergence now returns a dedicated `PostStateMismatch` error mapped to HTTP 500, while execution failures remain `ReplayFailed`. Rebased on `dev`; Casper and Node tests pass.

jeffrey-l-turner:
@a-k-l-sdao — from an open-issue audit (mapping issues to addressing PRs): could you add explicit `Fixes #N` / `Refs #N` lines to this PR's description for the issues it addresses? Cross-referencing had to be reconstructed thematically because the linkage is missing — and note closing keywords only auto-close when a PR merges to the DEFAULT branch, so dev-targeted fixes need a manual close at promotion time either way. Candidate issues for this PR (please confirm/correct): #48 (block_report replay panics on LFS-synced observers), #37 (LFS observer parent-walk failure).

a-k-l-sdao:
Confirmed the issue mapping: this PR references #48 and does not claim to fix the separate #37 parent-walk problem. Rebased onto current `dev`; the observer recovery regression, reporting regression, formatting, and strict Clippy pass.

spreston8 (commented):
Reviewed at `fe4d84e2` against `dev`. Requested changes below.

**HIGH**

1. Scope the escalation so it does not convert a tolerated condition into a permanent failure. `check_replay_data_with_fix` (`replay_runtime.rs:645`) is meant to tolerate an unused COMM when the deploy legitimately failed, but its guard is `err_msg.contains("unused")` while the message reads `"RSpace Bug Found Error: Unused COMM event: …"` — capital `U`, `contains` is case-sensitive, and no lowercase producer exists in the tree. The branch is dead, so that error arrives here unsuppressed.

   That is pre-existing and harmless on `dev`, because the old code swallowed the error into empty events regardless. This PR makes it consequential: a block containing a failed deploy now fails report generation permanently. That is the scenario raised as Major 1, whose dismissal addressed the status check at `:383` — a different error class one call later, and genuinely benign as described.

   `replay_runtime.rs` is not one of this PR's files, so the smaller fix is to not propagate that specific class from `reporting_casper.rs` and leave the comparison to whoever owns it. Fixing the comparison upstream — matching on the error variant rather than its rendered text — would stop it recurring, but is optional here. Same exposure for system deploys via `rig_with_check_system_deploy:592`.

   Green Integration Tests jobs do not contradict this. The pin predates the merge of the regression test named in the description both before and after the rebase (`62f752f4` on the run that went green, `a0c6fba` now), so that test is absent from the checkout CI uses; the suite's only report consumer, `test_web_api.py::test_get_event_data`, deploys `@2000!(0)` and so never reaches the `!eval_successful` branch; and the read-only node's per-finalization report pre-cache, which does exercise this path continuously, discards its error into a `debug!`. Items 4 and 5 are what change that.

**Medium**

2. Log the failure where it is now raised — the `?` points in `reporting_casper.rs`, using the `target: "f1r3fly.casper.reporting"` the removed `warn!` carried — and increment a counter on `PostStateMismatch` in `block_report_api.rs`. After this change no `warn!` or `error!` remains in any of the three changed files, and the only metrics there are two `block_report.lock.queue_size` gauges, so the newly-escalated error has no server-side record on the HTTP path. This is also what delivers Major 2's stated goal, which the variant alone does not: `into_service_error` flattens to a single string and HTTP returns the same 500 as `ReplayFailed`, so nothing outside the node can distinguish an integrity mismatch from a replay failure.

3. Add a test that a `PostStateMismatch` leaves the `ReportStore` untouched. This is the property the whole change reduces to, and it currently rests on statement order alone — `block_report_api.rs:185` propagates before the `put` at `:187`, so hoisting the `put` would restore the original bug with every existing test still green.

**Integration wiring before merge**

4. Bump `SYSTEM_INTEGRATION_REF` to `0eed22cfa5d33b08dcdd636314def0acb2d285c6` in both `.github/workflows/_integration-pipeline.yml:47` and `.github/oci-validation.env:17`. The rebase onto `dev` brought this branch to `a0c6fba`, so no further `dev` merge is needed — one move remains. `0eed22c` is the system-integration `main` commit containing the ten new stability tests, this PR's included. On the `old..new` diff the pin comment requires: `a0c6fba..0eed22c` is 18 commits touching only `integration-tests/` and `CLAUDE.md` — nothing under `ci/`, `.github/`, `shardctl/`, `compose/` or `conf/`, so no script the secret-bearing launch job executes changes. The pre-existing suite runs green at that commit (85 of 85, docker provider, this pipeline's own flags and deselect list).

5. Enable `integration-tests/test/tests/custom/test_observer_missing_block_retry.py` — the test named in this PR's description. Until the pin moves, that file does not exist in the checkout CI uses, so the regression this PR fixes has never been exercised by CI and the "94.34 seconds" figure is a local run. After the bump it should report green, and it is the only CI evidence that the fix works.

6. Deselect the other nine new test files, whose node fixes are not yet on `dev`, in the same commit as the bump. This pipeline runs the whole suite, so without them the bump turns nine unrelated failures into this PR's problem.

   | new test | node fix |
   |---|---|
   | `test_duplicate_signed_deploy.py` | #207 |
   | `test_expired_deploy_admission.py` | #208 |
   | `test_transient_peer_liveness.py` | #209 |
   | `test_observer_overload.py` | #211 |
   | `test_finality_stall_recovery.py` | #224 |
   | `test_readonly_catchup_bounded.py` | #225 |
   | `test_concurrent_bridge_locks.py` | #226 |
   | `test_slow_peer_notification.py` | #227 |
   | `test_cold_start_readiness.py` | none open |

   If #207 merges first it will already have added these entries and enabled its own test — in that case drop `test_observer_missing_block_retry.py` from the list rather than re-adding it, and leave `test_duplicate_signed_deploy.py` enabled.

**Low**

7. Drop the two new `get_report().map_err(...)` branches, or reword "report-collection failures" in the description. `collect_report()` (`reporting_rspace.rs:159-168`) returns `Ok(())` unconditionally, so `get_report()` has no error path and both new branches are unreachable. The `unwrap_or_default()` they replace was equally dead, so this is not a regression — only diff volume and a description naming a mode that cannot occur.

jeffrey-l-turner (approved):

Issue #203, opened 2026-08-05
Observer read API (status/explore-deploy/last-finalized-block) degrades under sustained polling despite PR #113 — likely `global_lock` RwLock contention


9Neechan:
**Affected:** f1r3node-rust — primarily `casper::rust::api::block_api::BlockAPI::exploratory_deploy` / `compute_parents_post_state` (unpinned `/api/explore-deploy`'s per-request DAG-merge recompute; confirmed dominant cause via A/B test below); secondarily `block-storage::rust::dag::block_dag_key_value_storage` (`global_lock: Arc<parking_lot::RwLock<()>>`), `node::rust::web::shared_handlers`/`web_api_routes` (`offload()`-wrapped handlers for `/api/status`, `/api/explore-deploy`, `/api/last-finalized-block`); `observer-http-poll` experiment (`experiment/observerpoll`)
**Severity:** medium — no consensus-safety violation observed (network-wide finalized spread and observer-vs-validator ordering checks both pass in every run), but the observer's read API becomes significantly slower/less reliable under sustained load, and the observer's own finalization visibly and continuously falls behind the validators under moderate load

---

## Background: this is a follow-up to #73, not a duplicate

Issue #73 ("Multi-threaded HTTP runtime for observer API") was root-caused to `servers_instances.rs` running the entire `:40403` HTTP server on a single OS thread via `tokio::runtime::Builder::new_current_thread()`. That was merged in PR #113 (2026-07-15, `fix/global_lock_mutex_staging`), which:

- switched the `:40403` HTTP server to `new_multi_thread().worker_threads(4)`
- introduced an `offload()` helper wrapping every handler's blocking storage call in `tokio::task::spawn_blocking`, applied to `status_handler`, `explore_deploy_handler`, `explore_deploy_by_block_hash_handler`, and 20 other routes

**Confirmed the tested image includes this fix.** The image was built from `dev` at commit [`aff2428b`](https://github.com/F1R3FLY-io/f1r3node-rust/commit/aff2428b6d3c3314b870189f9bdf6fd2154a8dbe) (2026-08-04). `git merge-base --is-ancestor <PR#113 merge commit f1faec19> aff2428b` returns true, and the file content at `aff2428b` shows both the `new_multi_thread()` runtime and the `offload()` helper in place. So what follows is **not** the literal bug described in #73 — it reproduces under a build that already contains that fix.

## What did you expect to see?

[...]

PR #211, opened 2026-08-07
fix(observer): bound exploratory deploy load

a-k-l-sdao:
## Summary

- Bound concurrent exploratory deploy execution on read-only observers.
- Reject excess work immediately with an explicit busy response.
- Abort and join timed-out exploratory tasks before releasing capacity.
- Apply a shared five-million-phlo hard execution limit and configurable best-effort wall-clock timeout.
- Execute unpinned exploratory reads against the current LFB state.

Refs #203

This addresses the unbounded exploratory-query load and observer starvation described in #203. It does not claim to resolve every independent read-path or DAG-lock performance concern in that issue.

## Regression coverage

`integration-tests/test/tests/custom/test_observer_overload.py::test_observer_exploratory_overload_recovers`

The regression runs one expensive exploratory query plus eight concurrent excess queries, requires bounded rejection of excess work, checks the observer memory ceiling, and verifies that normal exploratory queries recover afterward.

## Runtime behavior

- Capacity rejection is immediate; there is no request wait queue.
- The phlo limit is the authoritative hard execution bound.
- The wall-clock timeout is best-effort for cooperative work; capacity remains held until cancellation completes.
- Busy responses use HTTP 503 and timeout responses use HTTP 504.

## Validation

- Baseline `dev`: all eight excess requests returned HTTP 200 and observer memory reached approximately 4.63 GB.
- This branch: all excess requests returned HTTP 503, observer peak memory stayed approximately 1.14 GB, and a subsequent query returned HTTP 200.
- Permit, timeout-abort/join, HTTP mapping, configuration, and exploratory API regressions passed.
- Strict Clippy for `casper` and `node` passed.

## Dependency


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR adds sensible load-bounding to exploratory deploys on read-only observers: a configurable semaphore with queue timeout, an execution deadline, a reduced phlo cap, allocator trimming, metrics, and clean HTTP 503/504 mappings with unit tests for permit semantics and error classification. The implementation is largely solid and validated by a regression test, but there are three noteworthy behavioral/robustness concerns: (1) exploratory reads now target the last finalized block instead of the merged DAG-tip state, introducing user-visible staleness (including for dev_mode validators); (2) the tokio timeout is only effective if the interpreter future yields, and cancellation safety of dropping a mid-flight execution should be verified; (3) the default phlo limit drops 20x, which can break existing expensive queries and is tunable only via undocumented env vars. Minor items include a cancellation-unsafe metrics gauge, per-request malloc_trim overhead, and test module placement. Safe to merge once the semantic changes are confirmed as intended and documented.

The change introduces useful execution concurrency and cost controls and correctly maps busy and timeout errors to explicit HTTP responses. However, the admission path still has an unbounded waiter queue, and the Tokio timeout does not by itself guarantee that CPU-bound or cancellation-unsafe interpreter work terminates. Those properties should be addressed or demonstrated before relying on this implementation as a hard memory and execution bound.

This change correctly attacks observer overload with shared admission control (semaphore + short queue wait), cheaper LFB-only reads, a much lower default phlo ceiling, and explicit 503/504 mapping—matching the reported regression win (503s and ~1.14GB peak). The main residual risks are that tokio::time::timeout is cooperative and may not stop CPU-bound Rholang or hold the permit for full cleanup, so the timeout/busy guarantees can be weaker than the API suggests under pathological terms; phlo limiting is doing the heavy lifting. Smaller follow-ups: fix metrics accuracy, panic-safe active gauge, OpenAPI placement, and document the LFB semantic break for no-block-hash exploratory reads. Safe to merge with those caveats if the intentional LFB behavior is accepted.

The code changes introduce necessary bounds and timeouts for exploratory deploys, improving the system's resilience under load. The implementation is correct and the new error types are handled appropriately. Minor documentation improvements and a potential performance optimization suggestion are noted.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Semantic change: exploratory deploys now read stale finalized state instead of merged tip state</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1672)

On timeout the future is dropped and _permit is released immediately when the handler returns. If interpreter/runtime cleanup is non-trivial or cancellation is delayed until the next await, a subsequent request can start while prior exploratory work is still releasing memory. That weakens the concurrency/memory ceiling the semaphore is meant to enforce.

---

Previously, when no block hash was specified, the exploratory deploy computed a merged post-state from all DAG tips (with late-block filtering disabled), giving clients visibility into the latest unfinalized state. The new code reads directly from the last finalized block's post-state. This means observers will return data that lags behind the chain tip by the finalization delay, which is a user-visible behavioral change. Clients that deploy and then immediately explore state (a very common pattern in tests and dApps) may see stale results until finalization catches up. This also applies to validators running in dev_mode, which previously saw merged tip state.

---

The previous merged-tips / compute_parents_post_state path is removed. Queries that omitted block_hash and depended on seeing unfinalized merged tip state will silently change behavior. This is called out in the PR summary and helps bound load, but it is an API semantics change.

**Suggestion:**
```
Call out the LFB-only behavior in API/OpenAPI docs and release notes. If tip-merged reads are still required, gate them behind an explicit parameter so the expensive path is opt-in rather than accidental.
```

</details>

<details>
<summary><b>:red_circle: Critical: tokio::time::timeout may not cancel CPU-bound Rholang execution</b> (Reported by: anthropic, openai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1719)

tokio::time::timeout is cooperative: it cannot interrupt work while the inner future is executing without yielding, and dropping the future only stops underlying work if the interpreter path is cancellation-safe and does not leave blocking or spawned work running. Expensive Rholang reduction may therefore exceed the advertised deadline or continue consuming resources after a 504. If underlying work survives cancellation, the semaphore permit is released and another execution can start, defeating the concurrency bound.

---

tokio::time::timeout only takes effect when the wrapped future yields to the executor. If play_exploratory_deploy performs long synchronous CPU-bound interpretation without .await points (or runs on a blocking section), the timeout cannot fire until the future next yields, meaning a pathological term could still hold the permit and CPU well past the 15s deadline. Additionally, when the timeout does fire, the future is dropped mid-execution; the cancellation safety of the spawned runtime and RSpace state under mid-execution drop should be verified (e.g., partially written soft checkpoints, leaked runtime resources).

**Suggestion:**
```
Verify and document that the complete exploratory execution path yields and is cancellation-safe. Prefer enforcing the deadline inside the interpreter/reduction loop, or execute queries in an abortable isolated worker and retain the permit until the worker has actually terminated. Add a test proving that timed-out work stops consuming execution capacity before a subsequent request is admitted.
```

</details>

<details>
<summary><b>:orange_circle: Major: Execution timeout may not preempt CPU-bound interpreter work</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1720)

tokio::time::timeout only cancels at await yield points. If play_exploratory_deploy runs a long CPU-bound Rholang evaluation without polling the runtime, the 15s deadline will not fire, the task can stall the executor, and the semaphore permit stays held until the work eventually returns. The phlo limit is the real bound here; the timeout path may rarely produce ExploratoryDeployTimeoutError under actual overload.

**Suggestion:**
```
Run exploratory evaluation on a cancellable blocking pool / dedicated task with an explicit abort signal checked by the interpreter, or document that phlo_limit is the authoritative cap and treat execution timeout as best-effort. At minimum, ensure the eval path yields periodically or runs via spawn_blocking so the timeout can fire and release the permit promptly.
```

</details>

<details>
<summary><b>:orange_circle: Major: Default exploratory phlo limit reduced 20x (100M -> 5M) — breaking for expensive queries</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 168)

The effective default phlo limit for exploratory deploys via RuntimeManager drops from 100,000,000 to 5,000,000 (EXPLORATORY_DEPLOY_PHLO_LIMIT_DEFAULT). Any existing exploratory query legitimately consuming between 5M and 100M phlo will now fail with out_of_phlogistons. This is an intentional load-bounding measure per the PR description, but it silently changes behavior for existing clients and is only tunable via an undocumented environment variable.

**Suggestion:**
```
Document the new default and the F1R3_EXPLORATORY_DEPLOY_PHLO_LIMIT env var in operator/API documentation, and consider whether the response for out-of-phlo on exploratory queries clearly directs users to the limit change.
```

</details>

<details>
<summary><b>:orange_circle: Major: Semaphore wait queue remains unbounded under load</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 329)

The semaphore bounds active executions but not the number of requests waiting for a permit. Every request may remain queued for up to the configured two-second timeout while retaining its handler task, parsed term, and request-related allocations. A sufficiently high request rate or large terms can therefore still cause memory usage proportional to incoming load, contrary to the stated goal of keeping observer memory bounded.

**Suggestion:**
```
Reject immediately with try_acquire_owned(), or add a second bounded admission mechanism that caps the combined number of executing and queued requests. If queueing is required, reserve a bounded queue slot before retaining or parsing large request data.
```

</details>

<details>
<summary><b>:orange_circle: Major: Metrics gauge increment/decrement not cancellation-safe</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1714)

The active gauge is incremented before awaiting execution and decremented only on the normal continuation after the await. If the request or handler future is cancelled, such as after a client disconnect, control may never reach the decrement and the gauge will permanently over-report active executions.

---

The exploratory_deploy.active gauge is incremented and later decremented with plain statements. If the surrounding future is cancelled between the two (e.g., outer task cancellation, client disconnect propagating through the handler), the gauge will permanently leak an increment, making the 'active' metric drift upward over time.

---

exploratory_deploy.active is incremented then decremented manually. A panic between those points leaves the gauge permanently elevated until process restart.

**Suggestion:**
```
Use a small RAII guard (or scopeguard) that increments on create and decrements on Drop so panic and early-return paths stay balanced.
```

</details>

<details>
<summary><b>:orange_circle: Major: Test module placed at top of file, before the struct under test</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 65)

The #[cfg(test)] mod tests block is inserted between MergeableKey and the RuntimeManager struct definition. Conventionally in this codebase (and Rust generally), test modules live at the bottom of the file, and placing it mid-file hurts readability.

---

The exploratory permit test module is inserted above the RuntimeManager struct. It works, but is harder to discover than end-of-file cfg(test) modules used elsewhere.

**Suggestion:**
```
Move the tests module to the bottom of the file for consistency.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: completed metric counts failed exploratory executions</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1738)

exploratory_deploy.completed is incremented for any Ok(timeout result), including when play_exploratory_deploy returns Err. That conflates interpreter/validation failures with successful completions and makes success-rate dashboards misleading.

**Suggestion:**
```
Increment completed only after result is Ok, and add a separate exploratory_deploy.failed counter for Err cases.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: trim_allocator called on every exploratory deploy</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1735)

RuntimeManager::trim_allocator() (malloc_trim on Linux) is invoked after every single exploratory deploy, including cheap ones. malloc_trim can be relatively expensive and cause allocator contention under a stream of small queries, adding latency to each request.

**Suggestion:**
```
Consider trimming only after expensive executions (e.g., cost above a threshold, or elapsed time above a threshold), or on a periodic timer, rather than on every request.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Missing documentation for new constants</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 83)

The new constants for exploratory deploy limits should be documented to explain their purpose and default values.

**Suggestion:**
```
Add comments above the constants to describe their usage and default values.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: New tuning environment variables are undocumented</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 161)

Four new operator-facing environment variables are introduced (F1R3_EXPLORATORY_DEPLOY_MAX_CONCURRENT, F1R3_EXPLORATORY_DEPLOY_QUEUE_TIMEOUT_MS, F1R3_EXPLORATORY_DEPLOY_PHLO_LIMIT, F1R3_EXPLORATORY_DEPLOY_EXECUTION_TIMEOUT_MS) with no accompanying documentation in the diff. Operators tuning observer capacity have no discoverable reference for these knobs or their defaults.

**Suggestion:**
```
Add these variables and their defaults to the node configuration/operations documentation.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: 503/504 documented on deploy_handler as well as explore_deploy</b> (Reported by: xai)</summary>

**File:** `node/src/rust/web/shared_handlers.rs` (line 465)

OpenAPI annotations for observer_busy and exploratory_timeout were added to deploy_handler, but those error types are produced on the exploratory path. This can mislead API consumers about normal deploy responses.

**Suggestion:**
```
Keep 503/504 only on explore_deploy_handler unless deploy genuinely shares this admission-control path.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Busy error message assumes observer context</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 199)

ExploratoryDeployBusyError's message says 'Observer is busy processing another exploratory query', but the same path is reachable on a validator running in dev_mode, where 'observer' is misleading.

**Suggestion:**
```
Use wording that covers both cases, e.g., 'Node exploratory query capacity is exhausted; retry later'.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: timeout_ms as u128 is unnecessarily wide</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 206)

ExploratoryDeployTimeoutError stores timeout_ms as u128 (from Duration::as_millis). A u64 is more than sufficient for any realistic timeout and integrates better with serialization and metrics APIs.

**Suggestion:**
```
Use u64 and convert with u64::try_from(timeout.as_millis()).unwrap_or(u64::MAX) or store the Duration itself.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Consider using a more efficient data structure for latest messages</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1664)

The current implementation uses a BTreeMap to store latest messages. Depending on the size and frequency of updates, a more efficient data structure might be beneficial.

**Suggestion:**
```
Evaluate the use of a HashMap or another data structure that offers better performance for the specific use case.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Divergent default phlo limits between RuntimeOps and RuntimeManager</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/rholang/runtime.rs` (line 836)

RuntimeOps::play_exploratory_deploy still hardcodes 100_000_000 while RuntimeManager defaults to 5_000_000 via env. Direct RuntimeOps callers get a very different cost ceiling than the observer API path.

**Suggestion:**
```
Route all exploratory calls through the configurable limit, or make the RuntimeOps wrapper take the limit from the same constants/env helpers to avoid accidental bypass.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Consider Retry-After on observer busy responses</b> (Reported by: xai)</summary>

**File:** `node/src/rust/web/shared_handlers.rs` (line 232)

503 observer_busy is the right status for admission control, but clients have no machine-readable backoff hint tied to exploratory_deploy_queue_timeout.

**Suggestion:**
```
Add a Retry-After header (or include retry_after_ms in the JSON body) derived from the queue timeout default/config.
```

</details>
---

<details>

This PR adds sensible load-bounding to exploratory deploys on read-only observers: a configurable semaphore with queue timeout, an execution deadline, a reduced phlo cap, allocator trimming, metrics, and clean HTTP 503/504 mappings with unit tests for permit semantics and error classification. The implementation is largely solid and validated by a regression test, but there are three noteworthy behavioral/robustness concerns: (1) exploratory reads now target the last finalized block instead of the merged DAG-tip state, introducing user-visible staleness (including for dev_mode validators); (2) the tokio timeout is only effective if the interpreter future yields, and cancellation safety of dropping a mid-flight execution should be verified; (3) the default phlo limit drops 20x, which can break existing expensive queries and is tunable only via undocumented env vars. Minor items include a cancellation-unsafe metrics gauge, per-request malloc_trim overhead, and test module placement. Safe to merge once the semantic changes are confirmed as intended and documented.

The code changes introduce necessary bounds and timeouts for exploratory deploys, improving the system's resilience under load. The implementation is correct and the new error types are handled appropriately. Minor documentation improvements and a potential performance optimization suggestion are noted.

#### openai (gpt-5.6-sol)

The change introduces useful execution concurrency and cost controls and correctly maps busy and timeout errors to explicit HTTP responses. However, the admission path still has an unbounded waiter queue, and the Tokio timeout does not by itself guarantee that CPU-bound or cancellation-unsafe interpreter work terminates. Those properties should be addressed or demonstrated before relying on this implementation as a hard memory and execution bound.

#### xai (grok-4.5)

This change correctly attacks observer overload with shared admission control (semaphore + short queue wait), cheaper LFB-only reads, a much lower default phlo ceiling, and explicit 503/504 mapping—matching the reported regression win (503s and ~1.14GB peak). The main residual risks are that tokio::time::timeout is cooperative and may not stop CPU-bound Rholang or hold the permit for full cleanup, so the timeout/busy guarantees can be weaker than the API suggests under pathological terms; phlo limiting is doing the heavy lifting. Smaller follow-ups: fix metrics accuracy, panic-safe active gauge, OpenAPI placement, and document the LFB semantic break for no-block-hash exploratory reads. Safe to merge with those caveats if the intentional LFB behavior is accepted.
</details>

---

jeffrey-l-turner:
@a-k-l-sdao — from an open-issue audit (mapping issues to addressing PRs): could you add explicit `Fixes #N` / `Refs #N` lines to this PR's description for the issues it addresses? Cross-referencing had to be reconstructed thematically because the linkage is missing — and note closing keywords only auto-close when a PR merges to the DEFAULT branch, so dev-targeted fixes need a manual close at promotion time either way. Candidate issues for this PR (please confirm/correct): #203 (observer read API degrades under sustained polling).

a-k-l-sdao:
Updated in `a165ab34`: immediate capacity rejection, abort-and-join timeout handling, consistent phlo limits, accurate lifecycle metrics, configuration coverage, and clarified timeout semantics. The PR now references #203 with its partial scope stated explicitly. Targeted tests, formatting, and strict Clippy pass.

spreston8 (commented):
Reviewed at `2964d2388` against `dev`. Suggestions below.

This looks like the right shape, and the documentation is better than the PR description suggests — the visibility change and the best-effort nature of the deadline are both stated accurately in `defaults.conf`, `api-reference.md` and `exploratory-deploy.md`. Two things are worth correcting in the earlier review rather than leaving on the record: its Critical about a timed-out request releasing capacity while work is still in flight does not apply, because `await_exploratory_deploy_task` does `abort()` then `task.await` — a real join — and the permit is owned inside the spawned task, so capacity cannot be freed early. The join is what makes the bound hold. And the metrics concern is addressed properly rather than papered over: `ExploratoryDeployMetrics` is an RAII guard inside the task, so the gauge decrement and terminal counter run on unwind including abort.

**Medium**

1. Consider reading from the finalized floor rather than the LFB. Deleting the merged-tip path is the right call for an unbounded read path, but the LFB is node-local finality state, whereas `floor.rs` notes that "no node-local finality state participates" in the floor and that `finalized_floor` "stays node-identical". So two observers can answer the same unpinned query with different state under the LFB, and cannot disagree about the floor for the same block — which is the property a read API serving finalized data most needs. `floor.rs` also describes the floor as "the cut the block's merge builds on" — the base a merge is defined against — which a finalized block's recorded post-state is not, so this looks like the stronger notion of finalized rather than merely a different one.

   It looks implementable now and cheaper than what it replaces: `finalized_floor(dag, parents, latest_messages, ftt)` is on `dev` (`casper/src/rust/finality/floor.rs:86`), the floor cache is already persisted, and the code being deleted here already fetched `get_snapshot()` and `latest_messages` for the merger — so keeping the snapshot, computing the floor and dropping the merge gets the load win with node-identical state.

2. Worth pinning the new read contract, and fixing a comment that now contradicts it. `casper/tests/api/exploratory_deploy_api_test.rs` is not in this PR's changed files but exercises the rewritten path, and still reads `None, // block_hash: None means use current DAG tips` — which is no longer what `None` means. More substantively, `exploratory_deploy_should_get_data_from_read_only_node` deploys and then reads back through an unpinned query, so it passes only because finalization catches up within a 4-validator shard: a timing-dependent pass on exactly the semantics this PR changes. The contract is documented in three places and asserted in none; a test that deploys without finalizing and asserts the effect is *not* visible would state it as an assertion instead of prose.

**Low**

3. A migration note for the phlogiston ceiling would help. It was `100 * 1000 * 1000` (`runtime.rs:838`) and is now 5,000,000 — a 20× reduction. The docs give the new default but do not flag it as a cut, so a query between 5M and 100M stops working with nothing in the release notes to explain it.

4. Consider rejecting non-positive configuration instead of clamping it. `ExploratoryDeployConfig::new` applies `phlo_limit.max(1)` and `execution_timeout.max(1ms)`, so a `0` becomes 1 phlogiston or 1 ms — a node that answers nothing on this endpoint, silently. The existing `--readonly-history-blocks` check is the precedent for failing closed on a non-positive value.

5. Three smaller items: the 503 has no `Retry-After` (nothing in `node/` sets it), which with a default cap of 1 and capacity held until termination is the header most likely to prevent a retry storm; the timeout error string is now the only place asserting a deadline the code does not enforce as a response bound, where `defaults.conf` and both docs state it correctly; and `RuntimeOps::play_exploratory_deploy` uses the hardcoded 5M constant while `RuntimeManager::play_exploratory_deploy` uses the configured value, so a future caller reaching for the wrong one bypasses configuration silently.

**A note rather than a request**

On the gRPC side, `exploratory_deploy` returns `Ok(Response{…Error})` and maps through `into_service_error()`, so gRPC clients get one prose string where HTTP clients get `observer_busy` / `exploratory_timeout`. Worth noting that `Status::unavailable` and `Status::deadline_exceeded` express these two conditions exactly and need no proto change — the transport's own status channel is simply unused here. Arguably its own change rather than this PR's.

spreston8 (approved):
Reviewed at `82b42d079` against `dev`. Approving.

The suggestions from the earlier review are addressed on this branch, with one deferral, one decision worth recording, and two corrections.

**Not addressed here: the read basis.** `casper.last_finalized_block()` is not a pointer read — it runs `Finalizer::run` with a `record_directly_finalized` write, and the API path carries none of the rate-limit or in-progress guards the admission path has. A floor-derived read would take both off this path. That is larger than this PR should carry and is left for separate work; deleting the merged-tip path stands on its own.

**Supersedes the exploratory-deploy portion of #224.** That PR declared this one as a dependency and merged ahead of it, carrying this branch's first commit without the hardening that followed it. Merging `dev` here conflicts in `block_api.rs` and `runtime_manager.rs`; both resolve to this branch — HOCON and CLI configuration rather than `F1R3_EXPLORATORY_DEPLOY_*` env vars, the metrics guard owned inside the execution task so it runs on unwind including abort, immediate rejection rather than a 2000 ms queued acquire, and the finalized-only unpinned read. The superseded variant is removed rather than left unreferenced. Nothing outside that overlap is touched.

**The phlogiston ceiling stays at 5,000,000.** The earlier review of #224 raised the reduction from 100,000,000 and `dev` currently carries the larger value. Measured on this branch: one exploratory query running a memory-hostile term peaked at 3793 MB, 4969 MB and 5010 MB across three runners, against an observer ceiling of 1500 MB, with excess requests still rejected — so that is one query's cost rather than contention. The endpoint is unauthenticated and the term is caller-supplied, so this ceiling bounds arbitrary submitted work and the default has to hold against the worst term. Operators needing a heavier budget can raise `api-server.exploratory-deploy-phlo-limit` or the matching CLI flag, which this PR makes reachable for the first time.

**Two corrections to the earlier review.** Item 4 cited `--readonly-history-blocks` as the precedent for failing closed on a non-positive value; that flag does not exist in this repo, and the precedent is `parse_non_negative_i64` (`node/src/rust/configuration/commandline/options.rs:25`). Item 1's node-identity claim holds only when the floor's inputs are a block's own frozen parents and justifications, not the live tips an ad-hoc query must use, and its block-versus-state finality distinction does not hold, since the floor's state is also a block's recorded post-state.

PR #216, opened 2026-08-09
feat(cost-accounting): internalize Rholang metering with formal verification

dylon:
## Summary

This PR replaces external Rholang cost tracking with the cost-accounted Rho calculus: an internalized metering model in which each billable `COMM` event—a matching send/receive reduction—consumes authorized phlogiston from a signature-scoped supply.

The new `RuntimeBudget` and `MeteredMachine` remove the legacy `ChargingRSpace` precharge/refund path, preserve deterministic consensus cost under concurrent execution, and retain stable error aggregation. Existing Rholang source remains valid because metering is introduced by the compiler and runtime rather than through new required surface syntax.

The implementation is accompanied by mechanized proofs, bounded model checking, property tests, fuzz targets, concurrency models, threat analysis, and explicit documentation of the remaining implementation-refinement boundary.

### Runtime metering

- Introduces signature, token, signed-process, resource-logic, and runtime-budget abstractions.
- Recursively meters normalized Rholang reductions while preserving concurrent evaluation.
- Replaces broad RSpace tuple-space charging and per-`COMM` refund bookkeeping with source-token consumption.
- Reconciles schedule-dependent runtime events into a deterministic total cost and out-of-phlogiston verdict.
- Keeps per-operation trace digests diagnostic-only; consensus uses total cost, execution status, and post-state.
- Deterministically rejects supply overflow instead of panicking.

### Funding and settlement

- Removes per-deploy `phloPrice` and `phloLimit`; cost is now the number of billable source-token consumptions.
- Funds deploys from signature-scoped supply pools checked by the block-assembly acceptance gate.
- Prevents cross-group over-admission, unauthorized weakening, and replay-side supply divergence.
- Adds genesis and epoch phlogiston configuration, client funding allocations, fee extraction, and the blessed exchange path.
- Adds replay-symmetric close-block settlement, slashing effects, and validator redemption.

### Signatures and deploy interfaces

- Adds canonical multi-signature deploy envelopes with signature verification, public-key ordering, duplicate rejection, and mixed signature algorithms.
- Supports threshold signatures and the richer signature algebra used by token splitting, joining, delegation, and capabilities.
- Extends HTTP, gRPC, protobuf, proposer, execution, persistence, and replay paths for cosigned deploys.
- Preserves legacy single-signature behavior through a one-signer cosigned envelope.
- Reserves removed protobuf tags so legacy phlogiston fields cannot be reused with conflicting meanings.
- Pins `rholang-parser` to the published revision containing the required cost-accounting syntax support.

### Consensus and security integration

- Preserves play/replay equivalence across funding, settlement, merge, recovery, and system-deploy execution.
- Hardens deterministic merge ordering, finalized-floor selection, checked numeric-channel merging, and fork-choice failure handling.
- Prevents finalized deploy reproposal and exercises cost-accounted recovery through genuine value-draining conflicts.
- Eliminates observation-order-dependent equivocation-record pollution during unbonded validator windows.
- Strengthens slash authorization, current-evidence requirements, and redemption quorum verification.

## Verification

The formal suite separates unbounded mathematical claims from bounded executable models:

- Rocq proofs cover token conservation, cost and step determinism, funded-fragment normalization and confluence, gate safety, translation properties, settlement, slashing composition, and typed mergeable-channel accounting.
- TLA+ and Apalache models explore scheduling, compound signatures, bounded runtime reconciliation, replay attacks, admission, and merge behavior.
- Lean, SageMath, Loom, Verus, Tamarin, ProVerif, Why3, Isabelle/HOL, Iris, mCRL2, and rewriting-system artifacts provide independent cross-witnesses.
- Rust unit, integration, property, concurrency, fuzz, and benchmark coverage exercise the implementation-facing correspondence.

The branch verification campaign recorded:

- 3,424 of 3,424 `nextest` tests passing after integration with `dev`.
- Workspace check, formatting, Clippy with warnings denied, doctests, and fuzz-workspace compilation passing.
- The cost-accounting umbrella gate passing all 15 legs.
- Finalized-floor, fork-choice, merge-algebra, deploy-lifecycle, and slashing verification gates passing.
- Rocq kernel checks, TLA+ configurations, TLAPS proofs, and unbounded Apalache invariants passing.
- Citation audits validating 125 Rocq references and 167 source references.

CI now includes the pure evaluator and cost-accounting Loom model crates and derives the slashing axiom-freedom checks from the complete theorem set.

## Compatibility and rollout

- Existing Rholang contracts require no source changes.
- Deploy clients must migrate away from `phloPrice` and `phloLimit` and may optionally submit cosigners or a signature-algebra envelope.
- Single-signature deploys remain supported.
- The economic state changes are intended for a coordinated fresh-genesis or explicit protocol-upgrade boundary.
- Strict funding enforcement remains configurable for staged activation.

## Documentation

The rationale, semantics, security assumptions, decisions, and proof boundaries are covered in:

- [Migration design](docs/theory/cost-accounting-migration.md)
- [Formal-verification dossier](docs/theory/cost-accounted-rho-verification.md)
- [Threat model](docs/theory/cost-accounting-threat-model.md)
- [Decision records](docs/theory/cost-accounting-decision-records.md)
- [Conformance property catalog](docs/theory/cost-accounting-conformance-properties.md)

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/216"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1788891649&installation_model_id=426883&pr_number=216&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F216&signature=f2d932781d926a33097fa6620ed7b850f0f97e1d07e170dc5bc1b96f565ee139"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :rotating_light: Critical Vulnerabilities Found (25.0000% agreement)

**Reviewed by:**

| openai | `gpt-5.6-sol` | Critical Vulnerabilities | 0.93 |

---

### Summary

This is a very large (677 files, +123k lines) consensus-critical rework of Rholang metering that replaces per-deploy phlo pricing with an internalized, signature-scoped funding gate, adds multi-signature deploy envelopes, and introduces a deploy-occurrence index with source-aware rejection tracking. The visible code is carefully commented and accompanied by property tests, but several substantive concerns remain: the transitional gate mode admits all deploys unenforced/undebited until pools are provisioned (a DoS window compounded by removing submission-side phlo checks); the in-memory cosigner-metadata sidecar risks play/replay envelope divergence after restarts and appears to lack persistence/eviction; deploys in invalid blocks are silently dropped from all indices; and the deleted processed-vs-rejected packaging invariant (and its test) needs to be verified as enforced elsewhere. There are also lock-ordering and per-deploy write-amplification concerns in the new DAG index path, and notable code duplication in the cosigned deploy API path. Given the scale, the consensus-sensitivity, and the fact that 666 files (including the acceptance gate, replay, and crypto envelope implementations that many of these observations depend on) were omitted from the diff, this needs thorough human review by maintainers familiar with the play/replay and merge invariants before merging to dev.

The change introduces substantial consensus-critical behavior, but the shown implementation has two blocking security problems: compound deploy authorization can be silently downgraded after restart because cosigner metadata is non-persistent, and funding enforcement is fail-open by default despite removal of legacy phlogiston limits. The new occurrence index also permits unbounded quadratic storage work, and one API path hides storage failures. These issues should be addressed before merge, particularly by persisting complete cosigned envelopes and making metering enforcement fail closed.

This is a very large consensus-and-economics change: internalized Rholang metering, multi-sig envelopes, deploy occurrence indexing, and a block-assembly funding/settlement gate. The visible Rust core is thoughtfully structured around play/replay symmetry and deterministic ordering, but several high-impact risks remain: nested locking on the DAG deploy-index path, in-memory cosigner sidecar reconstruction, removal of rejected-vs-processed filtering without a clear local replacement, hard failures on ambiguous deploy disposition, and a default-off strict funding mode that leaves cost accounting economically unenforced. Because most admission, supply, replay, and merge implementations are outside the provided diff, this should not be merged on the partial review alone; it needs targeted human review plus play/replay and funding-gate integration tests before approval.

The code change introduces significant modifications to the Rholang metering and cost accounting system. While the changes seem to be well-designed and thoroughly tested, there are a few potential security and logic issues that need to be addressed. Additionally, there is a minor style issue related to inconsistent naming conventions. Overall, the code change requires further review and testing before it can be merged.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Missing in-memory cosigner metadata silently downgrades compound deploys</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 1095)

Multi-signature metadata is stored only in the in-memory pending_cosigner_metadata HashMap, while the primary Signed<DeployData> is stored separately in persistent deploy storage. After a node restart, the map is initialized empty and reconstruct_cosigned therefore falls back to Cosigned::from_single_signer. A previously admitted compound or threshold-signed deploy can consequently be proposed using only its primary signer, changing its signature algebra, funding-pool identity, and authorization semantics. This is an authentication weakening and can also cause play/replay or cross-node divergence when one node retained the sidecar and another restarted.

**Suggestion:**
```
Persist the complete canonical Cosigned<DeployData> envelope atomically with the deploy instead of using an in-memory sidecar. At minimum, persist a marker that a deploy was compound and fail closed when its cosigner metadata is unavailable; never reinterpret missing metadata as proof that the deploy was originally single-signed.
```

</details>

<details>
<summary><b>:red_circle: Critical: Funding enforcement defaults to fail-open after legacy limits are removed</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/casper_conf.rs` (line 178)

strict_funding_enforcement defaults to false, and the documented transitional behavior admits deploys with absent signature supply pools without enforcement or settlement debit. An attacker can continually create fresh signing identities whose pools are absent and submit unfunded deploys. Because this change also removes per-deploy phlo price and limit admission checks, the default configuration lacks the intended economic authorization boundary and exposes validators to free computation and storage exhaustion.

**Suggestion:**
```
Default new cost-accounted shards to strict enforcement and treat absent pools as zero. If backward compatibility requires a transition period, retain an independently enforced legacy execution/resource cap during that period and require an explicit, consensus-bound activation configuration rather than making fail-open behavior the default.
```

</details>

<details>
<summary><b>:red_circle: Critical: Deploys in invalid blocks are no longer indexed at all</b> (Reported by: anthropic, xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1126)

Deploy index and occurrence index updates are now skipped entirely when `invalid` is true. Previously deploys from invalid blocks were still indexed. Lookups/status for deploys that only landed in invalid blocks will now miss them, which may be intended but changes API observability and any repair/debug flows.

---

The deploy index update is now wrapped in `if !invalid`, so deploys contained in invalid blocks are excluded from both `deploy_index` and the new `deploy_occurrence_index`. Previously they were indexed unconditionally. Any consumer relying on `lookup_by_deploy_id` to locate a deploy that landed only in an invalid block (e.g., duplicate-deploy detection, client status queries) will now report the deploy as unknown. If this is an intentional semantic change it should be explicitly documented; if not, it is a regression.

**Suggestion:**
```
Document the new invariant and add tests for invalid-only deploy lookup. If diagnostics still need them, write occurrences even for invalid blocks while keeping canonical `deploy_index` valid-only.
```

</details>

<details>
<summary><b>:red_circle: Critical: Removal of filter_unprocessed_rejected_deploys drops a packaging invariant</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 3360)

`filter_unprocessed_rejected_deploys` (which guaranteed a deploy sig never appears in both `deploys` and `rejected_deploys` of the same block) was deleted, along with its regression test `processed_deploys_are_not_packaged_as_rejected_deploys`. The new `rejected_deploys: Vec<RejectedDeploy>` now flows straight from `checkpoint_data` into `package_block`. Unless the checkpoint path now enforces this invariant internally (not visible in this truncated diff), a block could package a deploy as both processed and rejected, confusing the source-aware finalization-status resolver and validators.

---

`filter_unprocessed_rejected_deploys` and the post-checkpoint filter that dropped rejected sigs already present in `processed_deploys` were deleted. `package_block` now takes full `RejectedDeploy` values directly. If upstream still emits a sig both as processed and rejected, blocks can carry contradictory disposition records and break finalization/status logic.

**Suggestion:**
```
Confirm the invariant is enforced in `compute_deploys_checkpoint_cosigned` and re-add an equivalent regression test at the packaging boundary; otherwise restore the filter adapted to the new `RejectedDeploy` struct.
```

</details>

<details>
<summary><b>:orange_circle: Major: Nested lock acquisition while holding deploy-index write locks</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1135)

While holding write locks on both `deploy_index` and `deploy_occurrence_index`, the insert path re-acquires `block_metadata_index.read()` per deploy to compare heights. If any other path takes `block_metadata_index` first and then a deploy-index lock, this is a deadlock hazard. Even without deadlock, serializing metadata reads under two write locks lengthens the critical section on a hot DAG-insert path.

**Suggestion:**
```
Read existing deploy→hash mappings and the needed block heights into locals before taking the deploy-index write locks (or take locks in a single documented global order and batch height lookups outside the per-deploy loop).
```

</details>

<details>
<summary><b>:orange_circle: Major: Per-deploy read-modify-write loop replaces batched put; nested lock acquisition</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1135)

The previous code did a single batched `put(deploy_entries)`. The new code performs, per deploy: a `get_one` on deploy_index, a `block_metadata_index.read()` acquisition + metadata fetch, a `put_one` on deploy_index, a `get_one` + full-set deserialize + insert + full-set re-serialize `put_one` on deploy_occurrence_index — all while holding BOTH the deploy_index and deploy_occurrence_index write locks. For blocks with many deploys this is O(n) round-trips with repeated whole-BTreeSet serialization, and acquiring `block_metadata_index.read()` while holding two other write locks creates a lock-ordering hazard if any other path acquires these locks in a different order.

**Suggestion:**
```
Batch reads/writes where possible, hoist the `block_metadata_index.read()` acquisition out of the loop, and document/verify a global lock-ordering convention (metadata → deploy_index → occurrence_index) across all call sites.
```

</details>

<details>
<summary><b>:orange_circle: Major: Deploy occurrence index has unbounded quadratic write amplification</b> (Reported by: openai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1161)

Every new occurrence reads, deserializes, clones, extends, serializes, and rewrites the entire BTreeSet<BlockHashSerde> stored under a deploy ID. A deploy repeated across many competing blocks therefore costs O(n) per insertion and O(n^2) total bytes processed, while holding both deploy-index write locks. The value also grows without an apparent finalization or retention bound, creating a storage and lock-contention denial-of-service vector.

**Suggestion:**
```
Store occurrences as individually keyed records such as (DeployId, BlockHash), with an index or prefix scan, and prune or compact entries once their disposition is finalized. Avoid holding both index locks during potentially large value serialization and persistent writes.
```

</details>

<details>
<summary><b>:orange_circle: Major: Potential security issue in deploy_cosigned function</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 337)

The deploy_cosigned function validates deploys based on the primary signer's fields. However, it does not explicitly check if the cosigners' signatures are valid or if they have been tampered with. This could potentially allow an attacker to submit a deploy with invalid cosigners, bypassing the validation checks.

**Suggestion:**
```
Add validation checks for cosigners' signatures to ensure they are valid and have not been tampered with.
```

</details>

<details>
<summary><b>:orange_circle: Major: Removal of per-deploy phlo admission checks creates a transitional unmetered-execution window</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 511)

The `validate_phlo` submission check (phlo price/limit) is removed from both `deploy` and `deploy_cosigned`. Per the block_creator comments, the acceptance gate operates in a TRANSITIONAL per-pool-presence mode by default: any deploy whose supply pool is absent (i.e., all deploys until Workstream C provisions pools) is admitted UNENFORCED and UNDEBITED. Until the economic producer lands and/or `strict_funding_enforcement` is enabled, deploys execute with no economic cost at all, which is a resource-exhaustion / spam DoS vector at the network boundary — worse than the pre-change state where at least a min phlo price gated submission.

**Suggestion:**
```
Consider retaining a minimal submission-side rate limit or size/complexity cap while the transitional gate is active, or default `strict_funding_enforcement` to true on new shards. Document the operational risk window explicitly for shard operators.
```

</details>

<details>
<summary><b>:orange_circle: Major: deploy_cosigned duplicates ~150 lines of deploy() including the full propose retry loop</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 570)

`casper_deploy_cosigned` copy-pastes the entire trigger-propose retry state machine (attempt counting, retry delay, ProposerResult matching, logging) from `casper_deploy`, plus most of the validation chain. Any future fix to the propose retry logic must now be made in two places, and they will inevitably drift.

**Suggestion:**
```
Extract the propose-trigger retry loop and the shared validation steps (read-only, shard-id, forbidden-key, expiration) into helpers shared by both `deploy` and `deploy_cosigned`.
```

</details>

<details>
<summary><b>:orange_circle: Major: Multiple active occurrences become a hard API error</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 518)

`finalize_sig_state` returns `DeployDispositionAmbiguity` when more than one source-aware occurrence remains active. Under forks, delayed finality, or partial tombstone propagation this can be a transient DAG state, turning status/lookup endpoints into hard failures instead of a defined pending/ambiguous client-visible state.

**Suggestion:**
```
Return an explicit non-terminal status (or deterministic primary occurrence plus ambiguity flag) for clients, and reserve hard corruption errors for invariants that cannot occur under honest protocol rules.
```

</details>

<details>
<summary><b>:orange_circle: Major: Cosigned reconstruction depends on in-memory sidecar integrity</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 1049)

`reconstruct_cosigned` rebuilds multi-sig envelopes from `pending_cosigner_metadata` keyed only by primary signature. If sidecar entries can be stale, partial, or replaced between admit and propose, the proposer may build a different envelope/supply key than the submitter or than replay (`ProcessedDeploy::to_cosigned`). `Cosigned::from_signed_data` re-verifies signatures, which helps, but metadata loss still drops cosigners and changes funding identity.

**Suggestion:**
```
Persist cosigner metadata with the deploy record (not only an in-memory map), fail proposal if a compound deploy is missing sidecar data, and test restart-between-submit-and-propose.
```

</details>

<details>
<summary><b>:orange_circle: Major: Potential logic error in create function</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2753)

The create function adds the close block deploy after computing the checkpoint data. However, it does not handle the case where the checkpoint data computation fails. This could result in an inconsistent state where the close block deploy is added but the checkpoint data is not computed correctly.

**Suggestion:**
```
Add error handling for the case where the checkpoint data computation fails, and ensure that the close block deploy is only added if the checkpoint data computation is successful.
```

</details>

<details>
<summary><b>:orange_circle: Major: Play-path settlement/fee state is easy to desync from replay</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 3177)

Close-block settlement debits, fee carve, and client fuel allocations are assembled on the play path and intentionally recomputed on replay from `block.body.deploys`. The retry loop re-admits prefixes and hard-errors on divergence, which is good, but any future change that packages deploys without going through this exact Close assembly will create play≠replay consensus failures. This coupling is high-risk in a 677-file change where acceptance/replay paths are mostly omitted from the review diff.

**Suggestion:**
```
Add golden play/replay tests for: full admit, prefix shrink retry, fee-only carve, empty admit + dummy-only, strict vs transitional gate, and multi-sig supply keys. Keep Close fields out of block serialization permanently and document that invariant next to both call sites.
```

</details>

<details>
<summary><b>:orange_circle: Major: In-memory cosigner-metadata sidecar risks envelope divergence after node restart</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/proposer.rs` (line 636)

`pending_cosigner_metadata` is an in-memory `HashMap` shared via `Arc<Mutex<...>>`. Deploys persist in `KeyValueDeployStorage`, but the cosigner sidecar does not appear to be persisted. After a node restart, a multi-sig deploy still in deploy storage will miss its sidecar entry in `reconstruct_cosigned` (block_creator.rs) and be silently uplifted via `Cosigned::from_single_signer`, producing a DIFFERENT envelope (and therefore a different supply key Σ⟦s⟧) than the one the client submitted and than replay derives from `ProcessedDeploy::to_cosigned`. This can cause play/replay divergence or debiting the wrong pool. Additionally, no eviction of sidecar entries is visible in the shown diff, suggesting unbounded memory growth.

**Suggestion:**
```
Persist cosigner metadata alongside the deploy in deploy storage (or reject multi-sig deploys whose sidecar is missing at proposal time with a hard error rather than a silent single-signer fallback), and evict entries when deploys are packaged, expired, or purged.
```

</details>

<details>
<summary><b>:orange_circle: Major: Funding gate defaults to unenforced admission</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/casper_conf.rs` (line 166)

`strict_funding_enforcement` defaults to false, so absent per-signature supply pools admit deploys with no debit. That is explicitly transitional, but it means cost-accounting is economically inert until pools are provisioned and operators flip the flag. A misconfigured or upgraded shard can process unpaid work while appearing to run the new metering stack.

**Suggestion:**
```
Require an explicit genesis choice for new shards, emit a startup warning when the transitional mode is active, and document that strict mode is mandatory once client pools exist. Consider failing closed in non-mainnet/dev profiles.
```

</details>

<details>
<summary><b>:orange_circle: Major: Filter-decision debug logging iterates five deploy collections per proposal</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 875)

The new `f1r3fly.merge.step` block emits one debug event per deploy across five collections. It is guarded by `tracing::enabled!`, which is good, but on busy shards with debug enabled this could produce a very large volume of per-deploy events per proposal cycle.

---

When `f1r3fly.merge.step` debug is enabled, the proposer emits one debug event per filtered/selected deploy. Under large deploy pools this becomes a significant propose-latency and log-volume cost.

**Suggestion:**
```
Consider aggregating to one event per reason with a joined list of truncated sigs, or sampling.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Occurrence sets can grow without bound across forks</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1155)

Each valid inclusion inserts into a persistent `BTreeSet` of block hashes per deploy id. There is no visible pruning when forks die or when a deploy is fully finalized/expired, so long-running nodes may retain ever-growing occurrence sets for repeatedly reproposed or widely forked deploys.

**Suggestion:**
```
Prune occurrence entries below finalized floors or after terminal disposition, and bound worst-case set size in tests.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Potential u8 overflow in benchmark signature synthesis</b> (Reported by: anthropic)</summary>

**File:** `casper/benches/multi_sig_fanout_bench.rs` (line 100)

`vec![0xCC + (i as u8); 32]` overflows u8 for i >= 52 (0xCC + 52 = 256). Under `cargo bench` this wraps silently, but if the bench is ever compiled with debug assertions (e.g., `cargo test --benches`) it panics.

**Suggestion:**
```
Use `0xCCu8.wrapping_add(i as u8)` to make the wrap explicit.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unused primary binding in cosigned deploy path</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 680)

Comments say validation is against the primary signer, but the code validates shard/expiry on `cosigned.data` and forbidden keys on all signers, then does `let _ = primary;`. This is misleading for future maintainers.

**Suggestion:**
```
Either use `primary` intentionally (e.g., deploy id / logging) or remove it and adjust the comment to describe whole-envelope validation.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Dead `primary` binding suppressed with `let _ = primary`</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 720)

`deploy_cosigned` binds `let primary = cosigned.primary();`, never uses it in the validation chain (the doc comment claims validation is against the primary signer's fields, but the checks actually use `cosigned.data` and `cosigned.signers()`), and then suppresses the warning with `let _ = primary;`. The comment and the code disagree.

**Suggestion:**
```
Remove the binding and correct the doc comment, or actually validate against the primary signer where intended.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Deploy status path loads every occurrence block</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1472)

`find_deploy`/status resolution fetches all occurrence hashes from storage, filters by body contents, then sorts. With the new multi-occurrence index, contested deploys can accumulate many block hashes across forks, making a previously O(1) index lookup into O(occurrences) block loads on a client API path.

**Suggestion:**
```
Prefer canonical finalization status first and only materialize occurrence blocks when needed; cap/page occurrences; or store height with the occurrence index to avoid full block fetches for ordering.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Block store errors silently swallowed when resolving deploy occurrences</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1493)

`occurrence_hashes.into_iter().filter_map(|hash| casper.block_store().get(&hash).ok().flatten())` converts storage errors into silent omissions. A transient store failure could cause the API to skip a legitimate occurrence block and fall back to a stale or absent answer without surfacing the error.

**Suggestion:**
```
Distinguish `Ok(None)` (block genuinely missing) from `Err` (storage failure) and propagate or at least log the latter.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Block-store failures are silently treated as missing occurrences</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1495)

The occurrence lookup uses get(&hash).ok().flatten(), discarding all block-store errors. Corruption or an I/O failure can therefore be reported as an absent occurrence, after which the API may select a fallback block and return misleading deploy information instead of exposing the storage failure.

**Suggestion:**
```
Collect occurrence blocks through a fallible iterator and propagate block-store errors. Only treat an explicit Ok(None) as a missing block, and consider reporting a DAG/index corruption error when an indexed occurrence has no corresponding stored block.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Eager evaluation of fallback lookup in run_prelude</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 196)

`.or(dag.lookup_by_deploy_id(&deploy_id).map_err(...)?)` uses `or` (eager) rather than a lazy fallback, so the legacy deploy-index lookup executes on every call even when the occurrence index already yielded a result. Same pattern appears in block_api.rs (`.or(dag.lookup_by_deploy_id(deploy_id)?)`). Cost is minor but the legacy lookup and its potential error path run unnecessarily.

**Suggestion:**
```
Use a lazy fallback (e.g., match on the min_by result and only perform the legacy lookup when it's None).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: DeployDispositionAmbiguity surfaces a possible-in-practice DAG condition as a hard API error</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 561)

When more than one active finalized occurrence survives source-aware reduction, `finalize_sig_state` returns an error. If concurrent branches can legitimately (even transiently) finalize the same deploy in two blocks before a rejection record lands, clients polling status will receive hard errors instead of a Pending/ambiguous state. Whether this state is truly unreachable depends on merge invariants not visible in this diff.

**Suggestion:**
```
Confirm via the merge invariants that double-active-finalized is impossible; if it can occur transiently, consider returning a dedicated `Ambiguous`/`Pending` state instead of an error.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: No-op self-rebinding of user_deploys</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2860)

`let user_deploys: HashSet<Signed<DeployData>> = user_deploys;` is a no-op rebinding whose only purpose seems to be hosting a comment. It adds noise and can confuse readers into thinking a transformation occurred.

**Suggestion:**
```
Remove the rebinding and attach the comment to the subsequent gate code.
```

</details>

<details>

**File:** `casper/src/rust/casper.rs` (line 105)

The deploy_cosigned function has a different naming convention compared to other functions in the Casper trait. It would be better to follow a consistent naming convention throughout the trait.

**Suggestion:**
```
Rename the deploy_cosigned function to match the naming convention used in other functions, such as deploy.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Benchmark name says 64-of-32 but threshold is 32-of-64</b> (Reported by: xai)</summary>

**File:** `casper/benches/multi_sig_fanout_bench.rs` (line 117)

`bench_cosigned_threshold_64_choose_32` labels the case `64-of-32` while constructing 64 signers with threshold 32. The inverted wording will confuse regression comparisons.

**Suggestion:**
```
Rename to `32-of-64` / `threshold_32_of_64` to match the API meaning.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Threshold benchmark name inverted vs semantics</b> (Reported by: anthropic)</summary>

**File:** `casper/benches/multi_sig_fanout_bench.rs` (line 120)

`bench_cosigned_threshold_64_choose_32` labels the case `"64-of-32"` while the construction is n=64 signers with threshold k=32 (i.e., 32-of-64). The naming will mislead anyone comparing bench output over time.

**Suggestion:**
```
Rename to `32-of-64` and align the function name.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Acceptance-gate activation semantics need operator-facing docs</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2895)

The gate’s per-pool-presence activation, dummy-deploy exemption, fee-vs-cost distinction, and genesis client-fuel seeding are subtle and mostly explained in in-code comments. Operators upgrading a live shard can easily misread this as “metering is on” when it is not strictly enforced.

**Suggestion:**
```
Add a short release-note/runbook covering: default transitional mode, when pools appear, how to enable strict enforcement safely, and how genesis `client-fuel-allocations` interact with unpaid deploys.
```

</details>
---

<details>

This is a very large (677 files, +123k lines) consensus-critical rework of Rholang metering that replaces per-deploy phlo pricing with an internalized, signature-scoped funding gate, adds multi-signature deploy envelopes, and introduces a deploy-occurrence index with source-aware rejection tracking. The visible code is carefully commented and accompanied by property tests, but several substantive concerns remain: the transitional gate mode admits all deploys unenforced/undebited until pools are provisioned (a DoS window compounded by removing submission-side phlo checks); the in-memory cosigner-metadata sidecar risks play/replay envelope divergence after restarts and appears to lack persistence/eviction; deploys in invalid blocks are silently dropped from all indices; and the deleted processed-vs-rejected packaging invariant (and its test) needs to be verified as enforced elsewhere. There are also lock-ordering and per-deploy write-amplification concerns in the new DAG index path, and notable code duplication in the cosigned deploy API path. Given the scale, the consensus-sensitivity, and the fact that 666 files (including the acceptance gate, replay, and crypto envelope implementations that many of these observations depend on) were omitted from the diff, this needs thorough human review by maintainers familiar with the play/replay and merge invariants before merging to dev.

The code change introduces significant modifications to the Rholang metering and cost accounting system. While the changes seem to be well-designed and thoroughly tested, there are a few potential security and logic issues that need to be addressed. Additionally, there is a minor style issue related to inconsistent naming conventions. Overall, the code change requires further review and testing before it can be merged.

#### openai (gpt-5.6-sol)

The change introduces substantial consensus-critical behavior, but the shown implementation has two blocking security problems: compound deploy authorization can be silently downgraded after restart because cosigner metadata is non-persistent, and funding enforcement is fail-open by default despite removal of legacy phlogiston limits. The new occurrence index also permits unbounded quadratic storage work, and one API path hides storage failures. These issues should be addressed before merge, particularly by persisting complete cosigned envelopes and making metering enforcement fail closed.

#### xai (grok-4.5)

This is a very large consensus-and-economics change: internalized Rholang metering, multi-sig envelopes, deploy occurrence indexing, and a block-assembly funding/settlement gate. The visible Rust core is thoughtfully structured around play/replay symmetry and deterministic ordering, but several high-impact risks remain: nested locking on the DAG deploy-index path, in-memory cosigner sidecar reconstruction, removal of rejected-vs-processed filtering without a clear local replacement, hard failures on ambiguous deploy disposition, and a default-off strict funding mode that leaves cost accounting economically unenforced. Because most admission, supply, replay, and merge implementations are outside the provided diff, this should not be merged on the partial review alone; it needs targeted human review plus play/replay and funding-gate integration tests before approval.
</details>

---

---

*Review scope note: GitHub's diff API rejects this PR (677 files > 300-file limit), so the review ran on a locally generated diff truncated to the tool's 100KB budget — 11 highest-priority core source files (block-storage DAG index, casper proposer/block_creator, block_api, deploy_finalization_status, casper_conf, benches). The other 666 files (including rholang metering internals, replay, crypto envelope, docs, and Cargo.lock) were not reviewed here.*

jeffrey-l-turner:
## Casper integration decisions — ratification meeting, 16 September 2026

The team ratified the following dispositions at the **16 September 2026 meeting**.

**Meeting record and ratification proof:** [D-01 through D-12 review on PR #390](https://github.com/F1R3FLY-io/f1r3node-rust/pull/390#pullrequestreview-5227717933).

PR #390 records the decisions. This comment states their consequences for PR #216. Ratification does not establish that the implementation or verification work is complete.

### Authority

Use current `dev` as the default Casper authority. Retain a PR #216 departure only for a demonstrated defect, required accounting invariant, or approved FIP.

The accounting paper controls accounting requirements. It does not independently authorize changes to Casper committees, fork choice, finality, recovery, or slashing.

### Required dispositions

| Entry | Disposition for PR #216 |
|---|---|
| **D-01 — Version authority** | Target Casper protocol **7**, with one authority chain from ceremony through reception. Activate through fresh genesis after FIPS approval. Keep accounting authority version **8** independent in the reusable node-level accounting module. |
| **D-02 — Committee and certificates** | Preserve upstream committee and stake provenance, exact justification-set equality, duplicate rejection, and validator block signatures. Do not require signed-floor commitments, certificate sidecars, or certificate retrieval. **Do not remove coupled certificate code before the required integration tests pass.** |
| **D-03 — Fork choice** | Preserve upstream GHOST, electorate, stake provenance, depth filtering, truncation, and progress policy. Bound the LCA traversal at the finalized floor. Prove that the bound preserves valid upstream head selection. |
| **D-04 — Finality** | Preserve upstream containment, cooperative budgets, disagreement rules, divergence telemetry, and the **inclusive** finality threshold. Do not adopt the second state certificate or universal certified advancement. Permit required exact accounting effect identities, including verified settlement from failed bodies. |
| **D-05 — Publication** | Require atomic publication, stale-result refusal, restart atomicity, terminal-verdict eviction, and separation of original FT evidence from display projections. Keep the complete ledger architecture optional and single-flight finalization as the default. Parallel evaluation requires equivalence evidence. |
| **D-06 — Heartbeat** | Adopt typed intents, local coalescing, permit revalidation, and stale-work rejection. Preserve all-eligible stale recovery, leader-only convergence, and frontier follow. Keep rotating stale leadership, frontier-follow removal, and progress-clock replacement deferred. |
| **D-07 — Deploy recovery** | Preserve current `dev` exact occurrences, tombstones, reason joining, carrier custody, causal closure, indexing, lifespan, and one-parent coverage. Test collective coverage and other unratified recovery policies separately. Experiments do not authorize activation. |
| **D-08 — Merge algebra** | Adopt exact effect identity, causal rejection closure, admission-effect alignment, checked arithmetic, and locally verified merge evidence. Select additive multiset composition conditionally for protocol 7. Require FIPS approval, fresh genesis, differential tests, compatibility analysis, and accounting-conservation evidence before activation. |
| **D-09 — Slashing** | Preserve upstream authorization, slash recovery, activation-epoch protection, and neglect policy. Canonical evidence reconstruction remains supplementary until equivalence is proved. Keep economic neglect slashing inactive and retain the Rust-to-Scala bisimilarity anchor. |
| **D-10 — Carrier index** | Preserve the dedicated index, watermark, retention, read-failure fallback, and narrow optimization role. Introduce a domain-separated protocol 7 lookup identity through the FIP-defined envelope commitment. This changes the lookup key, not deploy authentication. Keep `CLAIM-FINALITY-002` pending until its evidence gates pass. |
| **D-11 — Verification governance** | Preserve the five practice rules, named gate replacements, negative controls, mandatory CbC scope, and the slashing bisimilarity anchor. Use truthful 2,000-case PR, 10,000-case nightly, and at least 100,000-case extended tiers. |
| **D-12 — Deploy cost fields** | **Reject removal of `phloLimit` and `phloPrice`. Require both fields in protocol 7.** Preserve their signed consensus fields, prepayment, refund, exhaustion, minimum-price validation, protobuf tags, and APIs. Token accounting may add a FIPS-approved validity layer but cannot replace either field. Multi-wallet funding requires a normative mapping for both fields. |

Align PR #216 specifications, decision records, wire definitions, and implementation with these dispositions. Preserve the rejected proposals as history, not as current authority.

The linked meeting record controls where an earlier ledger proposal differs from the selected disposition. Protocol activation and pending correctness claims remain subject to their stated evidence conditions.

jeffrey-l-turner:
## Ratification follow-up — separate conformance and soak harness PR

This follow-up implements the evidence requirements from the [16 September 2026 ratification meeting](https://github.com/F1R3FLY-io/f1r3node-rust/pull/390#pullrequestreview-5227717933).

Prepare the harness additions on a **separate branch and PR**. Keep deterministic conformance tests distinct from long-running soak profiles. Do not activate deferred PR #216 policies merely because the harness can execute them.

### Required profiles

- [ ] **D-02 — Certificate removal:** Cover replay, settlement, restart, missing dependencies, and finalization before removing coupled certificate code.
- [ ] **D-03 — LCA equivalence:** Compare bounded and unbounded traversal on identical admissible DAGs. Fail if the floor bound changes a valid upstream head.
- [ ] **D-05 — Publication:** Test atomic publication, stale-result refusal, restart equivalence, publication order, and identical finalized results before enabling parallel evaluation.
- [ ] **D-06 — Recovery soak:** Compare frontier follow enabled and disabled, and all-eligible versus rotating stale recovery. Include paused and delayed validators, empty workloads, and deploy-bearing workloads. Measure recovery time, blocks per height, processing latency, and resource use. Prepare evidence for the deferred progress-clock change as well.
- [ ] **D-07 — Retry soak:** Compare one-parent and collective parent-set coverage. Compare current retry leadership with leader-free custody on joined and split frontiers. Measure duplicate occurrences, custody consistency, retry completion, and expiry rates.
- [ ] **D-08 — Merge conformance:** Cover changed merge results, repeated identical outputs, numeric overflow, admission-effect alignment, local evidence replay, and accounting conservation.
- [ ] **D-09 — Authorization conformance:** Compare both authorization paths for merge-lost slashes, evidence-order permutations, same-key rebond, stale epochs, missing dependencies, forged deploys, duplicate evidence, and restart. Use the soak profile for repeated concurrency, evidence-order, delayed-dependency, and restart variations.
- [ ] **D-10 — Index conformance:** Compare forced index-on and reference-scan verdicts. Cover identity-domain collisions, publication crashes, restart, watermark and pruning boundaries, missing history, and index read failures. Include fork-local, out-of-scope, valid, invalid, and approved carriers. Use soak runs to measure completed finalization and the claimed work bound.

### Evidence requirements

Retain exact node and harness revisions, deterministic seeds, configuration, run identifiers, metrics, and artifacts. Identify the baseline and experimental policy for each comparison.

Report failed, timed-out, and incomplete runs explicitly. Do not treat a soak result as formal proof or as authority to activate a deferred policy.

The team must review the evidence before reconsidering deferred policies. FIPS approval remains required where a selected change affects the protocol.

jeffrey-l-turner:
**Ledger entry D-07, sub-decisions 7.1 and 7.2.** @leithaus @metaweta , you are the final ratifiers for these decisions. Please confirm or correct the reading below in a reply.

The [ratification record of 2026-09-16](https://github.com/F1R3FLY-io/f1r3node-rust/pull/390#pullrequestreview-5227717933) lists exact occurrences, tombstones, and reason joining as current dev rules to preserve. Dev holds no occurrence store and no tombstone, so the record cannot mean they exist today.

Under the meeting's own rule, PR #216 supplies a rule for a demonstrated defect. DR-33 documents the duplicate-occurrence storm, and R-REASON-CONFLUENCE removes a block body that changed with parent arrival order. Both are protocol changes, since occurrence identity and record keys are protocol.

The consistent reading is the one the meeting applied to D-04, D-08, and D-10: sub-decisions 7.1 and 7.2 are ratified for protocol 7 under the D-08 activation conditions, and dev keeps signature-keyed records until then.

The D-07 entry on dev ([decision ledger](https://github.com/F1R3FLY-io/f1r3node-rust/blob/dev/docs/casper/design/decision-ledger/07-deploy-recovery-custody.md)) leaves the two sub-decisions unflipped until this confirmation. It will link this thread when they flip.

Issue #160, opened 2026-07-29
Heartbeat-driven networks stall permanently after hours of continuous empty-block production


9Neechan:
**Affected:** f1r3node-rust, heartbeat-based block production (no load), 3-validator + bootstrap + observer, `mid` profile (2/8 CPU, 4Gi/16Gi RAM)

---

## What did you expect to see?

The suite swept three heartbeat cadences — `ci1s/lfb2s/cd2s`, `ci3s/lfb5s/cd6s`, `ci5s/lfb10s/cd10s` (consensus-interval / last-finalized-block-check / catch-up-delay) — with **no deploy load**, just continuous heartbeat block production, to find parameters the chain could sustain indefinitely. All three were expected to run stably for the full suite duration, keeping finalization progressing at a roughly constant interval.

## What did you see instead?

**All three experiments were ABORTED by the watchdog** (`finalized block N has not advanced for ≥10m`, threshold 10m) — none passed:

| experiment | duration before abort | final block | avg block interval | recent-window interval | degradation |
|---|---|---|---|---|---|
| `hb-obs-ci1s-lfb2s-cd2s` | 3h28m | 4219 | 2798 ms | 9438 ms (last 30 heights) | **+237%** |
| `hb-obs-ci3s-lfb5s-cd6s` | 7h59m | 3774 | 7393 ms | 15646 ms | **+112%** |
| `hb-obs-ci5s-lfb10s-cd10s` | 11h06m | 3400 | 11650 ms | 26372 ms | **+126%** |

Every run shows the same shape: block/finalization interval degrades steadily for hours (2–2.4× slower than the starting cadence), small finalization gaps (10, then tens of blocks) start opening and closing, and eventually the network fails to recover from a gap and finalization freezes for good. The faster the configured heartbeat, the sooner it happens (3.5h vs 8h vs 11h) — i.e. it scales with **block count / DAG size**, not wall-clock time or configured interval.

Resource usage during all three runs stayed low (peak system CPU ≤5.5%, peak memory ≤1.3%, pod CPU/mem/disk all far under the `mid` profile's limits) — this rules out simple CPU/memory exhaustion as the trigger.

---

## Main logs errors

1. gRPC/TLS timeouts between nodes (ci3s, ci5s) — a precursor to the nodes losing network visibility of one another.

```
[00:07:15] observer-0   WARN: Request timeout: Protocol error. Request to rnode://…@validator1-0…:40400 timed out after 5000ms
[...]

Issue #23, opened 2026-04-17
finalization stalls when DAG grows wide from accumulated divergence


jeffrey-l-turner:
> **Migrated from F1R3FLY-io/f1r3node#463** — originally filed 2026-03-30 by @spreston8.
> This issue was moved as part of the f1r3node → f1r3node-rust deprecation (EPOCH-005).
> Original: https://github.com/F1R3FLY-io/f1r3node/issues/463

## Summary

When the DAG accumulates many unfinalized blocks from repeated divergence/convergence cycles, the finalizer stops advancing LFB even though validators continue proposing blocks. The clique oracle cannot find sufficient fault tolerance in the wide DAG.

## Observed Behavior

In CI integration tests (`test_replay_determinism`), tests 1-3 create divergence scenarios (bridge.rho deploys, validator pause). By the time test 4 runs (slow deploy), the DAG has ~20 unfinalized blocks above LFB. After the slow deploy completes:

- Validators propose blocks #14-#17 successfully
- Convergence recovery fires correctly
- Finalizer runs but finishes instantly without advancing LFB
- LFB stuck at #13, blocks #14+ never finalize
- 720s timeout expires with zero finalization progress

The same test passes locally on a faster machine — the DAG stays narrower because validators propose and finalize faster.

## CI Evidence

- arm64: https://github.com/F1R3FLY-io/f1r3node/actions/runs/23734904276/job/69140349431
- amd64: https://github.com/F1R3FLY-io/f1r3node/actions/runs/23732029104/job/69139093321
- arm64 (earlier): https://github.com/F1R3FLY-io/f1r3node/actions/runs/23732029104/job/69139093308

## Root Cause Hypothesis

The clique oracle's agreement traversal is sensitive to DAG width. With many parallel branches (from prior divergence/convergence cycles), the clique cannot find a large enough agreeing set for any block above the current LFB. The `DagMerger.merge` logs show `scope=32 blocks, actualBlocks (above LFB)=20`.

## Impact

Liveness concern for production shards under sustained load. If the DAG grows wide from repeated validator desynchronization, finalization can stall permanently even though block creation continues.

## Relevant Code

- `casper/src/rust/finality/finalizer.rs` — clique oracle traversal and work budget
- `casper/src/rust/safety/clique_oracle.rs` — agreement calculation across DAG
- `casper/src/rust/merging/dag_merger.rs` — merge scope calculation

## Related

- #437 — DAG tip divergence recovery
[...]

Issue #146, opened 2026-07-23
Finalization latency has a ~25-45s floor independent of block-replay cost, causing persistent TPS-suite backlog


9Neechan:
## What did you expect to see?

After the sharded-HotStore fix for [#43](https://github.com/F1R3FLY-io/f1r3node-rust/issues/43) collapsed per-deploy replay cost (EVAL ms 372 → 49.7 at cap=100, 7.5×), the TPS suite's sustained-backlog failure was expected to improve or disappear, since `submitted/finalized` ratio backlog should track replay/propose throughput. A single-deploy block with sub-second replay cost should finalize within roughly a few block-production intervals (~2-3s cadence under heartbeat-off/explicit-propose), not tens of seconds.

## What did you see instead?

The TPS suite (target 10 d/s, 2 workers, `mid` profile) hits `⚑ backlog` (`submitted/finalized > 2.0×`) in its very first phase, **identically before and after the sharded-HotStore fix**:

| run | submitted | finalized | ratio |
|---|---|---|---|
| pre-fix baseline | 1198 | 27 | 44× |
| broken new-v (invalid image, see [#43](https://github.com/F1R3FLY-io/f1r3node-rust/issues/43)) | 1198 | 20 | 60× |
| true-new-v (valid fix) | 1198 | 55 | 22× |

Finalization latency p95 stayed **24.3-25.2s across all three runs** — essentially unchanged by a fix that made replay 7.5× faster. Even more tellingly, `mntd` at **cap=1** (a single deploy per block, replay cost sub-millisecond, no batching cost at all) shows **p95 finalization latency of 42-45s in every run, old and new**. A single-deploy block cannot take 40+ seconds to finalize because of Rholang/rspace++ execution — that execution is over in under a second.

### Steps to reproduce

1. Deploy a 3-validator `mid` shard, heartbeat disabled, explicit propose.
2. Run the `tps` suite (target 10 d/s, 2 workers) or `mntd` at cap=1.
3. Observe `submitted`/`finalized`/`FinLatencyP95Ms` in the resulting report — backlog and the ~25-45s p95 floor appear regardless of block size or per-deploy replay cost.

## Supporting data

- `logs mntd-22-07-26` — cap=1: `P95 (ms) = 39274`; `Avg replay phase` for the same run stays in the 281.5-658 ms/block range across the whole sweep, confirming replay cost is not the bottleneck at any cap.
[...]

Issue #105, opened 2026-07-08
Finalization Latency Grows With Unfinalized DAG Frontier Size, Not CPU or Deploy Load


9Neechan:
## What did you expect to see?

With no deploys and no Rholang execution happening at all, finalization latency and block interval should stay roughly flat over time, bounded by consensus/gossip overhead alone. Raising the CPU limit per validator should not make timing behavior worse or more uniform in a way that suggests a shared bottleneck.

## What did you see instead?

Finalization latency climbed almost monotonically over a 15-minute window with no plateau, even though every block was empty (deploys disabled, zero Rholang execution). Per-block replay cost stayed flat and fast throughout, and doubling the CPU limit didn't help — if anything it made the degradation *more uniform* across validators while CPU usage stayed well under the new ceiling the entire time. This points at a cost that scales with the **size of the unfinalized DAG frontier** (how far finalization is lagging behind block production) rather than with CPU quota, deploy volume, or per-block replay work — most likely somewhere in Casper's fault-tolerance-oracle / last-finalized-block (LFB) decision path.

### Steps to reproduce the bug

1. Start a 4-validator network, all bonded.
2. Disable deploys entirely (no Rholang execution — every produced block is empty).
3. Let the network run for at least 15 minutes while it keeps producing blocks on its own.
4. Track block interval and finalization latency (time from a block being produced to it being finalized) over the run.
5. Optionally repeat with a higher CPU limit per validator to rule out CPU/cgroup throttling as the cause.

## Supporting data

- Finalization latency climbed from roughly 4s to roughly 30s over a 15-minute window (p50/p95/p99 over the tail of the run: ~35s/56s/77s), with no sign of leveling off.
- Block interval climbed from roughly 1.4s to 2.5s over the same window.
- Doubling the CPU limit (tested specifically to rule out CFS throttling) did not fix this — it made the degradation *more uniform* across all validators, while peak CPU usage stayed well under the new ceiling the entire time.
- Per-block **replay** cost — the thing CPU/cgroup limits directly gate — spiked once at genesis init and then stayed under ~20ms for the rest of the run. Replay itself is not getting slower.
- Since deploys were disabled, there is also no Rholang/rspace++ deploy-execution work happening per block.
- The unfinalized frontier (blocks produced but not yet finalized) grew steadily and did not shrink back down for the whole run — finalization was falling further and further behind block production.
[...]

PR #224, opened 2026-08-10
fix(consensus): recover stalled finality

a-k-l-sdao:
## Summary

- Track finalized-block progress independently from block age.
- Permit a deterministic, bounded convergence attempt only after finality stalls.
- Keep stale validators in Running while requesting fork-choice tips instead of forcing a destructive rejoin.
- Reset the recovery budget whenever finality advances.

## Regression coverage

`integration-tests/test/tests/custom/test_finality_stall_recovery.py::test_finality_stall_bounded_recovery`

The regression removes quorum after convergence, verifies bounded behavior during the stall, restores quorum, and requires pending work and finality to recover.

## Validation

- This exact branch passed the isolated finality-stall regression in 148.52 seconds; peak total shard memory was 1280 MB.
- Heartbeat unit tests, focused Casper recovery tests, formatting, clippy, and release build passed.

## Dependency

Depends on #211 and is the first new PR in the continuation stack. GitHub base remains `dev`.

Fix commit: `ec1c7dcb`.

## Issue references

Refs #160

Refs #23


jeffrey-l-turner:
@a-k-l-sdao — from an open-issue audit (mapping issues to addressing PRs): could you add explicit `Fixes #N` / `Refs #N` lines to this PR's description for the issues it addresses? Cross-referencing had to be reconstructed thematically because the linkage is missing — and note closing keywords only auto-close when a PR merges to the DEFAULT branch, so dev-targeted fixes need a manual close at promotion time either way. Candidate issues for this PR (please confirm/correct): #160 (heartbeat networks stall after hours), #146 (finalization latency floor), #105 (latency grows with frontier size), #23 (finalization stalls on wide DAG).

a-k-l-sdao:
Updated: #224 now references #160 and #23. I used `Refs` because this is a bounded recovery mitigation rather than complete root-cause closure; I omitted #146 and #105 because this PR does not change their underlying latency/frontier behavior.

jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

| anthropic | `claude-fable-5` | Needs Review | 0.75 |

---

### Summary

This PR makes substantial, intertwined changes to consensus recovery behavior and to the exploratory-deploy execution path. The heartbeat proposer changes (finality-progress tracking, bounded convergence recovery, new height-based leader election, removal of the genesis fast-path, and gating routine stale-LFB recovery on !stalled) alter liveness-critical logic; while unit and one integration regression pass, the leader-election view-dependence and the narrow convergence predicate pose real risks of concurrent recovery proposals or permanent stalls in edge cases not covered by tests. The exploratory-deploy changes (semaphore, queue/execution timeouts, 20x phlo-limit reduction, switch from merged-tip state to LFB-only state) are sensible resource-protection measures but include undocumented behavioral regressions for API consumers, and the timeout mechanism may not actually cancel CPU-bound Rholang work while releasing the concurrency permit. No security vulnerabilities were found, but the consensus-liveness implications and API semantic changes warrant careful human review before merge.

The change introduces useful finality-progress tracking, but the recovery accounting currently permits repeated convergence attempts for the same finalized block and simultaneously disables all routine stale-LFB recovery once the finality timeout expires. Together these behaviors conflict with the PR's bounded-recovery guarantees and can either create ongoing empty proposals or leave a node with no eligible recovery path. The unrelated change to exploratoryDeploy also appears to regress its default merged-tip query semantics and should be confirmed or reverted before merge.

The finality-progress tracker and bounded convergence attempt are a reasonable approach to idle finality stalls, and the regression/unit coverage around budget reset and leader selection is a good start. However, the heartbeat predicates become significantly more restrictive while stalled—routine stale/high-lag recovery is disabled and convergence still depends on `self_recently_proposed` plus leadership—so there is a plausible gap where a stalled node proposes nothing. Burning the recovery budget on failed proposes, requiring user deploys for frontier-follow, and removing ApprovedBlock rejoin all amplify that risk and deserve targeted human/scenario review before merge. Separately, the exploratory-deploy changes (LFB-only state, 20x phlo cut, timeouts/semaphore) are largely orthogonal API/DoS hardening with user-visible behavior changes and should be validated or split out.

The code changes look good overall. The new features and improvements are well-implemented, and the tests cover the changes adequately. There are a few minor style issues that could be improved for better readability and maintainability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Exploratory deploy no longer computes merged state from DAG tips</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1676)

The previous implementation merged post-states from multiple DAG parents (with late-block filtering disabled) when no block hash was specified; the new code always uses the last finalized block's post-state. Read queries will now observe strictly finalized state, which can be significantly stale — especially during the very finality stalls this PR is about, where the LFB stops advancing. Clients depending on near-tip reads (e.g., checking a recent deploy's effects) will see regressed behavior.

---

Unscoped exploratory deploys no longer compute a merged post-state from DAG parents/tips; they always execute against the last finalized block post-state. That is a user-visible API semantics change (read-only observers lose visibility into unfinalized tip state) and is largely unrelated to the consensus stall fix. Callers and dashboards that depended on pre-finalized state will silently see different results.

---

When no block hash is supplied, exploratoryDeploy previously computed the merged state of all DAG tips, but now unconditionally executes against the last finalized block's post-state. This changes externally visible query semantics and can return stale results that omit accepted but not-yet-finalized work. The change is unrelated to the documented finality-recovery behavior and has no regression test demonstrating that finalized-only state is now intended.

**Suggestion:**
```
Call this out explicitly in the PR/changelog, gate it behind a flag, or restore optional tip-merge for non-finalized reads if product still needs it. Keep the concurrency/timeout hardening either way.
```

</details>

<details>
<summary><b>:red_circle: Critical: frontier_follow_due redundancy</b> (Reported by: anthropic, xai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 597)

When `finality_progress.stalled` is true, `stale_lfb_recovery_due`, `stale_lfb_leader_recovery_due`, and `high_lag_recovery_due` are all forced off. The only stall-specific path left is `convergence_recovery_due`, which additionally requires `self_recently_proposed`, `has_new_parents`, `frontier_is_stale`, and lag-recovery leadership. If finality stalls while the local validator has not recently proposed (or lacks new parents / leadership), no recovery propose is scheduled at all for the idle/stale case. That is a narrower safety net than the PR summary suggests and can leave a bonded validator idle through the stall window.

---

`frontier_follow_due` gained `&& has_new_parent_with_user_deploys`. Combined with the genesis `inspect_parent_updates` change, empty-block frontier advancement and post-genesis 'no latest block' catch-up no longer propose while finality is fresh. The renamed unit test encodes this, but it may reintroduce idle empty-frontier stalls that previous heartbeat logic intentionally broke.

---

frontier_follow_due now requires both has_new_parents and has_new_parent_with_user_deploys; the latter implies the former in inspect_parent_updates' loop logic (except for the genesis seed case), making the conjunction subtle. A short comment explaining why both flags are checked would help future readers.

**Suggestion:**
```
Add a comment or simplify to the single sufficient condition if genesis handling permits.
```

</details>

<details>
<summary><b>:orange_circle: Major: Timeout cancellation may not stop the underlying Rholang execution</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1725)

tokio::time::timeout drops the play_exploratory_deploy future on expiry, but Rholang evaluation is CPU-bound and may run inside spawned/blocking contexts that are not cancel-safe. When the timeout fires, the semaphore permit (_permit) is released as the function returns, allowing a new exploratory deploy to be admitted while the timed-out one may still be consuming CPU/memory. Under sustained abusive queries this defeats the concurrency bound the semaphore is supposed to enforce.

**Suggestion:**
```
Verify play_exploratory_deploy is cancel-safe (aborts evaluation when the future is dropped). If not, thread a cancellation token/abort flag into the interpreter, or hold the permit until the underlying work is confirmed terminated.
```

</details>

<details>
<summary><b>:orange_circle: Major: Exploratory deploy phlo limit silently reduced from 100M to 5M</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 165)

play_exploratory_deploy previously hardcoded a 100,000,000 phlo limit; the RuntimeManager path now defaults to 5,000,000 (EXPLORATORY_DEPLOY_PHLO_LIMIT_DEFAULT). Existing read-only queries that legitimately consume more than 5M phlo will now fail with out-of-phlogistons. This is a 20x reduction in a user-visible limit that is not mentioned in the PR description and is only overridable via an undocumented environment variable.

**Suggestion:**
```
Call out this behavioral change explicitly, choose the default based on measured real-world query costs, and document the F1R3_EXPLORATORY_DEPLOY_* environment variables (or move them into HOCON config alongside the other casper settings).
```

</details>

<details>
<summary><b>:orange_circle: Major: Recovery budget is replenished without finality advancing</b> (Reported by: openai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 82)

FinalityProgress::observe marks recovery_round_due again once another finality_progress_timeout has elapsed since last_recovery_attempt_at. Consequently, a node can make one convergence proposal every timeout interval indefinitely while the same block remains finalized. This conflicts with the stated bounded behavior and with the requirement that the recovery budget reset only when finality advances.

**Suggestion:**
```
Treat last_recovery_attempt_at.is_some() as exhausting the recovery budget for the current finalized block. Clear it only in the branch where last_finalized_block changes. If periodic retries are intended, document that behavior explicitly and add a separate configured retry budget rather than implicitly granting unlimited rounds.
```

</details>

<details>
<summary><b>:orange_circle: Major: Finality stall disables routine stale-LFB recovery</b> (Reported by: openai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 601)

The new !finality_progress.stalled conditions disable stale_lfb_recovery_due, stale_lfb_leader_recovery_due, and high_lag_recovery_due after the finality timeout. At that point only the much narrower convergence_recovery_due path remains. If its prerequisites are not met—for example, there are no new parents, the validator has not recently proposed, or the frontier is not considered stale—the node performs no recovery at all for the remainder of the stall. This contradicts the configuration documentation stating that finality recovery does not delay or replace routine stale-LFB recovery governed by max_lfb_age.

**Suggestion:**
```
Keep the existing stale-LFB recovery paths active during a finality stall and add the convergence proposal as an independently budgeted extra path. Add regression coverage for a stalled node that is not eligible for convergence_recovery_due and verify routine stale recovery continues.
```

</details>

<details>
<summary><b>:orange_circle: Major: Routine stale-LFB recovery paths are disabled once finality is stalled</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 606)

stale_lfb_recovery_due, stale_lfb_leader_recovery_due, and high_lag_recovery_due are all gated on !finality_progress.stalled. Once the 30s finality-progress timeout elapses, the only remaining recovery path is convergence_recovery_due, which requires a narrow conjunction (has_new_parents && self_recently_proposed && !can_chase_frontier_while_ahead && frontier_is_stale && lag_recovery_leader && stale_recovery_window_open && !empty_frontier_backpressure). If any of those conditions never becomes true on the elected leader (e.g., frontier not stale, or no new parents), the network can remain permanently stalled with no recovery mechanism firing. This is a subtle liveness risk that warrants careful human review beyond the single integration regression cited.

**Suggestion:**
```
Consider a fallback: if the bounded convergence attempt has not fired after N finality-progress timeout periods, re-enable the routine stale-LFB recovery paths or escalate to fork-choice tip requests.
```

</details>

<details>
<summary><b>:orange_circle: Major: Recovery budget is consumed even when propose fails</b> (Reported by: xai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 647)

`finality_recovery_attempted` is set solely from `convergence_recovery_due` and is returned for Failure/Started/Success alike. `record_recovery_attempt()` therefore runs after a failed or no-op propose, suppressing further idle recovery for another full `finality_progress_timeout`. Under transient propose errors, a single failed attempt can extend a finality stall by 30s (default).

**Suggestion:**
```
Record the attempt only on Success/Started, or distinguish 'attempt scheduled' from 'attempt accepted' and retry sooner on clearly transient failures.
```

</details>

<details>
<summary><b>:orange_circle: Major: New lag-recovery leader selection is view-dependent and may not converge across nodes</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 1040)

select_lag_recovery_leader now picks the validator with the highest latest-message block number (per the local DAG view) instead of the previous deterministic round-robin on block number. Different nodes can have divergent views of latest_message heights during a stall (which is exactly when this code runs), so multiple validators may simultaneously believe they are the leader, or the intended leader may not recognize itself. This can cause concurrent convergence proposals or no proposal at all. Additionally, latest_height swallows DAG errors via unwrap_or(0), silently demoting a validator whose latest message failed to load.

**Suggestion:**
```
Use a selection input that is consistent across honest views (e.g., deterministic rotation keyed on the stalled LFB hash), or document why divergent leader election is acceptable/bounded here. Log rather than swallow latest_message errors.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Code duplication</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1658)

The code for computing the merged state from DAG tips is duplicated. Consider extracting it into a separate function to improve readability and maintainability.

**Suggestion:**
```
Create a function `compute_merged_state` that takes the necessary parameters and returns the merged state.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Metrics gauge not decremented on panic</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1718)

exploratory_deploy.active is incremented before the await and decremented after; if the future panics (or the task is cancelled between the two calls), the gauge leaks an increment permanently, skewing observability.

**Suggestion:**
```
Use a drop-guard struct that decrements the gauge in Drop to guarantee balance.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Execution timeout may not stop CPU-bound Rholang work promptly</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1724)

`tokio::time::timeout` drops the future on expiry, but if `play_exploratory_deploy` performs heavy CPU work without await points (or on blocking primitives), work can continue until the next yield while still holding runtime resources. The semaphore mitigates concurrency, but a single pathological term can still pin a runtime/worker for longer than `timeout_ms` suggests. Error mapping to 504 is good.

**Suggestion:**
```
Ensure interpreter cooperative cancellation or run exploratory play on a cancellable dedicated path; keep max concurrent at 1 by default (already done).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Redundant code</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/casper.rs` (line 484)

The block store is created twice in the `new` function. This can be simplified by creating the block store once and reusing it.

**Suggestion:**
```
Create the block store once and assign it to the `block_store` field.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Destructive rejoin removed; minority-fork healing relies only on tip gossip</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/running.rs` (line 116)

`recover_stuck_validator()` is no longer invoked on stuck detection. Validators remain Running and only emit fork-choice tip requests. This avoids destructive rejoin, but a true minority-fork / state-divergence case that previously healed via ApprovedBlock/Initializing no longer has that escape hatch in this path. If tip exchange is insufficient, the node can remain Running forever on a dead fork.

**Suggestion:**
```
Document the residual failure mode and consider a last-resort rejoin after N failed tip-based recovery intervals or after finality remains stalled beyond a larger bound.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Configuration via ad-hoc environment variables instead of HOCON config</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 65)

Four new tuning knobs (max concurrent, queue timeout, phlo limit, execution timeout) are configured via F1R3_* environment variables while the rest of this PR (finality-progress-timeout) correctly uses the HOCON config plumbing. This creates two parallel configuration mechanisms and leaves the env vars undocumented in defaults.conf.

---

The #[cfg(test)] mod tests block is inserted between MergeableKey and the RuntimeManager struct definition. Codebase convention (and general Rust convention) places test modules at the bottom of the file.

**Suggestion:**
```
Move mod tests to the end of runtime_manager.rs.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Default exploratory phlo limit reduced 20x</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 158)

Default phlo limit moves from a hard-coded 100_000_000 in RuntimeOps to 5_000_000 via RuntimeManager env defaults. Legitimate heavier read-only queries may start failing with out-of-phlogiston errors after upgrade without config changes.

**Suggestion:**
```
Document the new default and env override (`F1R3_EXPLORATORY_DEPLOY_PHLO_LIMIT`) in release notes; confirm explorer/UI queries fit under 5M.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded phlo limit</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 832)

The phlo limit is hardcoded in the `play_exploratory_deploy` function. Consider making it configurable.

**Suggestion:**
```
Add a configuration parameter for the phlo limit and use it in the `play_exploratory_deploy` function.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Test asserts non-empty requests, not ForkChoiceTipRequest</b> (Reported by: xai)</summary>

**File:** `casper/tests/engine/running_spec.rs` (line 455)

The updated stale-validator test only checks `!requests.is_empty()` and that ApprovedBlockRequest is absent. It does not assert that a ForkChoiceTipRequest (or equivalent) was actually sent, so a regression that sends some other message would still pass.

**Suggestion:**
```
Assert on the concrete tip-request message type/id, matching the minority_fork_recovery_spec change.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Genesis fast-path removal changes post-genesis proposal behavior</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 951)

inspect_parent_updates previously returned early for genesis with has_new_parents=true, explicitly breaking the post-genesis deadlock. Now genesis merely seeds has_new_parents=true and, combined with frontier_follow_due newly requiring has_new_parent_with_user_deploys, an idle network right after genesis will not propose until the 30s finality-progress timeout elapses and the convergence path fires. The updated unit test codifies this (expects 0 proposals), but it delays initial network convergence and depends on the fragile convergence_recovery_due predicate.

**Suggestion:**
```
Verify multi-validator startup convergence timing with the new gating; consider exempting the genesis-only DAG from the finality-progress gate.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: PR mixes consensus recovery with observer DoS hardening</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1)

Finality-stall recovery, stale-validator rejoin behavior, exploratory-deploy admission control/timeouts/phlo limits, and API status-code mapping are bundled together. That increases review risk and complicates revert/bisect if one concern misbehaves in soak.

**Suggestion:**
```
Prefer splitting observer hardening from consensus recovery in the stack, or clearly mark them as intentionally coupled with separate test evidence (which partially exists).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: ExploratoryDeployTimeoutError uses u128 for milliseconds</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 204)

timeout_ms: u128 comes from Duration::as_millis but u64 is more idiomatic for millisecond values and avoids awkward downstream formatting/serialization.

**Suggestion:**
```
Store the timeout as Duration or u64 milliseconds.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Leader election semantics changed from round-robin to most-advanced</b> (Reported by: xai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 1010)

`is_lag_recovery_leader` no longer selects `next_block_number % n`; it picks the validator with the highest latest-message height (byte-order tie-break). That is deterministic, but permanently prefers already-advanced nodes and can starve a slightly behind validator from ever acting as recovery leader. Worth an explicit design note because it affects who can open `idle_recovery_window_open` convergence.

**Suggestion:**
```
Document why height-based leadership is preferred for stall recovery, and add a multi-validator test where only the lagging node sees the stall conditions.
```

</details>
---

<details>

This PR makes substantial, intertwined changes to consensus recovery behavior and to the exploratory-deploy execution path. The heartbeat proposer changes (finality-progress tracking, bounded convergence recovery, new height-based leader election, removal of the genesis fast-path, and gating routine stale-LFB recovery on !stalled) alter liveness-critical logic; while unit and one integration regression pass, the leader-election view-dependence and the narrow convergence predicate pose real risks of concurrent recovery proposals or permanent stalls in edge cases not covered by tests. The exploratory-deploy changes (semaphore, queue/execution timeouts, 20x phlo-limit reduction, switch from merged-tip state to LFB-only state) are sensible resource-protection measures but include undocumented behavioral regressions for API consumers, and the timeout mechanism may not actually cancel CPU-bound Rholang work while releasing the concurrency permit. No security vulnerabilities were found, but the consensus-liveness implications and API semantic changes warrant careful human review before merge.

The code changes look good overall. The new features and improvements are well-implemented, and the tests cover the changes adequately. There are a few minor style issues that could be improved for better readability and maintainability.

#### openai (gpt-5.6-sol)

The change introduces useful finality-progress tracking, but the recovery accounting currently permits repeated convergence attempts for the same finalized block and simultaneously disables all routine stale-LFB recovery once the finality timeout expires. Together these behaviors conflict with the PR's bounded-recovery guarantees and can either create ongoing empty proposals or leave a node with no eligible recovery path. The unrelated change to exploratoryDeploy also appears to regress its default merged-tip query semantics and should be confirmed or reverted before merge.

#### xai (grok-4.5)

The finality-progress tracker and bounded convergence attempt are a reasonable approach to idle finality stalls, and the regression/unit coverage around budget reset and leader selection is a good start. However, the heartbeat predicates become significantly more restrictive while stalled—routine stale/high-lag recovery is disabled and convergence still depends on `self_recently_proposed` plus leadership—so there is a plausible gap where a stalled node proposes nothing. Burning the recovery budget on failed proposes, requiring user deploys for frontier-follow, and removing ApprovedBlock rejoin all amplify that risk and deserve targeted human/scenario review before merge. Separately, the exploratory-deploy changes (LFB-only state, 20x phlo cut, timeouts/semaphore) are largely orthogonal API/DoS hardening with user-visible behavior changes and should be validated or split out.
</details>

---

jeffrey-l-turner:
## Multi-agent review resolution status

Follow-up implementation is prepared in `db6897f3` (`fix(consensus): address finality recovery review findings`). The local branch is currently one commit ahead of the PR head, so this commit must still be pushed before the changes below are visible in this PR.

### Critical and major findings

| Finding | Resolution |
|---|---|
| Unscoped exploratory deploy lost merged DAG-tip semantics | Restored merged-parent state calculation with late-block filtering disabled; explicit block queries remain unchanged. |
| Stall predicates could leave no recovery path | Kept routine stale-LFB/high-lag recovery active during a finality stall and made the one-shot convergence proposal an independent fallback. |
| Frontier follow required a user-deploy parent | Restored ordinary frontier following and the post-genesis fast path. |
| Timeout could release admission while Rholang work continued | Exploratory execution now runs in an owned Tokio task; timeout aborts and awaits that task before the semaphore permit can be released. |
| Exploratory phlo default dropped from 100M to 5M | Restored the legacy `100_000_000` default and added a regression assertion. |
| Recovery budget replenished without finality advancing | A successful convergence attempt exhausts the budget until the finalized block changes. |
| Failed/deferred proposal consumed recovery budget | `Empty` and `Failure` results no longer consume it; only `Started`/`Success` for the selected convergence path do. |
| Recovery leader depended on each node's local latest-message heights | Selection is now deterministic from the sorted bond set and shared stalled-LFB hash, with order-invariance coverage. |

### Additional review notes resolved

- The exploratory active gauge and duration histogram now use a drop guard, so cancellation and early returns cannot leak the gauge.
- CPU-bound timeout handling is covered by the owned-task abort-and-await lifecycle.
- The 20x phlo regression finding is resolved by restoring the prior default while preserving the environment override.
- Genesis startup behavior is restored.
- The “hardcoded phlo limit” note is already addressed by `F1R3_EXPLORATORY_DEPLOY_PHLO_LIMIT`.
- The alleged duplicate block-store creation in `TestCasperWithSnapshot::new` is not present: one store is created, seeded with the LFB, and moved into the fixture.

### Validation completed locally

- `cargo check -p casper -p node --all-targets`
- `cargo test -p node heartbeat_proposer --lib` — 16 passed
- `cargo test -p casper exploratory_deploy --lib --tests` — all selected tests passed
- `cargo clippy -p casper -p node --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- Rust LSP and blocking diagnostics — clean

### Remaining resolutions before merge

1. Push `db6897f3`, then rerun the multi-agent review against the updated PR head.
2. Re-run the isolated finality-stall system-integration regression on both amd64 and arm64; the prior arm64 failure is the key acceptance signal for this PR.
3. Strengthen `casper/tests/engine/running_spec.rs` to assert an actual `ForkChoiceTipRequest`, not merely a non-empty request collection.
4. Document the `F1R3_EXPLORATORY_DEPLOY_*` controls or migrate them into the normal HOCON configuration surface.
5. Record the intentionally removed destructive-rejoin fallback as a follow-up design decision: tip gossip is the primary recovery path, but a separately bounded last-resort policy may still be warranted if repeated tip recovery cannot heal a minority fork.
6. Optional non-blocking cleanup: extract merged-state selection into a helper, move the local runtime-manager test module to the conventional file tail, and change timeout milliseconds from `u128` to `u64`/`Duration`.

The observer hardening and consensus recovery remain separate commits in the branch history, although they are reviewed together in this PR.

jltatbeach (approved):

jeffrey-l-turner:
## CI follow-up: node test flake identified

Run `31659326260` failed only `failed_convergence_propose_does_not_consume_recovery_budget` (`left: 0`, `right: 2`). The production recovery logic was not the cause. The test called `bond_validator_in_snapshot` on a parent produced by `get_random_block_default()`, which already contains 10 generated bonds. The LFB-based leader election could therefore select one of those unrelated generated validators instead of the test validator, so the expected proposal count was nondeterministically zero.

A local follow-up patch now retains only the intended validator bond in this fixture before exercising leader-only convergence recovery. Validation:

- failing release test: 50/50 repeated runs passed
- complete `cargo test -p node --release --lib`: 126 passed
- heartbeat suite: 16 passed
- node Clippy with `-D warnings`: passed
- Rust LSP diagnostics: clean

The patch still requires a follow-up commit and push before it appears on this PR.

PR #225, opened 2026-08-10
fix(observer): bound block report catch-up

a-k-l-sdao:
## Summary

- Serve ordinary block API reads from cached reports instead of synchronously replaying blocks.
- Serialize background report generation through a bounded queue and a single global report permit.
- Keep genesis deploys outside normal cost accounting and return service-unavailable responses while report state is unavailable.

## Regression coverage

`integration-tests/test/tests/custom/test_readonly_catchup_bounded.py::test_readonly_catchup_parallelism_keeps_api_responsive`

The regression builds 40 finalized blocks, starts a read-only observer, and verifies bounded catch-up, responsive APIs, correct block visibility, and bounded memory.

## Validation

- This exact branch passed the full 40-block regression in 419.52 seconds.
- Observer peak memory was 269 MB; peak total shard memory was 2737 MB.

## Dependency

Depends on #224, which depends on #211. GitHub base remains `dev`.

Fix commit: `ab073992`.


jeffrey-l-turner:
@a-k-l-sdao — from an open-issue audit (mapping issues to addressing PRs): could you add explicit `Fixes #N` / `Refs #N` lines to this PR's description for the issues it addresses? Cross-referencing had to be reconstructed thematically because the linkage is missing — and note closing keywords only auto-close when a PR merges to the DEFAULT branch, so dev-targeted fixes need a manual close at promotion time either way. Candidate issues for this PR (please confirm/correct): #48 (block_report replay panics on readonly observers), #203 (observer degradation).

spreston8 (approved):
Reviewed at `d014e1182` against `dev`. Approving.

Pushed to this branch rather than requested: the `dev` merge, a per-caller permit policy, removal of `F1R3_BLOCK_REPORT_QUEUE_TIMEOUT_MS`, removal of the `v2:` cache prefix, cache-first reads, drop counters and logging, and both regression gates.

The regression could not run: `test_readonly_catchup_bounded.py` was deselected and gated on a capability no manifest advertised. Both gates are now open and it passes on all four legs.

**Two of those reverse decisions in the PR.** The `v2:` prefix orphaned every stored report, and since nothing here changes the stored type it discarded the cache to correct at most the genesis report — with on-demand replay gone, nothing would have rebuilt it. Reads now check the cache before touching the permit and fall through to `block_report` on a miss, which is safe only because that call refuses rather than queues: a read during catch-up returns immediately instead of adding load. A read that does replay also caches, so ordinary traffic repopulates what pre-caching skipped.

spreston8:
Closing as superseded. Everything on this branch is on `dev`: `d0ef85f3d` (node-wide report permit, bounded prewarm queue, catch-up gate opened), then `f0c83df84` and `771a94415` on top of it.

Issue #104, opened 2026-07-08
Purse-Deposit Double-Produce Corrupts Number Channel, Halts Entire Network


9Neechan:
## What did you expect to see?

A bond deploy's stake transfer should complete its purse deposit exactly once. Regardless of any retries at the pre-charge/deposit layer, the underlying number channel should end up holding a single value, and the network should keep producing and finalizing blocks normally afterward.

## What did you see instead?

Six seconds after a bond deploy was accepted, a purse number channel ended up holding **two values simultaneously**, violating the single-value invariant that `IntegerAdd` (and presumably any other purse-balance operation) relies on. From that point on, **every subsequent block-proposal attempt on every validator** failed at the pre-charge step, because pre-charge tries to read/update that same channel. Since the tuplespace/purse state is part of shared chain state (not per-node), the corruption was global: the whole network halted simultaneously rather than just the node that triggered it, and it never recovered on its own. The bond deploy never landed in the finalized bonds map, and the run had to be aborted after a 5-minute timeout.

### Steps to reproduce the bug

1. Start a small network (e.g. 4 validators + 1 observer), all validators live and bonded.
2. Submit a bond deploy for a new validator (a stake transfer through the bonding purse contract).
3. Watch node logs on the already-bonded validators for the next ~10 seconds.

## Timeline

```
T+0s   bond deploy accepted
T+0s   new validator: propose fails with ReadOnlyMode (expected — not bonded yet)
T+6s   first "pre-charge evaluation failed: … purse deposit failed"
T+6s   first "propose call failed: … IntegerAdd single-value invariant violated"
T+~12s chain height metrics freeze (finalized/latest block heights stop advancing)
T+6s–T+~5min  same errors repeat every few seconds on every bonded validator,
              always citing the SAME channel ID
T+5min bond verification timeout — run aborted
```

### Node error log (representative excerpt, validator1)

```
ERROR pre-charge evaluation failed
      error="Deploy payment failed: BUG FOUND: purse deposit failed"
      target=casper::rust::rholang::runtime
      casper/src/rust/rholang/runtime.rs:484

ERROR propose call failed
      error="Runtime error: number channel
             75f4517d5dfb6ce750c7d4f2c32988558bf4e8f790d081afaf4e4263688731ce
             holds 2 values [0, 100000000]; IntegerAdd single-value
             invariant violated"
      target=node::rust::instances::proposer_instance
      node/src/rust/instances/proposer_instance.rs:268

[...]

Issue #47, opened 2026-05-20
Network desync and corrupted RSpace state after bridge-contract deploys


spreston8:
> Migrated from [F1R3FLY-io/f1r3node#506](https://github.com/F1R3FLY-io/f1r3node/issues/506) as part of the deprecation of that repository. Original author: @Timekiller7. Comments on the legacy issue are not mirrored here — see the linked legacy issue for the full thread.

---

### What did you expect to see?

The network should remain synchronized after bridge-contract deploys.  
All validators and the observer should be able to replay and validate blocks correctly, with the current block height staying consistent with or above the latest finalized block.

### What did you see instead?

After a series of bridge-contract deploys, the network entered a serious desync state.

Validators `v1`, `v2`, and the observer appear to have corrupted RSpace state and can no longer replay or validate blocks. At the same time, `v3` continues finalizing blocks above height `2374`.

Current block height became lower than the latest finalized block on `v1` and `v2`:

```text
=== Block / Finalized ===
validator1 — block=2280, finalized=2374
validator2 — block=2221, finalized=2368
validator3 — block=2375, finalized=2374
observer — block=2175, finalized=2172
```

Relevant observations from validator1.log:
```text
13:03:03: bridge deploys

13:04:58: Computation ran out of phlogistons

13:05:12: Replay block ......... got error SystemRuntimeError(ConsumeFailed), retries details: will retry, attempt 1

13:05:12: got error SystemRuntimeError(ConsumeFailed), retries details: giving up after 3 retries

13:05:12: raised BlockException:
Runtime error: Internal errors encountered while processing deploy:
System runtime error: Unable to consume results of system deploy

13:05:12: Recording invalid block b9f9ca460c... for InvalidTransaction

13:05:21: RootRepository validateAndSetCurrentRoot FAILED:
Blake2b256Hash(.........) not in roots store

13:05:21: level="ERROR", message="Replay block ......... got error InterpreterError(RSpaceError(HistoryError(RootError(UnknownRootError
```

### Steps to reproduce the bug
1. Start shard from the tested [commit](https://github.com/F1R3FLY-io/f1r3node/commit/6cba72cde9a0e7e7b6ec0fae58680a0b5ee05fa6)
2. Run a series of bridge-contract deploys (will share contract code in MM)
3. Observe validator logs after the deploys
4. Check current and LFB  on validators and observer.
5. Notice that some nodes stop replaying/validating blocks and end up with corrupted RSpace state

[...]

PR #226, opened 2026-08-10
fix(consensus): preserve concurrent bridge-lock accounting

a-k-l-sdao:
## Summary

- Traverse finalized blocks directly during mergeable-channel garbage collection.
- Delete finalized mergeable state only when the block is a strict main-chain ancestor shared by every validator latest message.
- Preserve concurrent multi-parent bridge-lock state through finalization cleanup.

## Regression coverage

`integration-tests/test/tests/custom/test_concurrent_bridge_locks.py::test_concurrent_bridge_locks_exact_accounting`

The regression submits 12 bridge locks concurrently, requires every deploy to finalize on every node, and reconciles the exact nonce, total locked, source balance, and vault balance.

## Validation

- This exact branch passed all 12 simultaneous locks with exact accounting in 361.45 seconds.
- Peak total shard memory was 13898 MB across five nodes under the repository-standard 28672 MB guard.
- Mergeable-channel GC units, formatting, clippy, and release build passed.

## Dependency

Depends on #225. Stack order is #211, #224, #225, then this PR. GitHub base remains `dev`.

Fix commit: `093a4aa5`.


jeffrey-l-turner:
@a-k-l-sdao — from an open-issue audit (mapping issues to addressing PRs): could you add explicit `Fixes #N` / `Refs #N` lines to this PR's description for the issues it addresses? Cross-referencing had to be reconstructed thematically because the linkage is missing — and note closing keywords only auto-close when a PR merges to the DEFAULT branch, so dev-targeted fixes need a manual close at promotion time either way. Candidate issues for this PR (please confirm/correct): #47 (network desync after bridge-contract deploys), #104 (purse-deposit double-produce corrupts number channel).

spreston8 (approved):
Reviewed at `5e9666e92` against `dev`. Approving.

Pushed to this branch rather than requested: `dev` merges, an anchor walk replacing the per-validator ancestor intersection, and a fork fixture.

**Most of what this PR carried is already on `dev` by another route.** The citation clause and the sweep that replaced the finalized-block enumeration both landed through #291, which adopted them during its own review. The merges here resolve the whole GC file to `dev` rather than re-applying them, so one thing is left: `common_strict_main_chain_ancestors` now finds the deepest block every validator has moved past and walks that one chain, instead of materialising each validator's chain and intersecting. Main-parent chains are linear, so the intersection of N strict-ancestor paths is the path below their deepest common point. `min_height` still bounds the walk, and `GcSweep`, `reaches_ancestor_check`, the metrics and the removal-after-storage ordering are untouched — the change is how many walks happen, not what they mean. `common_strict_ancestors` stays as the differential oracle.

Issue #18, opened 2026-04-17
Bonded validators that never run a node cause indefinite finalization stall


jeffrey-l-turner:
> **Migrated from F1R3FLY-io/f1r3node#373** — originally filed 2026-01-28 by @spreston8.
> This issue was moved as part of the f1r3node → f1r3node-rust deprecation (EPOCH-005).
> Original: https://github.com/F1R3FLY-io/f1r3node/issues/373

## Summary

When a validator bonds stake but never runs a node, block finalization (LFB advancement) halts indefinitely while block creation continues. This is a liveness issue - the network can never recover without manual intervention (restarting nodes to reset state).

## Current Behavior

1. Validator A is running with 1000 stake (100% of network)
2. Validator B bonds 1000 stake via PoS contract (now 50% each)
3. Validator B never runs a node
4. Blocks continue to be created by Validator A
5. Last Finalized Block (LFB) stops advancing
6. Network logs show: `Latest message is 20 blocks ahead of the last finalized block`
7. Deploys accumulate and eventually expire with "already in scope" errors
8. Network is permanently stuck until node restart

## Expected Behavior

The network should have a mechanism to detect validators that:
- Bond stake but never produce any blocks
- Go offline for an extended period after previously participating

These validators should be:
1. Quarantined after N blocks/epochs of inactivity
2. Excluded from finality calculations until they resume participation
3. Eventually slashed if inactivity persists

## Technical Details

The existing quarantine mechanism handles:
- Validators who propose invalid blocks
- Validators who equivocate (sign conflicting blocks)

But it does NOT handle:
- Validators who bond and never propose
- Validators who stop proposing entirely

The `synchrony-constraint-threshold = 0.0` setting allows block **creation** to proceed without seeing other validators, but finalization still requires justifications from validators representing sufficient stake.

## Reproduction Steps

1. Start a standalone node with a single validator
2. Run a deploy that bonds a new validator key (e.g., `bond-validator` command)
3. Do NOT start a node for the new validator
4. Observe that new blocks are created but LFB stops advancing
5. Subsequent deploys will eventually fail as the mempool fills with "already in scope" deploys

### Option 1: Inactivity Quarantine
Add a mechanism to quarantine validators who haven't proposed a block within N blocks of bonding or N blocks since their last proposal.

### Option 2: Epoch-based Active Validator Recalculation
At each epoch boundary, recalculate the active validator set based on actual participation in the previous epoch.

[...]

PR #227, opened 2026-08-10
fix(consensus): retain quorum during slow-peer recovery

a-k-l-sdao:
## Summary

- Select recovery leadership from bonded validators that have advanced beyond the LFB before falling back to the full bonded set.
- Prefer the greatest observed validator height with a deterministic validator-key tie-break.
- Preserve block processing and finality when a low-stake notification recipient is paused.

## Regression coverage

`integration-tests/test/tests/custom/test_slow_peer_notification.py::test_slow_peer_does_not_block_block_processing`

The regression pauses the 1-stake validator, submits 16 deploys through the active 200-of-201 quorum, requires continued finalization, restores the peer, and verifies convergence.

## Validation

- This exact branch passed the slow-peer regression in 85.07 seconds; peak total shard memory was 2254 MB.
- Fourteen heartbeat-focused unit tests passed.
- Formatting and clippy with warnings denied passed.
- The final stacked tip passed all four stack regressions in 1000.19 seconds; peak total shard memory was 14338 MB.

## Dependency

Depends on #226. Stack order is #211, #224, #225, #226, then this PR. GitHub base remains `dev`.

Fix commit: `d30fb58f`.

### DIFF
--- a/casper/src/rust/api/block_report_api.rs
+++ b/casper/src/rust/api/block_report_api.rs
@@ -1,9 +1,9 @@
 // See casper/src/main/scala/coop/rchain/casper/api/BlockReportAPI.scala
 
 use std::sync::Arc;
+use std::time::Duration;
 
 use block_storage::rust::key_value_block_store::KeyValueBlockStore;
-use dashmap::DashMap;
 use models::casper::{
     BlockEventInfo, DeployInfoWithEventData, ReportProto, SingleReport,
     SystemDeployInfoWithEventData,
@@ -11,9 +11,10 @@ use models::casper::{
 use models::rust::block_hash::BlockHash;
 use models::rust::casper::protocol::casper_message::{BlockMessage, SystemDeployData};
 use prost::bytes::Bytes;
+use rspace_plus_plus::rspace::hashing::blake2b256_hash::Blake2b256Hash;
 use rspace_plus_plus::rspace::reporting_transformer::ReportingTransformer;
 use shared::rust::store::key_value_typed_store::KeyValueTypedStore;
-use shared::rust::ByteString;
+use shared::rust::{env, ByteString};
 use tokio::sync::Semaphore;
 
 use crate::rust::api::block_api::BlockAPI;
@@ -22,6 +23,18 @@ use crate::rust::report_store::ReportStore;
 use crate::rust::reporting_casper::ReportingCasper;
 use crate::rust::reporting_proto_transformer::ReportingProtoTransformer;
 use crate::rust::safety_oracle::CliqueOracleImpl;
+use crate::rust::util::proto_util;
+
+const REPORT_CACHE_VERSION: &[u8] = b"v2:";
+const BLOCK_REPORT_QUEUE_TIMEOUT_MS_DEFAULT: u64 = 2_000;
+const BLOCK_REPORT_QUEUE_TIMEOUT_MS_ENV: &str = "F1R3_BLOCK_REPORT_QUEUE_TIMEOUT_MS";
+
+fn report_cache_key(block_hash: &BlockHash) -> ByteString {
+    let mut key = Vec::with_capacity(REPORT_CACHE_VERSION.len() + block_hash.len());
+    key.extend_from_slice(REPORT_CACHE_VERSION);
+    key.extend_from_slice(block_hash);
+    key
+}
 
 /// Domain-specific errors for BlockReportAPI operations
 #[derive(Debug, thiserror::Error)]
@@ -32,6 +45,10 @@ pub enum BlockReportError {
     ReadOnlyRequired,
     #[error("Block {0:?} not found")]
     BlockNotFound(BlockHash),
+    #[error("Block report pre-state is unavailable for block {0:?}")]
+    StateUnavailable(BlockHash),
+    #[error("Block reporter is busy; retry later")]
+    Busy,
     #[error("Failed to trace block: {0}")]
     ReplayFailed(String),
     #[error("Block info error: {0}")]
@@ -54,9 +71,8 @@ pub struct BlockReportAPI {
     block_store: KeyValueBlockStore,
     #[allow(dead_code)] // Part of constructor signature matching Scala, not directly used
     oracle: CliqueOracleImpl,
-    /// Thread-safe map of block hashes to semaphores for per-block locking
-    /// Equivalent to Scala's `blockLockMap: TrieMap[BlockHash, MetricsSemaphore[F]]`
-    block_lock_map: Arc<DashMap<BlockHash, Arc<Semaphore>>>,
+    block_report_semaphore: Arc<Semaphore>,
+    block_report_queue_timeout: Duration,
     /// Transformer for converting reporting events to protobuf format
     report_transformer: Arc<ReportingProtoTransformer>,
     /// When true, allows block reports on validator nodes (bypasses read-only check)
@@ -79,7 +95,14 @@ impl BlockReportAPI {
             engine_cell,
             block_store,
             oracle,
-            block_lock_map: Arc::new(DashMap::new()),
+            block_report_semaphore: Arc::new(Semaphore::new(1)),
+            block_report_queue_timeout: Duration::from_millis(
+                env::var_or(
+                    BLOCK_REPORT_QUEUE_TIMEOUT_MS_ENV,
+                    BLOCK_REPORT_QUEUE_TIMEOUT_MS_DEFAULT,
+                )
+                .max(1),
+            ),
             report_transformer: Arc::new(ReportingProtoTransformer::new()),
             dev_mode,
         }
@@ -122,27 +145,23 @@ impl BlockReportAPI {
         block: &BlockMessage,
         casper: &Arc<dyn crate::rust::casper::MultiParentCasper + Send + Sync>,
     ) -> ApiErr<BlockEventInfo> {
-        let block_hash = block.block_hash.clone();
-
-        let semaphore = self
-            .block_lock_map
-            .entry(block_hash.clone())
-            .or_insert_with(|| Arc::new(Semaphore::new(1)))
-            .clone();
-
         metrics::gauge!("block_report.lock.queue_size", "source" => "casper").increment(1.0);
-        let _permit = semaphore
-            .acquire()
-            .await
-            .map_err(|e| BlockReportError::SemaphoreError(e.to_string()))?;
+        let permit = tokio::time::timeout(
+            self.block_report_queue_timeout,
+            self.block_report_semaphore.acquire(),
+        )
+        .await;
         metrics::gauge!("block_report.lock.queue_size", "source" => "casper").decrement(1.0);
+        let _permit = match permit {
+            Ok(Ok(permit)) => permit,
+            Ok(Err(error)) => return Err(BlockReportError::SemaphoreError(error.to_string())),
+            Err(_) => {
+                metrics::counter!("block_report.busy", "source" => "casper").increment(1);
+                return Err(BlockReportError::Busy);
+            }
+        };
 
-        let result = self.block_report_inner(force_replay, block, casper).await;
-
-        // Remove semaphore entry to prevent unbounded growth of the lock map
-        self.block_lock_map.remove(&block_hash);
-
-        result
+        self.block_report_inner(force_replay, block, casper).await
     }
 
     /// Inner block report logic (separated to ensure lock map cleanup on all paths)
@@ -152,7 +171,7 @@ impl BlockReportAPI {
         block: &BlockMessage,
         casper: &Arc<dyn crate::rust::casper::MultiParentCasper + Send + Sync>,
     ) -> ApiErr<BlockEventInfo> {
-        let block_hash_bytes: ByteString = block.block_hash.to_vec().into();
+        let block_hash_bytes = report_cache_key(&block.block_hash);
         let cached = self
             .report_store
             .get(&vec![block_hash_bytes.clone()])
@@ -162,6 +181,16 @@ impl BlockReportAPI {
             if !force_replay {
                 return Ok(cached_report.clone());
             }
+        }
+
+        let pre_state_hash = Blake2b256Hash::from_bytes_prost(&proto_util::pre_state_hash(block));
+        let has_pre_state = casper
+            .runtime_manager()
+            .has_root(&pre_state_hash)
+            .map_err(|error| BlockReportError::ReplayFailed(error.to_string()))?;
+        if !has_pre_state {
+            metrics::counter!("block_report.state_unavailable", "source" => "casper").increment(1);
+            return Err(BlockReportError::StateUnavailable(block.block_hash.clone()));
         }
 
         let report = self.replay_block(block, casper).await?;
@@ -198,6 +227,16 @@ impl BlockReportAPI {
 
         self.block_report_within_lock(force_replay, &block, &casper)
             .await
+    }
+
+    pub fn cached_block_report(&self, hash: &BlockHash) -> ApiErr<Option<BlockEventInfo>> {
+        let key = report_cache_key(hash);
+        let cached = self
+            .report_store
+            .get(&vec![key])
+            .map_err(|e| BlockReportError::StoreError(e.to_string()))?;
+
+        Ok(cached.into_iter().next().flatten())
     }
 
     /// Create system deploy report from replay results
--- a/casper/src/rust/reporting_casper.rs
+++ b/casper/src/rust/reporting_casper.rs
@@ -5,7 +5,6 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use block_storage::rust::dag::block_dag_key_value_storage::BlockDagKeyValueStorage;
-use block_storage::rust::key_value_block_store::KeyValueBlockStore;
 use models::rhoapi::{BindPattern, ListParWithRandom, Par, TaggedContinuation};
 use models::rust::casper::protocol::casper_message::{
     BlockMessage, ProcessedDeploy, ProcessedSystemDeploy, SystemDeployData,
@@ -65,7 +64,6 @@ impl ReportingCasper for NoopReportingCasper {
 /// Real implementation using RhoReporter
 pub struct RhoReporterCasper {
     rspace_store: RSpaceStore,
-    block_store: KeyValueBlockStore,
     block_dag_storage: BlockDagKeyValueStorage,
     external_services: rholang::rust::interpreter::external_services::ExternalServices,
 }
@@ -94,16 +92,6 @@ impl ReportingCasper for RhoReporterCasper {
             .block_dag_storage
             .get_representation()
             .map_err(|e| format!("Failed to get DAG representation: {}", e))?;
-
-        let genesis = self
-            .block_store
-            .get_approved_block()
-            .map_err(|e| format!("Failed to get approved block: {}", e))?;
-
-        let is_genesis = genesis
-            .as_ref()
-            .map(|g| block.block_hash == g.candidate.block.block_hash)
-            .unwrap_or(false);
 
         let invalid_blocks_set = dag.invalid_blocks();
 
@@ -135,7 +123,7 @@ impl ReportingCasper for RhoReporterCasper {
             &pre_state_hash,
             &block.body.deploys,
             &block.body.system_deploys,
-            !is_genesis,
+            with_cost_accounting(&block.header.parents_hash_list),
             &block_data,
             seen_invalid_blocks,
         )
@@ -249,16 +237,29 @@ pub fn noop() -> Arc<dyn ReportingCasper> { Arc::new(NoopReportingCasper) }
 /// Factory function to create rho reporter with real reporting capability
 pub fn rho_reporter(
     rspace_store: &RSpaceStore,
-    block_store: &KeyValueBlockStore,
     block_dag_storage: &BlockDagKeyValueStorage,
     external_services: rholang::rust::interpreter::external_services::ExternalServices,
 ) -> Arc<dyn ReportingCasper> {
     Arc::new(RhoReporterCasper {
         rspace_store: rspace_store.clone(),
-        block_store: block_store.clone(),
         block_dag_storage: block_dag_storage.clone(),
         external_services,
     })
+}
+
+fn with_cost_accounting(parent_hashes: &[prost::bytes::Bytes]) -> bool { !parent_hashes.is_empty() }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn genesis_deploys_are_not_cost_accounted() {
+        assert!(!with_cost_accounting(&[]));
+        assert!(with_cost_accounting(&[prost::bytes::Bytes::from_static(
+            b"parent",
+        )]));
+    }
 }
 
 /// ReportingRuntime wraps RhoRuntimeImpl with ReportingRspace to enable event collection
--- a/casper/src/rust/util/mergeable_channels_gc.rs
+++ b/casper/src/rust/util/mergeable_channels_gc.rs
@@ -4,6 +4,8 @@
 //! This is required for multi-parent mode where immediate deletion during finalization
 //! can cause data races.
 
+use std::collections::HashSet;
+
 use block_storage::rust::dag::block_dag_key_value_storage::KeyValueDagRepresentation;
 use block_storage::rust::key_value_block_store::KeyValueBlockStore;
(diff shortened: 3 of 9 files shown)


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR delivers the stated slow-peer recovery fix (keeping stale validators in Running, height-aware recovery leader selection, finality-progress-gated idle recovery) with reasonable test coverage, but it bundles several substantial, undocumented behavioral changes well beyond the description: exploratory deploys no longer compute merged multi-parent state and lose 95% of their default phlo budget; a new timeout wrapper likely fails to cancel the underlying Rholang execution while releasing the concurrency permit; block reports move from per-block to global serialization with a busy-timeout; validators lose transfer report prewarming; and mergeable-channels GC is rewritten with an unbounded full-chain ancestor walk. None of these are security vulnerabilities, but the consensus-timing changes (30s stall gate replacing 5s LFB-age gate on all recovery paths) and the exploratory-deploy semantic/limit regressions warrant careful human review and explicit sign-off before merge.

The resource bounding and stale-validator Running-state changes are directionally useful, but there are significant recovery and API behavior risks. Most importantly, recovery leadership can become pinned to a paused validator that is only one block beyond finality, preventing the active quorum from recovering. The change also silently alters exploratory deploy state semantics and makes transfer enrichment permanently unavailable for several common cache-miss paths. These behaviors should be resolved or explicitly confirmed and covered before merging.

This PR meaningfully improves slow-peer behavior by keeping stale validators Running and requesting fork-choice tips, and it adds useful observer protections (report pre-state checks, cache versioning, exploratory busy/timeout). However, it also rewires core heartbeat recovery around finality-progress leadership with slower arming, inverts deploy-recovery chase-cap policy, removes approved-block rejoin escalation, and changes exploratory-deploy semantics/phlo defaults in ways that can regress API and catch-up behavior. I recommend human review focused on minority-fork escalation, deploy-burst lag caps, and the exploratory LFB-only/phlo-limit changes before merge.

The code changes look good overall. The new error handling for exploratory deploys and block reports is well-implemented. The changes to the heartbeat proposer logic and the addition of finality progress tracking are also well-thought-out. The tests cover the new functionality adequately. Minor improvements to error messages could enhance user experience.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Exploratory deploy no longer computes merged multi-parent state</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1678)

The previous implementation computed the merged post-state from all DAG tips when multiple parents existed (via compute_parents_post_state). The new code always uses the LFB post-state when no block hash is specified. Observers querying 'latest' state will now see data as of last finalization rather than the current fork-choice frontier, which is a functional regression for read-only queries against recent (unfinalized) state. This change is also not mentioned in the PR description, which focuses on slow-peer recovery.

---

When no block hash is provided, exploratory deploy now always uses the LFB post-state instead of computing the merged state across current DAG parents/tips. Multi-parent / unfinalized tip effects that were previously visible to read-only queries are no longer observable. This is a user-visible API semantics change bundled with consensus recovery work and may break clients or operator workflows that relied on tip-merged exploratory reads.

**Suggestion:**
```
Confirm this is intentional product behavior. If tip-merged reads are still required, gate the old merge path behind an explicit flag or only skip merge when under observer load; document the behavior change in release notes either way.
```

</details>

<details>
<summary><b>:red_circle: Critical: tokio::time::timeout does not stop underlying Rholang execution; permit released while work may continue</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1727)

Wrapping play_exploratory_deploy in tokio::time::timeout only drops the future. If the interpreter executes on spawn_blocking or a dedicated thread (typical for Rholang evaluation), the computation keeps consuming CPU/memory after the timeout fires. Because the semaphore permit (_permit) is dropped on early return, a new exploratory deploy can then be admitted concurrently, defeating the concurrency bound the semaphore was intended to enforce and allowing unbounded pile-up of orphaned executions under load.

---

tokio::time::timeout drops the awaiter on expiry, but play_exploratory_deploy/runtime execution may not cancel cooperatively. After timeout the OwnedSemaphorePermit is released, so another exploratory query can start while prior execution still consumes CPU/memory. The concurrency limit then under-protects the node.

**Suggestion:**
```
Keep the permit until the runtime op fully settles, or run exploratory work on a cancellable task/runtime with hard abort, and only release the permit in a finally path after teardown.
```

</details>

<details>
<summary><b>:red_circle: Critical: Default exploratory deploy phlo limit reduced 20x (100M -> 5M)</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 167)

EXPLORATORY_DEPLOY_PHLO_LIMIT_DEFAULT is 5,000,000, while the previous hardcoded limit in RuntimeOps::play_exploratory_deploy was 100,000,000. Existing exploratory queries that consume between 5M and 100M phlo will now fail with out-of-phlogistons. This is a silent breaking change for API consumers and is unrelated to the stated PR purpose.

---

Exploratory deploy phlo limit changes from a hardcoded 100_000_000 to an env-configurable default of 5_000_000. Queries that previously succeeded can now fail with out-of-phlogiston errors unless operators set F1R3_EXPLORATORY_DEPLOY_PHLO_LIMIT. This is easy to miss because it is not called out in the PR summary.

**Suggestion:**
```
Either keep the historical 100_000_000 default and rely on execution timeout/concurrency limits for DoS control, or explicitly document and version the new default with a migration note.
```

</details>

<details>
<summary><b>:orange_circle: Major: Default exploratory deploy semantics changed from DAG tips to the LFB</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1680)

When no block hash is supplied, exploratoryDeploy previously computed the merged state of all DAG tips. It now unconditionally executes against the last finalized block's post-state. This silently makes queries omit all accepted but not-yet-finalized state and can return substantially stale results. The behavior change is not described in the API documentation or covered by a compatibility test.

**Suggestion:**
```
If current-tip semantics are still required, retain the merged-tip calculation while applying the new concurrency and execution limits. If finalized-state semantics are intentional, expose it explicitly through the request/API contract, document the change, and add tests demonstrating the expected behavior.
```

</details>

<details>
<summary><b>:orange_circle: Major: Destructive rejoin removed; tip-only recovery may be insufficient on true minority forks</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/running.rs` (line 119)

update_fork_choice_tips_if_stuck no longer calls recover_stuck_validator, and tests now assert the node stays Running and only emits ForkChoiceTipRequest. For soft lag this is desirable, but a validator on a true minority/invalid fork may remain Running without the ApprovedBlock/LFS rejoin path that previously rebuilt state. Correctness now depends entirely on tip fetch + normal block processing converging.

**Suggestion:**
```
Add an explicit escalation path: stay Running and request tips for N intervals / until finality_progress_timeout budget is exhausted, then fall back to rejoin. Cover a case where tip requests alone cannot heal state.
```

</details>

<details>
<summary><b>:orange_circle: Major: Transfer enrichment is permanently unavailable on report cache misses</b> (Reported by: openai)</summary>

**File:** `node/src/rust/api/web_api.rs` (line 277)

Transfer enrichment now only calls cached_block_report and no longer generates a missing report. At the same time, the cache key version is changed to v2, invalidating all existing cached reports, and setup.rs only prewarms future finalized blocks after readiness through a bounded queue that deliberately drops entries when full. Consequently, historical blocks, blocks finalized during startup, and blocks dropped from the prewarm queue will report transfers as unavailable indefinitely unless an unrelated trace request happens to populate the cache first. The same issue exists in deploy_grpc_service_v1.rs.

**Suggestion:**
```
Provide a bounded on-demand cache-miss path, or schedule deduplicated background generation on misses. Also consider backfilling finalized blocks after startup/cache-version changes and retrying or recording skipped prewarm entries rather than permanently dropping them.
```

</details>

<details>
<summary><b>:orange_circle: Major: Idle recovery replaces LFB-age stale recovery and is much slower to arm</b> (Reported by: xai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 560)

stale_lfb_recovery_due, leader recovery, high-lag recovery, and convergence recovery now key off idle_recovery_window_open (finality hash unchanged for finality_progress_timeout, default 30s) rather than lfb_is_stale (max_lfb_age, default 5s). Clusters that previously self-healed on wall-clock LFB age will wait significantly longer, and recovery is also restricted to lag_recovery_leader. This is central consensus behavior and needs careful product/ops sign-off.

**Suggestion:**
```
Confirm 30s default against network finality SLOs. Consider arming idle recovery on max(finality_progress_timeout, max_lfb_age) or keeping a narrower non-leader tip-request path while only proposals are leader-gated.
```

</details>

<details>
<summary><b>:orange_circle: Major: Recovery proposals now gated on 30s finality-progress stall instead of max_lfb_age</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 610)

stale_lfb_recovery_due, stale_lfb_leader_recovery_due, high_lag_recovery_due, and convergence_recovery_due are all now conditioned on idle_recovery_window_open, which requires finality to have made no progress for finality_progress_timeout (default 30s), replacing the previous lfb_is_stale gate (max_lfb_age default 5s). This delays all idle recovery paths by up to 6x and rate-limits attempts to one round per timeout. This may be intended, but it materially changes cluster stall-recovery latency and interacts with liveness across all shards; it deserves careful review against the other recovery timers (stale_recovery_min_interval, max_lfb_age).

**Suggestion:**
```
Confirm the interaction between finality_progress_timeout (30s) and max_lfb_age (5s) is intentional; consider documenting the new state machine in defaults.conf, where finality-progress-timeout was added with no explanatory comment.
```

</details>

<details>
<summary><b>:orange_circle: Major: Deploy-recovery frontier chase cap inverted from widen to narrow</b> (Reported by: xai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 1048)

Previous soak-tested policy widened frontier chase during deploy recovery (max of configured lag and a recovery floor). The new effective_frontier_chase_cap uses min(frontier_chase_max_lag, 2) whenever deploy_recovery_hint or deploy_grace_active is set. That can stall catch-up under bursty deploy finalization and fights the PR goal of preserving progress when some peers are slow.

**Suggestion:**
```
Re-validate this against the slow-peer and deploy-burst scenarios. If narrowing is intentional, add a comment explaining why the prior widen policy was wrong; otherwise restore widen-on-deploy-recovery and keep leadership/idle-recovery as the main controls.
```

</details>

<details>
<summary><b>:orange_circle: Major: Recovery leader selection can permanently choose a paused validator</b> (Reported by: openai)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 1071)

select_lag_recovery_leader sorts eligible validators by latest height in ascending order and always selects the least-advanced validator above the LFB. A validator that produced one block beyond the LFB and was then paused remains eligible indefinitely and will be preferred over validators that continued advancing. Because all other validators gate recovery proposals on lag_recovery_leader, this can suppress recovery entirely—the exact failure mode this change is intended to prevent. The fallback is also a fixed leader rather than a rotating choice, so an unavailable fallback leader can similarly block recovery.

**Suggestion:**
```
Prefer validators at the highest observed height, or otherwise require recent progress, and use deterministic rotation among the eligible set so one paused validator cannot retain leadership indefinitely. Add a test where the lowest-height eligible validator is paused at LFB+1 while other validators are at higher heights.
```

</details>

<details>
<summary><b>:orange_circle: Major: Global single-permit semaphore serializes block reports for distinct blocks</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 98)

Per-block DashMap semaphores were replaced by one process-wide Semaphore(1) plus a 2s queue timeout. Any in-flight replay makes all other report/trace callers fail busy, including unrelated block hashes. Combined with cache-only transfer enrichment on validators, this can increase 503s under modest concurrent observer traffic.

---

The per-block DashMap of semaphores is replaced with a single global Semaphore::new(1). Previously, reports for different blocks could replay concurrently; now all block report requests are fully serialized, and callers waiting more than 2s (default) get Busy errors. Under bursty explorer traffic this reduces throughput significantly.

**Suggestion:**
```
Consider a small concurrency level (>1) or retain per-block locks with a separate global CPU/replay budget semaphore.
```

</details>

<details>
<summary><b>:orange_circle: Major: finality-progress-timeout added without explanatory comment</b> (Reported by: anthropic, xai)</summary>

**File:** `node/src/main/resources/defaults.conf` (line 559)

Every other heartbeat setting in defaults.conf carries a descriptive comment explaining its semantics and interactions; the newly added finality-progress-timeout has none, despite now being the primary gate for all idle recovery proposal paths.

---

finality-progress-timeout and multiple F1R3_* env vars (exploratory concurrency/timeout/phlo, block report queue timeout) change runtime behavior substantially but the PR description focuses only on slow-peer quorum retention.

**Suggestion:**
```
Add a comment describing what stalling means, the default rationale, and its interaction with max-lfb-age and stale-recovery-min-interval.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improve error handling for ExploratoryDeployBusyError</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 196)

Consider adding more context to the error message to help users understand why the exploratory deploy is busy.

**Suggestion:**
```
Add more details to the error message, such as suggesting users to retry later.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Active exploratory-deploy gauge is not cancellation-safe</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 1721)

The active gauge is incremented before awaiting execution and manually decremented afterward. If the request future is cancelled, such as when a client disconnects, control never reaches the decrement and the gauge remains inflated permanently. The block-report queue gauge uses the same manual increment/decrement pattern around an await.

**Suggestion:**
```
Use a small RAII guard whose Drop implementation decrements the gauge, ensuring metrics are restored on success, error, timeout, panic unwinding, and task cancellation.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improve error handling for BlockReportError::Busy</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 122)

Consider adding more context to the error message to help users understand why the block report is busy.

**Suggestion:**
```
Add more details to the error message, such as suggesting users to retry later.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: GC safety degrades open if latest-message set is empty</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 108)

common_strict_ancestors returns None when there are no latest messages. is_safe_to_delete treats that as 'do not block deletion' (is_some_and(...) is false), so finalized blocks past depth may be GC'd without the main-chain intersection guard. This is likely rare at runtime but is a sharp edge if called during early init/teardown.

**Suggestion:**
```
If common_strict_ancestors is None, return Ok(false) and skip GC for that pass.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: common_strict_ancestors walks the entire main chain per validator on every GC run</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 115)

common_strict_main_chain_ancestors traverses main parents from each validator's latest message all the way to genesis, materializing a HashSet of the full chain per validator, then intersecting. On long-running chains this is O(chain_length * num_validators) time and memory per GC invocation, growing unboundedly with chain height.

**Suggestion:**
```
Bound the walk at the last finalized block (blocks below the LFB horizon that are being GC'd are the only candidates), or cache/incrementally maintain the common-ancestor frontier.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Redundant duplicated condition in frontier_follow_due</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 609)

frontier_follow_due includes both frontier_follow_leadership_allowed(has_new_parent_with_user_deploys, idle_recovery_window_open, lag_recovery_leader) and a separate `&& (has_new_parent_with_user_deploys || idle_recovery_window_open)` clause. The second clause is already implied by frontier_follow_leadership_allowed, making it dead logic that obscures intent.

**Suggestion:**
```
Remove the duplicated `(has_new_parent_with_user_deploys || idle_recovery_window_open)` term.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: ProposerResult::Empty hardcodes idle_recovery_attempted: false</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/instances/heartbeat_proposer.rs` (line 756)

When an idle-recovery-driven propose results in ProposerResult::Empty, idle_recovery_attempted is hardcoded to false, so record_recovery_attempt is never called. If the proposer repeatedly returns Empty during a stall, the idle recovery window stays open and a recovery propose will be re-attempted on every heartbeat tick, bypassing the one-round-per-timeout budget the FinalityProgress struct is meant to enforce.

**Suggestion:**
```
Decide explicitly whether Empty should consume the recovery budget; if the current behavior is intentional, add a comment explaining why Empty results do not count as attempts.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Magic number 64 duplicated for prewarm queue capacity and gauge computation</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/runtime/setup.rs` (line 592)

The channel is created with capacity 64 and the queue-depth gauge computes `(64 - report_tx.capacity())` with a second hardcoded 64. If the capacity is changed in one place, the metric silently becomes wrong.

**Suggestion:**
```
Extract `const REPORT_PREWARM_QUEUE_CAPACITY: usize = 64;` and use it in both places.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Report prewarm now restricted to read-only nodes; validators lose transfer enrichment entirely</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/runtime/setup.rs` (line 600)

The BlockFinalised prewarm path is now gated on `is_node_read_only && is_ready_flag`. Combined with deploy_grpc_service_v1.rs and web_api.rs switching to cached_block_report (cache-read-only), validator nodes will never populate the report cache, so transfers will always be unavailable from validators. The code comment suggests this is intentional, but it is a behavioral change for API clients hitting validator endpoints and is not mentioned in the PR description.

**Suggestion:**
```
Confirm the intended behavior for validator-served transfer data and document it in the PR description / API docs.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: timeout_ms uses u128</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_api.rs` (line 205)

ExploratoryDeployTimeoutError stores timeout_ms as u128 (from Duration::as_millis). u64 is conventional for millisecond durations and avoids awkward interop; realistic timeouts never exceed u64.

**Suggestion:**
```
Use u64 and convert with `timeout.as_millis() as u64` or `u64::try_from(...)`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Cache key version bump orphans old v1 report entries</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/block_report_api.rs` (line 26)

report_cache_key prefixes hashes with 'v2:', so all previously cached reports (keyed by raw block hash) become unreachable but remain in the report store indefinitely, wasting disk.

**Suggestion:**
```
Add a one-time migration/cleanup that drops non-v2 keys, or note the expected storage growth.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Direct field access to finalized_blocks_set</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 145)

get_finalized_blocks now reads dag.finalized_blocks_set directly instead of a DAG API method. That couples GC to storage layout and may bypass invariants enforced by accessors.

**Suggestion:**
```
Prefer a public getter on KeyValueDagRepresentation if one exists, or add one.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: New tuning knobs use ad-hoc env vars instead of HOCON configuration</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 160)

F1R3_EXPLORATORY_DEPLOY_MAX_CONCURRENT, F1R3_EXPLORATORY_DEPLOY_QUEUE_TIMEOUT_MS, F1R3_EXPLORATORY_DEPLOY_PHLO_LIMIT, F1R3_EXPLORATORY_DEPLOY_EXECUTION_TIMEOUT_MS, and F1R3_BLOCK_REPORT_QUEUE_TIMEOUT_MS are read from environment variables, while the rest of the node uses defaults.conf/HOCON (e.g., the new finality-progress-timeout). This splits configuration surface across two mechanisms and none of the env vars are documented.

**Suggestion:**
```
Move these knobs into the casper HOCON configuration for consistency, or at minimum document the env vars in defaults.conf/README.
```

</details>
---

<details>

This PR delivers the stated slow-peer recovery fix (keeping stale validators in Running, height-aware recovery leader selection, finality-progress-gated idle recovery) with reasonable test coverage, but it bundles several substantial, undocumented behavioral changes well beyond the description: exploratory deploys no longer compute merged multi-parent state and lose 95% of their default phlo budget; a new timeout wrapper likely fails to cancel the underlying Rholang execution while releasing the concurrency permit; block reports move from per-block to global serialization with a busy-timeout; validators lose transfer report prewarming; and mergeable-channels GC is rewritten with an unbounded full-chain ancestor walk. None of these are security vulnerabilities, but the consensus-timing changes (30s stall gate replacing 5s LFB-age gate on all recovery paths) and the exploratory-deploy semantic/limit regressions warrant careful human review and explicit sign-off before merge.

The code changes look good overall. The new error handling for exploratory deploys and block reports is well-implemented. The changes to the heartbeat proposer logic and the addition of finality progress tracking are also well-thought-out. The tests cover the new functionality adequately. Minor improvements to error messages could enhance user experience.

#### openai (gpt-5.6-sol)

The resource bounding and stale-validator Running-state changes are directionally useful, but there are significant recovery and API behavior risks. Most importantly, recovery leadership can become pinned to a paused validator that is only one block beyond finality, preventing the active quorum from recovering. The change also silently alters exploratory deploy state semantics and makes transfer enrichment permanently unavailable for several common cache-miss paths. These behaviors should be resolved or explicitly confirmed and covered before merging.

#### xai (grok-4.5)

This PR meaningfully improves slow-peer behavior by keeping stale validators Running and requesting fork-choice tips, and it adds useful observer protections (report pre-state checks, cache versioning, exploratory busy/timeout). However, it also rewires core heartbeat recovery around finality-progress leadership with slower arming, inverts deploy-recovery chase-cap policy, removes approved-block rejoin escalation, and changes exploratory-deploy semantics/phlo defaults in ways that can regress API and catch-up behavior. I recommend human review focused on minority-fork escalation, deploy-burst lag caps, and the exploratory LFB-only/phlo-limit changes before merge.
</details>

---

jeffrey-l-turner:
@a-k-l-sdao — from an open-issue audit (mapping issues to addressing PRs): could you add explicit `Fixes #N` / `Refs #N` lines to this PR's description for the issues it addresses? Cross-referencing had to be reconstructed thematically because the linkage is missing — and note closing keywords only auto-close when a PR merges to the DEFAULT branch, so dev-targeted fixes need a manual close at promotion time either way. Candidate issues for this PR (please confirm/correct): #18 (never-running bonded validators stall finalization); possibly #21 (fork abandonment — this PR reworks that recovery area).

spreston8:
Reviewed at `d30fb58f3` against `dev`. Closing as superseded.

`select_lag_recovery_leader` no longer selects by height. `e5ab6f8d4` replaced its body with `validators.into_iter().min()`, to make the leader independent of the local LFB view — an earlier probe had shown the LFB-seeded form electing *no* leader at all when nodes hold different LFBs, which is the divergence case the mechanism exists to break. `dev` now pins that property with `lag_recovery_leader_is_stable_across_local_dag_and_lfb_views` and a 512-case `lag_recovery_leader_is_cross_view_deterministic`. Both LFB height and per-validator DAG heights are local inputs, so this branch would fail them by construction rather than by accident.

The gap this targets is narrower on current `dev` than when the branch was written, but not gone. Stale-LFB recovery is no longer leader-gated — it is open to every bonded validator on the temporal window — so a paused peer no longer blocks it. Leader selection survives only for the convergence one-shot, where `min()` yields one self-elector unless that validator is the one that is unavailable. That residual is recorded and will be handled separately, together with the observation that `slow-peer-notification-quorum` is advertised in the capability manifest while `test_slow_peer_notification.py` is still in the `--deselect` list, so nothing verifies it either way.

No rebase is worth attempting: the branch is 382 commits behind and the target function is gone.

Issue #147, opened 2026-07-23
Per-deploy replay cost is still 2-4× slower on 8/16-CPU pods than on 2-CPU pods, after the HotStore O(store-size) fix


9Neechan:
## What did you expect to see?

After fixing the dominant bottleneck behind [#43](https://github.com/F1R3FLY-io/f1r3node-rust/issues/43) (`create_soft_checkpoint()`/`HotStore::snapshot()` doing an O(store-size) clone on every deploy, fixed via a 256-way sharded `HotStore` — branch `issue-43`, commit `77b62a7a`), per-deploy replay cost across CPU profiles should be flat or improve with more CPU, not still show an inversion.

## What did you see instead?

Ran the `parallelism` suite (full MAX_DEPLOYS sweep: 1/5/10/50/100/250) on all three CPU profiles against the fixed image:

| MAX_DEPLOYS | minimum (2 CPU) s/depl | mid (8 CPU) s/depl | high (16 CPU) s/depl | mid/min | high/min |
|---|---|---|---|---|---|
| 1 | 0.168 | 0.387 | 0.234 | 2.30× | 1.39× |
| 5 | 0.108 | 0.354 | 0.293 | 3.28× | 2.71× |
| 10 | 0.097 | 0.383 | 0.359 | 3.95× | 3.70× |
| 50 | 0.129 | 0.290 | 0.373 | 2.25× | 2.89× |
| 100 | 0.139 | 0.377 | 0.351 | 2.71× | 2.53× |
| 250 | 0.071 | 0.186 | 0.193 | 2.62× | 2.72× |

`minimum` is faster than both `mid` and `high` at **every single N** in the sweep, by a consistent 1.4-4× margin — not noise, the pattern holds across all six independent block-size measurements.

The absolute magnitude improved massively versus the pre-fix baseline (mid @ N=1 went from 2.1-3.7s to 0.387s, ~6-10× faster) and the old super-linear (n^1.76) runaway growth with block size is gone on every profile — so the HotStore fix is real and worth shipping. But the literal "scales backwards with CPU" claim from the original bug report still reproduces, just at much smaller absolute cost.

### Steps to reproduce

Reproduced via the testbed's `parallelism` experiment (isolated-replay measurements) and `mntd-calibration` (concurrent-load validation of the production-relevant optimum, see below).

1. Deploy `9neechan/rust-node:issue-43-sharded-hotstore` (or any build containing commit `77b62a7a`).
[...]

PR #231, opened 2026-08-11
fix(node/rspace/block-storage): eliminate finalization-wall lock contention via jemalloc + batched DAG-insert writes (#146)

9Neechan:
**Summary**

Fixes the root cause identified for [#146](https://github.com/F1R3FLY-io/f1r3node-rust/issues/146) (finalization latency floor / zero-finalization wall at cap≥50 in `mntd-calibration`), and the same underlying mechanism behind [#147](https://github.com/F1R3FLY-io/f1r3node-rust/issues/147)'s residual CPU inversion.

Live `perf` profiling on a cap=50 cluster run showed **~40% of all sampled CPU cycles** spent inside glibc's futex wait/wake path (`__lll_lock_wait_private`/`__lll_lock_wake_private`) — threads blocked on a contended pthread mutex, not doing real work. Two app-level `Mutex` candidates were instrumented and ruled out beforehand (both sub-millisecond). By elimination, and confirmed by reading the LMDB env/DB layout, the contended lock is **LMDB's own internal single-writer lock**: `BlockDagKeyValueStorage::insert_internal` performs several independent `write_txn()`/commit round-trips per block against the same shared `dagstorage` LMDB environment (`deploy_index`, `latest_messages_index`, `block_metadata_index`, ... — 10 logical DBs share one lock). Under concurrent replay of multiple blocks per validator, all of these serialize on that one lock; more Tokio worker threads only feeds the convoy (matches the already-reported "more threads makes it worse" result).

**What changed**

- `shared/src/rust/store/lmdb_key_value_store.rs` — new `batched_put()`: detects when multiple writes target `LmdbKeyValueStore`s sharing the same `Env` and combines them into a single `write_txn()`/commit. Falls back to independent `put()` calls (unchanged behavior) for any non-LMDB or cross-env store.
- `shared/src/rust/store/key_value_store.rs` — added `as_any()` to `KeyValueStore` for the downcast check above (trivial impl across all 7 implementors, including test mocks).
- `shared/src/rust/store/key_value_typed_store_impl.rs` — added `raw_store()` getter.
- `block-storage/src/rust/dag/block_dag_key_value_storage.rs` — `insert_internal` now writes `deploy_index` and `latest_messages_index` (both on `dagstorage`) via one `batched_put()` instead of two separate `put()`s.
- `rspace++/src/rspace/history/history_repository_impl.rs` + `rspace++/src/rspace/metrics_constants.rs` — kept from the investigation: wait-time instrumentation on `HistoryRepositoryImpl`'s two `std::sync::Mutex`es, ruled out as the driver (0.001-0.002ms) but useful permanent observability given how close this bug's shape is to the already-fixed `LmdbKeyValueStore` mutex issue.

**Scope note**

`block_metadata_index` (the third writer into the same `dagstorage` env within `insert_internal`) is **not** batched in this PR — it's written through `BlockMetadataStore::add`, which carries slashing-critical in-memory `DagState` invariants; folding its write into the same transaction needs its own careful pass and is left for a follow-up.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/231"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789046991&installation_model_id=426883&pr_number=231&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F231&signature=cc2d8f5c8ef9bb82d293463460039d7fe44af576f7786a273e50b112c3d14d8c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


9Neechan:
@jeffrey-l-turner Its ready to be merged. I finished this fix.

jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a well-motivated and well-documented performance fix: batching same-environment LMDB writes in `insert_internal` directly targets the profiled single-writer-lock convoy, with a safe fallback path for non-LMDB/cross-env stores and mechanical `as_any` additions across implementors. However, there are a few real concerns before merge: the jemalloc LD_PRELOAD path is hardcoded to x86_64 and silently no-ops on the aarch64 images the Dockerfile explicitly supports; `batched_put` bypasses `LmdbKeyValueStore::put` and needs verification that no map-full/resize handling is lost; and the `deploy_index` write guard is now held across a much larger critical section including the LMDB write transaction, creating a longer hold time and a potential RwLock→LMDB-writer lock-ordering hazard. There are also smaller documentation issues (misleading `as_any` doc, misplaced `batched_put` doc comment) and missing unit tests for the new batching function. None of these are blocking security problems, but the LD_PRELOAD path and the put-semantics equivalence deserve fixes/verification.

The LMDB batching approach is reasonable and should reduce transaction-level writer-lock acquisitions, but two operational concerns should be addressed before merging: jemalloc is enabled through an x86_64-only path despite ARM64 image support, and the DAG insertion path now retains an application-level write lock across substantially more work, potentially undermining the intended contention reduction. The new persistence behavior should also receive focused transactional and fallback tests.

The core fix — detecting same-Env LmdbKeyValueStore instances and combining deploy_index + latest_messages_index writes into one write_txn via batched_put — is sound, with a correct fallback for non-LMDB/cross-env stores, and the surrounding metrics/Tokio/profiling changes are mostly additive observability. Before merge, fix the x86_64-only jemalloc LD_PRELOAD (silent no-op on aarch64) and shrink the deploy_index write-guard scope so it no longer covers new_latest_messages() and other unrelated work; also correct the as_any/raw_store docs and add direct unit tests for batched_put’s batch vs fallback behavior.

The code change addresses the identified performance issue by batching LMDB writes to reduce contention on the single-writer lock. There are minor performance and style improvements that could be made, and some documentation could be added to new functions for clarity.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: deploy_index write guard now held across a much larger critical section</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1115)

Previously `deploy_index.write()` was acquired, `put` completed, and the guard was dropped immediately. Now the write guard stays held through `invalid` handling, `new_latest_messages()?`, encoding of latest messages, and the LMDB `batched_put`. That widens critical-section scope on `deploy_index` far beyond what is required to serialize deploy-index mutations, and can re-introduce convoy behavior on the app-level RwLock even while LMDB single-writer acquisitions drop from 2 to 1.

---

Previously `deploy_index_guard` was dropped immediately after the `put()`. Now it is acquired before encoding and held across the `invalid_blocks_index` write, the equivocation logic, the entire `new_latest_messages()` computation, and the LMDB write transaction inside `batched_put`. This significantly extends the hold time of the `deploy_index` RwLock write guard, and introduces a lock-ordering pattern (app RwLock → LMDB writer lock) that other code paths acquiring these in a different order could deadlock against. It also partially works against the stated goal of reducing contention.

---

Previously, `deploy_index_guard` was dropped immediately after the deploy-index write. It is now retained while writing the invalid-block index, calculating latest messages, encoding their entries, acquiring LMDB's writer lock, writing both databases, and committing. Concurrent block inserts therefore serialize on this application-level write lock for substantially longer, potentially replacing part of the LMDB convoy this PR intends to remove. The longer nested-lock interval also increases lock-ordering risk with code that accesses these stores in another order.

**Suggestion:**
```
Encode deploy entries under the deploy_index write guard (or clone/encode then release), compute `new_latest_to_add` without holding that guard, then call `batched_put`. If mutual exclusion across the two indexes is required for invariants, take the minimal set of locks in a fixed order only around the batch write — not around `new_latest_messages()`.
```

</details>

<details>
<summary><b>:red_circle: Critical: LD_PRELOAD path is hardcoded to x86_64, silently breaking jemalloc on aarch64</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `node/docker-entrypoint-rust.sh` (line 21)

LD_PRELOAD is set to `/usr/lib/x86_64-linux-gnu/libjemalloc.so.2` unconditionally. The Dockerfile already supports aarch64 builds (`aarch64*) export RUSTFLAGS=...`), but on ARM this path does not exist. glibc will typically warn and ignore the preload, so the node still starts — but the jemalloc contention fix silently does not apply on aarch64, undermining the multi-arch runtime path and any A/B expectations around USE_JEMALLOC.

---

The Dockerfile explicitly supports aarch64, but the entrypoint always preloads `/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`. Debian ARM64 images install this library under `/usr/lib/aarch64-linux-gnu/`. On ARM64, the dynamic loader will reject or ignore this preload, so the node silently runs with glibc instead of the allocator this change enables by default, usually with a loader warning on every startup.

---

The entrypoint hardcodes `/usr/lib/x86_64-linux-gnu/libjemalloc.so.2`, but the Dockerfile explicitly supports aarch64 builds (the RUSTFLAGS case statement handles `aarch64*`). On ARM images the multiarch path is `/usr/lib/aarch64-linux-gnu/libjemalloc.so.2`. When LD_PRELOAD points at a nonexistent file, the dynamic loader emits a warning and continues with glibc malloc — so ARM deployments silently lose the entire jemalloc benefit this change is claiming (the ~2x TPS improvement cited in the comment), with only a stderr warning to indicate it.

**Suggestion:**
```
Resolve the architecture-specific library path dynamically, for example using `dpkg-architecture -qDEB_HOST_MULTIARCH`, or create a stable architecture-independent symlink during the image build. Also fail clearly or disable the preload when the resolved library does not exist.
```

</details>

<details>
<summary><b>:orange_circle: Major: batched_put bypasses LmdbKeyValueStore::put semantics via raw db.put</b> (Reported by: anthropic)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 178)

The batched path writes with `store.db.put(&mut wtxn, key, value)` directly instead of going through `LmdbKeyValueStore::put`. If `put()` implements any additional behavior — e.g. MDB_MAP_FULL detection with environment resize/retry, error translation, or bookkeeping — the batched path silently loses it, meaning failure behavior under a full map now differs between the batched and fallback code paths. This is exactly the kind of divergence that shows up only under load in production.

**Suggestion:**
```
Verify whether `LmdbKeyValueStore::put` performs map-resize/retry or other handling beyond a plain txn+put+commit; if it does, factor that logic into a shared helper used by both `put` and `batched_put`, and add a comment asserting equivalence otherwise.
```

</details>

<details>
<summary><b>:orange_circle: Major: No unit tests for batched_put</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 157)

The helper encodes the subtle contract of the whole PR (same Env → one write_txn; mixed/non-LMDB → independent puts; empty input; single store; cross-env fallback). Behavior is only exercised indirectly via DAG insert. Regressions in downcast/`Arc::ptr_eq` grouping would be easy to miss.

---

The new batching function has several behavioral branches — same-env batching, cross-env fallback, non-LMDB fallback, empty input, single store — none of which have direct test coverage in this PR, despite touching the block-insert hot path for every block.

---

The new helper changes same-environment LMDB writes from separately committed operations to one atomic transaction and introduces downcast and environment-identity fallback branches, but the diff contains no focused tests for those semantics. Regressions here could affect DAG index consistency.

**Suggestion:**
```
Add focused tests with two `LmdbKeyValueStore`s on one `Env`, two on different `Env`s, and a non-LMDB store, asserting one vs many write transactions (or observable put counts on mocks) and that all keys land correctly.
```

</details>

<details>
<summary><b>:orange_circle: Major: as_any documentation incorrectly claims a default implementation exists</b> (Reported by: openai, xai)</summary>

**File:** `shared/src/rust/store/key_value_store.rs` (line 10)

The comment says the default implementation is sufficient and no override is needed, but the trait method has no default body. Every implementor must explicitly define `as_any`, as demonstrated by the changes throughout the diff.

---

The doc comment says a default implementation is sufficient and no override is needed, but `fn as_any(&self) -> &dyn std::any::Any` has no default body. Every implementor must provide it (as the PR does across mocks). The new `'static` bound is also required for downcasting and is easy to miss when reading the comment alone.

**Suggestion:**
```
Either add a default body where feasible, or rewrite the comment to state that implementors should return `self`, and note that the trait now requires `'static` for `Any` downcasting.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Potential performance improvement</b> (Reported by: bedrock)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1112)

The encoding of keys and values for `deploy_entries` and `latest_messages_index` could be done in a more efficient manner by avoiding the intermediate `Vec` collection and directly encoding the keys and values within the `batched_put` function.

**Suggestion:**
```
Consider refactoring the encoding process to be more inline with the `batched_put` function to reduce memory allocations and improve performance.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Persistence ordering of deploy_index changed relative to other indices</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1170)

Before this change, `deploy_index` was persisted before `invalid_blocks_index` / equivocation-tracker writes; now it is deferred until after latest-messages computation and committed together with `latest_messages_index`. On a crash or error between those steps, the on-disk state differs from before (e.g. invalid-blocks entry persisted without the corresponding deploy_index entry). If any recovery/startup logic assumes the old ordering, this is a behavior change; if not, it should at least be stated explicitly since crash-recovery invariants are easy to break silently.

**Suggestion:**
```
Confirm and document that no recovery path depends on deploy_index being written first; consider batching the `invalid_blocks_index` write (same env?) into the same transaction in the follow-up mentioned for `block_metadata_index`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Trait definition improvement</b> (Reported by: bedrock)</summary>

**File:** `shared/src/rust/store/key_value_store.rs` (line 4)

The `KeyValueStore` trait should be defined with a `'static` bound to ensure that all implementors have a static lifetime, which is a common requirement for traits that are used with `Arc`.

**Suggestion:**
```
Add a `'static` bound to the `KeyValueStore` trait definition.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: as_any doc comment claims a default implementation that doesn't exist</b> (Reported by: anthropic)</summary>

**File:** `shared/src/rust/store/key_value_store.rs` (line 8)

The doc comment says 'Default implementation is sufficient for every implementor; no override needed', but `as_any` is declared without a default body — every one of the 7 implementors in this PR had to add `fn as_any(&self) -> &dyn Any { self }` manually. The comment is stale/misleading for future implementors, who will hit a compile error expecting a default.

**Suggestion:**
```
Fix the doc comment to say each implementor must provide the trivial `{ self }` body (a blanket default returning `self` isn't possible on the trait without Self: Sized tricks).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Adding 'static bound to KeyValueStore is a silent API-breaking change</b> (Reported by: anthropic)</summary>

**File:** `shared/src/rust/store/key_value_store.rs` (line 7)

The trait bound changed from `Send + Sync` to `Send + Sync + 'static`. All in-tree implementors satisfy this, but it is a breaking change to the trait's contract (any implementor holding non-'static borrows would no longer compile) that is not mentioned in the PR description and is only needed to make `as_any` downcasting sound.

**Suggestion:**
```
Mention the new `'static` requirement in the trait doc/PR description so downstream implementors understand why it exists.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Incorrect module path in raw_store docs</b> (Reported by: xai)</summary>

**File:** `shared/src/rust/store/key_value_typed_store_impl.rs` (line 37)

The doc references `key_value_store::lmdb_key_value_store::batched_put`, which is not a valid path in this crate layout. That will confuse readers and break any future intra-doc link checks.

**Suggestion:**
```
Point at `crate::rust::store::lmdb_key_value_store::batched_put` (or the equivalent public path used elsewhere).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: batched_put's doc comment is attached to the StoreWrite type alias, not the function</b> (Reported by: anthropic)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 143)

The long `/// Writes to multiple stores in a single LMDB write transaction...` doc block is immediately followed by `pub type StoreWrite<'a>` (with a second one-line doc comment stacked on top of it), so rustdoc attaches the entire explanation to the type alias while `batched_put` itself is undocumented.

**Suggestion:**
```
Move the main doc comment down so it sits directly above `pub fn batched_put`, leaving only the one-liner on `StoreWrite`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Scope note: third dagstorage writer still unbatched</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1174)

As the PR states, `block_metadata_index` still does its own write into the same `dagstorage` Env via `BlockMetadataStore::add`. Under concurrent insert this remains a third single-writer acquisition per block, so the contention floor may only partially move. Fine as incremental scope, but worth tracking so #146 is not considered fully closed by this PR alone.

**Suggestion:**
```
File/link the follow-up explicitly (issue or TODO near insert_internal) for folding metadata into the same txn without breaking DagState/slashing invariants.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add documentation for new functions</b> (Reported by: bedrock)</summary>

**File:** `rspace++/src/rspace/history/history_repository_impl.rs` (line 1)

The new functions `lock_current_history` and `lock_roots_repository` should have doc comments to explain their purpose and usage.

**Suggestion:**
```
Add doc comments to the new functions to improve code readability and maintainability.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Permanent per-acquisition metrics overhead on already-ruled-out locks</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/rspace/history/history_repository_impl.rs` (line 56)

Every acquisition of the `current_history` and `roots_repository` mutexes now performs an `Instant::now()` plus two `metrics::counter!` lookups/increments, even though the PR itself says these locks were ruled out (0.001–0.002ms wait). The metrics macros do a registry lookup per call; on hot paths this adds fixed overhead in exchange for observability the PR already concluded isn't needed.

**Suggestion:**
```
Consider gating these behind a feature flag or a cheap sampling scheme if they're kept as permanent instrumentation.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test assumes first thread-count entry is the minimum</b> (Reported by: anthropic)</summary>

**File:** `rspace++/tests/concurrent_rspace_test.rs` (line 300)

`let (min_threads, min_threads_ms) = results[0];` labels the 1-thread run as the minimum, but nothing guarantees 1 thread is fastest — the slowdown ratio is computed against index 0 rather than the actual minimum, which can under- or over-state the inversion. The test is #[ignore]d and diagnostic, so this is low stakes.

**Suggestion:**
```
Compute the actual min over `results` or rename the variables to `baseline_*` to reflect that it's the 1-thread baseline, not necessarily the minimum.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: batchable branch uses expect after equivalent match</b> (Reported by: xai)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 176)

After `batchable` is computed from `lmdb_stores`, the code `expect`s `Some` again. This is safe today but is noisier and more fragile than binding the stores once.

**Suggestion:**
```
Refactor to something like: if let Some(stores) = lmdb_stores if stores.len() > 0 && same_env(&stores) { ... } else { fallback }, avoiding expect.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Write txn opened even when all kv_pairs are empty</b> (Reported by: xai)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 180)

If every store is LMDB/same Env but all entry vectors are empty, the batch path still opens and commits a write transaction. Harmless but pure lock traffic.

**Suggestion:**
```
Early-return Ok(()) when total kv_pairs is zero before acquiring write_txn.
```

</details>
---

<details>

This is a well-motivated and well-documented performance fix: batching same-environment LMDB writes in `insert_internal` directly targets the profiled single-writer-lock convoy, with a safe fallback path for non-LMDB/cross-env stores and mechanical `as_any` additions across implementors. However, there are a few real concerns before merge: the jemalloc LD_PRELOAD path is hardcoded to x86_64 and silently no-ops on the aarch64 images the Dockerfile explicitly supports; `batched_put` bypasses `LmdbKeyValueStore::put` and needs verification that no map-full/resize handling is lost; and the `deploy_index` write guard is now held across a much larger critical section including the LMDB write transaction, creating a longer hold time and a potential RwLock→LMDB-writer lock-ordering hazard. There are also smaller documentation issues (misleading `as_any` doc, misplaced `batched_put` doc comment) and missing unit tests for the new batching function. None of these are blocking security problems, but the LD_PRELOAD path and the put-semantics equivalence deserve fixes/verification.

The code change addresses the identified performance issue by batching LMDB writes to reduce contention on the single-writer lock. There are minor performance and style improvements that could be made, and some documentation could be added to new functions for clarity.

#### openai (gpt-5.6-sol)

The LMDB batching approach is reasonable and should reduce transaction-level writer-lock acquisitions, but two operational concerns should be addressed before merging: jemalloc is enabled through an x86_64-only path despite ARM64 image support, and the DAG insertion path now retains an application-level write lock across substantially more work, potentially undermining the intended contention reduction. The new persistence behavior should also receive focused transactional and fallback tests.

#### xai (grok-4.5)

The core fix — detecting same-Env LmdbKeyValueStore instances and combining deploy_index + latest_messages_index writes into one write_txn via batched_put — is sound, with a correct fallback for non-LMDB/cross-env stores, and the surrounding metrics/Tokio/profiling changes are mostly additive observability. Before merge, fix the x86_64-only jemalloc LD_PRELOAD (silent no-op on aarch64) and shrink the deploy_index write-guard scope so it no longer covers new_latest_messages() and other unrelated work; also correct the as_any/raw_store docs and add direct unit tests for batched_put’s batch vs fallback behavior.
</details>

---

jeffrey-l-turner:
@9Neechan — a multi-agent review (Claude, GPT, Grok, Nova) was just posted on this PR: https://github.com/F1R3FLY-io/f1r3node-rust/pull/231#pullrequestreview — see the "Multi-Agent Code Review" review above.

The consensus agrees the diagnosis and the same-Env `batched_put` approach are sound, but no reviewer approved as-is. Before merge, please look at the two critical findings in particular (all three frontier reviewers flagged both independently):

1. **`deploy_index` write-guard scope** (`block_dag_key_value_storage.rs:1115`) — the guard is now held across `new_latest_messages()`, encoding, and the LMDB write transaction, which risks re-introducing a convoy on the app-level RwLock and creates an app-RwLock → LMDB-writer lock-ordering hazard.
2. **aarch64 jemalloc preload** (`node/docker-entrypoint-rust.sh:21`) — `LD_PRELOAD` hardcodes the x86_64 multiarch path, so ARM images silently run without jemalloc.

Also flagged as major: `batched_put` bypassing `LmdbKeyValueStore::put` semantics, missing unit tests for the batching branches, and a stale `as_any` doc comment. Details and suggested fixes are in the review body.

9Neechan:
## Rebuttal for the points not fixed

**"batched_put bypasses LmdbKeyValueStore::put semantics via raw db.put" (Major, anthropic)**
LmdbKeyValueStore::put has no additional behavior to lose — it's a plain write_txn() → loop of db.put() → commit(), with no map-full/resize handling, no error translation, no bookkeeping beyond the trivial commit. batched_put's inner loop does exactly the same db.put calls inside one shared write_txn, so the two are semantically equivalent by inspection, not by assumption. We added a comment in lmdb_key_value_store.rs stating this equivalence explicitly so the next reader doesn't have to re-derive it, but there's no behavior to fix.

**"Adding 'static bound to KeyValueStore is a silent API-breaking change" (Minor, anthropic)**
This is a documentation/PR-description ask, not a code defect — no implementor in this workspace holds non-'static borrows, so nothing actually breaks. We're noting it here in the review response rather than editing code for it.

**"KeyValueStore trait should have a 'static bound" (Minor, bedrock)**
Already true — pub trait KeyValueStore: Send + Sync + 'static at key_value_store.rs:7. This looks like the reviewer read a stale/pre-diff version of the trait; nothing to change.

**"Persistence ordering of deploy_index changed relative to other indices" (Minor, anthropic)**
Valid observation, but not something we're changing blind. deploy_index and invalid_blocks_index were already independent LMDB transactions before this PR — there was never an atomic commit spanning them, so a crash between the two could already leave inconsistent on-disk state pre-PR. This PR doesn't introduce a new failure mode, it just moves where in the sequence that pre-existing gap sits. Closing that gap for real (one transaction across all three indices) is a bigger change than this PR's scope and deserves its own investigation into what crash-recovery logic, if any, currently depends on write order — we're not doing that as a drive-by here.

**"Scope note: third dagstorage writer (block_metadata_index) still unbatched" (Suggestion, xai)**
Correct, and already called out in the PR description as a known follow-up. Folding block_metadata_index into the same transaction touches DagState/slashing invariants and warrants its own PR, not scope creep on this one.

**"Permanent per-acquisition metrics overhead on already-ruled-out locks" (Suggestion, anthropic)**
Intentional. lock_current_history/lock_roots_repository timing was added specifically to keep measuring these locks across future load profiles (per issues/13, /19 in the testbed investigation) — the "ruled out" conclusion was for the workload profiled so far, not a permanent guarantee. Removing the instrumentation would blind us to a regression here later. The overhead (an Instant::now() + two counter increments per lock acquisition) is negligible next to the LMDB/protobuf work each critical section already does.

**"Add documentation for lock_current_history/lock_roots_repository" (Suggestion, bedrock)**
Low value here — both functions are named for exactly what they do, and the surrounding module comment already explains why they exist and what they measure. Repo convention is no comments unless they carry non-obvious information; a "locks the mutex" docstring on a function called lock_current_history wouldn't clear that bar.

jeffrey-l-turner (approved):

PR #229, opened 2026-08-11
fix(node): clamp HTTP/2 frame size to the RFC 9113 legal range

jeffrey-l-turner:
## Summary

Fixes #20 — the configured `grpc-max-recv-message-size` was passed straight into tonic's `max_frame_size`, but HTTP/2's `SETTINGS_MAX_FRAME_SIZE` must lie in [2¹⁴, 2²⁴−1] (RFC 9113 §6.5.2). h2 asserts on out-of-range values, panicking the serving task and **resetting every inbound gRPC connection**.

- `http2_frame_size()` clamps the frame-size hint to the legal range (and handles >u32 sizes). The message size itself stays unclamped — frame size and message size are independent limits; the per-service codecs enforce the latter.

## Why production hasn't been burning

An accident of unit parsing: `defaults.conf`'s `16M` goes through `byte_unit` as **SI** 16,000,000 bytes — 777,215 under the cap. Any override in binary units (`16MiB`), a raw integer ≥ 2²⁴, or simply a larger size kills the entire gRPC API, one panic + connection reset per connection attempt. (Issue #20's reporter hit exactly this with the binary 16 MiB value; PR #195's rewrite of this file didn't touch the frame-size line.)

## Test evidence (verified red → green)

The obvious test (connect and succeed) **passes even on broken code** — tonic's lazy client masks the server-side panic. The regression test instead speaks raw HTTP/2: client preface + empty SETTINGS, then parses the server's own SETTINGS frame and asserts it advertises the clamped 16,777,215.

- **Red** (clamp removed): `h2-0.4.15/src/frame/settings.rs:91: assertion failed: DEFAULT_MAX_FRAME_SIZE <= val && val <= MAX_MAX_FRAME_SIZE` on the serving task, connection reset — the exact issue-#20 signature.
- **Green** (with clamp): server stays up and advertises `16777215`.
- Boundary unit tests: 16 MiB → cap, `usize::MAX` → cap, `0` → floor (2¹⁴), 1 MiB → passthrough.
- Full `node` lib suite + `fmt`/`clippy`/`deny` green.

## Note on deny.toml

Carries the same RUSTSEC-2026-0247 (`bitmaps` via `imbl`) ignore as #228: the advisory published 2026-08-10 and fails the pre-commit gate repo-wide; this branch is based on `dev`, which predates that entry. Identical text at the identical location, so #228 and this PR merge cleanly in either order.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
[...]

PR #228, opened 2026-08-11
fix(casper): stop block-expired recovered deploys from wedging the proposer

jeffrey-l-turner:
## Summary

Fixes #197 — the permanent finalization collapse under sustained deploy load (`ContainsExpiredDeploy` propose retry loop), reproduced in weekend-soak run [31404822218](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/31404822218) in 8/8 iterations (validators wedged at a fixed seqNum, finalization stalls up to 706 deploys, RSS ballooning to 31–37GB as work backed up). Full root-cause analysis: [#197 comment](https://github.com/F1R3FLY-io/f1r3node-rust/issues/197#issuecomment-5247848385).

**Root cause** — an expiry carve-out asymmetry: `block_creator`'s deploy selection admitted recovered / rejected-retry deploys past the `deploy_lifespan` window, but `Validate::transaction_expiration` has no recovery exemption. The proposer therefore built a block its own validation rejected, classified the failure as unrecoverable (`BugError`), never purged the deploy, and rebuilt the same invalid block on every heartbeat.

**Changes**

- `block_creator.rs` — block expiry now applies to recovered and rejected-retry deploys in selection; block-expired entries flow into the existing storage + rejected-buffer purge (their exclusion from that purge was the retention half of the wedge); the early buffered-deploy purge covers block expiry alongside time expiry.
- `proposer.rs` — `ContainsExpiredDeploy` reclassified as a recoverable self-validation failure: any residual expiry disagreement costs one skipped propose, never a permanent retry loop.
- `deny.toml` — accepts RUSTSEC-2026-0247 (`bitmaps` unmaintained, transitive via `imbl`); the advisory published today and blocks every commit repo-wide, so it rides in this commit by necessity.

## ⚠️ Reviewer attention: this PR inverts a designed contract

Three tests deliberately pinned the old behavior (block-expired recovered/rejected work is retried and retained):

- `block_expired_rejected_deploy_retries_after_source_leaves_scope` → now `…is_purged_not_retried`
- `recovered_buffered_deploy_is_selected_after_block_expiry` (inline) → now `…is_purged_after_block_expiry`
- `should_remove_expired_deploys_from_rejected_deploy_buffer`'s "block-expired sig must remain" assertion → inverted

[...]

PR #232, opened 2026-08-11
fix(ci): harden scheduler deploy, surface trigger provenance, dashboa…

jeffrey-l-turner:
Follow-through on the OCI scheduler outage (2026-08-04..11: every nightly Resource Scheduler dispatch 404'd on the function; the late GitHub cron fallback silently carried every soak — live IAM remediation applied 2026-08-11, this codifies it and makes a recurrence visible), plus the Charton chart suite for the soak dashboard.

## Scheduler hardening (codifies the 2026-08-11 live remediation)

- scripts/oci/deploy-soak-scheduler.sh: verify-after-write on every dynamic-group matching rule, via GET only — both list endpoints report matching-rule=null regardless of the real value in this tenancy, which is how the empty-group deploy went unnoticed; fail closed on mismatch. Policy statements grow to five: read functions-family for _02/_03 (use fn-invocation does not cover the scheduler's resource-resolution step, which 404s at 0%). The policy list replaces wholesale on update, so omitting them here would revert the live fix on the next redeploy.
- **End-to-end proof pending**: the first post-remediation Resource Scheduler slot is 2026-08-12T02:30Z. Success signal: a run named `Merge Recovery Soak [scheduled:…]` whose dashboard row shows trigger `oci` rather than `cron +Nm`.
- Scope decision (maintainer): a dedicated scheduler-health dashboard indicator was considered and **rejected** — the IAM fix should make the scheduler reliable, and the per-run trigger provenance below makes any regression visible without a bespoke widget.

## Trigger provenance end to end

The schedule gate emits trigger_source (oci-resource-scheduler | cron-fallback | manual) and slot_delay_seconds; soak-segment forwards them; run-merge-recovery-soak and write-soak-summary emit them in summary.json (fallback summary included); aggregate-perf-report carries them into history run entries, optional-by-construction for old rows. Dashboard: trigger on the tab provenance line ("oci", "cron +95m", "manual") and as a table column, dash for pre-field rows.

## Dashboard: mobile + Charton chart suite

- Mobile: tabs share the row as equal halves under 640px, tightened paddings, viewport-clamped interactions; light-mode series-3 #e87ba4 → #d5588e (the old pink sat at 2.36:1 against the chart surface, under the 3:1 mark-contrast floor; the new step passes all palette checks, CVD worst-adjacent ΔE 14.8, in both modes).
- Every chart is now a pre-rendered themed SVG pair from the standalone `scripts/soak-charts` crate (charton; deliberately not a workspace member; committed Cargo.lock is the supply-chain pin, built `--locked`). Failure map heatmap (neutral 0%, sequential red ramp, f/n volume text); metric panels pick marks from data density (layered line+points on a temporal axis when dense, value-labelled bars/points when sparse); all-zero too-far-ahead collapses to a "0 · target 0" badge; Peak CPU steps up from aggregate line → per-core facets → node × core cluster grid (Jet ramp clamped at 100% so red always means a saturated core, values printed on saturated cells) as richer telemetry appears.
- Publishers: chart SVGs ride `charts-manifest-<series>.json` (renderer-written) instead of hardcoded three-workflow lists. Manifest entries are treated as untrusted remote content (bare-filename allowlist; hostile names fail the publish). Per-SVG fetch/content failures are **fail-soft**: drop that chart with a warning and re-filter the manifest — charts are cosmetic and must never block the publish that makes soak history durable. Renders are staged outside the site tree and copied in only per fully-successful series, so a mid-render crash cannot publish a truncated SVG. Both policies are enforced by `check-workflow-invariants.sh`.
- CI: dedicated cargo-deny pass for the standalone crate (ci.yml + deny-schedule.yml); pre-commit/pre-push hooks gate its fmt/clippy/tests; deny.toml accepts **RUSTSEC-2024-0384** (`instant`, a wasm-only transitive of charton — never compiled on publisher targets). The RUSTSEC-2026-0247 (bitmaps/imbl) entry mentioned in earlier drafts arrived via dev and is not part of this diff.

## Merge order and chain

- Base: dev **with #228 (ContainsExpiredDeploy wedge fix) and #229 (HTTP/2 frame-size clamp) already merged** — merged into this branch at 968e1284 per the agreed ordering; the casper/grpc hunks visible in the branch history cancel against dev.
- Follows this PR: **#233** (`feature/enhance-soak-data-emission`, targets this branch) paired with **system-integration #103** — per-node and per-core CPU telemetry feeding the cluster grid. They land together; per-core data flows once the soak's SI pin advances past #103.

## Review record

Two multi-agent review passes (comments below). Confirmed findings fixed in `0c612be5`: per-SVG failures made fail-soft with manifest re-filtering, and staged rendering closing the partial-overwrite window. Determination on the escalated critical: the "missing ci-control checkout" half was a false positive (the checkout is the perf_report job's first step, outside the reviewed hunks); the in-place-render half was real and is fixed. Remaining minors (trigger_source validation, OCI rule-comparison normalization, carry-forward dedup) are noted as follow-ups.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/232"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789053765&installation_model_id=426883&pr_number=232&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F232&signature=8c418cb875c83eefb49c185d1c17142a87cdd079cd439e9781e22e0584cabbea"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a well-reasoned infrastructure PR: the OCI scheduler hardening (verify-after-write on dynamic-group rules, the load-bearing read-functions-family policy statements) directly codifies a live incident fix, trigger provenance is threaded end-to-end with an appropriate optional-by-construction stance for historical rows, and the workflow-side manifest handling shows real thought about remote-content validation (filename allowlisting, content sniffing, fail-closed on non-404 errors). The issues found are mostly robustness and maintainability concerns rather than blockers: the SLOT_DELAY_SECONDS numeric guard is applied inconsistently, the per-image onerror handler can hide a working chart when only one theme variant is missing, the OCI verification conflates GET failures with rule mismatches, and the ~55-line manifest-fetch block is duplicated across three workflows — recreating the exact multi-publisher drift hazard the docs warn about. The pre-commit hook's unconditional soak-charts test run adds meaningful friction (and an offline-commit failure mode) for a crate most commits don't touch. Note the diff was truncated, so the soak-charts Rust sources and some dashboard sections were not reviewed. Overall safe to merge after considering the feedback.

The visible changes generally improve scheduler verification, trigger provenance, dependency checks, and dashboard publishing, but the chart carry-forward logic can convert a transient 404 into a permanently incomplete publication, especially in the checkpoint workflow where no render follows. The dashboard also hides valid charts when only the inactive theme asset fails, and pre-push reporting miscounts skipped checks. Review coverage is limited to the provided, truncated portion of the diff.

Strong, well-motivated follow-through on the OCI scheduler outage: verify-after-write on dynamic-group matching rules, fail-closed policy updates that preserve the live `read functions-family` fix, and end-to-end trigger provenance into summaries, history, and the dashboard. Manifest-driven chart republish is carefully path-traversal hardened, and moving charts to pre-rendered themed SVGs with mobile/a11y tweaks is coherent. Main actionable concern is confirming the `ci-control/scripts/soak-charts` path in `merge-recovery-soak.yml`—if wrong, `continue-on-error` will silently freeze charts while history keeps updating. Remaining items are small robustness nits (theme onerror, slot_delay validation/clamp) plus out-of-scope crypto renames.

The PR introduces several improvements to the CI/CD pipeline, including hardening the scheduler deployment, adding trigger provenance, and upgrading the dashboard and charts. The changes look generally good, but there are a few areas that require further review, particularly around security and style consistency. The documentation should also be updated to reflect the new changes.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Confirm ci-control/ soak-charts path in soak publish job</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1755)

The new chart render step builds and runs `ci-control/scripts/soak-charts/...`, while `soak-dashboard-pages.yml` and the repo layout use `scripts/soak-charts/...`. If this job’s checkout/working directory is the repo root (not a `ci-control/` checkout), render will always fail. Because the step is `continue-on-error: true`, publish still succeeds and silently keeps stale carried-forward SVGs—so history updates while charts stop refreshing after soaks.

**Suggestion:**
```
Verify the publish job checkout path matches `ci-control/`. If the repo is at the workspace root, switch both the `--manifest-path` and binary path to `scripts/soak-charts/...` (same as soak-dashboard-pages). If `ci-control/` is correct, add a short comment at the step explaining the checkout layout so the asymmetry is intentional and reviewable.
```

</details>

<details>
<summary><b>:orange_circle: Major: A transient SVG 404 permanently republishes an incomplete chart set</b> (Reported by: openai)</summary>

**File:** `.github/workflows/soak-checkpoint-publish.yml` (line 205)

When a manifest-listed SVG returns 404, the workflow removes the local file and continues. The checkpoint workflow does not re-render charts, so it then publishes the manifest while omitting that SVG. This turns a transient CDN/cache inconsistency into persistent dashboard data loss. The same behavior exists in merge-recovery-soak.yml and soak-dashboard-pages.yml whenever their best-effort render step fails.

**Suggestion:**
```
Fail closed when any file listed by an existing manifest is unavailable or invalid, just as for other non-200 responses. Alternatively, remove the missing filename from the manifest before publishing, although aborting is safer for a carry-forward operation.
```

</details>

<details>
<summary><b>:orange_circle: Major: Potential security risk in dynamic group deployment</b> (Reported by: bedrock)</summary>

**File:** `scripts/oci/deploy-soak-scheduler.sh` (line 224)

The script deploys dynamic groups with specific matching rules. If the matching rule does not persist as expected, the script exits with an error. This is a good practice to ensure the correct configuration is applied. However, ensure that the `oci iam dynamic-group get` command is secure and does not expose sensitive information.

**Suggestion:**
```
Review the OCI CLI command to ensure it does not inadvertently expose sensitive data and consider adding additional logging or error handling to improve visibility into the deployment process.
```

</details>

<details>
<summary><b>:orange_circle: Major: SLOT_DELAY_SECONDS passed to --argjson without numeric validation</b> (Reported by: anthropic, xai)</summary>

**File:** `scripts/bench/write-soak-summary.sh` (line 30)

`SLOT_DELAY_SECONDS` is passed with `--argjson`, so any non-JSON numeric value aborts summary generation. `run-merge-recovery-soak.sh` validates digits, but this script is also a standalone entrypoint and will hard-fail on empty/invalid env (e.g. `SLOT_DELAY_SECONDS=""` after override, or a float).

---

run-merge-recovery-soak.sh validates SOAK_SLOT_DELAY_SECONDS with a regex before use, but write-soak-summary.sh passes it straight into `jq --argjson slot_delay "${SLOT_DELAY_SECONDS:-0}"`. If the script is invoked directly (or from a future caller) with a non-numeric value, jq fails and the entire summary write dies — the exact 'silent loss of summary' class of failure this PR is trying to eliminate. A well-formed but non-numeric JSON value (e.g. `{"x":1}`) would also be accepted and injected into the summary as the wrong type.

**Suggestion:**
```
Duplicate the `^[0-9]+$` guard from run-merge-recovery-soak.sh at the top of write-soak-summary.sh, defaulting to 0 (or failing loudly) on mismatch.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Skipped dashboard checks are omitted from the total count</b> (Reported by: openai)</summary>

**File:** `.githooks/pre-push` (line 177)

The dashboard suite increments total only inside run_dashboard_check, while skipped clippy, deny, and test checks increment skipped without incrementing total. Consequently, the aggregate summary can report counts where passed + failed + skipped exceeds total.

**Suggestion:**
```
Increment total before each dashboard check decision, including skipped checks, or move counting outside run_dashboard_check so every configured check contributes exactly once.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Input description formatting</b> (Reported by: bedrock)</summary>

**File:** `.github/actions/soak-segment/action.yml` (line 32)

The input descriptions use em dashes (—) instead of standard dashes (-), which may not be necessary and could be simplified for better readability.

**Suggestion:**
```
Use standard dashes (-) in the input descriptions for better readability and consistency.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: CSS variable naming consistency</b> (Reported by: bedrock)</summary>

**File:** `.github/dashboard/index.html` (line 12)

The CSS variables for series colors have been updated, but the variable names could be more consistent. For example, `--series-3` was changed to a darker shade, but the variable name remains the same.

**Suggestion:**
```
Consider renaming the CSS variables to reflect their updated values, e.g., `--series-3-dark` to improve clarity and maintainability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Theme-specific chart onerror hides both light and dark images</b> (Reported by: xai)</summary>

**File:** `.github/dashboard/index.html` (line 520)

Each chart/heatmap figure injects two `` tags (light/dark) whose `onerror` does `this.closest('figure').style.display='none'`. If only one theme variant is missing or corrupt, the handler hides the entire figure—including the still-valid other theme—so users in the working theme see no chart.

**Suggestion:**
```
Hide only the failed image (`this.style.display='none'`), or hide the figure only when both variants have failed (e.g. track errors on the figure and collapse after the second failure).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Failure of an inactive theme image hides the entire chart</b> (Reported by: openai)</summary>

**File:** `.github/dashboard/index.html` (line 531)

Both light and dark image elements use an onerror handler that hides their containing figure. Browsers generally fetch images even when CSS gives them display:none, so a missing dark SVG can hide a valid light chart and vice versa. This applies to both failure heatmaps and metric panels.

**Suggestion:**
```
Hide only the failed image in its onerror handler. Hide the entire figure only after confirming that both theme variants failed, or provide a fallback that displays the surviving variant.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: onerror on either theme variant hides the whole figure</b> (Reported by: anthropic)</summary>

**File:** `.github/dashboard/index.html` (line 560)

Both the light and dark `` variants carry `onerror="this.closest('figure').style.display='none'"`. If a partial publish leaves only one variant present (e.g. the light SVG published but the dark one 404s), the failing hidden variant's onerror hides the entire figure — including the working, currently-displayed chart. The renderer normally emits pairs, but the carry-forward path in the publishers fetches each file independently and drops individual 404s, so a half-pair on the site is a reachable state.

**Suggestion:**
```
Hide only the failing image (`this.style.display='none'`) and hide the figure only when both variants have failed, or have the publishers drop both variants of a pair when either is missing.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: cron-fallback slot_delay_seconds can be negative</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 377)

`slot_delay_seconds=$((now_epoch - slot_epoch))` assumes `now >= slot`. Clock skew or an unexpected schedule edge can yield a negative delay. The dashboard only formats delays `>= 60`, so negatives are hidden, but they still land in `summary.json` / history as misleading provenance.

**Suggestion:**
```
Clamp with `slot_delay_seconds=$(( now_epoch > slot_epoch ? now_epoch - slot_epoch : 0 ))` (or equivalent) before export.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: SVG content validation is presence-only, not sanitization</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1640)

The republish path accepts any body whose first 1024 bytes contain `<svg`. That correctly rejects CDN error pages, but an SVG can embed `<script>` and executes it when navigated to directly (not via ) — and these files are republished to the same Pages origin as the dashboard. The threat is low because the source is the project's own previously published site, but the check's comment ('a poisoned manifest cannot...') implies stronger guarantees than the content check provides. Also, a legitimate SVG with more than 1KB of XML prolog/comments before the `<svg` tag would be rejected, though that is renderer-controlled today.

**Suggestion:**
```
If stronger guarantees are wanted, grep for and reject `<script` in fetched SVG bodies, or note explicitly in the comment that content is trusted because it originated from a prior authenticated publish.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: ~55-line manifest-fetch/validation block duplicated across three workflows</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/soak-checkpoint-publish.yml` (line 173)

The charts-manifest fetch + filename validation + SVG content check block is copy-pasted nearly verbatim into merge-recovery-soak.yml, soak-checkpoint-publish.yml, and soak-dashboard-pages.yml (and a lighter variant in preview-soak-dashboard.sh). The docs explicitly call out the 'must be added to all three publishers' hazard for data files; this block now has the same hazard for its own validation logic — a security-relevant fix (e.g. tightening the filename allowlist) must land in four places or the weakest copy wins.

**Suggestion:**
```
Extract into a composite action or a checked-in shell script (e.g. scripts/oci or scripts/bench) that all publishers call, so validation logic has a single source of truth.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Documentation update needed</b> (Reported by: bedrock)</summary>

**File:** `docs/soak-benchmarks.md` (line 54)

The documentation mentions the addition of new data files to workflows but does not explicitly mention the changes related to the `scripts/soak-charts` crate and its impact on the dashboard.

**Suggestion:**
```
Update the documentation to include details about the `scripts/soak-charts` crate, its role in rendering charts, and how it affects the dashboard.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Verify-after-write GET swallows errors and does exact string comparison</b> (Reported by: anthropic)</summary>

**File:** `scripts/oci/deploy-soak-scheduler.sh` (line 234)

The verification GET uses `2>/dev/null || true`, so a transient network/auth failure yields an empty `actual` and the script exits with a misleading 'rule did not persist' error rather than surfacing the real failure. Separately, `[ "$actual" != "$rule" ]` is an exact byte comparison; if OCI normalizes whitespace, quoting, or case in the stored matching rule, every deploy will fail closed even when the rule bound correctly. Also note that `exit 1` inside this function only aborts the whole script if the caller's command substitution runs under `set -e` as a simple assignment — worth confirming that is the case.

**Suggestion:**
```
Distinguish 'GET failed' from 'rule mismatch' (capture and report the CLI error), and consider normalizing both sides (trim/collapse whitespace) before comparing, or comparing the rule after a round-trip through the same GET at write time.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Pre-commit now runs cargo test (and cold-builds soak-charts) on every commit</b> (Reported by: anthropic)</summary>

**File:** `.githooks/pre-commit` (line 116)

Adding `cargo test --manifest-path scripts/soak-charts/Cargo.toml` to the pre-commit hook means every commit pays a test run, and a cold cache pays a full dependency build (charton pulls nalgebra, wasm-bindgen, etc.) that also requires network access — commits will fail offline until the crate is cached. SKIP_TESTS=1 exists, but the default-path cost lands on every contributor for a crate most commits don't touch.

**Suggestion:**
```
Gate the soak-charts fmt/clippy/test steps on whether staged files touch scripts/soak-charts/ (e.g. `git diff --cached --name-only | grep -q '^scripts/soak-charts/'`), keeping the unconditional run in pre-push/CI.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: triggerLabel does not guard against negative slot_delay_seconds</b> (Reported by: anthropic)</summary>

**File:** `.github/dashboard/index.html` (line 434)

The cron-fallback path computes `slot_delay_seconds=$((now_epoch - slot_epoch))`, which can go negative under clock skew or a slot computed slightly in the future; `triggerLabel` only checks `d >= 60`, so a negative delay silently renders as an on-time trigger, and a hypothetical negative-minute value would never surface. Not incorrect today, but the field is 'remote data' by the PR's own optional-by-construction stance.

**Suggestion:**
```
Clamp with `Math.max(0, d)` or treat negative values like missing ones.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: SVG body check is intentionally shallow (acceptable with )</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1645)

Carried chart bodies are accepted after `head -c 1024 | grep -q '<svg'`. That blocks obvious HTML error pages but not crafted non-chart content containing the substring. Risk is low because the dashboard embeds charts as `` (scripts in SVG do not run in that context) and filenames are tightly allowlisted.

**Suggestion:**
```
Optional hardening: require a leading `<svg` after optional XML/doctype prologue, enforce a max file size, and/or Content-Type sniff. Not a merge blocker given current `<img>` usage and path allowlisting.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unrelated eyre Context → WrapErr churn in crypto</b> (Reported by: xai)</summary>

**File:** `crypto/src/rust/signatures/secp256k1.rs` (line 4)

The secp256k1/key_util `with_context` → `wrap_err_with` edits are behavior-neutral API renames and sit outside the CI/scheduler/dashboard scope described in the PR. They increase review surface and raise conflict risk with parallel crypto work.

**Suggestion:**
```
Prefer landing pure renames in a tiny dedicated commit/PR unless this branch already required them for a toolchain/clippy gate.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: wrap_err_with used with static strings</b> (Reported by: anthropic)</summary>

**File:** `crypto/src/rust/signatures/secp256k1.rs` (line 37)

Several call sites use `.wrap_err_with(|| "Invalid PEM format")` with a constant string. `wrap_err_with` exists to defer formatting cost; for static strings `.wrap_err("Invalid PEM format")` is simpler and what clippy's redundant_closure family generally prefers. Also, this Context→WrapErr rename in crypto/ appears unrelated to the PR's stated scope (scheduler/dashboard); worth a note in the description or a separate commit.

**Suggestion:**
```
Use `.wrap_err(...)` for static messages, keep `.wrap_err_with(|| format!(...))` for formatted ones.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: RUSTSEC-2024-0384 (instant) acceptance not mentioned in PR description</b> (Reported by: anthropic)</summary>

**File:** `deny.toml` (line 44)

The PR description covers RUSTSEC-2026-0247 but the diff also accepts RUSTSEC-2024-0384 for the soak-charts crate's wasm-only `instant` dependency. The inline rationale is good; the omission from the description just makes the policy change easy to miss in review, and since deny.toml is shared with the workspace check, the ignore is now global (the comment acknowledges it will sit unmatched there, emitting warnings).

**Suggestion:**
```
Mention the second advisory acceptance in the PR description, or scope soak-charts to its own deny config if per-crate ignores are preferred.
```

</details>
---

<details>

This is a well-reasoned infrastructure PR: the OCI scheduler hardening (verify-after-write on dynamic-group rules, the load-bearing read-functions-family policy statements) directly codifies a live incident fix, trigger provenance is threaded end-to-end with an appropriate optional-by-construction stance for historical rows, and the workflow-side manifest handling shows real thought about remote-content validation (filename allowlisting, content sniffing, fail-closed on non-404 errors). The issues found are mostly robustness and maintainability concerns rather than blockers: the SLOT_DELAY_SECONDS numeric guard is applied inconsistently, the per-image onerror handler can hide a working chart when only one theme variant is missing, the OCI verification conflates GET failures with rule mismatches, and the ~55-line manifest-fetch block is duplicated across three workflows — recreating the exact multi-publisher drift hazard the docs warn about. The pre-commit hook's unconditional soak-charts test run adds meaningful friction (and an offline-commit failure mode) for a crate most commits don't touch. Note the diff was truncated, so the soak-charts Rust sources and some dashboard sections were not reviewed. Overall safe to merge after considering the feedback.

The PR introduces several improvements to the CI/CD pipeline, including hardening the scheduler deployment, adding trigger provenance, and upgrading the dashboard and charts. The changes look generally good, but there are a few areas that require further review, particularly around security and style consistency. The documentation should also be updated to reflect the new changes.

#### openai (gpt-5.6-sol)

The visible changes generally improve scheduler verification, trigger provenance, dependency checks, and dashboard publishing, but the chart carry-forward logic can convert a transient 404 into a permanently incomplete publication, especially in the checkpoint workflow where no render follows. The dashboard also hides valid charts when only the inactive theme asset fails, and pre-push reporting miscounts skipped checks. Review coverage is limited to the provided, truncated portion of the diff.

#### xai (grok-4.5)

Strong, well-motivated follow-through on the OCI scheduler outage: verify-after-write on dynamic-group matching rules, fail-closed policy updates that preserve the live `read functions-family` fix, and end-to-end trigger provenance into summaries, history, and the dashboard. Manifest-driven chart republish is carefully path-traversal hardened, and moving charts to pre-rendered themed SVGs with mobile/a11y tweaks is coherent. Main actionable concern is confirming the `ci-control/scripts/soak-charts` path in `merge-recovery-soak.yml`—if wrong, `continue-on-error` will silently freeze charts while history keeps updating. Remaining items are small robustness nits (theme onerror, slot_delay validation/clamp) plus out-of-scope crypto renames.
</details>

---

jeffrey-l-turner:
### Multi-agent review: major finding follow-up

The four major findings from the multi-agent review have been triaged and addressed in the follow-up working tree:

- **`ci-control/scripts/soak-charts` path:** confirmed intentional. The `perf_report` job checks the repository out with `path: ci-control`, so both the renderer manifest and binary paths are correct. A workflow invariant now guards the checkout path and both renderer paths against drift.
- **Manifest-listed SVG returning 404:** fixed fail-closed behavior in all three publishers (`merge-recovery-soak.yml`, `soak-checkpoint-publish.yml`, and `soak-dashboard-pages.yml`). A manifest may be absent during bootstrap, but once a manifest lists an SVG, a 404 now aborts instead of publishing an incomplete chart set. The workflow invariant suite guards all three copies.
- **Dynamic-group verification:** OCI GET failures are now reported distinctly rather than being swallowed as an empty matching rule. Rule contents/OCIDs are no longer echoed on mismatch, while the deploy still fails closed.
- **`SLOT_DELAY_SECONDS` validation:** `write-soak-summary.sh` now independently validates the value and normalizes invalid input to `0`, matching the soak driver's behavior. A regression test exercises direct invocation with invalid input.

Validation completed successfully:

- `scripts/bench/test-write-soak-summary.sh`
- `scripts/bench/test-aggregate-perf-report.sh`
- `scripts/bench/test-run-merge-recovery-soak.sh`
- `.github/scripts/check-workflow-invariants.sh`
- shell syntax checks, YAML diagnostics, and `git diff --check`

These follow-up changes are prepared locally and will appear on the PR after the next commit and push.

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a well-reasoned CI/infrastructure hardening PR with unusually thorough inline documentation: the OCI dynamic-group verify-after-write and the additional read functions-family policy statements correctly codify the live outage remediation and fail closed; the trigger-provenance plumbing (schedule gate → soak-segment → summary → history → dashboard) is consistently optional-by-construction for old rows; and the dashboard/renderer split with manifest-driven SVG carry-forward includes sensible input validation (filename token whitelist, SVG sniffing, jq shape checks). The main concern is a logic tension: manifest-listed SVGs that 404 hard-fail all three publishers, which contradicts the PR's own 'charts are cosmetic, never block the durable publish' principle and creates a path where a transient CDN blip or manifest/SVG race blocks publishing real soak history. Secondary concerns are the byte-exact OCI matching-rule comparison (fragile to server normalization), the triplicated carry-forward block, the unvalidated trigger_source string, and the unconditional (and un-locked) soak-charts test run added to pre-commit. Note the diff was truncated, so the soak-charts Rust sources and preview binary changes were not reviewed.

The scheduler verification and trigger-provenance propagation appear carefully implemented, and the manifest filename validation avoids straightforward path traversal. The main concern is the chart publisher's failure handling: continue-on-error does not preserve prior charts because rendering mutates the publish directory in place, allowing partial output to be deployed. Rendering should be staged and committed only after complete validation. There is also a smaller dashboard issue where failure of either theme asset hides both variants. Review is limited to the visible, truncated portion of the diff.

Solid follow-through on the OCI scheduler outage: verify-after-write on dynamic groups, the extra functions-family read policy statements, and end-to-end trigger_source/slot_delay_seconds plumbing are well motivated and generally fail-closed. Dashboard work (pre-rendered SVG pairs, manifest-driven carry-forward, mobile tab layout, contrast fix) is thoughtful, and remote manifest entries are path-allowlisted before fetch. Main follow-ups: confirm the merge-recovery-soak perf_report job actually checks out ci-control before the new render step (the invariant expects it; it is not visible in the provided hunks), consider deduplicating the triplicated chart-fetch bash, and align trigger_source defaults so history does not grow an extra "unknown" state. No critical security vulnerability stood out given -based SVG embedding and the filename allowlist.

The PR introduces several changes to harden the scheduler deployment, surface trigger provenance, and upgrade the dashboard and charts. Most changes are functional improvements with minor style suggestions for consistency and readability. The changes look good overall, but a thorough review of the entire diff is recommended to ensure no critical issues are missed.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Failed chart rendering can publish partially overwritten output</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1748)

The new 'Render dashboard charts' step builds from ci-control/scripts/soak-charts/Cargo.toml. The invariant script now requires a 'Checkout CI control files' step with path: ci-control, but that checkout is not present in the visible diff hunks. If it is missing from the workflow, every perf_report render will fail (masked by continue-on-error), and publishers will permanently serve stale carried-forward SVGs after the first successful history publish on a runner without that tree.

---

The render step is marked continue-on-error but writes directly into site/data, which already contains the carried-forward manifests and SVGs. If soak-charts fails after creating, truncating, or replacing some files, the workflow continues and uploads that partially modified directory. Therefore the documented guarantee that the previous SVGs "stand" after a rendering failure is not actually provided. The same pattern exists in soak-dashboard-pages.yml. A failure between rendering the weekend and daily series can also publish a mixed or incomplete chart generation.

**Suggestion:**
```
Confirm the perf_report job checks out the workflow ref (or default branch) into path: ci-control before the render step. If it is already present outside the truncated diff, no change needed; otherwise add the checkout and keep the invariant.
```

</details>

<details>
<summary><b>:orange_circle: Major: Missing/404 chart SVG hard-fails the entire history publish, contradicting the 'charts are cosmetic' principle</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/soak-checkpoint-publish.yml` (line 216)

The PR (and docs/soak-benchmarks.md) states that chart rendering is cosmetic and 'must never block the publish that makes soak history durable' — the render step is even continue-on-error. But the new carry-forward loops in all three publishers (merge-recovery-soak.yml, soak-checkpoint-publish.yml, soak-dashboard-pages.yml) `exit 1` when a manifest-listed SVG returns 404 or fails the `<svg` sniff. A single transient CDN 404 or a stale manifest (e.g., a renderer change that suppressed a chart between publishes, then a render failure on the next run) will now block publishing the run's history/summary data — the exact data-loss scenario these workflows are designed to prevent. The manifest and the SVGs are fetched at different instants, so a publish racing a Pages deploy can legitimately see a manifest that names files not yet (or no longer) present.

**Suggestion:**
```
On SVG 404, drop the missing SVG from the carried manifest (or drop the whole chart set for that series) and warn, rather than exiting 1. Reserve hard failure for the history/summary JSON files, whose loss is genuinely unrecoverable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Environment variable naming consistency</b> (Reported by: bedrock)</summary>

**File:** `.githooks/pre-commit` (line 42)

The environment variable `SKIP_TESTS` is introduced in the pre-commit hook. Ensure that the naming convention for environment variables is consistent across the repository.

**Suggestion:**
```
Consider renaming `SKIP_TESTS` to `SKIP_SOAK_TESTS` to avoid potential conflicts with other tests.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Pre-commit runs the soak-charts test suite unconditionally on every commit</b> (Reported by: anthropic)</summary>

**File:** `.githooks/pre-commit` (line 115)

The new `cargo test --manifest-path scripts/soak-charts/Cargo.toml` executes on every commit regardless of whether any soak-charts file was staged. First-run compilation of the charton/nalgebra tree adds noticeable latency to unrelated node-workspace commits. It also omits `--locked`, unlike the pre-push and CI invocations, so a pre-commit run can silently update the crate's committed Cargo.lock — the very supply-chain pin the PR emphasizes.

**Suggestion:**
```
Gate the test (and the added fmt/clippy passes) on staged changes under scripts/soak-charts/, and add --locked for consistency with pre-push and the publishers.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: CSS variable naming consistency</b> (Reported by: bedrock)</summary>

**File:** `.github/dashboard/index.html` (line 12)

The CSS variables for series colors have been updated. Ensure that the naming convention for CSS variables is consistent across the repository.

**Suggestion:**
```
Consider prefixing all CSS variables with a common prefix, e.g., `--soak-series-1` instead of `--series-1`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Failure of an inactive theme image hides the usable chart</b> (Reported by: openai)</summary>

**File:** `.github/dashboard/index.html` (line 568)

Both light and dark images use onerror to hide their closest figure. Browsers load images even when CSS sets display:none, so a missing or invalid dark SVG will hide the entire figure in light mode even when the light SVG loaded successfully, and vice versa. This makes one broken theme variant unnecessarily remove the valid variant.

**Suggestion:**
```
Hide only the image that failed, then hide the figure only if both theme variants have failed. An explicit load/error state in JavaScript would make this behavior reliable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Chart manifest carry-forward logic duplicated in three workflows</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1605)

The same manifest download, filename allowlist, SVG sniff, and fail-closed 404 behavior is copy-pasted into merge-recovery-soak.yml, soak-checkpoint-publish.yml, and soak-dashboard-pages.yml (plus a softer variant in preview-soak-dashboard.sh). The new invariant only asserts the 404 error string exists—it will not catch behavioral drift between copies (e.g. one workflow keeps treating 404 as continue and another exits).

**Suggestion:**
```
Extract a shared script (e.g. scripts/bench/fetch-published-charts.sh) and invoke it from all publishers/preview. Keep the invariant, but point it at the shared script.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: SVG body check is a shallow heuristic</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1605)

Carried chart bodies are accepted after head -c 1024 | grep -q '<svg'. That blocks obvious HTML/JSON error pages, but any 200 response whose first 1KB contains the substring '<svg' will be republished. Risk is limited because the dashboard loads charts via  (scripts in SVG do not run), and filenames are tightly allowlisted, so this is defense-in-depth rather than an open XSS path.

**Suggestion:**
```
Prefer checking the start of the payload (e.g. optional XML declaration then <svg) or file(1)/a small magic check, and reject bodies above a sane size cap before writing into the Pages artifact.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: ~55-line manifest/SVG carry-forward block duplicated verbatim across three workflows</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1608)

The manifest fetch + filename validation + SVG fetch/sniff loop is copy-pasted into merge-recovery-soak.yml, soak-checkpoint-publish.yml, and soak-dashboard-pages.yml (and a softer variant in scripts/preview-soak-dashboard.sh). The invariant checker only greps for one error string, so the blocks can drift in validation logic (they already drift in error wording: 'no published series is lost' vs 'live dashboard data is not overwritten') without the check firing.

**Suggestion:**
```
Extract the carry-forward logic into a shared script under .github/scripts/ (or a composite action) sourced by all three workflows, and have check-workflow-invariants.sh assert its use.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: SVG content sniff limited to first 1024 bytes</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1650)

`head -c 1024 ... | grep -q '<svg'` rejects any valid SVG whose `<svg` element appears after 1 KiB of preamble (XML declaration, DOCTYPE, license comments, embedded metadata). Today the renderer controls the output, but a future charton version emitting a longer header would make every carry-forward publish fail (and per the issue above, fail the whole publish). The same pattern is duplicated in soak-checkpoint-publish.yml and soak-dashboard-pages.yml.

**Suggestion:**
```
Increase the sniff window (e.g., 8–16 KiB) or use `grep -m1 -q '<svg'` on the whole file; SVGs here are small.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Error handling consistency</b> (Reported by: bedrock)</summary>

**File:** `crypto/src/rust/signatures/secp256k1.rs` (line 29)

The error handling in `secp256k1.rs` has been updated to use `wrap_err_with` instead of `with_context`. Ensure that the error handling is consistent across the repository.

**Suggestion:**
```
Review other files for similar error handling patterns and update them to use `wrap_err_with` for consistency.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: trigger_source default differs across the pipeline</b> (Reported by: xai)</summary>

**File:** `scripts/bench/aggregate-perf-report.sh` (line 140)

write-soak-summary.sh and the schedule gate default missing trigger_source to 'manual', while aggregate-perf-report.sh uses ($passive.trigger_source // "unknown"). Pre-field rows therefore become "unknown" in history (dashboard correctly hides them), but any summary that omits the field after this ships will be stored as "unknown" rather than "manual", which is a third state beside the documented oci/cron/manual vocabulary.

**Suggestion:**
```
Align defaults: either emit null/omit when absent and let the dashboard treat missing as dash, or use the same fallback ("manual") everywhere and reserve "unknown" only for truly unreadable legacy rows if you still need that signal.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: trigger_source is unvalidated free-form input</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/write-soak-summary.sh` (line 8)

SLOT_DELAY_SECONDS is validated against ^[0-9]+$ and defaulted, but SOAK_TRIGGER_SOURCE is passed through verbatim into summary.json and onward into published history. The dashboard whitelist-maps it (unknown values render as a dash), and the table cell escapes it, so there's no injection today — but an arbitrary string silently degrading to '-' on the dashboard will hide misconfiguration (e.g., a typo'd value in a workflow edit) the same way the original outage was hidden.

**Suggestion:**
```
Validate against the three known values (oci-resource-scheduler | cron-fallback | manual) and fall back to 'manual' (or 'unknown') with a stderr warning, mirroring the slot-delay validation.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Script readability</b> (Reported by: bedrock)</summary>

**File:** `scripts/oci/deploy-soak-scheduler.sh` (line 224)

The script `deploy-soak-scheduler.sh` has a long block of code for verifying dynamic group matching rules. Consider breaking this into smaller functions for better readability and maintainability.

**Suggestion:**
```
Extract the verification logic into a separate function, e.g., `verify_dynamic_group_matching_rule`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Exact string comparison of matching rule may false-fail on server-side normalization</b> (Reported by: anthropic)</summary>

**File:** `scripts/oci/deploy-soak-scheduler.sh` (line 236)

The verify-after-write compares the GET's `matching-rule` byte-for-byte against the locally constructed rule (`[ "$actual" != "$rule" ]`). OCI IAM is known to normalize matching rules in some tenancies (whitespace, quoting, casing of keywords like `ALL`/`Any`). If normalization occurs, every deploy will fail closed even though the rule persisted correctly — turning the safety check into a permanent deploy blocker.

**Suggestion:**
```
Normalize both sides before comparing (e.g., collapse whitespace, case-fold keywords), or at minimum treat only `null`/empty as a hard failure and log a warning plus both values on textual mismatch so an operator can distinguish 'rule dropped' from 'rule reformatted'.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: triggerLabel shows delay for oci and manual runs too, not only cron</b> (Reported by: anthropic)</summary>

**File:** `.github/dashboard/index.html` (line 434)

The comment says 'an on-time oci dispatch stays a bare oci', which is accurate, but a late oci dispatch renders 'oci +Nm' and a manual run with a (hypothetical, currently always-0) recorded delay would render 'manual +Nm'. That's arguably the right behavior — late scheduler dispatches are exactly what this PR wants visible — but the PR description ('oci', 'cron +95m', 'manual') implies delay only annotates cron. Worth confirming this is intended and covered by the description.

**Suggestion:**
```
No code change needed if intentional; consider a one-line comment noting the delay suffix applies to any source, since a late oci dispatch is itself a scheduler-health signal.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unrelated eyre Context → WrapErr churn in crypto</b> (Reported by: xai)</summary>

**File:** `crypto/src/rust/signatures/secp256k1.rs` (line 1)

secp256k1.rs and key_util.rs only rename Context/with_context to WrapErr/wrap_err_with. That is unrelated to scheduler hardening, provenance, or dashboard charts and widens review/bisect surface on a security-sensitive crate.

**Suggestion:**
```
Split into a separate chore commit/PR unless this branch cannot compile without it (e.g. eyre API break from a lockfile bump).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: wrap_err_with with constant strings; unrelated to PR scope</b> (Reported by: anthropic)</summary>

**File:** `crypto/src/rust/signatures/secp256k1.rs` (line 37)

The Context→WrapErr migration in crypto/ uses `wrap_err_with(|| "Invalid PEM format")` with closures over string constants — `wrap_err("Invalid PEM format")` is the idiomatic form when there is no formatting cost to defer. Also, these crypto edits (and key_util.rs) are unrelated to the CI/dashboard/scheduler scope of this PR and are not mentioned in the description.

**Suggestion:**
```
Use `wrap_err(...)` for constant messages; consider splitting incidental crypto refactors into their own commit/PR or noting them in the description.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: PR description advisory ID does not match deny.toml hunk</b> (Reported by: xai)</summary>

**File:** `deny.toml` (line 41)

The PR body says it accepts RUSTSEC-2026-0247 (bitmaps via imbl, same as #228/#229), but the diff adds RUSTSEC-2024-0384 (instant via soak-charts). If 2026-0247 is already on dev, fine; otherwise the commit gate story in the description is misleading for reviewers.

**Suggestion:**
```
Adjust the PR description to name RUSTSEC-2024-0384 (and note the imbl ignore only if it is actually part of this branch).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Repository-wide advisory exception can mask future reachable uses</b> (Reported by: openai)</summary>

**File:** `deny.toml` (line 44)

RUSTSEC-2024-0384 is ignored in the repository-wide deny configuration based on its current reachability only through the standalone chart crate's wasm dependency graph. If the workspace later acquires a native or otherwise reachable dependency on instant, the global exception will continue suppressing the advisory without drawing attention to the changed exposure.

**Suggestion:**
```
Use a separate cargo-deny configuration for scripts/soak-charts if practical, or add an invariant/dependency-tree check that ensures instant remains confined to the documented target-specific path.
```

</details>
---

<details>

This is a well-reasoned CI/infrastructure hardening PR with unusually thorough inline documentation: the OCI dynamic-group verify-after-write and the additional read functions-family policy statements correctly codify the live outage remediation and fail closed; the trigger-provenance plumbing (schedule gate → soak-segment → summary → history → dashboard) is consistently optional-by-construction for old rows; and the dashboard/renderer split with manifest-driven SVG carry-forward includes sensible input validation (filename token whitelist, SVG sniffing, jq shape checks). The main concern is a logic tension: manifest-listed SVGs that 404 hard-fail all three publishers, which contradicts the PR's own 'charts are cosmetic, never block the durable publish' principle and creates a path where a transient CDN blip or manifest/SVG race blocks publishing real soak history. Secondary concerns are the byte-exact OCI matching-rule comparison (fragile to server normalization), the triplicated carry-forward block, the unvalidated trigger_source string, and the unconditional (and un-locked) soak-charts test run added to pre-commit. Note the diff was truncated, so the soak-charts Rust sources and preview binary changes were not reviewed.

The PR introduces several changes to harden the scheduler deployment, surface trigger provenance, and upgrade the dashboard and charts. Most changes are functional improvements with minor style suggestions for consistency and readability. The changes look good overall, but a thorough review of the entire diff is recommended to ensure no critical issues are missed.

#### openai (gpt-5.6-sol)

The scheduler verification and trigger-provenance propagation appear carefully implemented, and the manifest filename validation avoids straightforward path traversal. The main concern is the chart publisher's failure handling: continue-on-error does not preserve prior charts because rendering mutates the publish directory in place, allowing partial output to be deployed. Rendering should be staged and committed only after complete validation. There is also a smaller dashboard issue where failure of either theme asset hides both variants. Review is limited to the visible, truncated portion of the diff.

#### xai (grok-4.5)

Solid follow-through on the OCI scheduler outage: verify-after-write on dynamic groups, the extra functions-family read policy statements, and end-to-end trigger_source/slot_delay_seconds plumbing are well motivated and generally fail-closed. Dashboard work (pre-rendered SVG pairs, manifest-driven carry-forward, mobile tab layout, contrast fix) is thoughtful, and remote manifest entries are path-allowlisted before fetch. Main follow-ups: confirm the merge-recovery-soak perf_report job actually checks out ci-control before the new render step (the invariant expects it; it is not visible in the provided hunks), consider deduplicating the triplicated chart-fetch bash, and align trigger_source defaults so history does not grow an extra "unknown" state. No critical security vulnerability stood out given -based SVG embedding and the filename allowlist.
</details>

---

PR #233, opened 2026-08-11
feat(soak): emit per-node CPU utilization

jeffrey-l-turner:
- extract per-node CPU peaks and aggregate them into the dashboard cluster grid
- preserve saturation semantics while carrying grid data through performance reports
- cover the soak summary pipeline and document the per-core monitoring follow-up

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/233"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789072137&installation_model_id=426883&pr_number=233&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F233&signature=ad27a45b7e39984e9fb3d9dc51c173de7d45d5cb6f5751f2f1786c2d11fe9980"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a well-constructed change that threads per-node CPU peaks from the soak harness's resource-timeseries CSV through iteration metrics, the summary rollup (`cpu_peak_core_grid_pct` under an 'all' core row), the performance report passthrough, and the dashboard heatmap renderer, with the color-domain clamp fixing a real rendering problem (multi-core outliers washing out the ramp). Documentation and backlog entries are thorough, tests cover the happy path, null semantics, and non-numeric filtering, and fail-soft behavior is deliberate throughout. The main concerns are minor robustness gaps: awk's locale-sensitive `%.1f` can produce invalid JSON that the jq guard silently discards, the sanitization of node names can silently merge keys, and the claim that saturated heatmap cells print their unclamped value should be verified against the (unshown) text layer now that the dataset column carries clamped values. None of these block merge.

The current node-level CPU peak extraction, cell-wise aggregation, report propagation, renderer clamping, and associated tests are coherent for normal harness-generated container names. The main concerns are that the documented claim of a fully per-core-ready pipeline does not match the newly hard-coded flat-map/"all" schema, and that manually constructing JSON in awk is unnecessarily fragile and can silently discard all node metrics on malformed input.

Solid, well-tested plumbing to surface per-node CPU peaks as cpu_peak_core_grid_pct (under a synthetic "all" core), thread them through summary/aggregate reports, and fix heatmap color semantics so multi-core "all" outliers no longer wash out the scale. Main follow-up is confirming that the new clamp in render_cpu_grid affects only color and not the printed cell values the comments promise; secondary nits are awk-built JSON robustness and documenting why overall cpu_peak_pct can disagree with the per-node grid in tests. No security concerns for this soak/bench path.

The code change introduces new functionality for emitting per-node CPU utilization data and updates related scripts and documentation. There are some minor logic and style issues that need to be addressed before merging. Overall, the changes look good with minor adjustments needed.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Invalid JSON value</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/test-write-soak-summary.sh` (line 24)

The JSON object contains an invalid value 'not-a-number' for the key 'weird'. This should be a number or removed if it's not relevant.

**Suggestion:**
```
Ensure all values in JSON objects are valid and relevant. Remove or correct invalid values.
```

</details>

<details>
<summary><b>:orange_circle: Major: Clamped color values may also clamp cell labels</b> (Reported by: xai)</summary>

**File:** `scripts/soak-charts/src/main.rs` (line 885)

render_cpu_grid now writes only the clamped vector into the dataset column "% of one core" and pins the domain to [0, 100]. The updated comment says saturated cells still print the true value so readers can see how far past one core a reading went (e.g. 260%). If the text/label layer reads that same dataset column (likely, since the previous code put raw pct there), labels will show 100 for every saturated cell and the comment/invariant is false. Confirm labels still use the unclamped pct values (separate column or parallel array); if not, add an unclamped column for text.

**Suggestion:**
```
Keep raw values available for labels, e.g. .with_column("% of one core", pct.clone())? for text and a separate clamped column (or map in the text mark) for color encoding on a fixed [0, 100] domain.
```

</details>

<details>
<summary><b>:orange_circle: Major: Key sanitization can silently merge or empty node names</b> (Reported by: anthropic, openai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 243)

The awk implementation constructs JSON manually and handles backslashes and quotes by deleting them. This can alter identifiers, cause distinct node names to collide, and does not escape other JSON control characters. If any generated key makes the object invalid, the later jq validation replaces the entire per-node result with {}, silently discarding valid readings from every node.

---

`gsub(/[\"]/, "", k)` strips quotes/backslashes to keep the JSON valid, but two distinct container names differing only by those characters would collide into one key (peak of whichever wins), and a pathological name could sanitize to an empty string key `""`. Additionally, other JSON-hostile characters (control chars) are not stripped, though the downstream jq validation would catch invalid output. The risk is low since names come from the harness's own Docker containers, but the failure mode is a silent data merge rather than a rejection.

**Suggestion:**
```
Consider a stricter allowlist (e.g., replace anything outside `[A-Za-z0-9._-]` with `_`) so sanitization is predictable and collisions/empty keys can't occur.
```

</details>

<details>
<summary><b>:orange_circle: Major: awk printf '%.1f' is locale-sensitive and can silently drop all per-node data</b> (Reported by: anthropic, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 247)

The awk END block uses printf "%.1f" to serialize CPU values into JSON. Under locales where LC_NUMERIC uses a comma decimal separator (e.g., de_DE), awk emits values like '10,5', producing invalid JSON. The caller's `jq -e 'type == "object"'` guard then fails and silently replaces the whole map with '{}', so per-node CPU metrics vanish with no diagnostic on such systems.

---

iteration_cpu_peak_per_node_percent builds JSON with printf and only strips backslash and double-quote from names. Other characters (newlines, tabs, control chars) can yield invalid JSON. The caller fail-softs to {} after a jq type check, so this will not crash the run, but it can silently drop an entire iteration's per-node CPU map if any container name is odd.

**Suggestion:**
```
Prefer emitting TSV/lines from awk and assembling the object with jq -n --rawfile/--argjson, or escape via jq --arg, so invalid names cannot invalidate the whole object.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Per-core pipeline is not yet schema-ready as documented</b> (Reported by: openai)</summary>

**File:** `docs/Backlog.md` (line 46)

The documentation says arbitrary core IDs are already handled downstream and that real core IDs will simply replace the "all" row when the monitor emits per-core counters. However, iteration_cpu_peak_per_node_percent currently produces a flat node-to-number map, emit_iteration_metrics stores that as cpu_peak_per_node_pct, and write-soak-summary.sh always wraps each value under a hard-coded "all" key. There is currently no path for a monitor-provided core ID to reach cpu_peak_core_grid_pct.

**Suggestion:**
```
Either revise the documentation to state that driver and summary schema changes will also be needed for per-core telemetry, or introduce a nested node-to-core map now and make the summary aggregation preserve supplied core IDs while using "all" only for aggregate samples.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent comment formatting</b> (Reported by: bedrock)</summary>

**File:** `scripts/soak-charts/src/main.rs` (line 852)

The comments in the code have inconsistent formatting, which can make the code harder to read.

**Suggestion:**
```
Standardize the formatting of comments for better readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Verify printed cell values still use unclamped percentages</b> (Reported by: anthropic)</summary>

**File:** `scripts/soak-charts/src/main.rs` (line 890)

The rect layer's dataset column now carries `clamped` values (capped at 100). The doc comment claims 'the printed value carries how far beyond' 100%, but that only holds if the text/label layer (not shown in this diff) is built from the original `pct` vector or a separate dataset. If the text layer reads the same `% of one core` column from this dataset, saturated cells will print '100.0' instead of the real value (e.g., 260), breaking the documented invariant that saturated cells carry their actual value.

**Suggestion:**
```
Confirm the text-annotation layer sources unclamped `pct` values. If it shares the clamped dataset column, add a second column (e.g., 'actual %') for the label encoding while keeping the clamped column for color.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Grid rollup jq expression is dense; comment carries the weight</b> (Reported by: anthropic, xai)</summary>

**File:** `scripts/bench/write-soak-summary.sh` (line 106)

The `cpu_peak_core_grid_pct` computation (flatten entries → group_by(.key) → cell-wise max → wrap under {all: ...}) is correct and well-tested (including the non-numeric 'weird' value filter), but the single-expression pipeline is hard to scan. The extensive comment mitigates this well.

---

cpu_peak_core_grid_pct always nests peaks under {all: max}. That matches today's container-level samples and the backlog note, but the rollup will need a real merge path when iterations start carrying genuine core ids (or mixed shapes).

**Suggestion:**
```
No change required; optionally factor the rollup into a named jq function (like the existing `numeric_values`/`max_or_null` helpers) for readability, e.g. `def per_node_grid: ...`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Good test coverage for the new pipeline</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/test-run-merge-recovery-soak.sh` (line 121)

The tests cover the full path well: per-node extraction in the driver (cpu_peak_pct == 30 aggregate vs {validator1:10, validator2:20} per-node), the summary rollup with cell-wise max across iterations, null semantics for absent data, non-numeric value filtering, and report passthrough. Object equality via jq `==` correctly avoids awk's nondeterministic key ordering.

**Suggestion:**
```
Consider one additional case: an iteration whose CSV yields no matching rows (all `__system__` or malformed cpu column), asserting `cpu_peak_per_node_pct == null`, to lock in the empty-map → null behavior in `emit_iteration_metrics`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Assert consistency between cpu_peak_pct and per-node peaks</b> (Reported by: xai)</summary>

**File:** `scripts/bench/test-run-merge-recovery-soak.sh` (line 121)

The test expects cpu_peak_pct == 30 while cpu_peak_core_grid_pct is validator1=10, validator2=20. That is only coherent if overall peak is not defined as max(per-node peaks) (e.g. max of per-timestamp sums, or different filters). Worth an explicit comment in the test or script so future readers do not "fix" one of the numbers.

**Suggestion:**
```
Document the intended relationship between iteration_cpu_peak_percent and iteration_cpu_peak_per_node_percent next to both helpers and in the test fixture setup.
```

</details>
---

<details>

This is a well-constructed change that threads per-node CPU peaks from the soak harness's resource-timeseries CSV through iteration metrics, the summary rollup (`cpu_peak_core_grid_pct` under an 'all' core row), the performance report passthrough, and the dashboard heatmap renderer, with the color-domain clamp fixing a real rendering problem (multi-core outliers washing out the ramp). Documentation and backlog entries are thorough, tests cover the happy path, null semantics, and non-numeric filtering, and fail-soft behavior is deliberate throughout. The main concerns are minor robustness gaps: awk's locale-sensitive `%.1f` can produce invalid JSON that the jq guard silently discards, the sanitization of node names can silently merge keys, and the claim that saturated heatmap cells print their unclamped value should be verified against the (unshown) text layer now that the dataset column carries clamped values. None of these block merge.

The code change introduces new functionality for emitting per-node CPU utilization data and updates related scripts and documentation. There are some minor logic and style issues that need to be addressed before merging. Overall, the changes look good with minor adjustments needed.

#### openai (gpt-5.6-sol)

The current node-level CPU peak extraction, cell-wise aggregation, report propagation, renderer clamping, and associated tests are coherent for normal harness-generated container names. The main concerns are that the documented claim of a fully per-core-ready pipeline does not match the newly hard-coded flat-map/"all" schema, and that manually constructing JSON in awk is unnecessarily fragile and can silently discard all node metrics on malformed input.

#### xai (grok-4.5)

Solid, well-tested plumbing to surface per-node CPU peaks as cpu_peak_core_grid_pct (under a synthetic "all" core), thread them through summary/aggregate reports, and fix heatmap color semantics so multi-core "all" outliers no longer wash out the scale. Main follow-up is confirming that the new clamp in render_cpu_grid affects only color and not the printed cell values the comments promise; secondary nits are awk-built JSON robustness and documenting why overall cpu_peak_pct can disagree with the per-node grid in tests. No security concerns for this soak/bench path.
</details>

---

jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.93 |

---

### Summary

This PR wires per-node (and optionally per-core) CPU peak telemetry from the soak harness through iteration metrics, the summary rollup, and into the dashboard's node × core heatmap, with a sensible 'all' fallback row for nodes lacking per-core data. The design is fail-soft throughout (JSON validation with '{}' fallbacks, null when no data), documentation and test coverage are strong, and the jq rollup semantics match the tests. The main concerns are consistency and robustness gaps in the new awk extractors: the per-core extractor omits the __system__ filter present in its per-node sibling, the core-id field is unvalidated, JSON keys are sanitized by stripping rather than escaping, and locale-sensitive printf could silently discard metrics. On the Rust side, the color-clamp change is correct in intent, but it should be verified that the printed-value text layer still uses unclamped values as the doc comment promises. None of these block merge, but the __system__ filter and text-layer clamping deserve a look before landing.

The implementation consistently carries per-node and per-core CPU peaks through iteration metrics, soak summaries, aggregate reports, and chart rendering, with useful coverage for mixed real-core and fallback data. No security, correctness, or significant performance problems are evident in the diff. The only actionable concern is stale commentary that still presents per-core telemetry as future work despite this change consuming the paired harness output.

This is a solid, well-tested plumbing change: per-node and optional per-core CPU peaks are extracted fail-soft from harness CSVs, rolled up cell-wise into `cpu_peak_core_grid_pct` with a clean real-core-vs-`all` fallback policy, passed through the perf-report aggregator, and covered by focused summary/soak tests plus docs. The main concerns are a likely-wrong `cpu_peak_pct == 30` soak assertion (max of 10/20 should be 20), and the chart color clamp possibly also clamping the printed ≥100% cell labels if they share the same column—worth verifying before merge so multi-core `all` rows do not all read as exactly 100. Smaller nits: align `__system__` filtering in the per-core awk path and consider numeric core ordering later.

The code changes look good overall. The new feature for emitting per-node CPU utilization is well-implemented. Minor improvements include updating the backlog item status, adding comments for new fields and logic, and ensuring documentation is up-to-date.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Per-core extractor does not filter out __system__ rows</b> (Reported by: anthropic, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 268)

`iteration_cpu_peak_per_node_percent` filters `$2 != "__system__"`, but `iteration_cpu_peak_per_node_core_percent` only validates that cpu_percent is numeric. If the per-core CSV ever includes a synthetic `__system__` (or similar) node, it will appear as a grid column. Today this may be harmless if the monitor never emits that row in the per-core file, but the two extractors are inconsistent.

---

iteration_cpu_peak_per_node_percent explicitly excludes rows where $2 == "__system__", but the new iteration_cpu_peak_per_node_core_percent awk program has no such guard. If the harness monitor ever emits per-core rows for the __system__ pseudo-node (as it does for the aggregate CSV), a spurious '__system__' node column will appear in the dashboard grid, and because a node with per-core data drops its 'all' fallback in the rollup, this could distort the grid semantics.

**Suggestion:**
```
Add `$2 != "__system__"` to the per-core awk row predicate, mirroring the per-node extractor.
```

</details>

<details>
<summary><b>:orange_circle: Major: cpu_peak_pct assertion likely wrong (sum vs max)</b> (Reported by: xai)</summary>

**File:** `scripts/bench/test-run-merge-recovery-soak.sh` (line 133)

The new assertion expects `.cpu_peak_pct == 30` from resource rows with per-node peaks 10 and 20. `iteration_cpu_peak_percent` reduces with a max (and `write-soak-summary.sh` also maxes across iterations), so the expected value should be 20, not 30. 10+20=30 strongly suggests the test accidentally encoded a sum. If this assertion is currently passing, something else is feeding cpu_peak and should be explained; otherwise CI should fail.

**Suggestion:**
```
Change the assertion to `and .cpu_peak_pct == 20` (or whatever the extractor actually emits), and keep per-node expectations as the source of truth.
```

</details>

<details>
<summary><b>:orange_circle: Major: Clamping color values may also clamp printed saturation labels</b> (Reported by: xai)</summary>

**File:** `scripts/soak-charts/src/main.rs` (line 885)

The heatmap now builds the `% of one core` column from values clamped with `v.min(100.0)` and pins the domain to [0, 100]. That correctly stops multi-core `all` outliers (260%+) from washing out the ramp. However, if the ≥100% text labels still read from this same dataset column, saturated cells will print `100` instead of the true peak, contradicting the updated comment that 'the printed value carries how far beyond'. With `all` fallback rows this loses the only signal that distinguishes 110% from 260%.

**Suggestion:**
```
Keep an unclamped column (e.g. raw pct) for label/text marks and tooltip, and use the clamped column only for the color encoding/domain.
```

</details>

<details>
<summary><b>:orange_circle: Major: Clamped values may also be used by the printed-value text layer</b> (Reported by: anthropic)</summary>

**File:** `scripts/soak-charts/src/main.rs` (line 890)

The dataset column "% of one core" is now built from `clamped` instead of `pct`. The doc comment says saturated cells (>=100%) print their real value so "the printed value carries how far beyond", but if the text-annotation layer (not shown in this diff) reads the same dataset column, it will print 100.0 for every saturated cell, silently losing the actual peak (e.g. 260%). Please verify the text layer is built from the unclamped `pct` vector.

**Suggestion:**
```
Ensure the annotation/text layer uses the original `pct` values (e.g. a separate column such as "printed %" or a second dataset) so saturated cells still show the true peak.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Update backlog item status</b> (Reported by: bedrock)</summary>

**File:** `docs/Backlog.md` (line 32)

The backlog item BACKLOG-FI-003 should be updated to 'implemented' once both branches are merged.

**Suggestion:**
```
Update the status field to 'implemented' once the feature is fully merged.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Backlog status says implemented_pending_merge across two repos</b> (Reported by: xai)</summary>

**File:** `docs/Backlog.md` (line 41)

BACKLOG-FI-003 is marked implemented on paired branches and asks to remove the entry only after both merges. That is clear, but the item still reads a bit like active work product inside Backlog rather than a short-lived merge checklist.

**Suggestion:**
```
No code change required; just ensure the paired system-integration PR is linked in the merge checklist so this entry does not rot at `implemented_pending_merge`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Add comment for new field</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/aggregate-perf-report.sh` (line 177)

Add a comment explaining the purpose of the new `cpu_peak_core_grid_pct` field.

**Suggestion:**
```
Add a comment above the line to describe what `cpu_peak_core_grid_pct` represents.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Add comment for new logic</b> (Reported by: bedrock)</summary>

**File:** `scripts/bench/write-soak-summary.sh` (line 97)

Add a comment explaining the new logic for handling `cpu_peak_core_grid_pct`.

**Suggestion:**
```
Add a comment above the block to describe the new logic for aggregating per-core CPU peaks.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: CPU telemetry comment still describes per-core sampling as future work</b> (Reported by: openai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 231)

The comment above iteration_cpu_peak_per_node_percent says the grid remains at node granularity and that per-core sampling is a harness-side follow-up. This change immediately adds consumption of resource-percore-timeseries.csv, and the accompanying documentation says the paired harness implementation already emits it. The stale description makes the two extraction paths and current deployment behavior unclear.

**Suggestion:**
```
Describe cpu_peak_per_node_pct as the aggregate fallback for harnesses or providers without per-core telemetry, and state that iteration_cpu_peak_per_node_core_percent supplies real core rows when resource-percore-timeseries.csv is available. Consider making the corresponding Passive field comment in scripts/soak-charts/src/main.rs use the same wording.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: awk printf %.1f is locale-sensitive and can silently drop metrics</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 235)

Both new extractors build JSON with awk `printf "%.1f"`. Under a locale with a comma decimal separator (LC_NUMERIC), this emits `42,5`, producing invalid JSON. The caller's `jq -e 'type == "object"'` validation then falls back to '{}', so the run silently loses all per-node/per-core CPU data with no diagnostic.

**Suggestion:**
```
Prefix the awk invocations with `LC_ALL=C` (or set LC_NUMERIC=C) to guarantee dot-decimal output regardless of environment.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: JSON key sanitization strips characters instead of escaping, risking key collisions</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 240)

`gsub(/[\"]/ , "", k)` removes quotes and backslashes from node/core names rather than escaping them. Two distinct node names differing only by such characters would collapse into one key, and the peaks would silently merge via max. Other control characters (e.g. tabs) also pass through unescaped into the JSON string.

**Suggestion:**
```
Escape quotes/backslashes instead of stripping them, or restrict node/core names to a safe character class and drop non-conforming rows explicitly.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Core id field ($3) is not validated</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 274)

The per-core awk validates $4 (cpu_percent) but accepts any $3 as the core id, including empty strings or garbage, which become JSON keys like "" in the grid. Malformed rows in the CSV would then flow into the dashboard as phantom core rows.

**Suggestion:**
```
Add a predicate such as `$3 ~ /^[0-9]+$/` (or at least `$3 != ""`) to the row filter.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Snapshot loop now waits for a file older harnesses never emit</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 291)

resource-percore-timeseries.csv was added to the filename list in snapshot_iteration_monitor_outputs. If the surrounding `while :` loop's exit condition requires all listed files to be found (or retries until a timeout keyed off started_epoch), runs against a harness that predates per-core emission — explicitly a supported scenario per the fallback comments — will spin until the timeout on every iteration, adding latency to each soak iteration.

**Suggestion:**
```
Confirm the loop exits promptly when only a subset of files exists, or treat resource-percore-timeseries.csv as optional (best-effort copy outside the wait condition).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Complex grid rollup embedded as inline jq</b> (Reported by: anthropic)</summary>

**File:** `scripts/bench/write-soak-summary.sh` (line 107)

The cpu_peak_core_grid_pct rollup is a dense ~20-line jq expression inside an already large jq program. The comment block is thorough (good), but the group_by/from_entries pipeline with the $core_nodes exclusion is hard to modify safely. Test coverage in test-write-soak-summary.sh does exercise the mixed real/fallback case, which mitigates this.

**Suggestion:**
```
Consider extracting the grid computation into a named jq function (def grid: ...) at the top of the program to make the cell/fallback semantics self-documenting.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: awk JSON builder is fail-soft but brittle on odd keys</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 248)

Both new awk emitters build JSON by stripping only `\"` and `\` from keys. That is acceptable for Docker-style names and is gated by a post-hoc `jq -e 'type == "object"'` check, so bad output becomes `{}` rather than breaking metrics emission. Still, empty core ids, commas in fields, or unexpected CSV shape will silently drop data.

**Suggestion:**
```
Optional hardening: emit TSV/array pairs from awk and let `jq` build the object (`jq -n --slurpfile ...`), which avoids hand-rolled JSON entirely.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: O(nodes × cells) nested scan in per-core END block</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 280)

The END block iterates all peak cells once per node to group cores under nodes, giving quadratic behavior. Harmless at current scale (a few nodes × a few dozen cores), but it can be done in one pass by accumulating per-node core lists during the main loop.

**Suggestion:**
```
Optional: build a per-node core index during the row scan to keep output emission linear; only worth doing if cluster/core counts grow.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Core id sort is lexicographic</b> (Reported by: xai)</summary>

**File:** `scripts/soak-charts/src/main.rs` (line 855)

`id_sorted` orders core ids as strings, so once real per-core data lands you get `0,1,10,2,...` instead of numeric order. Fine for the current `all`/single-digit fixtures; slightly awkward for wider grids.

**Suggestion:**
```
When all keys parse as integers, sort numerically and keep `all` (or other non-numeric keys) pinned first/last.
```

</details>
---

<details>

This PR wires per-node (and optionally per-core) CPU peak telemetry from the soak harness through iteration metrics, the summary rollup, and into the dashboard's node × core heatmap, with a sensible 'all' fallback row for nodes lacking per-core data. The design is fail-soft throughout (JSON validation with '{}' fallbacks, null when no data), documentation and test coverage are strong, and the jq rollup semantics match the tests. The main concerns are consistency and robustness gaps in the new awk extractors: the per-core extractor omits the __system__ filter present in its per-node sibling, the core-id field is unvalidated, JSON keys are sanitized by stripping rather than escaping, and locale-sensitive printf could silently discard metrics. On the Rust side, the color-clamp change is correct in intent, but it should be verified that the printed-value text layer still uses unclamped values as the doc comment promises. None of these block merge, but the __system__ filter and text-layer clamping deserve a look before landing.

The code changes look good overall. The new feature for emitting per-node CPU utilization is well-implemented. Minor improvements include updating the backlog item status, adding comments for new fields and logic, and ensuring documentation is up-to-date.

#### openai (gpt-5.6-sol)

The implementation consistently carries per-node and per-core CPU peaks through iteration metrics, soak summaries, aggregate reports, and chart rendering, with useful coverage for mixed real-core and fallback data. No security, correctness, or significant performance problems are evident in the diff. The only actionable concern is stale commentary that still presents per-core telemetry as future work despite this change consuming the paired harness output.

#### xai (grok-4.5)

This is a solid, well-tested plumbing change: per-node and optional per-core CPU peaks are extracted fail-soft from harness CSVs, rolled up cell-wise into `cpu_peak_core_grid_pct` with a clean real-core-vs-`all` fallback policy, passed through the perf-report aggregator, and covered by focused summary/soak tests plus docs. The main concerns are a likely-wrong `cpu_peak_pct == 30` soak assertion (max of 10/20 should be 20), and the chart color clamp possibly also clamping the printed ≥100% cell labels if they share the same column—worth verifying before merge so multi-core `all` rows do not all read as exactly 100. Smaller nits: align `__system__` filtering in the per-core awk path and consider numeric core ordering later.
</details>

---

jeffrey-l-turner:
## Review triage — second multi-agent pass

Maintainer triage of the 17 findings; fixes land in the next commit on this branch.

### Fixed

- **`__system__` filter missing from the per-core extractor** (critical, Claude + Grok — confirmed): `iteration_cpu_peak_per_node_core_percent` now carries the same `$2 != "__system__"` guard as its per-node sibling, so host-state rows can never become a phantom grid node (which would also have dropped that "node"'s fallback semantics).
- **Core-id validation** (Claude): per-core rows now require `$3 ~ /^[0-9]+$/` per the telemetry contract — malformed rows are rejected, not sanitized into phantom cores.
- **Locale-sensitive `printf "%.1f"`** (Claude, third recurrence): all three CPU awk extractors now run under `LC_ALL=C`. The failure mode was real — a comma decimal would have cost the iteration its whole `metrics.json` via the `--argjson` path, not just one field.
- **Sanitization by deletion** (Claude + ChatGPT, from pass 1): node names are now sanitized to an allowlist (`[^A-Za-z0-9._-]` → `_`) instead of stripping quotes/backslashes, so distinct hostile names can no longer collide or produce empty keys.
- **Stale "per-core is future work" comment** (ChatGPT — confirmed): the per-node extractor's header now describes it as the aggregate fallback beside the per-core extractor.
- The driver test fixture now includes a `__system__` per-core row and a non-numeric core id, and its grid assertion proves both filters.

### False positives (with evidence, to close the recurring threads)

- **"`cpu_peak_pct == 30` should be 20 (sum vs max)"** (Grok): the aggregate extractor deliberately sums `cpu_percent` across nodes **per timestamp** and peaks that total — "how hot did the whole shard run at once". Two nodes at 10% and 20% in the same sample correctly yield 30. This is now documented on the helper so the two questions (shard-wide peak vs per-node peak) can't be conflated again.
- **"Clamped color values also clamp the printed labels"** (Claude + Grok, both passes): the ≥100% labels are built from the untouched `pct` vector in a separate dataset — only the color channel sees the clamp; a 263% cell renders max-red and prints 263. A comment now says so at the clamp site.
- **"Snapshot loop waits on the per-core file"** (Claude): `snapshot_iteration_monitor_outputs` is a best-effort copy loop — a missing filename is `continue`d, never awaited, so pre-emission harnesses add no latency.
- **"Core sort is lexicographic"** (Grok): `id_sorted` sorts numerically whenever both ids parse as integers; `0,1,2,10` is the actual order.

Remaining nits (jq rollup as a named function, awk→jq JSON assembly, O(nodes×cells) END block) are noted and deliberately deferred — all are style/scale concerns mitigated by tests at current shard sizes.

jltatbeach (approved):

PR #242, opened 2026-08-12
fix(ci): gate unreleased system-integration regressions

jeffrey-l-turner:
## Summary

- introduce a validated manifest of node capabilities implemented by the branch under test
- pass those capabilities to the pinned system-integration pytest suite
- skip forward regression tests until their corresponding node behavior is available
- enforce capability parsing and all three system-integration pin sites in required CI checks

## Why

Advancing the system-integration pin for per-core telemetry also exposed forward-looking regression tests whose node fixes had not yet reached the target branch. That made unrelated PRs fail deterministically. This change makes compatibility explicit: baseline-supported regressions run, unreleased regressions skip, and malformed capability declarations fail closed.

## Synchronized dependency

This PR is synchronized with F1R3FLY-io/system-integration#105:
https://github.com/F1R3FLY-io/system-integration/pull/105

The two PRs must be merged concurrently. This PR pins system-integration commit `fe91755ed16774a5f56f33da9712f7df1f52ca9d`; merging only this PR before that commit remains reachable from the trusted system-integration branch would break the immutable checkout contract, while merging only the system-integration PR without this consumer manifest would leave all capability-gated regressions disabled by default.

## Validation

- repository pre-commit checks
- repository pre-push checks: 15/15 passed
- workflow invariant and capability-contract tests
- system-integration capability unit tests and Ruff
- YAML diagnostics and diff checks
- verified six unreleased regressions skip and three baseline regressions remain enabled

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/242"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789130304&installation_model_id=426883&pr_number=242&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F242&signature=24c0ca1415f449400bc5e1b5d24b950248084621a9076def8af2baf1e67afee3"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/.github/oci-validation.env
+++ b/.github/oci-validation.env
@@ -14,4 +14,4 @@
 # .github/workflows/_integration-pipeline.yml — a reusable workflow cannot
 # source this file at env-resolution time, so the pin is duplicated there.
 # Bump both in the same commit.
-SYSTEM_INTEGRATION_REF=755301662e4f86ffb55f51077b88f32212a1f3ed
+SYSTEM_INTEGRATION_REF=483d7d4f1aa51d87c3eb3064531b630a36257639
--- a/.github/oci-validation.env
+++ b/.github/oci-validation.env
@@ -14,4 +14,4 @@
 # .github/workflows/_integration-pipeline.yml — a reusable workflow cannot
 # source this file at env-resolution time, so the pin is duplicated there.
 # Bump both in the same commit.
-SYSTEM_INTEGRATION_REF=483d7d4f1aa51d87c3eb3064531b630a36257639
+SYSTEM_INTEGRATION_REF=fe91755ed16774a5f56f33da9712f7df1f52ca9d
--- a/.github/scripts/read-system-integration-node-capabilities.sh
+++ b/.github/scripts/read-system-integration-node-capabilities.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+capabilities_file=${1:-.github/system-integration-node-capabilities.txt}
+seen='|'
+
+while IFS= read -r capability || [ -n "$capability" ]; do
+	capability=${capability%$'\r'}
+	case "$capability" in
+	'' | \#*) continue ;;
+	esac
+	if [[ ! "$capability" =~ ^[a-z0-9]+(-[a-z0-9]+)*$ ]]; then
+		printf 'invalid system-integration node capability: %s\n' "$capability" >&2
+		exit 1
+	fi
+	case "$seen" in
+	*"|$capability|"*)
+		printf 'duplicate system-integration node capability: %s\n' "$capability" >&2
+		exit 1
+		;;
+	esac
+	seen="${seen}${capability}|"
+	printf '%s\n' "$capability"
+done <"$capabilities_file"
--- a/.github/scripts/test-system-integration-node-capabilities.sh
+++ b/.github/scripts/test-system-integration-node-capabilities.sh
@@ -1,0 +1,53 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+ROOT=$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd)
+READER="$ROOT/.github/scripts/read-system-integration-node-capabilities.sh"
+TMP=$(mktemp -d)
+trap 'rm -rf "$TMP"' EXIT
+
+expect_failure() {
+	if "$READER" "$1" >"$TMP/stdout" 2>"$TMP/stderr"; then
+		printf 'expected capability validation to fail for %s\n' "$1" >&2
+		exit 1
+	fi
+}
+
+: >"$TMP/empty"
+test -z "$("$READER" "$TMP/empty")"
+
+cat >"$TMP/valid" <<'EOF'
+# released node behavior
+expired-deploy-admission
+observer-missing-block-retry
+EOF
+printf '%s\n' expired-deploy-admission observer-missing-block-retry >"$TMP/expected"
+"$READER" "$TMP/valid" >"$TMP/actual"
+cmp "$TMP/expected" "$TMP/actual"
+
+printf '%s\n' 'Invalid-Capability' >"$TMP/invalid"
+expect_failure "$TMP/invalid"
+
+printf '%s\n' duplicate-capability duplicate-capability >"$TMP/duplicate"
+expect_failure "$TMP/duplicate"
+
+expect_failure "$TMP/missing"
+
+for workflow in \
+	"$ROOT/.github/workflows/_integration-pipeline.yml" \
+	"$ROOT/.github/workflows/reusable-oci-validation.yml"; do
+	grep -Fq 'read-system-integration-node-capabilities.sh' "$workflow"
+	grep -Fq '"${NODE_CAPABILITY_ARGS[@]}"' "$workflow"
+done
+
+"$READER" "$ROOT/.github/system-integration-node-capabilities.txt" >/dev/null
+
+expected_ref=$(grep '^SYSTEM_INTEGRATION_REF=' "$ROOT/.github/oci-validation.env" | cut -d= -f2-)
+for workflow in \
+	"$ROOT/.github/workflows/_integration-pipeline.yml" \
+	"$ROOT/.github/workflows/merge-recovery-soak.yml"; do
+	actual_ref=$(grep -E '^[[:space:]]*SYSTEM_INTEGRATION_REF:' "$workflow" | head -1 | sed -E 's/^[^:]*:[[:space:]]*//; s/[[:space:]]*(#.*)?$//; s/"//g')
+	test "$actual_ref" = "$expected_ref"
+done
+
+printf 'system-integration node capability contract checks passed\n'
--- a/.github/scripts/test-system-integration-node-capabilities.sh
+++ b/.github/scripts/test-system-integration-node-capabilities.sh
@@ -36,8 +36,13 @@ expect_failure "$TMP/missing"
 for workflow in \
 	"$ROOT/.github/workflows/_integration-pipeline.yml" \
 	"$ROOT/.github/workflows/reusable-oci-validation.yml"; do
-	grep -Fq 'read-system-integration-node-capabilities.sh' "$workflow"
+	grep -Fq 'NODE_CAPABILITIES=$(bash "$GITHUB_WORKSPACE/.github/scripts/read-system-integration-node-capabilities.sh"' "$workflow"
+	grep -Fq 'done <<< "$NODE_CAPABILITIES"' "$workflow"
 	grep -Fq '"${NODE_CAPABILITY_ARGS[@]}"' "$workflow"
+	if grep -Fq 'done < <(bash' "$workflow"; then
+		printf 'capability reader failure is masked by process substitution in %s\n' "$workflow" >&2
+		exit 1
+	fi
 done
 
 "$READER" "$ROOT/.github/system-integration-node-capabilities.txt" >/dev/null
--- a/.github/system-integration-node-capabilities.txt
+++ b/.github/system-integration-node-capabilities.txt
@@ -1,0 +1,3 @@
+concurrent-bridge-lock-accounting
+finality-stall-recovery
+slow-peer-notification-quorum
--- a/.github/system-integration-node-capabilities.txt
+++ b/.github/system-integration-node-capabilities.txt
@@ -2,6 +2,5 @@
 # capability-gated tests at SYSTEM_INTEGRATION_REF. Add a kebab-case identifier
 # only when its behavior is present on the target branch; the reader rejects
 # malformed and duplicate entries before pytest starts.
-concurrent-bridge-lock-accounting
 finality-stall-recovery
 slow-peer-notification-quorum
--- a/.github/system-integration-node-capabilities.txt
+++ b/.github/system-integration-node-capabilities.txt
@@ -1,3 +1,7 @@
+# Node behaviors implemented by the current f1r3node baseline and required by
+# capability-gated tests at SYSTEM_INTEGRATION_REF. Add a kebab-case identifier
+# only when its behavior is present on the target branch; the reader rejects
+# malformed and duplicate entries before pytest starts.
 concurrent-bridge-lock-accounting
 finality-stall-recovery
 slow-peer-notification-quorum
--- a/.github/workflows/_integration-pipeline.yml
+++ b/.github/workflows/_integration-pipeline.yml
@@ -44,7 +44,7 @@ env:
   # the old..new range in system-integration; never replace with a tag/branch.
   # MUST be kept identical to .github/oci-validation.env (the Full OCI
   # Validation workflow sources that file) — bump both in the same commit.
-  SYSTEM_INTEGRATION_REF: 755301662e4f86ffb55f51077b88f32212a1f3ed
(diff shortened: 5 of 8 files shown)


jeffrey-l-turner:
### Synchronized merge coordination

Companion PR: F1R3FLY-io/system-integration#105
https://github.com/F1R3FLY-io/system-integration/pull/105

Please treat these as one compatibility rollout. Merge the system-integration capability gate and this consumer manifest/pin together; after either merge, verify the immutable SHA remains reachable and rerun the f1r3node heavy integration pipeline.

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.97 |

---

### Summary

This is a well-constructed CI change: the system-integration pin bump is applied consistently across all three pin sites, the new capability manifest is validated with a strict kebab-case regex and duplicate detection (which also prevents argument injection into the pytest invocation), and the contract test covers the reader's failure modes plus workflow invariants and pin consistency. The main flaw is that both workflow consumers read the manifest via process substitution (`done < <(reader ...)`), which discards the reader's exit status under bash: a malformed manifest would silently produce a partial or empty `--node-capability` set instead of failing the job, undermining the PR's own 'fail closed' guarantee at the point where it matters most. The standalone contract test in ci.yml mitigates this for typical PRs, but the in-band gap should be fixed by capturing the reader output through a command substitution assignment (which does propagate failure under set -e), ideally in a single shared script to eliminate the duplicated loop. Remaining findings are minor consistency and documentation polish. Safe to merge after addressing the fail-open reader invocation.

The capability manifest validation and pin synchronization are generally well structured, but both workflow consumers accidentally suppress validation failures because the reader is invoked through process substitution. That undermines the central fail-closed guarantee of this change and should be corrected in both pytest invocation sites before merge.

This is a solid CI compatibility change: validated capability manifest, pin synchronization across the three SYSTEM_INTEGRATION_REF sites, and contract tests are the right shape for gating unreleased system-integration regressions. The main functional gap is that both pytest invocation sites load capabilities through bash process substitution, which drops the reader’s non-zero exit status—so malformed/missing manifests will not fail those jobs despite the stated fail-closed intent (the separate contract job mitigates this only when it runs and is required). Fixing error propagation in _integration-pipeline.yml and reusable-oci-validation.yml should be done before relying on this as a hard gate; also confirm soak does not need the same capability flags after the pin bump.

The code changes look good and address the identified issues with the CI pipeline. The new scripts for reading and testing system-integration node capabilities are well-implemented. Minor style suggestions have been provided for better readability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Reader failure is masked by process substitution — capability gating fails open, not closed</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 518)

NODE_CAPABILITY_ARGS is populated via process substitution: `done < <(bash .../read-system-integration-node-capabilities.sh ...)`. In bash, the exit status of a process substitution is discarded and is not honored by `set -e` / `pipefail`. If the capabilities file is missing or invalid, the reader can exit non-zero while the while-loop simply gets empty input and pytest continues with no `--node-capability` flags. That contradicts the PR’s fail-closed contract for malformed declarations (relying only on a separate CI check that may run in parallel or be bypassed in partial workflow runs).

---

The capability list is consumed via `done < <(bash .../read-system-integration-node-capabilities.sh ...)`. In bash, the exit status of a process substitution is not checked by `set -e` (and is not part of any pipeline), so if the reader exits 1 (invalid or duplicate capability, missing file), the workflow step continues silently. Worse, because the reader streams valid lines before failing, a manifest with a bad line at the end would result in a *partial* capability set being passed to pytest. This directly contradicts the PR's stated invariant that 'malformed capability declarations fail closed'. The standalone CI contract test mitigates this for the checked-in manifest on PRs that run `ci.yml`, but the integration pipeline and reusable OCI validation workflows themselves have no in-band guarantee (e.g., if the reusable workflow is invoked from a context where the contract test job did not run, or if the manifest is modified in the same run).

---

The reader runs inside `< <(...)`, and Bash does not propagate the process substitution command's exit status to the surrounding `while` loop. If the manifest is missing, unreadable, malformed, or contains a duplicate, the reader exits nonzero but the workflow continues and invokes pytest with an empty or partially populated capability list. This contradicts the stated fail-closed behavior. The separate contract check catches committed malformed manifests, but it does not make this execution path itself fail closed.

**Suggestion:**
```
Capture output first so failure propagates, then build the args array. Example:
NODE_CAPABILITIES=$(bash "$GITHUB_WORKSPACE/.github/scripts/read-system-integration-node-capabilities.sh" "$GITHUB_WORKSPACE/.github/system-integration-node-capabilities.txt") || exit 1
NODE_CAPABILITY_ARGS=()
while IFS= read -r capability; do
  [ -z "$capability" ] || NODE_CAPABILITY_ARGS+=("--node-capability=$capability")
done <<< "$NODE_CAPABILITIES"
Apply the same pattern in reusable-oci-validation.yml.
```

</details>

<details>
<summary><b>:red_circle: Critical: Same fail-open process substitution pattern duplicated in reusable OCI validation workflow</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `.github/workflows/reusable-oci-validation.yml` (line 188)

As in the integration pipeline, a nonzero exit from the capability reader inside process substitution does not cause the shell step to fail. Pytest can therefore run with no capabilities or only the valid capabilities emitted before an invalid line, rather than rejecting the declaration.

---

Identical issue as in `_integration-pipeline.yml`: `done < <(bash ... reader ...)` swallows the reader's non-zero exit status, so an invalid manifest yields a silently partial or empty `--node-capability` argument list rather than a hard failure. Since this is a reusable workflow potentially invoked outside the context where the capability contract test runs, this pin site is the least protected of the two consumers.

---

This workflow duplicates the same `< <(bash ...read-system-integration-node-capabilities.sh...)` pattern. Reader validation failures will not abort the job here either, so capability gating can silently degrade to “no capabilities declared” (skipping gated regressions instead of failing the run).

**Suggestion:**
```
Apply the same fix as in `_integration-pipeline.yml` (command substitution into a variable, or explicit exit-status check), ideally by factoring the read-into-args logic into a single shared shell script so the two workflows cannot drift.
```

</details>

<details>
<summary><b>:orange_circle: Major: Manifest lacks a header comment explaining its contract</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/system-integration-node-capabilities.txt` (line 1)

The capabilities file is the compatibility contract between this repo and system-integration, but it has no header comments describing when to add/remove entries, naming rules, or the link to the synchronized system-integration gating mechanism.

---

The new capabilities file is the load-bearing artifact of this change (adding a line enables previously-skipped regressions; removing one disables baseline coverage), but it has no comment describing its format, its consumer (the pinned system-integration pytest suite), or the rule that entries must correspond to node behavior actually present on the branch. The reader script already supports `#` comments, so a header is free.

**Suggestion:**
```
Add a brief header comment documenting the kebab-case format, the validation script, the synchronized system-integration pin, and when to add/remove entries.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Shebang line</b> (Reported by: bedrock)</summary>

**File:** `.github/scripts/read-system-integration-node-capabilities.sh` (line 1)

Consider adding a space after the shebang for better readability: `#!/usr/bin/env bash`

**Suggestion:**
```
Add a space after `#!/usr/bin/env`
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Shebang line</b> (Reported by: bedrock)</summary>

**File:** `.github/scripts/test-system-integration-node-capabilities.sh` (line 1)

Consider adding a space after the shebang for better readability: `#!/usr/bin/env bash`

**Suggestion:**
```
Add a space after `#!/usr/bin/env`
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Pin-extraction from oci-validation.env does not trim trailing whitespace</b> (Reported by: anthropic)</summary>

**File:** `.github/scripts/test-system-integration-node-capabilities.sh` (line 48)

`expected_ref=$(grep '^SYSTEM_INTEGRATION_REF=' ... | cut -d= -f2-)` does not strip trailing whitespace, while the workflow-side `actual_ref` extraction does (`s/[[:space:]]*(#.*)?$//`). A stray trailing space in `oci-validation.env` (which ci.yml's `envval` tolerates via its own sed) would make this contract test fail with a confusing mismatch even though the pins are logically identical.

**Suggestion:**
```
Pipe the env value through the same trailing-whitespace strip used for the YAML side, e.g., append `| sed -E 's/[[:space:]]*$//'`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Capability-to-args logic duplicated verbatim across two workflows</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/reusable-oci-validation.yml` (line 185)

The four-line NODE_CAPABILITY_ARGS construction block is copy-pasted in `_integration-pipeline.yml` and `reusable-oci-validation.yml`. The contract test only greps for the presence of the script invocation and the array expansion, so the loop bodies themselves can silently diverge (e.g., one gets the fail-closed fix and the other does not).

**Suggestion:**
```
Move the read-and-emit-args logic into `.github/scripts/` (e.g., a script that prints ready-to-use `--node-capability=` args or is sourced to populate the array) so both workflows invoke a single implementation.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Duplicate detection via string containment is brittle long-term</b> (Reported by: xai)</summary>

**File:** `.github/scripts/read-system-integration-node-capabilities.sh` (line 16)

Duplicate checks use `case "$seen" in *"|$capability|"*)`, which is correct today because the regex restricts names to `[a-z0-9-]+`. It is harder to reason about than an explicit set and will get clumsier if capability naming rules ever expand.

**Suggestion:**
```
Prefer a bash 4+ associative array (`declare -A seen`) for membership tests, keeping the same validation regex and exit-on-duplicate behavior.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Empty-array expansion relies on bash >= 4.4 semantics</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 538)

`"${NODE_CAPABILITY_ARGS[@]}"` with an empty array errors under `set -u` on bash < 4.4. GitHub-hosted ubuntu runners ship bash 5 so this is safe today, but the pattern is a latent portability trap if these steps are ever run on self-hosted runners with an older bash (macOS ships 3.2).

**Suggestion:**
```
Either note the bash >= 4.4 requirement in a comment or use the `${NODE_CAPABILITY_ARGS[@]+"${NODE_CAPABILITY_ARGS[@]}"}` idiom for defensive portability.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Soak pin updated but capabilities not passed</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 170)

SYSTEM_INTEGRATION_REF is advanced to the capability-aware system-integration commit, but this workflow does not invoke the capabilities reader or pass `--node-capability` flags. If soak executes any of the newly capability-gated pytest regressions, they will remain skipped by default even when this branch implements the behavior.

**Suggestion:**
```
Confirm soak does not run capability-gated shared/custom regressions. If it does, wire in the same manifest → `--node-capability` path (and extend the contract test to cover this workflow).
```

</details>
---

<details>

This is a well-constructed CI change: the system-integration pin bump is applied consistently across all three pin sites, the new capability manifest is validated with a strict kebab-case regex and duplicate detection (which also prevents argument injection into the pytest invocation), and the contract test covers the reader's failure modes plus workflow invariants and pin consistency. The main flaw is that both workflow consumers read the manifest via process substitution (`done < <(reader ...)`), which discards the reader's exit status under bash: a malformed manifest would silently produce a partial or empty `--node-capability` set instead of failing the job, undermining the PR's own 'fail closed' guarantee at the point where it matters most. The standalone contract test in ci.yml mitigates this for typical PRs, but the in-band gap should be fixed by capturing the reader output through a command substitution assignment (which does propagate failure under set -e), ideally in a single shared script to eliminate the duplicated loop. Remaining findings are minor consistency and documentation polish. Safe to merge after addressing the fail-open reader invocation.

The code changes look good and address the identified issues with the CI pipeline. The new scripts for reading and testing system-integration node capabilities are well-implemented. Minor style suggestions have been provided for better readability.

#### openai (gpt-5.6-sol)

The capability manifest validation and pin synchronization are generally well structured, but both workflow consumers accidentally suppress validation failures because the reader is invoked through process substitution. That undermines the central fail-closed guarantee of this change and should be corrected in both pytest invocation sites before merge.

#### xai (grok-4.5)

This is a solid CI compatibility change: validated capability manifest, pin synchronization across the three SYSTEM_INTEGRATION_REF sites, and contract tests are the right shape for gating unreleased system-integration regressions. The main functional gap is that both pytest invocation sites load capabilities through bash process substitution, which drops the reader’s non-zero exit status—so malformed/missing manifests will not fail those jobs despite the stated fail-closed intent (the separate contract job mitigates this only when it runs and is required). Fixing error propagation in _integration-pipeline.yml and reusable-oci-validation.yml should be done before relying on this as a hard gate; also confirm soak does not need the same capability flags after the pin bump.
</details>

---

jeffrey-l-turner:
### Multi-agent review follow-up

The two critical findings and the major documentation finding have been addressed in the working tree and validated:

- both integration consumers now capture the capability reader with checked command substitution before parsing, so missing, malformed, duplicate, or partially emitted manifests abort the shell step instead of degrading to an empty/partial capability set;
- the contract test now requires that checked pattern in both workflows and rejects the old process-substitution form;
- `.github/system-integration-node-capabilities.txt` now documents its compatibility role, kebab-case format, and add/remove rule.

Validation passed: capability contract tests, explicit invalid/missing-manifest fail-closed probes, workflow invariants, Bash syntax, YAML diagnostics, and `git diff --check`.

These corrections are local and still require a follow-up commit/push before the PR is merge-ready. The synchronized merge requirement with F1R3FLY-io/system-integration#105 remains unchanged.

Review triage: the shebang findings are false positives (`#!/usr/bin/env bash` is already correct), and the soak workflow runs only unmarked `test_load.py`, so it does not require capability arguments.

jeffrey-l-turner:
### Review fixes committed

The critical and major multi-agent review fixes are now committed as `8724ee96` (`fix(ci): fail closed on capability manifest errors`).

The branch is one commit ahead of the currently published PR head; push this commit to make the fixes visible to CI and reviewers. After push, rerun the required checks and keep the synchronized merge with F1R3FLY-io/system-integration#105.

jeffrey-l-turner:
### Correction: review-fix commit is published

The remote branch and PR head are now at `8724ee96703a10a2065193d6e468e328966e1afe`; the preceding note saying a push was still required is stale. CI has started on the corrected head. The synchronized merge requirement with F1R3FLY-io/system-integration#105 remains unchanged.

jltatbeach (approved):

PR #248, opened 2026-08-12
fix(casper): exactly-once at the merge

spreston8:
Layer 1 of the ground-up consensus stack: exactly-once at the merge.

- Settled-in-base dedup sentinel: a sig whose effect is already in the merge base drops every scope copy — the base/scope-boundary blindness that double-applied sibling inclusions is closed.
- Create-side pre-state guard: a block never executes a deploy whose effect its own pre-state already carries (the reinstated-effect re-execution class).
- Also bounds the whole RhoSpec pipeline so a wedged pos_spec fails fast and labeled instead of hanging the suite.

### DIFF
--- a/casper/src/rust/api/block_api.rs
+++ b/casper/src/rust/api/block_api.rs
@@ -1693,18 +1693,17 @@ impl BlockAPI {
                             .iter()
                             .map(|j| (j.validator.clone(), j.latest_block_hash.clone()))
                             .collect();
-                        let (merged_state_hash, _rejected, _rejected_slashes) =
-                            crate::rust::util::rholang::interpreter_util::compute_parents_post_state(
-                                casper.block_store(),
-                                parents.clone(),
-                                &snapshot,
-                                &runtime_manager,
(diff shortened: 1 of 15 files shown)


spreston8:

On `dev` via `a21ad5f4`: settled-in-base dedup and the `MergedPreState` plumbing; `slash_recovery_spec`, `compute_parents_post_state_regression_spec`, `block_creator_memory_profile_spec` at parity.

Not on `dev`, carried forward:

- **`deploy_effect_in_state`** — `dev` gates on `settled_user_sigs` (record-derived) rather than probing the block's actual pre-state. The record plane carries `included` and `rejected` but not `applied`, so a chain reinstated by a merge has no won record and isn't caught.
- `rho_spec.rs` whole-pipeline execution bound with phase-named failure.
- `pos_spec.rs` bound, still 1800s on `dev`.
- `exactly_once_spec.rs` — `dev` has the tracer test only; the race and floor-coverage scenarios are absent.

PR #249, opened 2026-08-12
feat(casper): rejection records name the carrier they adjudicated

spreston8:
Layer 2: rejection records are full consensus content — each names the carrier block it adjudicated and carries a formation-time duplicate flag.

- Records travel merge → checkpoint → block body structurally (the strip sites are gone, not patched); validation compares full records and the same-block sig-set exemption is closed.
- The height-blind rejection suppressor is replaced by exact (sig, carrier, flag) record matching.
- Format break: `RejectedDeployProto` gains fields 2–3. proto3 default-omission keeps the genesis encoding byte-stable, so no fixture regeneration.

### DIFF
--- a/block-storage/src/rust/key_value_block_store.rs
+++ b/block-storage/src/rust/key_value_block_store.rs
@@ -131,9 +131,12 @@ impl KeyValueBlockStore {
         Ok(has_any)
     }
 
-    /// Fetch rejected deploy signatures for a block without decoding a full BlockMessage.
-    /// Returns the `body.rejected_deploys[*].sig` values. Most blocks have none; only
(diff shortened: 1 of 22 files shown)


spreston8:

On `dev`: the `carrier` field on `RejectedDeploy` (proto + `casper_message.rs`), the checkpoint-equality consensus check, `proto_util`, `key_value_block_store`, and both `interpreter_util_test.rs` / `runtime_manager_test.rs` at parity.

Not on `dev`, carried forward: most of `carrier_record_spec.rs` (49 of 229 lines present). The `stage_contest` scenarios that pin carrier equality against a forged record are the gap.

PR #250, opened 2026-08-12
fix(rspace): merge-apply hardening — absent-removal error, netted no-op

spreston8:
Two self-contained rspace merge-apply hardenings (no dependency on the layers below — placed here so they merge early).

- Removing a datum that is absent from the base is now a hard error naming the channel and counts — applied diffs incoherent with the base state must never silently no-op.
- A fully-netted channel change (adds cancel removes) is a no-op instead of a spurious merge failure; pinned end-to-end by a spec where the netted COMM lands exactly once.

### DIFF
--- a/casper/tests/batch2/mod.rs
+++ b/casper/tests/batch2/mod.rs
@@ -9,6 +9,7 @@ pub mod limited_parent_depth_spec;
 pub mod lmdb_key_value_store_spec;
 pub mod map_cell_convergence_spec;
 pub mod multi_validator_recovery_spec;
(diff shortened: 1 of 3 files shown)


spreston8:

On `dev`: roughly two-thirds of the `state_change_merger.rs` hardening.

Not on `dev`, carried forward: the remainder of the merge-apply changes, and most of `netted_noop_merge_spec.rs` (37 of 136 lines) — including the three-validator choreography that keeps the install chain above the merge base.

PR #251, opened 2026-08-12
fix(casper): mergeable-channel GC measures depth from the floor

spreston8:
Mergeable-channel GC measures deletion depth from the floor of the last finalized block, not from this node's tip. The data GC deletes is what merges read, and a merge reads it for the blocks above its base — the floor; a tip-anchored bound put the whole floor-to-tip span into the deletable set exactly when a finality stall makes that span widest. Floor-anchoring is strictly more conservative (the floor never leads the tip).

The module had no tests; it now has four over a real DAG fixture, two of which fail with the tip anchor restored.


spreston8:

On `dev`: `mergeable_channels_gc` measures depth from the floor rather than the tip.

Not on `dev`, carried forward: the `is_safe_to_delete` unit tests covering the floor-relative boundary. The behavior landed without its pinning tests.

PR #252, opened 2026-08-12
fix(casper): one finality clock — floor-keyed merge, validity, and eviction

spreston8:
Layer 3: one finality clock — every irreversible consensus edge reads the floor (the per-block, justification-derived finalized cut), never a node-local clock or marker.

- `FloorContext`: floor + floor state derived once per propose and once per validate; consumers read it and never re-derive.
- Re-keyed onto the floor: buffer retain (pure window arithmetic), terminal purge (effect-in-floor-state evidence), deploy expiry (removal, retry admission, and block validity on the block's own frozen floor), the merge validity window (late carriers past their window are rejected with records), parent-depth and progress-walk validation anchors (frozen block data, ending per-node verdict divergence), and deploy-storage eviction (a floor-keyed sweep, measured before/after under load).
- TestNode's parent-depth default becomes a realistic 30 (zero fixture casualties after the spread re-anchor), and the scheduled reinstatement falsifier runs green: a floor-covered effect survives a late-arriving rejection record.


spreston8:

On `dev`: `validate.rs` and `validate_test.rs` at parity, `block_creator` and `dag_merger` substantially, `floor_context.rs`, `test_node.rs`, `block_creator_spec.rs`.

Not on `dev`, carried forward:

- **`finalization_runner.rs`** — the floor-keyed deploy-storage sweep is essentially absent (4 of 54 lines). Eviction still keys on finalization markers.
- `merge_window_spec.rs` (60 of 144) and `exactly_once_spec.rs` (7 of 121).
- The `block_api.rs` per-query "no per-operation floor context" path.

PR #260, opened 2026-08-13
perf(casper): deploy-index fast path for repeat-deploy validation (soak finalization-lag root cause)

jeffrey-l-turner:
## What

First node-side fix on the soak finalization-lag track: a deploy-index fast path in `Validate::repeat_deploy` that eliminates the only block-validation cost that grows with chain length.

`repeat_deploy` currently BFS-walks every in-window ancestor of a candidate block, doing a block-store deploy-sig scan per visit. In daily soak run [31563121791](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/31563121791) this cost grew from **4.3ms to 124ms per validated block within a single iteration** as the DAG filled its expiration window — O(DAG-in-window) per block, quadratic over a run.

The fix: probe the persistent `deploy_index` (maintained by `insert_internal` for **every** deploy of **every** inserted block — valid, invalid, and approved modes — before any child of that block can be validated) ahead of both the parent-scope rejection scan and the ancestor BFS. If **none** of the candidate's sigs are indexed, there is provably no prior inclusion anywhere — on any fork, inside or outside the window — and no canonical rejection either (rejected deploys were themselves carried by inserted blocks), so validation returns `Valid` immediately. Fresh deploys are the overwhelmingly common case, so this is the hot path.

**Any index hit — even a stale one from a fork — falls through to the existing exact logic unchanged.** The index is last-write-wins and single-valued, so it can never answer *which* inclusion is canonical for a given parent scope; it only proves absence. The fork-safety semantics documented at the top of `repeat_deploy` (deterministic recovery-exemption, the fix for the roaming `InvalidRepeatDeploy` forks) are untouched. An index read error also falls through: the fast path is an optimization, never a second opinion.

## Why (root cause of the soak failures)

The recurring `N deploy(s) not finalized within 45s` soak failure is **not a flake, not an E6.Flex/32-core shape effect, and not a recent dev regression**:

- Daily run 31563121791 (dev@`71c4d11a`, E6.Flex): **19/19 iterations failed** over 13.9h, both providers.
- Weekend run 31404822218 (**master**@`438f4724`, **pre-E6 shape**): identical signature, 8/8 failed.
- Published history (both series): **zero passing iterations ever recorded** — earlier runs died to infra (RSS ceilings, cert mounts) within 1–2 iterations, masking the pattern. `test_load.py` is deselected in CI, so the soak is the only place it runs; it has plausibly never passed since its sustained phase was added (SI `f7de50f`, 2026-06-22).

Mechanism (per-block node internals from the run): the un-finalized cone never builds — `tip − LFB = 0` after every phase, merge costs are milliseconds. The shard simply sustains **~3.4 deploys/s against the test's 4/s**, so inclusion latency spirals (p50 50–80s) and late-phase deploys miss the 45s drain window. Validation saturates at ~1 block/s arrival × ~1.1s/block, of which:

| Cost | Sustained-phase value |
|---|---|
| precharge+refund system-deploy Rholang, per deploy per node | ≈128ms (user code itself: ~0ms) |
| `repeat_deploy` per validated block | 4.3ms → 124ms (grows with in-window DAG) |
| propose path per block created | ~2.8s (`prepare_user_deploys` 1.3s + `compute_deploys_checkpoint` 1.2s) |

## Scope

This PR removes the *growing* cost. The dominant *flat* cost — precharge/refund overhead at ~128ms/deploy/node, ~65× the user-code cost — is the follow-up that closes most of the 3.4→4 deploys/s gap and is deliberately not attempted here. BACKLOG-TD-001 is updated with the attribution and unblocked (its deterministic test should pin sustained-rate deploy throughput, not cone depth).

## Tests

- **New:** `repeat_deploy_accepts_fresh_deploys_in_block_not_yet_inserted` — production order (candidate built, not inserted): fresh deploys clear via the fast path.
- **New:** `repeat_deploy_fast_path_short_circuits_on_deploy_index_absence` — pins the contract from both sides: with the sig indexed, the fallback scan flags the repeat; after deleting the index entry (via the test-only accessor), the verdict flips to Valid — proving the probe is consulted.
- **Unchanged and green:** all 15 pre-existing repeat-deploy specs — recovery exemptions, divergent-local-view agreement, the proposer/validator agreement property suite, and slashing integration (`uc_07`, `integration_t_invalid_repeat_deploy`).
- **Completeness invariant:** already pinned by block-storage's `dag_storage_should_be_able_to_restore_deploy_index_on_startup` property test, which asserts every deploy sig of every inserted block (including `InsertMode::Invalid`) is retrievable via `lookup_by_deploy_id`.

Support change: `IndexedBlockDagStorage` gains a `test-internals`-gated passthrough to the existing `deploy_index_for_tests` accessor.

## Reviewer flag

The negative filter is sound **iff `deploy_index` is complete for every stored block**. The index write has been part of `insert_internal` since the Scala port, so any database produced by this codebase should satisfy it — but a data directory originating from a node version that did not maintain the index would silently weaken repeat-deploy detection. If such databases exist in the wild, this PR should gain a startup check or index backfill before merging.

## Expected effect

`block_summary.repeat_deploy` in the soak's per-iteration "Node internals" telemetry should drop from tens-of-ms-and-growing to ~constant microseconds. This alone does not close the throughput gap; it removes the superlinear term and is independently verifiable in the next soak run.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/260"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789172278&installation_model_id=426883&pr_number=260&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F260&signature=c2d09064f26f217ddd433f2eff4d20cf53ba17c725da518d7561e78c4bf836c5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/block-storage/src/rust/test/indexed_block_dag_storage.rs
+++ b/block-storage/src/rust/test/indexed_block_dag_storage.rs
@@ -34,6 +34,24 @@ impl IndexedBlockDagStorage {
         self.underlying.get_representation_internal()
     }
 
+    /// Passthrough to the underlying storage's test-only deploy-index
+    /// accessor, so tests built on this fixture can manipulate index
+    /// entries (same rationale as `BlockDagKeyValueStorage`'s accessor).
+    #[cfg(any(test, feature = "test-internals"))]
+    #[doc(hidden)]
+    pub fn deploy_index_for_tests(
+        &self,
+    ) -> Arc<
+        parking_lot::RwLock<
+            shared::rust::store::key_value_typed_store_impl::KeyValueTypedStoreImpl<
+                crate::rust::dag::block_dag_key_value_storage::DeployId,
+                models::rust::block_hash::BlockHashSerde,
+            >,
+        >,
+    > {
+        self.underlying.deploy_index_for_tests()
+    }
+
     pub fn insert(
         &mut self,
         block: &BlockMessage,
--- a/casper/src/rust/validate.rs
+++ b/casper/src/rust/validate.rs
@@ -538,6 +538,41 @@ impl Validate {
             return Either::Right(ValidBlock::Valid);
         }
 
+        // Fast path on the persistent deploy index. `insert_internal` writes
+        // every deploy sig of every inserted block (valid, invalid and
+        // approved alike) into `deploy_index` before any child of that block
+        // can be validated, so a sig with no index entry has no prior
+        // inclusion anywhere — on any fork, inside or outside the expiration
+        // window — and cannot carry a canonical rejection either (rejected
+        // deploys were themselves carried by an inserted block). When none of
+        // this block's sigs are indexed, the parent-scope scan and the
+        // ancestor traversal below have nothing to find and are skipped —
+        // that traversal's cost grows with the in-window DAG (4ms -> 124ms
+        // per block within one soak iteration, run 31563121791), and fresh
+        // deploys are the overwhelmingly common case. Any hit falls through
+        // to the exact logic below, which stays the sole authority on
+        // whether a prior inclusion is canonical for THIS block's parent
+        // scope: the index is last-write-wins and single-valued, so it can
+        // never answer that question, only prove absence. An index read
+        // error also falls through — the fast path is an optimization,
+        // never a second opinion.
+        let any_sig_indexed = block.body.deploys.iter().any(|pd| {
+            match s.dag.lookup_by_deploy_id(&pd.deploy.sig.to_vec()) {
+                Ok(hit) => hit.is_some(),
+                Err(e) => {
+                    tracing::warn!(
+                        target: "f1r3fly.casper",
+                        "repeat-deploy fast path: deploy index lookup failed ({}); falling back to ancestor scan",
+                        e
+                    );
+                    true
+                }
+            }
+        });
+        if !any_sig_indexed {
+            return Either::Right(ValidBlock::Valid);
+        }
+
         let block_metadata = BlockMetadata::from_block(block, false, None, None);
 
         tracing::debug!(target: "f1r3fly.casper", "before-repeat-deploy-get-parents");
--- a/casper/tests/batch2/validate_test.rs
+++ b/casper/tests/batch2/validate_test.rs
@@ -992,6 +992,119 @@ async fn repeat_deploy_validation_should_not_accept_blocks_with_a_repeated_deplo
     .await
 }
 
+/// The deploy-index fast path, production order: the candidate is validated
+/// BEFORE insertion, so its own sigs are not yet indexed, and deploys never
+/// included in any inserted block clear validation without the parent-scope
+/// scan or the ancestor traversal.
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn repeat_deploy_accepts_fresh_deploys_in_block_not_yet_inserted() {
+    with_storage(|mut block_store, mut block_dag_storage| async move {
+        let genesis_deploy = construct_deploy::basic_processed_deploy(0, None).unwrap();
+        let genesis = create_genesis_block(
+            &mut block_store,
+            &mut block_dag_storage,
+            None,
+            None,
+            None,
+            Some(vec![genesis_deploy]),
+            None,
+            None,
+            None,
+            None,
+        );
+
+        let fresh_deploy = construct_deploy::basic_processed_deploy(1, None).unwrap();
+        let candidate = build_block(
+            vec![genesis.block_hash.clone()],
+            None,
+            1786500000000,
+            None,
+            None,
+            Some(vec![fresh_deploy]),
+            None,
+            None,
+            None,
+            Some(1),
+        );
+
+        let dag = block_dag_storage
+            .get_representation()
+            .expect("dag representation");
+        let mut casper_snapshot = mk_casper_snapshot(dag);
+
+        let result = Validate::repeat_deploy(&candidate, &mut casper_snapshot, &block_store, 50);
+        assert_eq!(result, Either::Right(ValidBlock::Valid));
+    })
+    .await
+}
+
+/// Pins the fast path's contract from both sides. Same repeat shape as
+/// `repeat_deploy_validation_should_not_accept_blocks_with_a_repeated_deploy`
+/// but in production order (candidate built, not inserted): while the sig is
+/// indexed, the fallback ancestor scan flags the repeat; deleting the index
+/// entry flips the verdict to Valid, proving the index probe is consulted
+/// and that the skip rests on the completeness invariant (every inserted
+/// block's sigs reach `deploy_index` — see `insert_internal`).
+#[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+async fn repeat_deploy_fast_path_short_circuits_on_deploy_index_absence() {
+    use shared::rust::store::key_value_typed_store::KeyValueTypedStore;
+
+    with_storage(|mut block_store, mut block_dag_storage| async move {
+        let deploy = construct_deploy::basic_processed_deploy(0, None).unwrap();
+        let genesis = create_genesis_block(
+            &mut block_store,
+            &mut block_dag_storage,
+            None,
+            None,
+            None,
+            Some(vec![deploy.clone()]),
+            None,
+            None,
+            None,
+            None,
+        );
+
+        let candidate = build_block(
+            vec![genesis.block_hash.clone()],
+            None,
+            1786500000000,
+            None,
+            None,
+            Some(vec![deploy.clone()]),
+            None,
+            None,
+            None,
+            Some(1),
+        );
+
+        let dag = block_dag_storage
+            .get_representation()
+            .expect("dag representation");
+        let mut casper_snapshot = mk_casper_snapshot(dag);
+        let result = Validate::repeat_deploy(&candidate, &mut casper_snapshot, &block_store, 50);
+        assert_eq!(
+            result,
+            Either::Left(BlockError::Invalid(InvalidBlock::InvalidRepeatDeploy))
+        );
+
+        {
+            let deploy_index_handle = block_dag_storage.deploy_index_for_tests();
+            let deploy_index_guard = deploy_index_handle.write();
+            deploy_index_guard
+                .delete(vec![deploy.deploy.sig.to_vec()])
+                .expect("delete deploy-index entry");
+        }
+
+        let dag = block_dag_storage
+            .get_representation()
+            .expect("dag representation");
+        let mut casper_snapshot = mk_casper_snapshot(dag);
+        let result = Validate::repeat_deploy(&candidate, &mut casper_snapshot, &block_store, 50);
+        assert_eq!(result, Either::Right(ValidBlock::Valid));
+    })
+    .await
+}
+
 /// Regression test for `repeat_deploy`'s `rejected_in_scope` exemption.
 ///
 /// Without the exemption, validation rejects any block that re-includes a
--- a/docs/Backlog.md
+++ b/docs/Backlog.md
@@ -35,8 +35,8 @@ title: "Encode the 'N deploy(s) not finalized within 45s' soak failure as a dete
 category: technical_debt
 priority: p2
 added_at: 2026-08-11
-blocked_by: root cause unknown (flake vs VM-shape behavior)
-repo_scope: node-side (casper finalization) and/or system-integration (test_load assertion)
+blocked_by: none (root-caused 2026-08-12)
+repo_scope: node-side (casper deploy throughput) and/or system-integration (test_load assertion)
 ---
 ```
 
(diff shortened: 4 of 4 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Needs Review | 0.82 |
| xai | `grok-4.5` | Needs Review | 0.86 |

---

### Summary

This is a well-motivated and carefully documented performance fix: the deploy-index probe is a legitimate absence-proof that skips the O(in-window DAG) ancestor BFS for the common fresh-deploy case, with conservative fall-through on any hit or read error, and the two new tests pin the contract from both directions. However, the change quietly makes a consensus validation rule depend on an external invariant — completeness and crash-atomicity of `deploy_index` writes in `insert_internal`, and its completeness for pre-existing on-disk DAG state — and the diff neither tests nor documents that invariant at its write site. The test that deletes an index entry to flip a repeat verdict to Valid is itself a demonstration of the failure mode if the invariant is ever violated. Recommend human review to confirm the index write is atomic with block insertion and that no deployed database predates index maintenance (or add a backfill/guard); the remaining issues are minor style and logging concerns.

The fast path is performance-motivated and safely falls back on hits and read errors, but an index miss now determines consensus validation. The supplied change does not demonstrate that deploy-index absence is authoritative across crash recovery, partial persistence, and database upgrades. That durability invariant should be verified or explicitly enforced before this optimization is merged.

This is a well-motivated consensus-adjacent optimization: probe deploy_index for absence, short-circuit to Valid only when no candidate sig is present, and otherwise preserve the existing parent-scope/BFS authority, with errors failing open to the slow path. The control flow and tests match that design, and fresh-deploy short-circuit should remove the O(in-window DAG) growth seen in soak. The main residual risk is not the branch itself but the cross-module completeness invariant it elevates to a safety dependency—if any insert path skips indexing, repeat deploys can be accepted. Human review should confirm insert_internal (all modes) before merge; optional metrics and avoiding sig.to_vec() are non-blocking improvements.

The code change introduces a performance optimization for the `repeat_deploy` validation by adding a fast path that checks the persistent deploy index. The implementation looks correct and the new tests cover the intended behavior. The documentation update is minor and can be addressed post-merge. Overall, the change appears safe and beneficial for performance.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Safety depends on unshown deploy_index completeness invariant</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/validate.rs` (line 541)

The fast path treats index absence as a proof that a deploy was never included or rejected on any fork. That is only sound if every insert path (valid, invalid, approved, genesis, replay/recovery) writes every deploy sig in insert_internal before any child can be validated, and if rejected deploys are always carried by an inserted block. This diff does not show those write paths. If any path omits indexing, repeat_deploy will return Valid for a true repeat — a consensus-safety failure, not a mere performance miss. The new test that deletes an index entry and observes Valid explicitly demonstrates this failure mode.

**Suggestion:**
```
Before merge, verify (or add regression tests for) all BlockDag insert/approve/invalid paths writing deploy_index, including genesis and restart/replay. Consider a debug assertion or metric when the slow path finds a repeat that the index missed, to detect invariant breakage in soak.
```

</details>

<details>
<summary><b>:orange_circle: Major: Consensus soundness now depends on deploy_index completeness/durability invariant not enforced in this diff</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/validate.rs` (line 559)

If `deploy_index` was introduced (or its maintenance in `insert_internal` for invalid/approved blocks was added) after some deployed nodes already persisted DAG state, blocks inserted before that point will have unindexed deploy sigs. On such a node, a repeat of one of those deploys would take the fast path and validate as fresh, while freshly-synced nodes would reject it. The PR description asserts the index is maintained for 'every deploy of every inserted block' but the diff contains no backfill or version/migration guard proving that holds for existing databases.

---

The fast path converts `repeat_deploy` — a consensus validation rule — into a check whose correctness rests entirely on the invariant that every deploy sig of every inserted block is present in `deploy_index` at validation time. If that invariant can ever be violated (e.g., a crash between the block-store/DAG insert and the deploy-index write if they are not committed atomically in `insert_internal`, or a partial write), a node would accept a block that repeats an in-window deploy, while other nodes (or the same node after re-sync) reject it — a consensus split. The second test even demonstrates that a single missing index entry flips a repeat verdict to Valid. The diff provides no evidence (test or comment referencing the atomicity guarantee) that index writes are crash-atomic with block insertion.

**Suggestion:**
```
Confirm the deploy index has been populated for all insertion paths since the storage format's inception, or add a backfill/migration step (or gate the fast path behind a marker indicating the index is complete for this database).
```

</details>

<details>
<summary><b>:orange_circle: Major: Consensus correctness now depends on deploy-index completeness</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/validate.rs` (line 575)

Returning Valid on an index miss changes deploy_index from an optimization into a consensus-critical negative index. The insertion-order argument covers normal in-process operation, but the diff does not establish that every persisted or recoverable DAG block must have a corresponding index entry after crashes, interrupted multi-store writes, upgrades from older databases, index deletion, or failed historical writes. A missing entry in any of those cases causes a genuinely repeated deploy to bypass the exact parent/ancestor checks. The new test demonstrates this failure mode directly by deleting an entry and observing an otherwise invalid repeat become Valid.

**Suggestion:**
```
Before using absence as proof, establish and enforce a durable completeness invariant. For example, write the index before making a block discoverable with crash-safe ordering, add a persisted schema/completeness marker or watermark, rebuild/verify the index during migration or recovery, and disable the fast path whenever completeness is uncertain. Add restart/fault-injection and upgrade tests showing that interruption between storage writes cannot expose an indexed-DAG block without its deploy entries.
```

</details>

<details>
<summary><b>:orange_circle: Major: Per-deploy warn logging on index read errors can flood logs</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/validate.rs` (line 562)

Fresh blocks still pay one Vec allocation per deploy for lookup_by_deploy_id. This is tiny next to the old BFS, but it is the new common path and may be avoidable if the API accepts &[u8] or a borrowed DeployId.

---

If the deploy index becomes unreadable (e.g., store corruption or I/O errors), `tracing::warn!` fires once per deploy per validated block. Under sustained load this produces high-volume repeated warnings for the same underlying condition. Note also that on the first error `any()` short-circuits with `true`, so subsequent sigs are not probed — the log volume is bounded per block, but still one warn per block indefinitely.

---

The PR’s soak rationale hinges on fresh deploys being the common case. Without a metric, it will be hard to confirm in CI/soak that the fast path is actually taken and that residual lag is the precharge/refund cost called out in the backlog.

**Suggestion:**
```
Add cheap counters: repeat_deploy_fast_path_absent vs repeat_deploy_fast_path_fallback (hit/error), for soak dashboards and the follow-up throughput work.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Fast-path invariant should be cross-referenced from insert_internal</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/validate.rs` (line 541)

The 20-line comment on the fast path is thorough, but the invariant it depends on ('insert_internal writes every deploy sig before any child can be validated') lives in another crate. Nothing at the `insert_internal` write site warns maintainers that weakening that write (e.g., making it best-effort, async, or pruning entries) would silently break consensus validation here.

**Suggestion:**
```
Add a comment at the deploy-index write in `insert_internal` (and on any future pruning logic) pointing back to `Validate::repeat_deploy`'s fast path so the invariant is protected from both sides.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Error fallthrough is correct but collapses distinct failure modes</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/validate.rs` (line 561)

On lookup Err, the code forces any_sig_indexed=true and falls back to the ancestor scan. That is the right fail-open choice for an optimization. However, a persistent index error will silently disable the fast path for every block (one warn per short-circuit) while still relying on the slow path remaining correct and affordable — the original soak failure mode.

**Suggestion:**
```
Ensure the warn is high-signal (rate-limited or counter) and that ops can see fast-path disablement; optional counter for fallback_due_to_index_error vs fallback_due_to_hit.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Empty-deploy blocks silently take the fast path</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/validate.rs` (line 571)

For a block with zero deploys, `iter().any(...)` is vacuously false and the function returns Valid immediately. This is behaviorally correct (nothing can repeat), but it is an implicit consequence of the iterator semantics rather than an intentional, documented case, and it also skips the parent-scope rejection scan for empty blocks.

**Suggestion:**
```
Add a brief comment (or an explicit early-return for empty deploy lists) so future readers know the vacuous case was considered.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Test proves unsafe index incompleteness by construction</b> (Reported by: xai)</summary>

**File:** `casper/tests/batch2/validate_test.rs` (line 1045)

repeat_deploy_fast_path_short_circuits_on_deploy_index_absence deliberately corrupts deploy_index and expects Valid for a repeated deploy. That is a good contract pin, but readers may miss that this is not a supported production state.

**Suggestion:**
```
Keep the test; add one sentence that this Valid outcome is intentionally incorrect w.r.t. chain history and exists only to prove the probe is necessary and load-bearing.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Update documentation</b> (Reported by: bedrock)</summary>

**File:** `docs/Backlog.md` (line 35)

Consider updating the documentation to reflect the current state of the backlog item, including the root cause and the new scope of work.

**Suggestion:**
```
Ensure the documentation is clear and up-to-date with the latest changes and understanding of the issue.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test-only index mutator is appropriately gated</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/test/indexed_block_dag_storage.rs` (line 37)

deploy_index_for_tests is cfg(test/test-internals) and doc(hidden), matching the stated rationale. No issue; worth keeping gated so production code cannot “fix” validation by deleting index entries.

</details>

<details>
<summary><b>:white_circle: Suggestion: Verbose fully-qualified return type on test accessor</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/test/indexed_block_dag_storage.rs` (line 42)

The passthrough `deploy_index_for_tests` spells out a deeply nested fully-qualified type (`Arc<parking_lot::RwLock<KeyValueTypedStoreImpl<DeployId, BlockHashSerde>>>`). If the underlying store type changes, both accessors must be edited in lockstep.

**Suggestion:**
```
Introduce a `pub type DeployIndexHandle = ...` alias next to the underlying accessor in `block_dag_key_value_storage.rs` and reuse it here.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Magic timestamp literal in new tests</b> (Reported by: anthropic)</summary>

**File:** `casper/tests/batch2/validate_test.rs` (line 1019)

Both new tests use the raw timestamp `1786500000000` in `build_block` with no explanation of its significance (presumably a far-future timestamp keeping the deploy inside the expiration window).

**Suggestion:**
```
Name the constant (e.g., `const IN_WINDOW_TIMESTAMP: i64 = ...`) or add a comment explaining why this value is chosen relative to the expiration window and the deploy timestamps.
```

</details>
---

<details>

This is a well-motivated and carefully documented performance fix: the deploy-index probe is a legitimate absence-proof that skips the O(in-window DAG) ancestor BFS for the common fresh-deploy case, with conservative fall-through on any hit or read error, and the two new tests pin the contract from both directions. However, the change quietly makes a consensus validation rule depend on an external invariant — completeness and crash-atomicity of `deploy_index` writes in `insert_internal`, and its completeness for pre-existing on-disk DAG state — and the diff neither tests nor documents that invariant at its write site. The test that deletes an index entry to flip a repeat verdict to Valid is itself a demonstration of the failure mode if the invariant is ever violated. Recommend human review to confirm the index write is atomic with block insertion and that no deployed database predates index maintenance (or add a backfill/guard); the remaining issues are minor style and logging concerns.

The code change introduces a performance optimization for the `repeat_deploy` validation by adding a fast path that checks the persistent deploy index. The implementation looks correct and the new tests cover the intended behavior. The documentation update is minor and can be addressed post-merge. Overall, the change appears safe and beneficial for performance.

#### openai (gpt-5.6-sol)

The fast path is performance-motivated and safely falls back on hits and read errors, but an index miss now determines consensus validation. The supplied change does not demonstrate that deploy-index absence is authoritative across crash recovery, partial persistence, and database upgrades. That durability invariant should be verified or explicitly enforced before this optimization is merged.

#### xai (grok-4.5)

This is a well-motivated consensus-adjacent optimization: probe deploy_index for absence, short-circuit to Valid only when no candidate sig is present, and otherwise preserve the existing parent-scope/BFS authority, with errors failing open to the slow path. The control flow and tests match that design, and fresh-deploy short-circuit should remove the O(in-window DAG) growth seen in soak. The main residual risk is not the branch itself but the cross-module completeness invariant it elevates to a safety dependency—if any insert path skips indexing, repeat deploys can be accepted. Human review should confirm insert_internal (all modes) before merge; optional metrics and avoiding sig.to_vec() are non-blocking improvements.
</details>

---

Issue #267, opened 2026-08-13
CI: shared push concurrency can cancel protected-branch unit tests


jeffrey-l-turner:
## Summary

Protected-branch push unit tests can be cancelled before any job starts. The workflow-level concurrency group puts every push into one `CI-push` lane. GitHub keeps only one running and one pending run in a concurrency group. Each new push replaces the pending run.

PR #224 did not change the CI workflow. However, its merge exposed this failure mode:

- The push run for merge commit `4d49965a` queued with zero jobs: https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/31664412506
- The next `dev` push cancelled that run.
- Several later protected-branch push runs were cancelled in the same way.

Run `31626632776` caused the blockage: https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/31626632776

All tests and integration jobs in that run completed. The `Release Docker Image` job remains in `waiting` because `release_docker_image` uses the reviewer-gated `ephemeral-launch` environment. That waiting job keeps the workflow-level `CI-push` lane occupied.

PR #224's tests did run before merge and on the automatic `dev -> master` PR run for the same tree. Thus, this issue does not prove that PR #224 caused the soak regression. The issue is still a branch-push coverage gap, especially for direct pushes and merge results that differ from tested PR heads.

## Expected behavior

- Each protected-branch push must run the per-crate unit-test matrix.
- A publishing approval or heavy-pipeline queue must not prevent unit tests from starting.
- Heavy OCI work and image publication must remain bounded and serialized where required.

## Proposed fix

1. Remove `environment: ephemeral-launch` from `release_docker_image`, or use a dedicated internal publishing environment that has no fork-approval rule.
2. Do not put fast checks and tests in the shared push concurrency lane.
3. Give each push SHA an independent fast-CI run.
4. Apply job-level concurrency only to heavy OCI work and publication.
5. Add a workflow invariant that rejects a configuration where a reviewer-gated job can block protected-branch unit tests.

## Acceptance criteria

- Two rapid `dev` pushes both start and complete all `Test (<crate>)` jobs.
- A waiting heavy or publication job does not cancel or delay those unit-test jobs.
- Internal `dev` and `master` image publication does not wait for fork launch approval.
- OCI launch and image publication do not overlap beyond configured limits.
- CI tests validate these concurrency and environment rules.

PR #158, opened 2026-07-29
Hotfix/soak checkpoint publishes

jeffrey-l-turner:
---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/158"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787884828&installation_model_id=426883&pr_number=158&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F158&signature=137ee9cba18892bc72224732c863b5efdb48a7027f64ffffd6de168819911b91"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>

PR #269, opened 2026-08-14
fix(ci): preserve tests across protected pushes

jeffrey-l-turner:
## Summary

- Give each push ref and SHA an independent CI run for checks and unit tests.
- Keep a branch push separate from a version tag at the same SHA.
- Let a newer branch or pull request head replace obsolete heavy work.
- Give each version tag an independent heavy and publication path.
- Prevent an older branch SHA from replacing mutable image tags.
- Keep reviewer approval for version-tag image publication.

## Problem

All protected-branch pushes shared one workflow concurrency group. GitHub permits only one running run and one pending run in each group.

Run `31626632776` waited for `Release Docker Image` approval. Later pushes replaced pending runs before their unit-test jobs started.

The merge push for PR #224 showed this failure. Its push run queued no jobs and the next push cancelled it.

## Concurrency model

Push workflows use the Git ref and commit SHA in the workflow group. A branch and tag at the same SHA use different groups.

Pull request workflows keep one group for each pull request ref. A new head cancels only the obsolete head run.

Heavy jobs use a pull-request-specific or Git-ref-specific group. New branch work replaces obsolete heavy work, but version tags run to completion.

Image and package jobs use Git-ref-specific groups. Different version tags cannot evict each other from one shared pending slot.

Intermediate branch image jobs can leave the pending slot. `Release Docker Image` is not a required status check, and only the newest branch image is authoritative.

## Publication safety

The image publisher checks the current remote branch tip before checkout. The API check retries three times and then fails closed.

The publisher repeats the tip check before each mutable Docker Hub and Oracle Cloud Infrastructure Registry update. Immutable version images remain independent.

A stale branch SHA skips mutable publication. The final job summary reports `published`, `skipped-stale-sha`, or `failed`.

Registry APIs do not provide an atomic ref comparison and push. Ref-specific queues and repeated checks reduce this unavoidable interval.

`dev` and `master` use `protected-branch-image-publish`. GitHub restricts this environment to those two branches.

All other refs use `ephemeral-launch`. Therefore, every version tag keeps the required reviewer approval.

## Invariants

The workflow invariant checker requires:

- Ref-and-SHA workflow groups for pushes
- Pull-request or Git-ref groups for heavy work
- Git-ref groups for image and package publication
- Unit-test dependencies for publication
- Branch-restricted automatic publication
- Reviewer approval as the default environment
- Three branch-tip API attempts
- A tip check before every mutable registry update
- A publication result in the job summary

The checker handles scalar and mapping-style environment declarations. A Ruby or YAML failure produces an explicit invariant error.

## Validation

- All workflow invariants passed.
- YAML and shell syntax checks passed.
- Error-level diagnostics passed.
- Same-SHA branch and tag group simulation passed.
- Current-tip, stale-tip, and version-tag gate tests passed.
- Mutable-boundary current and stale tests passed.
- Transient API failures recovered on the third attempt.
- Persistent API failure stopped publication after three attempts.
- Eight mutation tests confirmed fail-closed behavior.
- GitHub shows only `dev` and `master` in the automatic publication environment policy.

Two rapid `dev` pushes require post-merge verification. Both push SHAs must start and complete every `Test (<crate>)` job.

Refs #267

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/269"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789309888&installation_model_id=426883&pr_number=269&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F269&signature=77d5b16c4b8302aa2bdd91cdee0e46ffe5f60f1dfab5845b66324fad3d594a45"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR restructures CI concurrency so each push SHA gets an independent workflow run while heavy pipeline and publication jobs serialize through dedicated job-level concurrency groups, backed by a new Ruby-based invariant check. The intent is sound, but two things warrant human review: (1) GitHub's one-pending-slot-per-concurrency-group semantics means burst pushes can still evict and cancel queued heavy/publish jobs despite `cancel-in-progress: false` — the very failure mode the deleted comment documented — potentially undermining the PR's stated goal; (2) moving `release_docker_image` off the reviewer-gated `ephemeral-launch` environment removes a human approval gate on image publication with no stated rationale, and the invariant script now enforces that removal. The invariant script itself has minor robustness gaps (mapping-style `environment` declarations evade detection; ruby failure handling). Recommend confirming the new environment's protections and documenting the eviction tradeoff before merge.

The workflow-level SHA-specific concurrency correctly prevents one push run from superseding another, while retaining cancellation for stale PR runs. However, the newly added static job-level concurrency groups are not durable queues: GitHub can replace an existing pending job with a newer one regardless of `cancel-in-progress: false`. This can skip heavy pipelines and potentially omit image or tagged package publications during bursts of pushes, so the serialization design needs revision before merge.

The change correctly addresses collapsed/cancelled protected-branch test runs by giving each push SHA its own workflow concurrency group while keeping PR cancel-in-progress behavior, and it adds solid invariant checks plus non-cancelling job-level queues for heavy pipeline and publish jobs. The main risk is loss of push ordering: shared publish/pipeline mutexes serialize readiness order, not git history, so an older commit can still overwrite newer branch images/artifacts unless you add a tip-of-ref guard. The environment move to ephemeral-launch-internal looks deliberate but should be confirmed as an accepted trust-model change; restoring brief concurrency documentation would help prevent regressions.

The code changes in this PR appear to be well-implemented and address the intended issue of preserving tests across protected pushes. The new concurrency settings ensure that each push SHA runs independently while heavy work and publication remain serialized. The only minor issue is the outdated comments in the `ci.yml` file, which should be updated to reflect the new behavior.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Job-level concurrency queue eviction can still cancel push jobs despite cancel-in-progress: false</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `.github/workflows/ci.yml` (line 359)

GitHub Actions concurrency does not provide an unbounded FIFO queue. A concurrency group permits at most one running and one pending job; when another job enters the same group, the existing pending job can be canceled even when `cancel-in-progress` is `false`. Consequently, three rapid pushes can cause an intermediate `pipeline` job to be discarded. The same issue affects `release_docker_image` and, more seriously, `release_packages`, where rapid tag pushes could cause a tag's package publication to never run. This contradicts the invariant's claim that these queues do not cancel work and partially reintroduces the protected-push problem this change is intended to solve.

---

GitHub Actions concurrency groups hold at most one running and one pending item per group, regardless of `cancel-in-progress: false`. With the new per-SHA workflow groups, three pushes in quick succession will each spawn a `pipeline` (or publish) job contending for the shared `ci-heavy-pipeline` / `ci-image-publish` / `ci-package-publish` groups. The middle push's queued job gets evicted ('Canceled by higher priority waiting request') when the third arrives, leaving that SHA's required checks cancelled or never reported. This is exactly the starvation/eviction failure mode the deleted comment block documented (PR #158 incident), now reintroduced at the job level. The PR title says 'preserve tests across protected pushes', but under burst pushes it does not.

---

pipeline now fans out across push SHAs and only serializes on group ci-heavy-pipeline. Whichever commit becomes ready first runs first. On busy protected branches, a newer commit can enter the heavy pipeline before an older in-flight one, then the older one runs afterward—odd for integration/release side effects tied to branch state.

**Suggestion:**
```
Consider a per-branch serialization group (e.g. `ci-heavy-pipeline-${{ github.ref }}`) to reduce contention, or accept and document the eviction behavior explicitly. At minimum, restore an explanatory comment describing the one-pending-slot eviction semantics so future maintainers understand the tradeoff.
```

</details>

<details>
<summary><b>:red_circle: Critical: Docker image publication moved off the reviewer-gated environment</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/ci.yml` (line 489)

`release_docker_image` switches from `environment: ephemeral-launch` to `ephemeral-launch-internal`, and the new invariant script actively asserts this must NOT be the reviewer-gated environment. This removes a human-approval gate on image publication for dev-branch pushes. It may be intentional (internal images only), but it is a security-policy change buried in a CI concurrency fix with no PR description explaining the rationale, whether the new environment exists, what secrets it exposes, and whether it has its own protection rules (branch restrictions, deployment protection).

---

release_docker_image switches from ephemeral-launch to ephemeral-launch-internal, and the new invariant forbids ephemeral-launch on CI jobs. That removes human approval from the image publish path. Intentional per the script, but it changes the trust boundary: anyone who can land commits on the triggering refs can publish without the previous review gate.

**Suggestion:**
```
Confirm `ephemeral-launch-internal` exists with the required secrets and appropriate deployment branch restrictions, and document in the PR/commit why the reviewer gate is being removed for this job.
```

</details>

<details>
<summary><b>:orange_circle: Major: Parallel push runs can publish older commits over newer ones</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 22)

Workflow concurrency no longer serializes pushes (each push SHA gets its own group), while publish jobs only use a shared mutex with cancel-in-progress: false. That mutex preserves “one at a time”, not “git order”. If push B’s test/build path finishes before push A, B can acquire ci-image-publish first; when A later publishes it may overwrite branch tags (e.g. dev/latest) with older code. The previous workflow-level push group avoided this by running entire push workflows strictly in sequence.

**Suggestion:**
```
Keep the per-SHA workflow group for tests, but make publish order tip-aware: e.g. concurrency group per ref (ci-image-publish-${{ github.ref }}) plus an explicit “still HEAD of this ref?” guard before tag/push, or compare against the branch tip and skip stale publishes. Same consideration applies to any branch-mutable artifact from pipeline/release_packages.
```

</details>

<details>
<summary><b>:orange_circle: Major: Detailed concurrency rationale comment deleted without replacement</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/ci.yml` (line 21)

The long comment explaining PR cancel-in-progress vs non-cancelling pushes and the GitHub pending-run eviction failure mode was deleted. The new model (per-SHA push groups + job-level heavy/publish mutexes) is non-obvious and easy to “simplify” back into the old bug.

---

The removed ~20-line comment documented hard-won operational knowledge (the PR #158 queue-eviction incident, why PRs must cancel, why pushes must not). The replacement concurrency model is materially different and now has zero inline explanation; the rationale lives only implicitly in the invariant script's error messages.

**Suggestion:**
```
Add a concise comment above the `concurrency:` block explaining the new model: per-SHA push groups for parallel feedback, job-level serialization queues for heavy/publish work, and the known eviction caveat.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Ruby subprocess failure produces an unhelpful hard exit</b> (Reported by: anthropic)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 127)

If `ruby` is not installed on the runner, or the YAML fails to parse, the command substitution exits non-zero and (under `set -e`) aborts the script with no context, or worse, if `set -e` is not in effect on assignments here, `ci_concurrency_errors` could be empty and the check would falsely report `ok`. Silent false-pass on tool failure defeats the purpose of an invariant checker.

**Suggestion:**
```
Capture and check the ruby exit status explicitly, e.g. `if ! ci_concurrency_errors="$(ruby ...)"; then err "invariant checker failed to run"; fi`, and verify `command -v ruby` up front.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Reviewer-gate detection misses mapping-style environment declarations</b> (Reported by: anthropic)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 152)

The Ruby check `job["environment"] == "ephemeral-launch"` (and the equality check for `ephemeral-launch-internal`) only matches the scalar form. GitHub Actions also permits `environment: { name: ephemeral-launch, url: ... }`, which would silently evade the invariant (or falsely fail the internal-environment assertion).

**Suggestion:**
```
Normalize the value first, e.g. `env_name = job["environment"].is_a?(Hash) ? job["environment"]["name"] : job["environment"]`, then compare.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Update comments to reflect new concurrency behavior</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/ci.yml` (line 18)

The comments in the `ci.yml` file still describe the old concurrency behavior. They should be updated to reflect the new changes made in this PR.

**Suggestion:**
```
Update the comments to accurately describe the new concurrency settings for pushes and pull requests.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: New check breaks the script's numbered-section convention</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 127)

The new checks are valuable and match the PR intent, but the heredoc is getting large and duplicates strings from ci.yml. Failures are also flattened to a single semicolon-separated err line, which is harder to read in logs than one failure per line.

---

The surrounding script uses numbered comment headers ('# 5. The CI runner compartment...'). The new concurrency-invariant block is inserted between sections with no numbered header or explanatory comment, making the file's structure inconsistent.

**Suggestion:**
```
Add a numbered/section comment above the new block describing what invariant it enforces and why.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Per-SHA push groups allow unbounded parallel push runs for non-heavy jobs</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 22)

Previously all pushes shared one 'push' group, serializing entire runs. Now each push SHA runs independently, so lint and the unit-test matrix from many rapid pushes to dev/master/staging/trying can run fully in parallel, multiplying runner consumption. Heavy pipeline and publish jobs are serialized, but everything upstream is not.

**Suggestion:**
```
If runner cost/capacity matters, confirm the expected push cadence makes this acceptable, or scope push groups per-branch with queueing for the light jobs too.
```

</details>
---

<details>

This PR restructures CI concurrency so each push SHA gets an independent workflow run while heavy pipeline and publication jobs serialize through dedicated job-level concurrency groups, backed by a new Ruby-based invariant check. The intent is sound, but two things warrant human review: (1) GitHub's one-pending-slot-per-concurrency-group semantics means burst pushes can still evict and cancel queued heavy/publish jobs despite `cancel-in-progress: false` — the very failure mode the deleted comment documented — potentially undermining the PR's stated goal; (2) moving `release_docker_image` off the reviewer-gated `ephemeral-launch` environment removes a human approval gate on image publication with no stated rationale, and the invariant script now enforces that removal. The invariant script itself has minor robustness gaps (mapping-style `environment` declarations evade detection; ruby failure handling). Recommend confirming the new environment's protections and documenting the eviction tradeoff before merge.

The code changes in this PR appear to be well-implemented and address the intended issue of preserving tests across protected pushes. The new concurrency settings ensure that each push SHA runs independently while heavy work and publication remain serialized. The only minor issue is the outdated comments in the `ci.yml` file, which should be updated to reflect the new behavior.

#### openai (gpt-5.6-sol)

The workflow-level SHA-specific concurrency correctly prevents one push run from superseding another, while retaining cancellation for stale PR runs. However, the newly added static job-level concurrency groups are not durable queues: GitHub can replace an existing pending job with a newer one regardless of `cancel-in-progress: false`. This can skip heavy pipelines and potentially omit image or tagged package publications during bursts of pushes, so the serialization design needs revision before merge.

#### xai (grok-4.5)

The change correctly addresses collapsed/cancelled protected-branch test runs by giving each push SHA its own workflow concurrency group while keeping PR cancel-in-progress behavior, and it adds solid invariant checks plus non-cancelling job-level queues for heavy pipeline and publish jobs. The main risk is loss of push ordering: shared publish/pipeline mutexes serialize readiness order, not git history, so an older commit can still overwrite newer branch images/artifacts unless you add a tip-of-ref guard. The environment move to ephemeral-launch-internal looks deliberate but should be confirmed as an accepted trust-model change; restoring brief concurrency documentation would help prevent regressions.
</details>

---

jeffrey-l-turner:
## Multi-Agent Review Remediation

Commit `93da6a8e1` addresses the critical and major findings from the multi-agent review.

### Critical findings

1. **Shared job queues could evict unrelated work.**
   - Heavy work now uses a pull-request-specific or Git-ref-specific group.
   - New branch work replaces only obsolete work for the same ref.
   - Each version tag has an independent heavy, image, and package group.

2. **Image publication removed reviewer approval.**
   - `dev` and `master` pushes use `ephemeral-launch-internal`.
   - Version tags use the reviewer-gated `ephemeral-launch` environment.
   - Thus, trusted branch publication stays automatic and release publication keeps human approval.

### Major findings

1. **An older branch SHA could replace newer mutable image tags.**
   - The publisher checks the remote branch tip after it gets the publication slot.
   - A stale SHA skips every image publication step.

2. **The workflow removed the concurrency rationale.**
   - `ci.yml` now explains the per-SHA test model and the scoped heavy-work model.
   - The invariant script now has a numbered section for these controls.

### Additional fixes

- The invariant checker now reports Ruby and YAML failures explicitly.
- Environment checks now support scalar and mapping-style declarations.
- New invariants require the branch-tip gate on all image publication steps.

### Validation

- All workflow invariants passed.
- YAML and shell syntax checks passed.
- Error-level diagnostics passed.
- Current-tip, stale-tip, and version-tag gate tests passed.
- Eight mutation tests confirmed fail-closed behavior.
- All 15 pre-push checks passed.

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a well-reasoned CI concurrency fix: per-SHA workflow groups for pushes, PR/ref-scoped heavy-work groups, ref-scoped non-cancelling publication queues, a branch-tip publication gate, and a Ruby-based invariant checker that enforces the model with fail-closed error handling. The design is sound and the gate correctly fails closed on API errors. The main concerns are edge cases rather than blockers: a tag and branch push of the same commit still share a workflow group (recreating the eviction risk for tag releases), a TOCTOU window remains between the tip check and the actual manifest pushes, and non-`v` tags would route to the no-approval internal environment if the job's `if` permits them. None of these are showstoppers for merge, but the same-SHA group collision in particular deserves a follow-up since it can silently drop a version-tag release run — the precise class of failure this PR set out to eliminate.

The per-SHA workflow concurrency and ref-scoped publication groups address the pending-run eviction problem, and the version-tag approval path appears preserved. However, the new branch-tip gate does not fully enforce the stated publication-safety invariant because it is separated from the mutable image pushes by several potentially lengthy steps. A newer branch push during that interval can still allow an older SHA to publish mutable tags, so the gate should be moved or repeated at each publication boundary, with an atomic design considered if strict prevention is required.

The concurrency redesign correctly addresses the reported failure mode: push runs no longer share one workflow group with approval-gated release work, so unit tests are not evicted while a prior release waits. Job-level groups for heavy/pipeline and ref-scoped non-cancelling publish queues are coherent, and the branch-tip publish gate plus per-step `publish == 'true'` conditions are a solid fail-closed guard against stale SHAs moving mutable tags. The invariant script improvements (expression-aware environment handling, gate completeness checks) reinforce those rules. Main follow-ups are operational/security confirmation that `ephemeral-launch-internal` is fully wired for secrets and accepted as an approval bypass on dev/master, plus small robustness/efficiency nits around `gh api` retries and unnecessary checkout on skipped publishes.

The code change effectively addresses the concurrency issues in the CI workflow by ensuring that each push SHA runs tests independently and release side effects use ref-scoped controls. The invariants have been updated to reflect these changes, and all checks have passed. The code looks good and is safe to merge.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Tag and branch pushes of the same commit share a workflow concurrency group</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/ci.yml` (line 24)

The workflow-level group is `${{ github.workflow }}-${{ github.event_name == 'push' && github.sha || github.ref }}`. A version tag push resolves `github.sha` to the commit the tag points to, so tagging the current dev tip produces the *same* group as the dev branch push run. Since cancel-in-progress is false for pushes, the tag run queues behind the branch run, and GitHub keeps only one pending run per group — the exact eviction failure mode this PR is fixing could recur for a tag release created on a just-pushed commit (e.g., if another same-SHA event lands, or the pending tag run is evicted).

---

Workflow concurrency is now unique per push SHA, which fixes lost unit-test runs, but rapid pushes to protected branches can run many full test workflows in parallel. Heavy work is constrained by the job-level `ci-heavy-*` group; fast checks are not. That is likely intentional, but it can increase runner spend and queue pressure.

**Suggestion:**
```
After merge, watch concurrent `Test (<crate>)` runs under bursty dev pushes. If cost/queueing hurts, consider a separate lightweight workflow for unit tests or a bounded test-only concurrency group that still does not share a group with approval-gated release jobs.
```

</details>

<details>
<summary><b>:orange_circle: Major: Branch image publishes no longer require reviewer environment</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 493)

dev/master image publication now uses `ephemeral-launch-internal` instead of the reviewer-gated `ephemeral-launch`. That matches the PR intent for trusted protected branches, but it is a real security-posture change: any merge to dev/master can publish mutable image tags without human approval. Confirm that `ephemeral-launch-internal` exists, has the same registry secrets (Docker Hub/OCIR), and still has appropriate environment restrictions (branch filters, required reviewers if any, secret scope).

**Suggestion:**
```
Before merge, verify in GitHub Environment settings that `ephemeral-launch-internal` is configured with the needed secrets and deployment branch policy limited to dev/master, and that this approval bypass is accepted for supply-chain risk on mutable tags.
```

</details>

<details>
<summary><b>:orange_circle: Major: Branch-tip check has a race with mutable image publication</b> (Reported by: openai)</summary>

**File:** `.github/workflows/ci.yml` (line 511)

The branch tip is checked only once near the beginning of the job. A newer push can update the branch after this check but before the Docker Hub or OCIR publication steps, which may occur much later after artifact downloads, image imports, and logins. In that case the older SHA still publishes mutable tags despite no longer being the branch tip. The ref-scoped job concurrency does not prevent this because branch updates are not synchronized with the publication group, and the newer publication job may not become eligible until its dependencies finish.

**Suggestion:**
```
Revalidate the remote branch tip immediately before every operation that updates mutable tags, especially before Docker Hub and OCIR pushes. For a strict guarantee, use a publication mechanism with an atomic compare-and-publish or equivalent monotonic guard; a separate check followed by a push will always retain a small TOCTOU window. The invariant checker should also require these checks at the actual publication boundaries rather than merely requiring one earlier gate.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Minor formatting issue</b> (Reported by: bedrock)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 124)

The comment at line 124 is very long and could be split into multiple lines for better readability.

**Suggestion:**
```
Break the comment into multiple lines to improve readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Pending-slot eviction still cancels intermediate publish jobs on rapid pushes</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 476)

With `cancel-in-progress: false` on `ci-image-publish-${{ github.ref }}`, GitHub still keeps only one pending job per group: a third rapid dev push evicts the older pending publish job, which will report cancelled. This is acceptable for mutable branch tags (latest wins), but intermediate push runs will show a cancelled `Release Docker Image` job, which could be confusing or break status expectations if that job is ever made a required check.

**Suggestion:**
```
Document this expected eviction behavior in the concurrency comment, and ensure `Release Docker Image` is never listed as a required status check for branch protection.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Non-`v` tags fall into the no-approval internal environment</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 493)

The environment expression `startsWith(github.ref, 'refs/tags/v') && 'ephemeral-launch' || 'ephemeral-launch-internal'` routes any tag not starting with `v` to `ephemeral-launch-internal` (no reviewer gate). Whether this is reachable depends on the job's `if` condition (not fully visible in the diff). If the job's `if` allows any `refs/tags/*`, a maliciously or accidentally named tag (e.g., `version-1.0`) would publish images without reviewer approval.

**Suggestion:**
```
Ensure the job `if` restricts tags to `refs/tags/v*`, or invert the expression so the reviewer-gated `ephemeral-launch` is the default and `ephemeral-launch-internal` is only selected for the explicit allow-list of protected branches (`refs/heads/dev`, `refs/heads/master`).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Publish tip check has no retry on transient API failure</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 499)

The branch-tip gate uses a single `gh api` call under `bash -euo pipefail`. A transient GitHub API error fails the whole `release_docker_image` job rather than retrying. Because the job fails before setting a successful skip path, this can create noisy failures on otherwise valid tip commits without publishing.

**Suggestion:**
```
Add a short retry loop around the `gh api` ref lookup (e.g. 3 attempts with backoff), and keep fail-closed behavior only after retries are exhausted.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: TOCTOU window remains between branch-tip check and image publication</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 501)

The `publish_gate` step checks the remote branch tip once, before artifact download, login, and multi-arch manifest pushes. If a newer push lands and its (faster) run publishes between the stale run's gate check and its `docker manifest push`, the stale SHA can still overwrite mutable `dev-latest`/branch tags. The `ci-image-publish-${{ github.ref }}` queue with cancel-in-progress: false narrows but does not eliminate this race, since the gate check and the actual push are separated by several minutes of artifact handling.

**Suggestion:**
```
Re-verify the branch tip immediately before the `docker manifest push` / OCIR push steps (a second tip check inside the Publish steps), or perform the tip check as the last action before pushing manifests.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent shell invocation between steps</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 505)

The new `publish_gate` step uses `shell: bash -euo pipefail {0}` while neighboring steps use `shell: bash -ex {0}`. The stricter flags are better, but the inconsistency invites divergent failure semantics (e.g., pipeline failures silently ignored in `-ex` steps that use pipes such as `zcat ... | docker image load`).

**Suggestion:**
```
Standardize on `bash -euo pipefail {0}` (optionally with `-x` for tracing) across the job's run steps, or set a job-level `defaults.run.shell`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Invariant checker relies on brittle exact-string and token matching</b> (Reported by: anthropic)</summary>

**File:** `.github/scripts/check-workflow-invariants.sh` (line 165)

The Ruby checker compares full `${{ ... }}` expression strings byte-for-byte and validates the publish gate body by presence of tokens like `publish=false` and `GITHUB_OUTPUT`. Legitimate reformatting (whitespace, quote style) will fail the check, while a semantically broken gate that still contains the tokens (e.g., `publish=false` in a comment) would pass. Mutation testing mitigates this today, but the coupling is fragile.

**Suggestion:**
```
Consider normalizing whitespace before comparison and, for the gate body, asserting on structural patterns (e.g., regex requiring `publish=false` on an assignment line and the `GITHUB_OUTPUT` append) rather than bare substring inclusion.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Reduced concurrency-model documentation loses institutional context</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 21)

The old comment block documented the PR #158 incident and the reasoning behind cancel-vs-queue semantics in detail; the replacement is two lines. Given how subtle GitHub's one-pending-slot-per-group behavior is (it is the root cause of this very PR), losing that rationale increases the risk of a future regression by someone 'simplifying' the concurrency keys.

**Suggestion:**
```
Keep a condensed version of the incident history and the one-pending-run-per-group constraint in the comment, or link to issue #267 and the invariant checker section that enforces the model.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Checkout always runs even when publication will be skipped</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 495)

The full checkout (`fetch-depth: 0`) runs before `publish_gate`. Stale branch SHAs that later set `publish=false` still pay for a clone they do not need for the gate (the gate only uses `gh api` + env vars). Later publish steps already appear to rely on `build_base` outputs for version/branch metadata.

**Suggestion:**
```
Move checkout below the gate (or add the same `if: steps.publish_gate.outputs.publish == 'true'` if checkout is only needed for publish), unless some publish step still requires the workspace contents.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Stale SHA release jobs will go green without publishing</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 508)

When the tip gate sets `publish=false`, remaining steps are skipped and the job succeeds. That correctly avoids failed required checks on superseded SHAs, but the UI will show a green `Release Docker Image` even though nothing was published. Operators may misread that as “this SHA was released.”

**Suggestion:**
```
Keep the skip log line prominent (already present) and optionally add a final summary step that always runs and prints `published` vs `skipped-stale-sha` for clearer run annotations.
```

</details>
---

<details>

This is a well-reasoned CI concurrency fix: per-SHA workflow groups for pushes, PR/ref-scoped heavy-work groups, ref-scoped non-cancelling publication queues, a branch-tip publication gate, and a Ruby-based invariant checker that enforces the model with fail-closed error handling. The design is sound and the gate correctly fails closed on API errors. The main concerns are edge cases rather than blockers: a tag and branch push of the same commit still share a workflow group (recreating the eviction risk for tag releases), a TOCTOU window remains between the tip check and the actual manifest pushes, and non-`v` tags would route to the no-approval internal environment if the job's `if` permits them. None of these are showstoppers for merge, but the same-SHA group collision in particular deserves a follow-up since it can silently drop a version-tag release run — the precise class of failure this PR set out to eliminate.

The code change effectively addresses the concurrency issues in the CI workflow by ensuring that each push SHA runs tests independently and release side effects use ref-scoped controls. The invariants have been updated to reflect these changes, and all checks have passed. The code looks good and is safe to merge.

#### openai (gpt-5.6-sol)

The per-SHA workflow concurrency and ref-scoped publication groups address the pending-run eviction problem, and the version-tag approval path appears preserved. However, the new branch-tip gate does not fully enforce the stated publication-safety invariant because it is separated from the mutable image pushes by several potentially lengthy steps. A newer branch push during that interval can still allow an older SHA to publish mutable tags, so the gate should be moved or repeated at each publication boundary, with an atomic design considered if strict prevention is required.

#### xai (grok-4.5)

The concurrency redesign correctly addresses the reported failure mode: push runs no longer share one workflow group with approval-gated release work, so unit tests are not evicted while a prior release waits. Job-level groups for heavy/pipeline and ref-scoped non-cancelling publish queues are coherent, and the branch-tip publish gate plus per-step `publish == 'true'` conditions are a solid fail-closed guard against stale SHAs moving mutable tags. The invariant script improvements (expression-aware environment handling, gate completeness checks) reinforce those rules. Main follow-ups are operational/security confirmation that `ephemeral-launch-internal` is fully wired for secrets and accepted as an approval bypass on dev/master, plus small robustness/efficiency nits around `gh api` retries and unnecessary checkout on skipped publishes.
</details>

---

jltatbeach (approved):

Issue #102, opened 2026-07-08
Deploys from finalized noncanonical blocks can be lost


a-k-l-sdao:
A deploy can appear in a block that becomes finalized but is not on the canonical state path. In that case the deploy is DAG-final, but its effects are not included in canonical state.

Today the node can still treat that deploy as already seen/finalized and filter it as a duplicate instead of re-proposing it. This can leave the deploy stuck or expired even though the user submitted a valid deploy.

Observed impact in bridge testing: parallel lock deploys can finalize, but the canonical bridge state may not include all locks, causing missing bridge messages or incorrect accounting.

spreston8:
This should be resolved with recent changes. Reopen or open new issue if seen again.

Issue #197, opened 2026-08-04
Sustained deploy load to a single validator permanently stalls finalization via an unrecoverable `ContainsExpiredDeploy` retry loop


9Neechan:
**Affected:** f1r3node-rust, `casper::rust::blocks::proposer::block_creator` (ordinary-deploy selection) / `node::rust::instances::heartbeat_proposer`, `deploy-api-stress` experiment (`experiment/deploystress`), `mid` profile (2/8 CPU, 4Gi/16Gi RAM, 100Gi disk)
**Severity:** high — finalization stops permanently and does not self-recover; on at least one run the target validator's own process became CPU-pegged with unbounded memory growth and stopped logging entirely for the rest of the run

---

## What did you expect to see?

Three isolated chains (3 genesis validators + bootstrap + observer, default heartbeat params, no tuning) each received a **constant-rate** stream of transfer deploys sent to `validator2` only, for 15 minutes, then a 15-minute drain wait:

- `slow`: ~0.2 deploys/sec (1 per 5s)
- `medium`: 10 deploys/sec
- `high`: 100 deploys/sec

Expected: blocks keep producing and finalizing throughout, `validator2`'s GET API stays responsive (checked separately every 5s), and most of the submitted deploys finalize within the 15-minute drain window. A backlog building up under `medium`/`high` and draining down afterward would be an acceptable outcome; a **permanent, unrecoverable halt** would not.

## What did you see instead?

`slow` was healthy in all 3 runs (100% of submitted deploys finalized, zero errors). `medium` and `high` **both froze finalization permanently, every single time**, at a low block height early in the 15-minute load window, and never advanced again — not during the remaining load time, not during the 15-minute drain, not ever (confirmed by the harness's `chain_alive` watchdog firing at exactly the 10-minute-no-advance threshold in every case):

| run | shard (rate) | froze at block | submitted before freeze detected |
|---|---|---|---|
| `04-08-26-124932` | medium (10/s) | 71 | 6792 |
| `04-08-26-124932` | high (100/s) | 82 | 76858 |
| `04-08-26-090436` | medium (10/s) | 68 | 5950 |
| `04-08-26-090436` | high (100/s) | 79 | 81063 |
| `04-08-26-110353` | medium (10/s) | 72 | 6913 |
| `04-08-26-110353` | high (100/s) | 80 | 86766 |

[...]

PR #271, opened 2026-08-14
fix(casper): harden merge recovery and finality windows

jeffrey-l-turner:
- enforce exactly-once merge effects and carrier-specific records
- use floor evidence for recovery, expiry, validation, and garbage collection
- add RED/GREEN regressions for issue #102 and PRs #248 through #252

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/271"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789329203&installation_model_id=426883&pr_number=271&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F271&signature=6dba005c6deac9c834d2340cf94172efc557bf5975d283d1c1138c2a67b05341"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a large, consensus-critical rework of Casper's merge recovery and finality logic: RejectedDeploy records gain consensus-visible `duplicate`/`carrier` fields, deploy-storage eviction moves from node-local finalization markers to floor-state evidence, a merge-time floor-keyed validity-window rule is added, and a new FloorContext memoizes per-operation floor derivations. The design rationale is unusually well documented and new regression tests are included. However, several points need human judgment before merge: the protocol-format change to RejectedDeploy has no visible version gating or compatibility story; `FloorContext::derive` failures (including transient local-store misses) are converted into block invalidity during validation; the finalization-time sweep hard-fails on a missing floor block and performs O(stored_deploys) state probes per floor advance; and the deletion of both `resolve_batch` and the `filter_unprocessed_rejected_deploys` invariant removes performance and safety backstops the codebase previously relied on. The diff is also truncated (45 files, only a portion shown), so validation-side enforcement of the new record fields and the FloorContext test coverage could not be fully verified.

The visible portion introduces a coherent floor-derived context and improves carrier-aware rejection handling, but two significant concerns should be resolved before merging: the advertised exactly-once protection is incomplete for deploys invisible to the state probe, and finalization now performs an unbounded full-store scan with potentially expensive per-deploy state queries on every LFB update. The floor-window arithmetic should also be centralized to avoid boundary and overflow inconsistencies. This assessment covers only the non-truncated portion of the diff.

This is a substantial consensus-adjacent hardening of merge recovery: floor-keyed expiry/purge, settled-in-base dedup, carrier/duplicate rejection records, double-apply prevention, and a shared FloorContext to stop divergent floor derivations inside one propose/validate. The direction matches the stated goals and the new regressions are encouraging, but several irreversible paths (deploy-storage sweep vs operation-floor buffer/merge logic, window-boundary equivalence, and cost of effect probes) need careful human review before merge. I would not approve on the partial diff alone; verify caller updates for the removed resolve_batch API, align all window/floor predicates on one helper, and bound or meter the finalization storage sweep.

The code changes introduce significant improvements to the merge recovery and finality windows, with a focus on enforcing exactly-once merge effects and using floor evidence for various operations. While the changes are generally sound, there are a few areas that could benefit from further review and optimization, particularly around ensuring no deploy is executed more than once and optimizing the deploy storage eviction process.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential double-application of deploys</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 331)

The code ensures that a deploy is not executed twice by checking if its effect is already in the pre-state. However, this check might not cover all edge cases, especially when deploys are reinstated from scope. It's crucial to ensure that no deploy is executed more than once to avoid state corruption.

**Suggestion:**
```
Review the logic to ensure that all possible paths where a deploy might be executed are covered, including those where deploys are reinstated from scope.
```

</details>

<details>
<summary><b>:orange_circle: Major: Pre-state double-apply filter can hide selection/merge coordination bugs</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2898)

Dropping selected deploys whose effects are already in the merged pre-state is a good safety net against toxic double-apply. Because it is provenance-agnostic and only logs, a systematic selection bug (recovery + reinstatement both delivering the same work) becomes silent delay rather than a loud invariant failure. That makes regressions harder to detect in production.

**Suggestion:**
```
Keep the safety net, but also assert/metric when applied_from_scope and deploy_effect_in_state disagree, and add a counter/alert for dropped-already-applied deploys so the underlying dual-delivery path cannot rot unnoticed.
```

</details>

<details>
<summary><b>:orange_circle: Major: Exactly-once guard has known false negatives</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2910)

The new pre-state guard claims to prevent every deploy whose effect is already present from executing again, but its fallback relies on deploy_effect_in_state. Elsewhere in this change, the purge documentation explicitly states that this probe cannot detect deploys that create no number cells. applied_from_scope only covers effects applied by the current merge, so an effect already inherited through the base or a merge fast path can be missed by both checks. Such a selected deploy can then execute again and duplicate arbitrary Rholang effects, contradicting the exactly-once invariant this change is intended to enforce.

**Suggestion:**
```
Do not use a number-cell presence probe as the authoritative exactly-once test. Persist an execution marker for every user deploy in RSpace, or use complete execution/disposition metadata committed in the state. Add a regression using a deploy with observable effects but no created number cells, with its effect inherited from the base and the deploy selected for recovery.
```

</details>

<details>
<summary><b>:orange_circle: Major: Per-deploy state probe loop in create() (N+1 pattern on the propose hot path)</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2925)

The new pre-state dedup in `create` iterates every selected user deploy and, unless covered by `applied_from_scope`, calls `interpreter_util::deploy_effect_in_state` per deploy against the merged pre-state. Similarly, the settled-buffered purge in `prepare_user_deploys_with_policy` calls `ctx.effect_settled_in_floor` per buffered deploy. Each probe is a history-repository lookup; with large buffers or catch-up-heavy selections this is a significant per-propose cost. The deleted `resolve_batch` docstring quantified exactly this problem (200 fetches vs 10,000).

**Suggestion:**
```
Batch the probes (single scan over the relevant state segment for all sigs), or at minimum share results via FloorContext's memo for both the purge and the pre-state dedup in the same propose (the pre-state dedup currently probes the merged pre-state directly, bypassing the memo).
```

</details>

<details>
<summary><b>:orange_circle: Major: Removal of filter_unprocessed_rejected_deploys drops a defensive packaging invariant</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 3246)

The old code filtered rejected_deploys to exclude sigs that also appear in processed_deploys before packaging the block, with a unit test asserting the invariant. Both the filter and the test were deleted. The new design relies on the merge's returned records being consistent with processed deploys, but the code itself notes probe-blind deploys ('a deploy that creates no number cells is invisible to the probe'). If any path still yields a sig both processed and rejected, the block now packages contradictory records, which downstream validators or disposition readers may treat as invalid or misinterpret.

**Suggestion:**
```
Keep a cheap debug_assert or explicit filter (with logging) that no processed sig appears in the packaged rejected_deploys, or add a validation-side check with a regression test replacing the deleted one.
```

</details>

<details>
<summary><b>:orange_circle: Major: Deploy-storage sweep probes every stored deploy on each LFB advance</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 309)

When a new LFB is found, the sweep reads the entire deploy storage and may call deploy_effect_in_state for every deploy whose validity window is still open. That turns finalization into O(stored_deploys × probe_cost) work on the hot path. Under backlog or slow inclusion this can stall finalization and propose pipelines, and the cost grows with uncleared storage rather than with newly finalized blocks.

**Suggestion:**
```
Bound the work: only re-check deploys whose valid_after is still open, index/memoize settled sigs across sweeps, batch probes, or cap per-sweep work and resume on later floor advances. At minimum add metrics for sweep size/duration and a regression test for large storage.
```

</details>

<details>
<summary><b>:orange_circle: Major: Every LFB update performs a full deploy-store scan with per-deploy state probes</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 319)

Whenever new_lfb_found is true, the finalization path reads every deploy from persistent storage and invokes deploy_effect_in_state separately for every non-expired deploy. This makes finalization cost proportional to the entire live deploy pool on every finalized-block update, even when the derived floor advances only slightly. Since deploy storage is populated by externally submitted work and each probe may require history/RSpace reads, a large backlog can substantially delay finalization and provides a resource-exhaustion vector. The proposal path separately performs similar probes, and the FloorContext memo does not apply to this sweep.

**Suggestion:**
```
Make eviction incremental. Persist the last swept floor, index deploys by expiry height, and only inspect deploys newly covered by the advancing floor or newly eligible canonical wins. If state probes remain necessary, provide a batched lookup and enforce a bounded amount of sweep work per finalization cycle, with the cursor persisted for restart safety.
```

</details>

<details>
<summary><b>:orange_circle: Major: Storage eviction floor may diverge from propose/validate operation floor</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 320)

Storage eviction uses floor_of_block(LFB), while buffer purge, retry admission, merge windowing, and bonds packaging use FloorContext derived from the operation's (parents, justifications). If those floors diverge, a deploy can be deleted from persistent storage before operation-floor logic considers it settled/window-closed, or retained long after merge logic has permanently rejected it. This is consensus-adjacent operational coupling and is hard to validate from the partial diff alone.

**Suggestion:**
```
Document the intended lattice relationship between LFB-floor and justification-derived floor, and add tests where LFB lag / justification lag differ. If they are not proven equivalent for eviction decisions, key both paths on the same floor definition or make storage eviction strictly more conservative.
```

</details>

<details>
<summary><b>:orange_circle: Major: Floor-keyed sweep aborts entire finalization on a missing floor block</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 325)

In `compute_last_finalized_block`, the new sweep does `block_store.get(&floor.hash)?.ok_or_else(...)` and propagates an error, aborting the whole LFB computation. On a node where the floor block is temporarily unavailable (fresh sync, restore from snapshot), this makes finalization fail repeatedly rather than skipping the sweep for that round. The sweep is explicitly best-effort ('delay, never loss'), so a hard failure here is stricter than its own design requires.

**Suggestion:**
```
If the floor block is absent, log and skip the sweep for this round instead of failing `compute_last_finalized_block`.
```

</details>

<details>
<summary><b>:orange_circle: Major: Full deploy-storage scan with per-deploy state probes on every floor advance</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 330)

The sweep calls `deploy_storage.lock().read_all()` and then runs `deploy_effect_in_state` (a state/history probe involving I/O) for every stored deploy whose window is still open, on every new LFB. Combined with the removal of the old per-finalized-block eviction, deploy storage can accumulate a large backlog (e.g., after restart or during catch-up), making each finalization pass O(stored_deploys) state probes. The PR also deleted `resolve_batch`, which existed precisely to avoid this class of O(N·M) fetch patterns.

**Suggestion:**
```
Bound the sweep (batching, incremental cursor, or only re-probing deploys whose status could have changed since the last floor), and cache probe results keyed by (floor_state, sig) across floor advances.
```

</details>

<details>
<summary><b>:orange_circle: Major: Transient local-store misses during floor derivation are converted into block invalidity</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/validation_dispatcher.rs` (line 118)

`FloorContext::derive` returns an error when the floor block is 'not in block store' (a RuntimeError in floor_context.rs). In `run_validation_steps`, any `Err` from `derive` is returned as `BlockError::BlockException(ex)`, i.e. the incoming block is judged invalid. During catch-up or after pruning, the floor block referenced by an honest block may simply not be locally present yet — treating that as block invalidity (potentially triggering slashing/quarantine paths) conflates local state absence with protocol violation and could be exploited to partition or stall honest validators.

**Suggestion:**
```
Distinguish 'floor block missing locally' (retry/defer validation, request the block) from genuine derivation failures that indicate an invalid block. Only the latter should map to BlockException.
```

</details>

<details>
<summary><b>:orange_circle: Major: Window-rule boundary must stay bit-identical to admission/expiry filters</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 555)

split_window_closed_chains keeps chains only when every valid_after > floor - lifespan, and tests claim this matches not_expired_deploy / selection. Any off-by-one between merge rejection and proposer admission recreates the issue #197 wedge (admit work the merge must reject, or reject work selection still offers). The definitions now live in multiple files.

**Suggestion:**
```
Extract one shared helper (e.g. window_open_at(floor, lifespan, valid_after)) used by merge, prepare_user_deploys, buffer retain, and storage sweep. Add a table-driven test covering boundary, boundary-1, and boundary+1 for all call sites.
```

</details>

<details>
<summary><b>:orange_circle: Major: Consensus-visible protocol change to RejectedDeploy with no visible migration/compatibility handling</b> (Reported by: anthropic)</summary>

**File:** `models (RejectedDeploy protobuf) / block-storage/src/rust/key_value_block_store.rs` (line 437)

RejectedDeploy gains `duplicate: bool` (tag 2) and `carrier: BlockHash` (tag 3) fields that are explicitly 'consensus content — it travels to the block body as-is'. This changes what goes into the block body (and therefore block hashes / signatures) for merge blocks. Nodes running the previous version will produce and validate blocks without these fields, while new nodes populate them, creating a hard-fork risk. The diff shows no protocol version gating, feature flag, or migration path for blocks already on disk / in flight.

**Suggestion:**
```
Gate the new fields behind a shard/protocol version or activation height, and document the upgrade path. At minimum add cross-version tests: old-format block validated by new node and vice versa.
```

</details>

<details>
<summary><b>:orange_circle: Major: deploy_windows HashMap cloned per deploy chain</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/merging/block_index.rs` (line 197)

`deploy_windows.clone()` is executed once per deploy chain in the loop; for blocks with many chains this clones the full sig→window map repeatedly, even though `DeployChainIndex::new` then filters it down to the chain's own deploys.

---

deploy_windows is built once, then cloned for every chain and filtered inside DeployChainIndex::new. For blocks with many chains this is unnecessary allocation/churn.

**Suggestion:**
```
Pass a borrowed map or pre-filter the window entries for the chain's deploy ids before construction.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: resolve_batch removal may regress rejected-buffer admit path</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 638)

The batched finalized-window resolver is deleted. Comments said should_admit_to_rejected_buffer depended on it to avoid O(N·M) scans. The caller update is not visible in the truncated diff; if it now loops resolve() per sig, catchup merges with many rejections will slow sharply.

**Suggestion:**
```
Confirm the admit path no longer needs batch resolution (e.g. replaced by floor/effect probes). If it still resolves many sigs, restore batching or reuse FloorContext dispositions.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Without a derivable floor, buffered deploys never block-expire</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 470)

`expired_buffered` now only applies the block-expiry bound when `floor_expiry_bound.is_some()`. Combined with the removal-side filter (`floor_expiry_bound.is_some() || !is_retry_sig(...)`), a node that repeatedly fails to derive a floor (or fixture-like shapes) will retain retry work indefinitely. The code documents this as 'delay, never loss', but there is no bound or metric on how long retry work can accumulate if floor derivation degrades in production.

**Suggestion:**
```
Add a metric/log for buffered entries retained past the tip-clock window due to missing floor, so operators can detect a stuck-floor condition.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: rejected_buffer_has_recoverable_deploys tightened for rejected-in-scope entries</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2117)

The old predicate allowed rejected-in-scope entries regardless of block expiry (`rejected_in_scope || not_expired_deploy(...)`); the new one requires `not_expired_deploy(window_bound, ...)` for everything. With `floor_ctx = None` (window_bound falls back to the tip clock), a tip-expired but floor-live rejected-in-scope entry no longer counts as recovery backlog, which is inconsistent with the admission-side behavior that intentionally keeps such entries retryable. This can flip the recovery-mode admission decision in the no-floor case.

**Suggestion:**
```
Confirm whether the no-floor fallback here should mirror the admission side (defer rather than tighten), and add a test covering the `floor_ctx = None` + tip-expired rejected-in-scope combination.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Per-deploy pre-state effect probes during propose</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 2920)

After parent merge, every selected user deploy not in applied_from_scope triggers deploy_effect_in_state against the pre-state. On large blocks or slow history reads this adds propose latency on top of the merge itself, and short-circuit merge shapes are exactly when applied_from_scope is empty so the expensive path is taken.

**Suggestion:**
```
Memoize probe results for the propose (similar to FloorContext.effect_memo), or limit probes to recovery/rejected-retry sigs that are the only dual-delivery sources.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Missing documentation</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/finality/floor_context.rs` (line 1)

The new file `floor_context.rs` lacks module-level documentation. Adding a brief description of the module's purpose and functionality would improve code readability and maintainability.

**Suggestion:**
```
Add a module-level documentation comment at the top of `floor_context.rs` describing its purpose and the functionality it provides.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Disposition/effect memo uses check-then-insert without single critical section</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/finality/floor_context.rs` (line 96)

dispositions() and effect_settled_in_floor() release the lock between cache miss and insert. Concurrent callers can duplicate walks/probes. Results are deterministic so this is wasted work rather than corruption, but propose/validate paths that share a context across tasks could amplify cost.

**Suggestion:**
```
Use the entry API under one lock, or compute outside and insert with entry().or_insert_with only if still vacant.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Validity-window boundary uses unchecked signed subtraction</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 574)

split_window_closed_chains computes floor_block_number - deploy_lifespan with ordinary i64 subtraction, while the finalization sweep uses saturating_add for the equivalent boundary test. Invalid or extreme shard configuration values can therefore panic in overflow-checked builds or produce behavior inconsistent with the eviction path.

**Suggestion:**
```
Validate deploy_lifespan as a non-negative bounded configuration value and use a single shared window-closure helper with saturating or checked arithmetic across merge, admission, and finalization.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Empty-BlockHash sentinel for settled-in-base sigs is fragile</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 770)

Settled sigs are seeded into `latest_for_deploy` with `(i64::MAX, BlockHash::new())` and the comment asserts 'the sentinel hash is empty, which no real chain source can equal'. This is an implicit invariant enforced only by convention; any future code that constructs a chain with a default/empty source_block_hash (tests do use synthetic hashes) would silently collide with the sentinel semantics.

**Suggestion:**
```
Use an explicit enum or Option-based sentinel (e.g., `enum Freshest { SettledInBase, Chain(i64, BlockHash) }`) instead of a magic (i64::MAX, empty-hash) tuple.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Lightweight rejected-deploy decode now depends on protobuf defaulting</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/key_value_block_store.rs` (line 437)

BlockDeploySigsRejectedDeploy gains duplicate: bool tag 2. Old bytes without the field default to false in prost, which is the safe 'counts as real rejection' interpretation. Worth making that compatibility contract explicit so future fields do not accidentally change disposition filtering.

**Suggestion:**
```
Comment that missing duplicate must default false, and cover decode of pre-change block bodies in storage tests.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: prepare_user_deploys_with_policy gains two more Option parameters</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 429)

The function now takes `runtime_manager: Option<&RuntimeManager>` and `floor_ctx: Option<&FloorContext>` alongside a policy struct and six other args, with test call sites passing bare `None, ... None` positionally. The Options exist primarily so fixtures can omit them, which pushes production-vs-test divergence into the type signature.

**Suggestion:**
```
Bundle the optional operation context (runtime + floor) into a single `OperationContext` struct, or provide a test-only builder, so call sites don't juggle positional Nones.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Optimize deploy storage eviction</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 229)

The deploy storage eviction process involves iterating over all stored deploys and checking their validity window and effect settlement. This could be optimized by maintaining a more efficient data structure or indexing mechanism to quickly identify deploys that need to be evicted.

**Suggestion:**
```
Consider using a more efficient data structure or indexing mechanism to optimize the deploy storage eviction process.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Strong invariants; consider a single architecture note</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/finality/floor_context.rs` (line 1)

The PR carefully encodes 'delay, never loss', floor-clock expiry, duplicate-flagged rejection records, and settled-in-base dedup. Those invariants are spread across module docs and long inline comments, which helps reviewers but will drift.

**Suggestion:**
```
Add a short design doc (or top-level casper finality/merge README section) listing the irreversible operations and which floor/evidence each one may use, and link the RED/GREEN tests for #102 / #248–#252 from there.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: merge() signature has grown to 11 positional parameters</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 601)

`merge` now takes floor_block_number, deploy_lifespan, and a `&dyn Fn` probe on top of nine existing parameters, and returns a 4-tuple whose members need long inline comments to explain. This is hard to call correctly and easy to misorder (two adjacent i64s).

**Suggestion:**
```
Introduce a `MergeParams`/`MergeOutcome` struct so floor inputs and outputs (rejected records, applied sigs) are named fields rather than positional tuple members.
```

</details>
---

<details>

This is a large, consensus-critical rework of Casper's merge recovery and finality logic: RejectedDeploy records gain consensus-visible `duplicate`/`carrier` fields, deploy-storage eviction moves from node-local finalization markers to floor-state evidence, a merge-time floor-keyed validity-window rule is added, and a new FloorContext memoizes per-operation floor derivations. The design rationale is unusually well documented and new regression tests are included. However, several points need human judgment before merge: the protocol-format change to RejectedDeploy has no visible version gating or compatibility story; `FloorContext::derive` failures (including transient local-store misses) are converted into block invalidity during validation; the finalization-time sweep hard-fails on a missing floor block and performs O(stored_deploys) state probes per floor advance; and the deletion of both `resolve_batch` and the `filter_unprocessed_rejected_deploys` invariant removes performance and safety backstops the codebase previously relied on. The diff is also truncated (45 files, only a portion shown), so validation-side enforcement of the new record fields and the FloorContext test coverage could not be fully verified.

The code changes introduce significant improvements to the merge recovery and finality windows, with a focus on enforcing exactly-once merge effects and using floor evidence for various operations. While the changes are generally sound, there are a few areas that could benefit from further review and optimization, particularly around ensuring no deploy is executed more than once and optimizing the deploy storage eviction process.

#### openai (gpt-5.6-sol)

The visible portion introduces a coherent floor-derived context and improves carrier-aware rejection handling, but two significant concerns should be resolved before merging: the advertised exactly-once protection is incomplete for deploys invisible to the state probe, and finalization now performs an unbounded full-store scan with potentially expensive per-deploy state queries on every LFB update. The floor-window arithmetic should also be centralized to avoid boundary and overflow inconsistencies. This assessment covers only the non-truncated portion of the diff.

#### xai (grok-4.5)

This is a substantial consensus-adjacent hardening of merge recovery: floor-keyed expiry/purge, settled-in-base dedup, carrier/duplicate rejection records, double-apply prevention, and a shared FloorContext to stop divergent floor derivations inside one propose/validate. The direction matches the stated goals and the new regressions are encouraging, but several irreversible paths (deploy-storage sweep vs operation-floor buffer/merge logic, window-boundary equivalence, and cost of effect probes) need careful human review before merge. I would not approve on the partial diff alone; verify caller updates for the removed resolve_batch API, align all window/floor predicates on one helper, and bound or meter the finalization storage sweep.
</details>

---

jltatbeach (approved):

PR #274, opened 2026-08-15
Hotfix/repin 735a7b95 / regression on shared continuation cursor

jeffrey-l-turner:
Repins to current SI integration test fixes for soaks

Fixes node:
Updates lfs_horizon_requester so a late root can reopen a completed shared continuation cursor.

Adds  regression test with this order:

 1. Root A reaches shared cursor C.
 2. Cursor C completes.
 3. Root B later reaches cursor C.
 4. Cursor C reopens and completes for Root B.
 5. Both roots exist before replay starts.


jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.88 |

---

### Summary

This PR repins SYSTEM_INTEGRATION_REF consistently across all three pin sites (735a7b95), re-enables previously deselected convergence/bonding integration tests, adds an --rss-ceiling-mb 10000 guard, and fixes a real regression in lfs_horizon_requester where a late root could not reopen a completed (Done) shared continuation cursor. The one-line state-machine fix is plausible and is backed by a well-constructed multi-root regression test that verifies the reopen ordering end-to-end. Main concerns are minor: the reopen semantics apply to all Done keys (not just shared cursors) and rely on importer idempotency without documentation, the fix itself lacks an inline comment explaining the non-obvious Done→Init transition, and the test's early synchronization loops lack timeouts so a regression would hang rather than fail crisply. No security issues; safe to merge with the suggested improvements.

The state transition change directly addresses the shared-continuation regression by allowing a completed cursor to return to `Init` when discovered by a later root, and the pinned integration-test updates are consistent across the documented pin sites. No security or production performance problems are evident in the diff. The main improvement requested is to bound all synchronization waits in the new asynchronous regression test so a failure does not deadlock until the workflow-wide timeout.

The core hotfix—allowing ST::add to reinitialize ReqStatus::Done so a late root can reopen a completed shared continuation—matches the stated regression and is covered by a focused async test; CI repins and re-enables previously deselected SI tests are consistent with that intent. The main risk is that Done→Init is a global semantic change to add(), not a narrowly scoped reopen path, so call-site safety and idempotency under re-fetch should be confirmed before merge. The new test is valuable but somewhat timing-dependent and should explicitly assert reopen rather than only final finished/recorded state. No security issues identified in this diff.

The pull request appears to be a straightforward update to pin the system integration reference to the current fixes and adds a regression test. The code changes are minimal and focused, with no apparent security, logic, or performance issues. The test added seems to cover the regression scenario adequately. Minor style suggestions have been provided for improved readability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Done-state reopening applies to all keys, not just shared continuation cursors</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 79)

ST::add previously inserted only missing keys. It now also resets ReqStatus::Done back to Init. That fixes the late-root shared-continuation case, but it changes global behavior for every add() caller. If any path re-feeds already-completed keys (retries, overlapping root sets, duplicate horizon reports), work can be re-queued after successful completion, risking duplicate imports, extra network fetch, or non-termination depending on downstream idempotency.

---

The change to `ST::add` resets ANY key in `ReqStatus::Done` back to `Init`, not just shared continuation cursors. If a caller re-adds a path that legitimately completed (e.g., a duplicate root reference or a path re-derived from a re-received response), the state machine will re-request and re-import already-imported items. This is safe only if the importer is idempotent and the remote peer will re-serve completed paths; if the peer prunes served items, the reopened request can stall until the stream timeout fires. Consider whether the reopening should be scoped (e.g., only when the re-add originates from a distinct root's traversal) or at least documented as relying on importer idempotency.

**Suggestion:**
```
Add an explicit invariant comment on `ST::add` explaining why Done entries may be reopened (late root reaching a completed shared cursor) and confirm/document that the importer path is idempotent for duplicate imports.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improve readability of condition</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 76)

The condition `matches!(self.d.get(&key), None | Some(&ReqStatus::Done))` could be simplified for better readability.

**Suggestion:**
```
Consider using `!self.d.contains_key(&key) || self.d.get(&key) == Some(&ReqStatus::Done)` for clarity.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Regression test is timing/order coupled and can false-pass</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 1245)

stream_reopens_shared_continuation_for_late_root relies on yield loops and a 5s timeout to observe a second request for the shared continuation. If reopen does not happen, the feeder simply skips the second terminal send and drops tx; depending on stream error/finish handling, the test may fail opaquely or miss the intended assertion that continuation was requested twice. It also feeds resp2 only after r1 is recorded, which encodes one interleaving, not the full race space.

**Suggestion:**
```
Assert request_count >= 2 (or that continuation_reopened is true) before drain completes, and fail fast if reopen is not observed. Consider deterministic mock sequencing instead of spin-wait orchestration where possible.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Regression test can hang indefinitely before reaching its bounded assertion</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 1271)

Several feeder loops wait indefinitely for requests or imports, while only the continuation-reopen loop has a five-second timeout. If the requester fails to send either initial request, fails to request the first continuation, or fails to import r1, the feeder retains the channel sender and `drain(stream)` can wait until the external test timeout (currently up to 1200 seconds). This makes failures in the behavior under test slow and difficult to diagnose.

**Suggestion:**
```
Wrap the entire feeder task, or each wait loop, in `tokio::time::timeout` and use descriptive `expect` messages. Alternatively, implement a shared bounded `wait_until` helper so every synchronization point fails promptly.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unbounded busy-wait loops in feeder task can hang the test suite</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 1287)

The first three synchronization loops in the feeder task (`both_sent`, `contains(&feeder_cont)`, `contains(&feeder_r1)`) spin on `yield_now()` with no timeout. Only the final 'continuation reopened' check is wrapped in `tokio::time::timeout`. If the requester regresses such that one of the earlier conditions never becomes true, the test hangs indefinitely rather than failing, which is painful in CI (relies on the outer pytest/CI job timeout rather than a crisp test failure).

**Suggestion:**
```
Wrap each wait loop in `tokio::time::timeout(Duration::from_secs(5), ...)` (or extract a small `wait_until` helper with a deadline) so a regression produces a clear panic instead of a hung test.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Pin-history comment block is growing unbounded</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 191)

The SYSTEM_INTEGRATION_REF comment block accumulates multi-paragraph historical notes. Useful for audit, but it increases noise in a workflow file and duplicates docs/ToDos.md.

---

The `SYSTEM_INTEGRATION_REF` comment now carries three dated changelog entries inline in the workflow file. This history duplicates git blame/commit messages and will keep growing with each repin, drifting from the terser comments in the other two pin sites.

**Suggestion:**
```
Keep a short 'why this pin' note in the workflow and leave full chronology in docs/ToDos.md or a work-log.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Semantic change to state machine lacks doc comment</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 79)

The `matches!(self.d.get(&key), None | Some(&ReqStatus::Done))` change silently alters the request state machine's contract (Done is no longer terminal for a key). The PR description explains the shared-continuation-reopen rationale well, but the code itself carries no comment, making it easy for a future refactor to 'fix' this back to `!contains_key` and reintroduce the regression the new test guards against.

---

The behavioral change is subtle and central to LFS horizon progress. Without a short comment on why Done is re-initialized, future readers may treat it as accidental and regress it, or may extend it unsafely.

**Suggestion:**
```
Add a brief comment above the matches! condition describing the late-root shared continuation cursor case and the intended idempotency expectations after reopen.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: ST::add clones the entire map even when no keys change</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/lfs_horizon_requester.rs` (line 76)

`add` unconditionally clones `self.d` before checking whether any key actually needs insertion. For large horizons this copy-on-every-add pattern is wasteful when the incoming key set is entirely already-tracked (Requested/Received). This is pre-existing behavior, but the new reopen semantics may increase how often `add` is called with already-known keys.

**Suggestion:**
```
Short-circuit: first collect the keys that need `Init`, and return `self.clone()`-equivalent (or the same state) unchanged if the set is empty before cloning the map.
```

</details>
---

<details>

This PR repins SYSTEM_INTEGRATION_REF consistently across all three pin sites (735a7b95), re-enables previously deselected convergence/bonding integration tests, adds an --rss-ceiling-mb 10000 guard, and fixes a real regression in lfs_horizon_requester where a late root could not reopen a completed (Done) shared continuation cursor. The one-line state-machine fix is plausible and is backed by a well-constructed multi-root regression test that verifies the reopen ordering end-to-end. Main concerns are minor: the reopen semantics apply to all Done keys (not just shared cursors) and rely on importer idempotency without documentation, the fix itself lacks an inline comment explaining the non-obvious Done→Init transition, and the test's early synchronization loops lack timeouts so a regression would hang rather than fail crisply. No security issues; safe to merge with the suggested improvements.

The pull request appears to be a straightforward update to pin the system integration reference to the current fixes and adds a regression test. The code changes are minimal and focused, with no apparent security, logic, or performance issues. The test added seems to cover the regression scenario adequately. Minor style suggestions have been provided for improved readability.

#### openai (gpt-5.6-sol)

The state transition change directly addresses the shared-continuation regression by allowing a completed cursor to return to `Init` when discovered by a later root, and the pinned integration-test updates are consistent across the documented pin sites. No security or production performance problems are evident in the diff. The main improvement requested is to bound all synchronization waits in the new asynchronous regression test so a failure does not deadlock until the workflow-wide timeout.

#### xai (grok-4.5)

The core hotfix—allowing ST::add to reinitialize ReqStatus::Done so a late root can reopen a completed shared continuation—matches the stated regression and is covered by a focused async test; CI repins and re-enables previously deselected SI tests are consistent with that intent. The main risk is that Done→Init is a global semantic change to add(), not a narrowly scoped reopen path, so call-site safety and idempotency under re-fetch should be confirmed before merge. The new test is valuable but somewhat timing-dependent and should explicitly assert reopen rather than only final finished/recorded state. No security issues identified in this diff.
</details>

---

jeffrey-l-turner:
## Multi-agent review follow-up

Commit `327a1001de0dfa74946af1e9dbd3716e23741d70` addresses the actionable cursor and test findings.

- Generic `ST::add()` keeps completed entries terminal.
- Private `add_continuation()` limits reopening to pagination continuation paths.
- A unit test confirms that generic `add()` does not reopen completed entries.
- The code documents idempotent writes for repeated continuation imports.
- The regression feeder has a bounded total timeout.
- The regression test explicitly requires a second shared-continuation request.

Local verification completed successfully:

- `cargo test -p casper --lib`: 263 passed.
- `cargo clippy -p casper --lib --tests -- -D warnings`: passed.
- `cargo fmt --all -- --check`: passed.
- The pre-push suite passed all 15 checks.

The pin history remains because this PR must record immutable harness pin changes. The pre-existing map clone remains outside this fix.

PR #280, opened 2026-08-16
Hotfix/repin 7bff4d1f si

### LINKED ISSUE #23: finalization stalls when DAG grows wide from accumulated divergence

jeffrey-l-turner:
## Summary

This hotfix prevents rejected deploy loss and duplicate application during concurrent multi-parent recovery.

The change also protects consensus replay from reporting replay contention. Structured lifecycle events now identify deploy selection, rejection, recovery, and finalization transitions.

## Why this is a hotfix

This PR is incident-driven work from the integration preflight recovery effort. It does not add an unrelated product feature.

Earlier harness defects prevented the preflight from completing its full custom test path. `system-integration` PR #120 repairs those harness defects.

Once the tests reached the concurrency paths, they exposed rejected deploy loss, duplicate application, and replay contention in the node.

The consensus-adjacent changes in this PR directly address those release blockers. The tracing records were necessary to identify failures that occurred only during concurrent execution.

Several smaller recovery policies were tested and reverted because they caused expiry, repeat-deploy errors, finality stalls, or duplicate application.

The final policy passed the full 11-test concurrency module. The combined preflight and uninterrupted 60-hour soak remain required before promotion.

## Changes

- Give consensus replay priority over reporting replay through a cancellation-safe `ReplayLock`.
- Queue reporting replay while consensus replay waits for the shared replay permit.
- Select one validator in stable order for rejected deploy recovery.
- Retain rejected deploys until a clean recovery carrier reaches finalization.
- Keep deploy finalization pending when a later visible rejection can invalidate an earlier carrier.
- Preserve all relevant clean, failed, and rejected lifecycle events during status resolution.
- Prefer the oldest unresolved deploy branch during parent promotion.
- Emit structured events under `f1r3fly.casper.deploy_lifecycle` and `f1r3fly.casper.finalization_lifecycle`.
- Add regression coverage for replay priority, cancellation, recovery leadership, buffer retention, and deploy finalization.
- Align recovery fixtures with the deterministic recovery leader and the correct validator count.

## System-integration relationship

The focused acceptance work used immutable `system-integration` PR #120 head `7bff4d1fa4d438e4b3ff5b6ad1db4e817d08c1ac`.

PR #120 is now merged as `22121713d8874ac84f94f4b3ddb30dda78fa7f88`.

Commit `9d25655a5` pins the merged SHA in these three locations:

- `.github/oci-validation.env`
- `.github/workflows/_integration-pipeline.yml`
- `.github/workflows/merge-recovery-soak.yml`

The repin helper verified the remote commit and updated the pin history. The workflow invariant check confirms that all three values match.

## Validation

- `cargo test --release -p casper`: 846 passed, 0 failed, 10 ignored.
- Repository pre-push checks: 15 passed, 0 failed.
- Full concurrency module: 11 passed with no forbidden lifecycle signatures.
- Focused recovery and deploy-finalization regression tests passed repeatedly.
- Rust language-server diagnostics are clean.

## Review focus

This PR changes consensus-adjacent recovery and finalization behavior. Review the following areas carefully:

- rejection order across sibling finalized branches
- deterministic recovery-leader liveness
- `ReplayLock` notification and cancellation behavior
- per-request deploy-status scan cost
- lifecycle log volume under sustained load

## Convergence fixes (added 2026-08-17)

The repinned SI convergence test exposed a permanent finality stall after a phlo-exhausting deploy. Root cause chain (debugger-verified) and fixes, all on this branch:

1. **Quadratic replay starves block validation** (primary): `ReplayRSpace::locked_consume` cloned a persistent contract's entire recorded-COMM multiset on every replayed fire (~100k COMMs for a 20M-phlo loop), turning a ~30s replay into hours; validation serialized behind it, peer-block admission froze, proposals carried stale justifications, and the clique oracle's mutual-visibility edges failed. Fixed with an empty-store short-circuit plus a consume/produce COMM index (full-scan fallback preserves semantics).
2. **Deploy-branch main-parent promotion pinning**: equal-sig-set branches re-promoted a stale branch to the main-parent slot on every snapshot, keeping validators' main chains disjoint. Fixed with a covered-sig gate (a sibling competes only with deploy sigs the ghost main's ancestry does not already carry).
3. **View-dependent lag-recovery leader election**: hashing the local LFB elected incompatible recovery leaders across divergent views. Fixed with LFB-independent selection.

Result: `test_network_converges_after_slow_deploy` goes from a 280s timeout (~20/20 local failures) to 3/3 passes in 80–86s. Each fix carries a TLA+ model with a pre-fix refutation config, plus Rocq/CbC evidence (`scripts/ci/check-formal-invariants.sh`).

Closes #23.

**Temporary resource exception**: with replay fixed, all replicas replay the deploy concurrently at approximately 3 GB of live heap each. The measured five-node sum is 16.6–18.5 GB. This branch sets both standard integration guardians to 22,000 MB and keeps the convergence test active. Issue #282 tracks the interpreter working-set reduction. Follow-on PR #281 will restore the 10,000 MB ceiling only after exact-commit evidence passes at that limit. The soak preflight keeps its separate 45,056 MB ceiling.

Related to #282.


jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a large, consensus-sensitive change disguised as a hotfix: it rewrites the deploy finalization resolver (replacing ancestry-based canonical invalidation with height-only rejection comparison plus a new unfinalized-rejection scan), changes recovery leadership to a static first validator (removing the self-chain replay exemption), inverts parent-selection scoring to prefer the oldest unfinalized deploy, adds a priority ReplayLock serializing consensus vs reporting replays, and adds extensive lifecycle logging. The test coverage is good and the ReplayLock design is reasonable, but there are real concerns: a lost-wakeup race in acquire_reporting (Notify::notified is not registered until polled) that can hang reporting; the height-only rejection invalidation reintroduces exactly the same-height sibling / same-block edge cases the deleted comments documented as bugs; and static recovery leadership creates a single point of stall. Given the empty PR description and the depth of consensus-behavior changes, this needs careful human review by someone familiar with the finalization/recovery invariants before merge.

The added replay synchronization and tests are useful, but the finalization-status rewrite introduces a significant correctness regression by replacing canonical ancestry checks with a height-only rejection rule. The expanded per-request DAG scans and extensive info-level lifecycle logging also raise operational performance concerns. The rejection semantics should be corrected and the API traversal cost bounded before merge.

This hot-fix mixes a solid ReplayLock (consensus-over-reporting) and extensive deploy-lifecycle observability with several consensus-sensitive behavior changes: finalization status invalidation drops main-chain ancestry for height-only checks and no longer requires clean inclusions to be canonical; rejected-buffer populate stops suppressing visible non-source wins; block creator removes floor-settled buffer purge and self-chain recovery leadership; parent promotion prefers oldest unfinalized deploy over deploy count. Tests were updated to match the new policies, but the deleted multi-parent safety rationale and the breadth of recovery/finality impact mean this needs careful human review before merge rather than a clean approve.

The code changes introduce several improvements and fixes, but there are some logic issues that need to be addressed to ensure correct behavior. Additionally, there are opportunities to improve code style and documentation to make the code more maintainable.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Height-only rejection invalidation replaces ancestry-based checks the old code explicitly warned against</b> (Reported by: anthropic, openai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 512)

The new `rejection_invalidates` closure invalidates a finalized clean/failed inclusion whenever ANY finalized non-duplicate rejection has `reject_height >= inclusion_height`. The deleted code (and its extensive comments) explicitly explained why height comparison alone is wrong in a multi-parent DAG: (1) a rejection in a sibling block at the same or higher height on a different chain does not affect a canonical inclusion, and (2) counting sibling rejections as 'after' the clean inclusion creates a positive feedback loop keeping deploys Pending while the buffer re-proposes them. Additionally, the `>=` comparison means a rejection in the SAME block as the clean inclusion now invalidates it, which the old code explicitly excluded ('Same-block ... must not invalidate'). The old `is_in_main_chain` canonical-descendant gates were removed entirely. This may be intentional (the new `scan_visible_unfinalized_rejections` pass and new tests suggest a redesign), but it re-introduces exactly the false-Pending / recovery-loop failure modes the previous hotfix documented, and needs careful human review against those scenarios.

---

The new rejection_invalidates closure treats any non-duplicate finalized rejection at an equal or greater height as invalidating an inclusion. bfs_finalized_window traverses secondary-parent branches as well as the canonical chain, so this includes unrelated sibling rejections. This removes the previous ancestry and canonical-chain checks and can incorrectly turn a valid finalized inclusion into Pending. The >= comparison also invalidates a clean or failed inclusion when the rejection is recorded in the same block, despite the previous implementation explicitly treating same-block rejection as non-descendant.

**Suggestion:**
```
Either restore an ancestry/canonicality gate (e.g., require the rejection block be a DAG/main-chain descendant of the inclusion block, as is already done via `is_dag_ancestor` for the unfinalized-rejection pass), or add regression tests covering same-height sibling rejections and same-block clean+rejection cases and document why height-only comparison is now safe.
```

</details>

<details>
<summary><b>:red_circle: Critical: recovered_deploy_leader collapsed to static first validator; removes self-chain replay exemption</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 1867)

`recovered_deploy_leader` now always returns `current_proposal_validators(...).first()`, dropping both the main-parent-sender preference and (in `create`) the self-chain rejected-buffer exemption that allowed a validator to replay its OWN merge-rejected deploys without being leader. Consequences: (1) recovery leadership becomes a fixed single validator — if that validator is offline or lagging, rejected-deploy recovery stalls for the whole network with no fallback/rotation; (2) the removed self-chain exemption comment claimed owner replay was 'deterministic and duplicate-safe', so its removal changes liveness characteristics of recovery. Tests were updated to match, so this appears deliberate, but the liveness implication of a static leader with no failover deserves explicit review.

---

recovered_deploy_leader now always returns current_proposal_validators(...).first() (stable order), dropping main-parent sender preference. allow_recovered_deploys no longer ORs self_chain_rejected_buffered, so a validator cannot re-propose its own merge-rejected work unless it is the set leader. This is a consensus-facing policy change that can delay or strand recovery under leader skew; tests were updated to match but production impact needs explicit sign-off.

**Suggestion:**
```
Call out the policy change in the PR description and validate multi-validator recovery latency; consider retaining owner-replay if merge dedup still makes it safe.
```

</details>

<details>
<summary><b>:orange_circle: Major: Incorrect handling of clean finalized events</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 328)

The logic for handling clean finalized events seems to be incorrect. The code pushes events into `clean_finalized_events` but then only checks the latest event in `clean_finalized_event`. This might lead to incorrect state resolution.

**Suggestion:**
```
Ensure that all events in `clean_finalized_events` are considered when resolving the final state.
```

</details>

<details>
<summary><b>:orange_circle: Major: Each status request can traverse the visible DAG twice</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 413)

resolve_from_state now performs both bfs_finalized_window and scan_visible_unfinalized_rejections. The latter starts from every latest message and walks all ancestors down to min(rolling_floor, active_sig_floor), loading complete blocks along the way. The finalized scan uses the same expanded floor. For an old deploy, this can approach a traversal of the full retained DAG for every API request, creating substantial storage I/O and CPU cost and making the endpoint susceptible to request-driven resource exhaustion.

**Suggestion:**
```
Use indexed rejection/inclusion metadata or a bounded/cached DAG view instead of rescanning blocks per request. If historical scanning is unavoidable, enforce a request work limit and reuse one traversal for finalized and unfinalized classification.
```

</details>

<details>
<summary><b>:orange_circle: Major: Canonical ancestry gates replaced by height-only rejection invalidation</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 504)

finalize_sig_state previously required both (a) rejection downstream via is_in_main_chain and (b) rejection itself canonical on LFB's main chain before invalidating clean/failed inclusions. The new rejection_invalidates closure only checks reject_height >= inclusion_height over finalized_rejected_events. The long-form comments explaining why height alone is wrong in multi-parent DAGs (sibling forks, recovery-cycle false-Pending, non-canonical failed pinning) were deleted. This can mis-classify Finalized/Pending/Failed when rejections or inclusions sit on finalized-but-non-canonical branches.

**Suggestion:**
```
Restore canonical-descendant checks (or an equivalent main-chain filter) when selecting clean_canonical/failed_canonical, or document and prove with multi-parent tests why height-only is now safe under the new event collection model.
```

</details>

<details>
<summary><b>:orange_circle: Major: Clean inclusion no longer required to be canonical</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 520)

failed_candidates still gate on canonical_block(&block), but clean_candidates only require !rejection_invalidates(height). A non-canonical clean inclusion in the finalized BFS window can therefore become clean_canonical and drive DeployFinalizationState::Finalized even when its effects are not on LFB's main chain—the exact failure mode the removed comments warned about.

**Suggestion:**
```
Apply the same canonical_block check to clean candidates before accepting them, unless a later step already guarantees main-chain membership.
```

</details>

<details>
<summary><b>:orange_circle: Major: Floor-settled buffered deploy purge removed</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 547)

The settled_buffered path that removed rejected-buffer entries whose sigs are in floor_won_sigs before packing a block was deleted entirely. Only settled_stored is still purged from deploy storage. Stale buffer entries with already floor-settled effects can remain and be treated as recovery candidates again.

**Suggestion:**
```
Confirm whether buffer retention is deliberate under the new lifecycle; if not, restore buffer removal for floor-won sigs (with the new lifecycle logs).
```

</details>

<details>
<summary><b>:orange_circle: Major: Parent promotion priority inverted to earliest deploy</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 53)

better_deploy_branch_score now prefers smaller earliest_deploy_block_number over higher deploy_sig_count (and inverts root_block_number comparison). A branch with one old deploy beats a branch with many newer deploys. This changes which parent becomes main and therefore canonical support / finality shape.

**Suggestion:**
```
Treat as an intentional fork-choice-adjacent policy change; ensure multi-parent integration tests cover starvation of high-volume younger branches.
```

</details>

<details>
<summary><b>:orange_circle: Major: visible_non_source_win no longer suppresses rejected-buffer admission</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1386)

Previously floor_won OR rejected_sig_has_visible_non_source_win skipped buffer populate. The new path only continues on floor_won_sig; visible_non_source_win is computed solely for logging. The test rename from still_blocks_recovery_admission to is_detected confirms the behavioral drop. Deploys that already have a visible non-source win can re-enter the recovery buffer and be re-proposed, increasing recovery churn and double-execution risk (even if merge dedup later collapses copies).

**Suggestion:**
```
If intentional, add integration coverage that a visible non-source win still cannot be re-executed as live effects. If not, restore the OR suppress condition.
```

</details>

<details>
<summary><b>:orange_circle: Major: Lost-wakeup race in ReplayLock::acquire_reporting can hang reporting requests</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 158)

`acquire_reporting` creates `let ready = self.consensus_ready.notified();`, re-checks `consensus_waiters`, then awaits `ready`. However, tokio's `Notify::notified()` future does NOT register with the Notify until it is first polled (or explicitly enabled). If the last consensus waiter decrements the counter and calls `notify_waiters()` in the window between the re-check and the first poll of `ready.await`, the notification is lost and the reporting task blocks indefinitely (until another consensus waiter arrives, which may never happen). The existing test `cancelled_consensus_waiter_releases_reporting` will usually pass because the race window is small, but the bug is real under load.

**Suggestion:**
```
Pin and enable the notified future before the condition re-check: `let notified = self.consensus_ready.notified(); tokio::pin!(notified); notified.as_mut().enable(); if self.consensus_waiters.load(...) > 0 { notified.await; }` — this is the pattern the tokio docs recommend for avoiding lost wakeups with `notify_waiters`.
```

</details>

<details>
<summary><b>:orange_circle: Major: Redundant single-event fields kept alongside new event vectors</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 102)

ResolverState now keeps both failed_finalized_event/clean_finalized_event/latest_rejected_event and the new Vec counterparts. Singular fields appear largely superseded by the vectors + post-sort selection, increasing update-risk if one path is maintained and the other is not.

---

`ResolverState` now carries both `failed_finalized_event`/`clean_finalized_event` (single highest) and `failed_finalized_events`/`clean_finalized_events` (full lists). The single-event fields are no longer used for the final decision — `latest_rejected_event` is only used for logging, and `clean_finalized_event`/`failed_finalized_event` appear entirely unused after the refactor. The doc comments on these fields still describe the old canonical-descendant ancestry comparison that was deleted.

**Suggestion:**
```
Collapse to vectors (or derived best-of helpers) once call sites are migrated.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Scan floor now extends to the deploy's valid_after height; resolver does two full BFS passes</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 269)

`scan_floor` is now `min(rolling_floor, active_sig_floor)`, so a query for an old deploy (small `valid_after_block_number`) causes the BFS to traverse potentially the entire finalized DAG down to that height, plus a second full BFS in `scan_visible_unfinalized_rejections` over the same window. `resolve` is invoked per-signature, so each status query repeats both scans. This fixes the correctness issue tested by `resolve_keeps_old_finalized_inclusion_after_rolling_window_advances`, but on long chains a status query for an old deploy becomes an unbounded DAG walk.

**Suggestion:**
```
Consider capping the extra depth, caching per-sig scan results, or sharing the visited/height data between the two BFS passes (the second pass re-fetches every block from the block store even for blocks already visited in the finalized pass).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Extra tip-to-floor BFS on every status resolve</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 406)

scan_visible_unfinalized_rejections walks from all latest_message_hashes down to scan_floor (now also extended by min valid_after among active sigs), cloning block bodies and doing ancestry checks in finalize_sig_state. Coupled with retaining full event vectors and sorting clones per sig, API/status paths can get significantly heavier on large DAGs.

**Suggestion:**
```
Bound the unfinalized scan, reuse visited sets from the finalized BFS where possible, and avoid cloning full event vecs when only a best candidate is needed.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: High-volume info logging on deploy lifecycle hot paths</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 483)

Per-deploy tracing::info! was added across buffer remove/add, selection, carrier creation, rejection, finalization, merge, and parent promotion. Under load this will dominate log I/O and may include large parent hash lists. Same pattern appears in finalization_runner, interpreter_util, snapshot, and finalizer.

**Suggestion:**
```
Default these to debug/trace, sample, or gate behind a feature/config flag; keep info for aggregate counts only.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Redundant logging</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 483)

There are multiple logging statements that seem to be redundant or could be combined to reduce verbosity.

**Suggestion:**
```
Consolidate logging statements where possible to improve readability and reduce noise.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Per-deploy info logging can create excessive log volume</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/blocks/proposer/block_creator.rs` (line 3332)

The change adds info-level lifecycle events for every selected, created, rejected, finalized, buffered, expired, and removed deploy. Some events also allocate and format complete parent or deploy-signature vectors. On busy nodes this can materially increase CPU, allocation, disk usage, and log ingestion costs, especially because several stages log the same deploy.

**Suggestion:**
```
Use debug or trace level for per-deploy events, sample or aggregate routine events, and reserve info level for exceptional state transitions. Avoid collecting full vectors solely for logging unless the target is enabled.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent variable naming</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 195)

The variable `directly_finalized_hash` is assigned but not consistently used throughout the function.

**Suggestion:**
```
Ensure consistent use of variable names or remove unused assignments.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Per-deploy INFO-level logging in hot paths</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 241)

The PR adds `tracing::info!` per deploy and per rejection in block creation (`carrier_created`, `selected`), finalization (`finalized_inclusion`, `finalized_rejection`), merge (`merge_rejected`, `buffer_candidate`), and status resolution (`status_resolved`). At INFO level (typically enabled by default) this produces one log line per deploy per block per event, which can be substantial log volume and I/O on busy nodes, and each line pays hex-encoding costs for sigs/hashes/parents lists.

**Suggestion:**
```
Demote these lifecycle events to `debug!`/`trace!`, or document that operators are expected to filter the `f1r3fly.casper.deploy_lifecycle` target and verify default subscriber config excludes it.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: New test inserted between a doc comment and the function it documents</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (line 1804)

`deploy_support_prioritizes_oldest_unfinalized_deploy` was inserted directly between the multi-line comment describing 'DELIBERATELY tiny domains' (which documents `arb_scored_branch`) and `arb_scored_branch` itself, so the comment now reads as documentation for the wrong item.

**Suggestion:**
```
Move the new test above the comment block or below `arb_scored_branch`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: acquire_reporting can starve indefinitely and busy-loops under contention</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 175)

Consensus priority is enforced via a retry loop: reporting acquires the permit, re-checks `consensus_waiters`, drops the permit, and `yield_now`s. Under sustained consensus replay load, reporting requests can be starved without bound, and the drop-permit/yield/retry pattern is effectively a spin loop when consensus waiters continually arrive. This is likely acceptable by design (consensus must never wait on reporting), but there's no fairness bound or timeout for reporting callers.

**Suggestion:**
```
Document the intentional starvation, and consider a timeout on `acquire_reporting` in `RhoReporterCasper::trace` so API callers get an error instead of hanging arbitrarily long.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Hotfix PR with empty description for consensus-behavior changes</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/api/deploy_finalization_status.rs` (line 504)

The PR deletes the detailed invariants for invalidation and recovery-cycle prevention without replacing them with an updated model description. Given the algorithm change, future regressions will be hard to reason about.

---

This PR substantially changes deploy finalization semantics, recovery leadership election, and parent-selection scoring — all consensus-adjacent behavior — plus deletes ~80 lines of carefully written invariant documentation, yet the PR description is empty and the title ('Hotfix/repin 7bff4d1f si') gives no rationale.

**Suggestion:**
```
Add a PR description explaining the motivating incident/bug, why the ancestry-based invalidation was replaced with height comparison, and why the self-chain recovery exemption was removed.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Missing documentation for new struct</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 117)

The new `ReplayLock` struct lacks documentation explaining its purpose and usage.

**Suggestion:**
```
Add documentation to describe the `ReplayLock` struct and its methods.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: ReplayLock priority design looks sound</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 117)

Consensus waiters take priority over reporting via waiter count + Notify, with drop safety on cancelled consensus waiters and unit/integration coverage. acquire_reporting's loop correctly rechecks after acquiring the semaphore. No functional defect spotted; worth watching for livelock only under extreme reporting spam (yield_now backoff is present).

**Suggestion:**
```
Optional: cap reporting spin/yield or use a fairer queue metric for observability.
```

</details>
---

<details>

This is a large, consensus-sensitive change disguised as a hotfix: it rewrites the deploy finalization resolver (replacing ancestry-based canonical invalidation with height-only rejection comparison plus a new unfinalized-rejection scan), changes recovery leadership to a static first validator (removing the self-chain replay exemption), inverts parent-selection scoring to prefer the oldest unfinalized deploy, adds a priority ReplayLock serializing consensus vs reporting replays, and adds extensive lifecycle logging. The test coverage is good and the ReplayLock design is reasonable, but there are real concerns: a lost-wakeup race in acquire_reporting (Notify::notified is not registered until polled) that can hang reporting; the height-only rejection invalidation reintroduces exactly the same-height sibling / same-block edge cases the deleted comments documented as bugs; and static recovery leadership creates a single point of stall. Given the empty PR description and the depth of consensus-behavior changes, this needs careful human review by someone familiar with the finalization/recovery invariants before merge.

The code changes introduce several improvements and fixes, but there are some logic issues that need to be addressed to ensure correct behavior. Additionally, there are opportunities to improve code style and documentation to make the code more maintainable.

#### openai (gpt-5.6-sol)

The added replay synchronization and tests are useful, but the finalization-status rewrite introduces a significant correctness regression by replacing canonical ancestry checks with a height-only rejection rule. The expanded per-request DAG scans and extensive info-level lifecycle logging also raise operational performance concerns. The rejection semantics should be corrected and the API traversal cost bounded before merge.

#### xai (grok-4.5)

This hot-fix mixes a solid ReplayLock (consensus-over-reporting) and extensive deploy-lifecycle observability with several consensus-sensitive behavior changes: finalization status invalidation drops main-chain ancestry for height-only checks and no longer requires clean inclusions to be canonical; rejected-buffer populate stops suppressing visible non-source wins; block creator removes floor-settled buffer purge and self-chain recovery leadership; parent promotion prefers oldest unfinalized deploy over deploy count. Tests were updated to match the new policies, but the deleted multi-parent safety rationale and the breadth of recovery/finality impact mean this needs careful human review before merge rather than a clean approve.
</details>

---

jeffrey-l-turner:
## Design response and remediation status

The review correctly identifies that this PR changes consensus-adjacent policy. The following changes are intentional.

### Intentional decisions

1. **Later finalized rejection events remain effective across sibling LFB transitions.**
   Main-parent ancestry alone missed a traced rejection after a clean secondary carrier. That gap caused premature terminal status and deploy loss.

2. **Clean secondary-parent carriers can become effective finalized carriers.**
   Multi-parent merge state can include secondary-parent effects. Main-chain membership alone cannot classify these carriers.

3. **One stable validator performs rejected-deploy recovery.**
   Owner replay and rotating leadership produced duplicate application during integration tests. Stable leadership removed concurrent recovery proposals.

4. **The rejected buffer retains work after an earlier visible or floor win.**
   A later finalized rejection can invalidate that earlier carrier. Early buffer removal caused permanent deploy loss.

5. **Oldest unresolved deploy branches receive parent-promotion priority.**
   Deploy-count-first scoring starved old carriers during sustained load.

6. **A visible non-source win is diagnostic, not terminal.**
   Suppression on that signal lost high-fanout deploys before final rejection disposition.

### Finding classification

| Finding | Classification | Response |
|---|---|---|
| Equal-height or same-block rejection invalidates a clean carrier | Valid defect | Change `>=` to `>` and add an equal-height sibling regression test. |
| Static recovery leader can become unavailable | Accepted design risk | Keep stable leadership. A failover design must prevent concurrent recovery before implementation. |
| Resolver ignores clean-event history | False positive | Resolution sorts and checks all entries in `clean_finalized_events`. |
| Status resolution traverses the DAG twice | Valid performance issue | Stop the unfinalized scan at the finalized frontier, then measure remaining work. |
| Height rules replace main-parent ancestry | Intentional policy | Preserve later-height invalidation across sibling LFB transitions. Equal-height invalidation is not intentional. |
| Clean inclusion does not require main-chain membership | Intentional policy | Secondary-parent effects can enter merged finalized state. |
| Floor-settled rejected-buffer purge was removed | Intentional policy | Retain recovery work until final rejection disposition or expiry. |
| Parent promotion favors the oldest unresolved deploy | Intentional policy | Prevent starvation of old deploy carriers. |
| Visible non-source wins no longer suppress buffer admission | Intentional policy | Preserve recovery after a visible carrier later loses effect. |
| `ReplayLock` can lose a `Notify` wakeup | Valid defect | Register and enable the notification before the waiter-count check. |
| Singular resolver fields duplicate event vectors | Valid maintenance issue | Remove unused clean and failed singular fields. Keep derived logging data only. |

### First remediation

The local patch now uses strict later-height invalidation. An equal-height sibling rejection no longer invalidates a clean carrier.

The new regression test passes. All 14 deploy-finalization API tests pass.

I will address the remaining valid findings separately so each change keeps focused evidence.

jeffrey-l-turner:
## Why this work is a hotfix

The phrase “disguised as a hotfix” does not match the development evidence.

This PR comes from the integration preflight recovery effort. It does not contain an unrelated product feature.

Earlier harness defects stopped the preflight before the full custom concurrency path could complete. `system-integration` PR #120 repairs those harness defects.

Once the preflight reached those paths, the node lost rejected deploys and applied one high-fanout write twice. Reporting replay also contended with consensus replay.

The recovery, finalization, and replay changes directly address those release blockers. Structured tracing was necessary because the failures occurred only under concurrent multi-parent execution.

We tested and reverted smaller alternatives. Those alternatives caused expiry, `InvalidRepeatDeploy`, finality stalls, or duplicate application.

The retained policy passed the complete 11-test concurrency module twice. The latest run passed in 15 minutes with no forbidden lifecycle signatures.

This evidence does not remove the need for careful consensus review. It explains why the hotfix has a larger internal scope than a simple line change.

The release remains blocked on these gates:

- the exact merged `system-integration` repin
- the combined sequential preflight
- the uninterrupted 60-hour soak

The PR description now records this context and the remaining gates.

jltatbeach (approved):

PR #291, opened 2026-08-19
fix: stop worker starvation that stalls heartbeat networks (#160)

9Neechan:
Four fixes for the worker-starvation chain that eventually collapses consensus, found and verified across the investigation logged on #160.

1. Bound the connection-accept path (comm/src/rust/transport/f1r3fly_server.rs)
F1r3flyServer::incoming() spawned a detached, untimed task per accepted TCP connection to run the TLS handshake. A peer that stalled or went silent mid-handshake leaked the accepted socket (and its task) for the life of the process. Wrapped the handshake in tokio::time::timeout(HANDSHAKE_TIMEOUT, ...) (30s), matching the client-side connector's existing timeout.
The handshake result was then handed to a bounded mpsc::channel(10) via tx.send(result).await with no timeout. If the receiver (tonic's own accept loop) fell behind draining it, a task whose handshake had already succeeded blocked on send() forever, holding its now-established socket open. Wrapped the send in tokio::time::timeout(CHANNEL_SEND_TIMEOUT, ...) (5s) and drop the connection on timeout instead of blocking.
Verified in isolation: two independent soak reruns (fix 1 alone, then fix 1+2) each measurably shrank the open_fd_count/tcp_conn_count leak and eliminated the original "tls handshake eof" log signature entirely — necessary, but not sufficient on its own to prevent the eventual stall.
2. Root cause: the mergeable-channels GC pass monopolized the worker thread
With the connection-accept leaks ruled out, the network still stalled the same way, now surfacing as generic gRPC request timeouts on already-established, healthy connections — pointing at something making the responding node slow to answer, not a connection problem.

Root cause: the mergeable-channels garbage collector ran every 5 minutes and re-scanned the entire DAG from genesis, and for every candidate block walked the main-parent chain of every validator against it — cost that grows with the age of the whole chain, roughly quadratically. collect_garbage was declared async fn, but its body contained not a single .await. Tokio only yields a worker to another task at an .await point; with none present, the worker ran the GC pass's synchronous LMDB reads and O(chain × validators × depth) main-chain walk to completion, unable to service anything else — including answering a peer's RPC within its 5s timeout. One missed response pushed the pending-block frontier past the cap, after which block proposal stopped outright.

Confirmed in the logs: the end of each stall lined up exactly with the GC pass's own completion log line, and the peer's "Request timeout" fired exactly 5s later — its configured per-RPC timeout — directly tying this pass to the stall.

Fix A (shared/src/rust/store/lmdb_key_value_store.rs, node/src/rust/runtime/setup.rs): wrapped the blocking LMDB calls and the GC pass itself in tokio::task::block_in_place, so tokio can explicitly hand the worker off to another task for the duration of the blocking work. Verified with a 6-hour Friday heartbeat run: the node did not fall over for the full 6 hours — but the GC pass's own latency kept degrading over the run, so its algorithmic cost still needed fixing.

Fix B (casper/src/rust/util/mergeable_channels_gc.rs, casper/src/rust/metrics_constants.rs): rebuilt the traversal. The pass now walks a bounded window of max_parent_depth + mergeable_channels_gc_depth_buffer (~110 blocks in this shard's config) above a persisted watermark (GcState), instead of the whole chain, and never revisits already-cleared history between passes. Each validator's main chain is now computed once per pass into a HashSet (build_main_chains) instead of being walked once per candidate block. Added a mergeable_channels_gc.pass.time histogram so pass cost stays observable going forward.

Deletion safety is unchanged: eligibility still anchors to the finalized floor (floor_of_block), not the tip — a tip-anchored bound was previously found (and is covered by a regression test, a_block_above_the_floor_is_never_collected_however_far_the_tip_has_run) to delete merge data still needed while the floor lagged the tip during exactly this kind of stall.

Final verification: an image with both fixes ran with no noticeable degradation — the bounded-window pass no longer grows with DAG size.


9Neechan:
@jeffrey-l-turner It seems like there is  **flaky test: `finalized_win_pending_rejection_spec::finalized_noncanonical_deploy_is_reproposed_after_canonical_rejection`**

This test fails intermittently (~1 in 5–6 runs) even in complete isolation, running as the only test in the binary with no parallelism (`cargo test --release -p casper --test mod finalized_noncanonical_deploy_is_reproposed_after_canonical_rejection`), so it is not purely a resource-contention issue with the shared genesis-block cache under nextest's parallel execution. It has no relation to the files changed in this PR.

Failure:

```
panicked at casper/tests/batch2/finalized_win_pending_rejection_spec.rs:325:5:
deploy ... must be re-proposed after the canonical merge rejects its finalized noncanonical carrier
```

On the passing runs, the log always shows a `pre-charge evaluation failed: Insufficient funds` line right after the `floor-of-view does not capture the current LFB; holding` warning, before the assertion at line 325. On the one failing run this line was missing, which suggests a timing-dependent race inside the test's own choreography — the re-propose does not always land before the assertion checks for it.

Is this a known flake? If not, is there a preferred way to reproduce/investigate it (e.g., a seed to pin, or a suspected race in `TestNode`/`block_creator` propagation timing)?

spreston8:
> @jeffrey-l-turner It seems like there is **flaky test: `finalized_win_pending_rejection_spec::finalized_noncanonical_deploy_is_reproposed_after_canonical_rejection`**
> 
> This test fails intermittently (~1 in 5–6 runs) even in complete isolation, running as the only test in the binary with no parallelism (`cargo test --release -p casper --test mod finalized_noncanonical_deploy_is_reproposed_after_canonical_rejection`), so it is not purely a resource-contention issue with the shared genesis-block cache under nextest's parallel execution. It has no relation to the files changed in this PR.
> 
> Failure:
> 
> ```
> panicked at casper/tests/batch2/finalized_win_pending_rejection_spec.rs:325:5:
> deploy ... must be re-proposed after the canonical merge rejects its finalized noncanonical carrier
> ```
> 
> On the passing runs, the log always shows a `pre-charge evaluation failed: Insufficient funds` line right after the `floor-of-view does not capture the current LFB; holding` warning, before the assertion at line 325. On the one failing run this line was missing, which suggests a timing-dependent race inside the test's own choreography — the re-propose does not always land before the assertion checks for it.
> 
> Is this a known flake? If not, is there a preferred way to reproduce/investigate it (e.g., a seed to pin, or a suspected race in `TestNode`/`block_creator` propagation timing)?

Resolved in https://github.com/F1R3FLY-io/f1r3node-rust/pull/293

jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Needs Review | 0.95 |

---

### Summary

This is a well-motivated, well-documented fix for a real worker-starvation chain, with good test coverage of the deletion-safety invariants and the new watermark behavior. The transport-layer timeouts are sound, and the GC rewrite (per-pass main-chain precomputation + watermark) is the right algorithmic direction. However, there are two substantive concerns: (1) the watermark can stall permanently on orphaned/never-finalized blocks below the floor, which over time re-introduces the unbounded per-pass scan and processed-set growth the fix aims to eliminate — this needs a regression test with an orphaned fork; and (2) wrapping every LMDB call (including microsecond-scale hot-path gets) in block_in_place carries meaningful runtime overhead and worker churn that should be benchmarked or scoped down to bulk operations. Additionally, GcState is in-memory only despite the description calling it 'persisted', so restarts pay a full-chain first pass. The code is safe to merge functionally, but the orphan-stall edge case should be addressed promptly.

The TLS and blocking-work changes improve failure duration and Tokio worker availability, but two important boundedness problems remain. Handshake tasks are still unlimited under connection floods, and the new GC watermark can be permanently pinned by a normal stale DAG fork, causing traversal cost and processed state to grow with chain age. These should be addressed and covered by soak-style or targeted regression tests before merge.

The code changes look good overall. The fixes for worker starvation and the mergeable-channels GC pass seem well-implemented and thoroughly tested. There are a few minor documentation improvements that could be made to enhance clarity.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: A permanently unfinalized fork can pin the watermark and restore unbounded GC work</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 87)

The watermark advances only when every block in a topological level is safe to delete. Since is_safe_to_delete always rejects unfinalized blocks, any stale fork that never becomes finalized can leave level_complete false forever. The pass nevertheless continues over all subsequent levels and records safe blocks in state.processed. As the canonical chain grows, next_height remains pinned, topo_sort(next_height, None) and build_main_chains continue traversing an ever-growing range, and processed can grow without bound. This recreates the chain-age-dependent pass cost and adds an in-memory growth path, contrary to the stated bounded-window behavior. The added test only covers an unfinalized block that is eventually finalized, not a permanently losing fork, which is expected in a DAG.

**Suggestion:**
```
Redesign watermark progression so permanently unfinalized losing branches do not pin a height-based watermark. Consider tracking finalized candidates independently of all blocks at a height, or maintaining a bounded retry set tied to the finalized frontier. Also explicitly cap the candidate traversal at the highest currently eligible height and add a regression test with a stale unfinalized fork while the finalized main chain continues growing.
```

</details>

<details>
<summary><b>:orange_circle: Major: Watermark can stall permanently on never-finalized (orphaned) blocks, re-introducing unbounded per-pass cost</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 95)

The watermark (`state.next_height`) only advances when every block at a level is either already processed or `is_safe_to_delete`. `is_safe_to_delete` returns false for any non-finalized block. Blocks that are orphaned (below the finalized floor but never finalized) will never become finalized, so a level containing one blocks the watermark forever. From that point on, every pass re-runs `topo_sort(state.next_height, None)` over an ever-growing suffix of the chain, and `state.processed` grows without bound with chain age — exactly the quadratic/unbounded behavior this PR set out to fix. The old implementation avoided this by simply filtering to finalized blocks rather than requiring level completeness. In any shard that ever produces an orphan or equivocation, this fix degrades back to the original pathology over time.

**Suggestion:**
```
When a block is deep below the floor (depth_from_floor > max_allowed_depth) and not finalized, treat it as permanently orphaned for watermark purposes (it can never be finalized once the floor has passed it), i.e., count it toward `level_complete` (and optionally still skip deleting its data, or delete it since it can never be merged). Add a regression test with an orphaned fork below the floor verifying the watermark still advances.
```

</details>

<details>
<summary><b>:orange_circle: Major: TLS handshake concurrency remains unbounded</b> (Reported by: openai)</summary>

**File:** `comm/src/rust/transport/f1r3fly_server.rs` (line 173)

The timeout bounds the lifetime of each accepted connection, but the server still spawns one task per accepted socket without any concurrency limit. A remote peer can open connections faster than the 30-second timeout expires, creating a large number of sockets, tasks, TLS states, and CPU-intensive handshakes. The bounded mpsc channel applies only after the handshake and therefore does not protect the expensive accept/handshake stage. This remains a practical resource-exhaustion and denial-of-service path.

**Suggestion:**
```
Guard accepted handshakes with a configurable tokio::sync::Semaphore or another bounded admission mechanism. Acquire a permit before spawning the handshake task, or immediately reject/drop new sockets when the limit is reached, and retain the permit through channel handoff.
```

</details>

<details>
<summary><b>:orange_circle: Major: Wrapping every LMDB operation (including hot-path get/put) in block_in_place is expensive</b> (Reported by: anthropic)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 24)

`in_blocking` now wraps every single `get`, `put`, `delete`, etc. in `tokio::task::block_in_place`. `block_in_place` is not free: it demotes the current worker, migrates its run queue to another thread, and later restores it. LMDB point reads are typically microseconds, so on hot paths (block store lookups, DAG metadata reads, which happen thousands of times per second) the block_in_place overhead can dominate the actual I/O and cause significant worker churn and throughput regression. The GC pass needed this because it does long, tight loops of LMDB calls — but that is already covered by the `block_in_place` wrapper in `setup.rs` around the whole pass.

**Suggestion:**
```
Either (a) restrict `in_blocking` to the bulk/iteration operations (`iterate`, `iterate_while`, `to_map`, large batch `get`/`put`/`delete`), (b) only wrap when the batch size exceeds a threshold, or (c) rely on the pass-level `block_in_place` in setup.rs and drop the per-call wrapping. Benchmark block-processing throughput before/after to confirm no regression.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Add documentation for GcState struct</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 12)

The GcState struct should have a doc comment explaining its purpose and fields.

**Suggestion:**
```
Add a doc comment above the GcState struct definition.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Duplicate main chains built for validators sharing the same latest message</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 164)

`latest_message_hashes().values()` can contain the same hash for multiple validators; each duplicate rebuilds and stores an identical HashSet of the full window.

**Suggestion:**
```
Dedupe latest-message hashes (e.g., collect into a HashSet first) before building chains; the safety check only needs the distinct chains.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: build_main_chains fails the entire pass if a main-parent hash is missing from the DAG</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 168)

`dag.block_number_unsafe(&hash)?` inside the walk propagates an error for any main-parent hash not present in DAG metadata (e.g., a pruned or missing ancestor). The old per-candidate `is_in_main_chain` path likely tolerated this differently. A single missing ancestor aborts the whole GC pass every time, permanently disabling GC.

**Suggestion:**
```
Treat a missing main-parent as the end of that validator's chain (break) instead of propagating the error, or at least log and skip that validator's chain.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Add documentation for HANDSHAKE_TIMEOUT and CHANNEL_SEND_TIMEOUT constants</b> (Reported by: bedrock)</summary>

**File:** `comm/src/rust/transport/f1r3fly_server.rs` (line 20)

The HANDSHAKE_TIMEOUT and CHANNEL_SEND_TIMEOUT constants should have doc comments explaining their purpose and values.

**Suggestion:**
```
Add doc comments above the constant definitions.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: GcState is process-lifetime only, but PR description says 'persisted watermark'</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/runtime/setup.rs` (line 884)

The PR description claims the pass walks above a 'persisted watermark (GcState)', but `GcState` lives in an in-memory `Arc<Mutex<..>>` created at setup. On every node restart the watermark resets to 0 and the first GC pass re-scans the entire chain from genesis (mitigated only by `block_in_place`, not by the algorithmic fix). On a long-lived shard this makes the first post-restart pass arbitrarily expensive.

**Suggestion:**
```
Either persist `next_height` to the key-value store between restarts, or initialize the watermark on startup from the finalized floor minus the retention window (anything below floor - max_allowed_depth that was already collected is idempotent to re-scan, but skipping it saves the full-chain walk).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: in_blocking can panic when called from a tokio LocalSet</b> (Reported by: anthropic)</summary>

**File:** `shared/src/rust/store/lmdb_key_value_store.rs` (line 24)

`block_in_place` panics when invoked from within a `tokio::task::LocalSet`, even on a multi-thread runtime. The `runtime_flavor() == MultiThread` guard does not protect against that case. If any current or future code path calls LMDB from a LocalSet task, this will panic at runtime rather than degrade gracefully.

**Suggestion:**
```
Document this constraint on `in_blocking`, or catch the panic risk by checking whether the code runs inside a LocalSet is not detectable — at minimum add a comment and ensure no LocalSet usage touches the store.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: build_main_chains doc comment references 'floor_height' but callers pass the GC watermark</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/mergeable_channels_gc.rs` (line 160)

The parameter is named/documented as `floor_height`, but `run_pass` passes `state.next_height` (the watermark), and the test helper passes 0. The invariant that makes this correct (candidates ≥ watermark, children strictly higher) is only implicit.

**Suggestion:**
```
Rename the parameter to `min_height` or `watermark` and spell out the invariant in the doc comment: children of any candidate are strictly above the watermark, so chains truncated at the watermark still contain every child that matters.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Channel-send timeout silently drops fully-established connections with no observability beyond a warn log</b> (Reported by: anthropic)</summary>

**File:** `comm/src/rust/transport/f1r3fly_server.rs` (line 218)

When the bounded channel is backed up for 5s, a connection whose TLS handshake already succeeded is dropped. The remote peer sees an unexplained close and will retry, potentially amplifying load exactly when the receiver is already behind. There is only a local warn log; no metric to detect this in production.

**Suggestion:**
```
Add a counter metric (e.g., `f1r3fly.server.conn_dropped_on_backpressure`) so operators can distinguish backpressure drops from network flakiness, and consider whether 5s is aggressive enough to matter under GC-induced stalls that this PR also fixes.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Nested block_in_place → block_on → (in_blocking's) block_in_place pattern is subtle</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/runtime/setup.rs` (line 955)

The GC path now nests `block_in_place(|| handle.block_on(collect_garbage(...)))`, where `collect_garbage` awaits `floor_of_block` which in turn hits LMDB calls that call `block_in_place` again via `in_blocking`. Nested block_in_place is supported by tokio, but the interaction (including holding `blocking_lock()` across `block_on`) is non-obvious and fragile to future refactors.

**Suggestion:**
```
Consider extracting a small documented helper (e.g., `run_blocking_async`) shared between setup.rs and the LMDB store, and add a comment noting that nested block_in_place is intentional and safe.
```

</details>
---

<details>

This is a well-motivated, well-documented fix for a real worker-starvation chain, with good test coverage of the deletion-safety invariants and the new watermark behavior. The transport-layer timeouts are sound, and the GC rewrite (per-pass main-chain precomputation + watermark) is the right algorithmic direction. However, there are two substantive concerns: (1) the watermark can stall permanently on orphaned/never-finalized blocks below the floor, which over time re-introduces the unbounded per-pass scan and processed-set growth the fix aims to eliminate — this needs a regression test with an orphaned fork; and (2) wrapping every LMDB call (including microsecond-scale hot-path gets) in block_in_place carries meaningful runtime overhead and worker churn that should be benchmarked or scoped down to bulk operations. Additionally, GcState is in-memory only despite the description calling it 'persisted', so restarts pay a full-chain first pass. The code is safe to merge functionally, but the orphan-stall edge case should be addressed promptly.

The code changes look good overall. The fixes for worker starvation and the mergeable-channels GC pass seem well-implemented and thoroughly tested. There are a few minor documentation improvements that could be made to enhance clarity.

#### openai (gpt-5.6-sol)

The TLS and blocking-work changes improve failure duration and Tokio worker availability, but two important boundedness problems remain. Handshake tasks are still unlimited under connection floods, and the new GC watermark can be permanently pinned by a normal stale DAG fork, causing traversal cost and processed state to grow with chain age. These should be addressed and covered by soak-style or targeted regression tests before merge.

#### xai (grok-4.5)

</details>

---

jeffrey-l-turner:
**Cross-reference: PR #299 and the starvation framework** (reviewer context, supplements the multi-agent review above)

This fix and PR #299 address the same failure class at two different layers. PR #299 fixes deploy starvation in consensus adjudication. This PR fixes worker starvation in the runtime. Both match the principle now recorded in `docs/casper/CONSENSUS_PHILOSOPHY.md` (Section 3, via #299): individually-correct local decisions can compose into a global liveness failure.

That framework makes the two watermark majors from the panel sharper, because they describe the same hazard shape #299 analyzed:

- The panel found that a permanently unfinalized (orphaned) fork can pin the GC watermark, which restores unbounded per-pass cost. That is the "escape that waits on an event that may never occur" class. The floor machinery solved the same class for retries: the escape derives from an on-chain bound (the validity window and citability horizon), never from waiting for an orphan to finalize (`casper/src/rust/finality/floor_context.rs` states this rule).
- A concrete option in that spirit: bound the watermark pin by block number against the current floor (for example, ignore never-finalized branches older than `deploy_lifespan` below the floor), so the watermark advances on consensus evidence instead of stalling on orphan fate.

No action required for this PR to land — the panel verdicts stand on their own. This note records the connection so the watermark follow-up can reuse the #299 design rule instead of rediscovering it.

See: https://github.com/F1R3FLY-io/f1r3node-rust/pull/299 and `docs/casper/CONSENSUS_PHILOSOPHY.md` (remedy ladder, Principles P2 and P5).

spreston8 (changes requested):
Reviewed at `31ab8b10d` against `dev`. Suggestions below.

**High**

1. The watermark-pinning finding above is confirmed empirically. A probe on this PR's own fixtures — a never-finalized block at height 6, finalized chain extended to tip 13, two passes — leaves `next_height` at 6 with five entries stuck in `processed`. `level_complete` is an AND over the level and `is_safe_to_delete` refuses anything unfinalized, so a losing fork freezes enumeration for good; `watermark_holds_at_a_height_that_is_not_yet_finalized` only covers a block that later finalizes.

2. Worth resolving that by aligning the GC rewrite with #226 rather than patching the watermark in place. #226 rewrites this same file and does not pin: its sweep advances the watermark on enumeration and keeps undeletable blocks in a pending retry set, so an orphan costs one entry instead of freezing the scan. It also replaces `build_main_chains` with a single anchor walk. This PR will probably land first, so adopting that shape here lets #226 rebase onto it instead of the two rewrites having to be reconciled afterwards.

**Medium**

3. Consider reading `HANDSHAKE_TIMEOUT` from `rp_config.default_timeout`, already a parameter of the function that calls `F1r3flyServer::builder` (`grpc_transport_receiver.rs:618`). It carries `network-timeout`, which the deployed configs set to 5 seconds. The constant mirrors `DEFAULT_CONNECT_TIMEOUT`, which a running client never uses — `GrpcTransportClient` passes the configured value through `new_with_timeout` — so as written the server holds a stalled handshake six times longer than the peer that opened it waits.

9Neechan:
Updated Root Cause in #160 

Fixed Requested Chages:

**High**
1. Replaced the level-completion watermark with `GcSweep` from #226. `swept_height` now tracks enumeration progress, while unresolved blocks remain in `pending` for later retries. A permanently unfinalized fork no longer pins the watermark or causes repeated scans of a growing DAG suffix.

2. Kept the floor-based GC ceiling and the strict ancestor predicate from #226. The main-parent walk is additionally bounded by the oldest pending candidate that passes the finality, depth, and child checks. Ineligible orphans do not extend the walk toward genesis.

**Medium**
3. Added a configurable server handshake timeout. `GrpcTransportReceiver` now passes `rp_config.default_timeout` to `F1r3flyServer`, so the server timeout matches the configured client network timeout. The 30-second value remains a fallback for other callers.

**Additional**
Found and fixed an ordering bug in the new retry semantics while reviewing this myself: a candidate was removed from `pending` before the storage delete was attempted, so an error from `block_store.get` or `delete_mergeable_channels` would leave it permanently untracked with its deletion status unknown. Removal now happens only after the storage attempt completes without error. Verified with a minimal standalone repro isolating the exact control-flow pattern (`&mut` state mutated before a fallible call inside a loop, propagated via `?`) — the old order loses the failing candidate from tracking, the new order retains it.

Both the global LMDB and whole-GC-pass `block_in_place` protections remain unchanged because they provide the worker-starvation liveness fix.

spreston8 (approved):

Issue #270, opened 2026-08-14
feat: expose pending-deploy queue via `getPendingDeploys()` RPC (gRPC + HTTP)


semyon-dev:
### Summary

Add a bulk `getPendingDeploys()` API to the node that returns the deploys currently queued in `deploy_storage` (submitted, not yet proposed) and in the `rejected_deploy_buffer` (recovering after a merge conflict). This completes the deploy lifecycle observable by external consumers, which today stops at "included in a block".

### Why this is needed

1. **Indexer cannot represent deploy lifecycle.** The indexer learns about a  deploy only after it is included in a block (it syncs block bodies;
   `deployments.status` in the DB is effectively always `"included"`). It has no way to see submitted-but-not-yet-included deploys or their failure /
   rejection / expiry. A bulk RPC lets the indexer poll once per sync cycle (~5 s) and track `pending -> included -> finalized`, or
   `pending -> failed/expired`, and surface merge rejections via `isRejected`.

2. **Wallet has no post-submit feedback.** After `doDeploy`, a wallet returns  only the signature. There is no status visible until a block lands (which can be many seconds on a quiet network). A wallet can call  `getPendingDeploys(deployerPubkey=...)` to show "pending" immediately,
   list its in-flight transactions, and detect rejected deploys that it should re-submit instead of silently waiting.

3. **Eliminates N+1 per-signature polling.** The only status mechanism today is `deployFinalizationStatus`/`findDeploy`, queried one signature at a time (`rust-client` `deploy-and-wait`, `wait_for_deploy_finalization`, and the load-test harness all poll per signature). A single bulk read is O(1) per  polling cycle instead of O(pending count), reducing load on validators and observers.

### The data already exists on the node

- `KeyValueDeployStorage::read_all()` — all fresh pending deploys  (`block-storage/src/rust/deploy/key_value_deploy_storage.rs:78`).
- `KeyValueRejectedDeployBuffer::read_all()` — the rejected-recovery backlog.
- Both are currently internal-only; the proposer/heartbeat already inspects them but nothing exposes them over the network.

### Proposed API

**gRPC** (proto declaration draft):

```proto
rpc getPendingDeploys(PendingDeploysQuery) returns (PendingDeploysResponse) {}

message PendingDeploysQuery {
  bytes deployerPubkey = 1; // empty bytes = all deployers
}
message PendingDeployInfo {
  DeployDataProto deploy = 1;
[...]

PR #292, opened 2026-08-19
feat: expose pending-deploy queue via getPendingDeploys() RPC (gRPC + HTTP)

semyon-dev:
Closes: issue #270

## Summary
Add a bulk getPendingDeploys() API that returns deploys currently queued in deploy_storage (submitted, not yet proposed) and in the rejected_deploy_buffer (recovering after a merge conflict). Each entry is paired with an isRejected flag so consumers can distinguish fresh deploys from recovery-backlog deploys.
This completes the deploy lifecycle observable by external consumers, which today stops at "included in a block".
Why this is needed
1. Indexer cannot represent deploy lifecycle. The indexer learns about a deploy only after it is included in a block (deployments.status is always "included"). A bulk RPC lets it poll once per sync cycle and track pending -> included -> finalized, or pending -> failed/expired, and surface merge rejections via isRejected.
2. Wallet has no post-submit feedback. After doDeploy, a wallet gets only the signature. A wallet can call getPendingDeploys(deployerPubkey=...) to show "pending" immediately and detect rejected deploys it should re-submit.
3. Eliminates N+1 per-signature polling. Today the only status mechanism is deployFinalizationStatus/findDeploy, queried one signature at a time. A single bulk read is O(1) per polling cycle instead of O(pending count).
API
gRPC — DeployService.getPendingDeploys:
```
rpc getPendingDeploys(PendingDeploysQuery) returns (PendingDeploysResponse) {}

message PendingDeploysQuery      { bytes deployerPubkey = 1; }  // empty = all deployers
message PendingDeployInfo        { DeployDataProto deploy = 1; bool isRejected = 2; }
message PendingDeploysResponsePayload {
  repeated PendingDeployInfo deploys = 1;
  uint32 totalAvailable = 2;  // pre-cap count; compare with deploys.size() to detect truncation
}
```
HTTP — GET /api/pending-deploys[?deployer=<hex>]
Returns PendingDeploysJson:
```{
  "deploys": [
    {
      "term": "...", "timestamp": 1770028092477,
      "phloPrice": 1, "phloLimit": 100000,
      "validAfterBlockNumber": 0, "shardId": "",
      "deployer": "0487def456", "sig": "aa11", "sigAlgorithm": "secp256k1",
      "expirationTimestamp": null, "isRejected": false
    }
  ],
  "totalAvailable": 1
}
```
Design decisions
- Cap & truncate: the response is capped at PENDING_DEPLOYS_MAX_RESULTS = 1000 entries to protect consumers from unbounded responses. totalAvailable reports the pre-cap count so callers detect truncation by comparing deploys.len() < totalAvailable.
- Deterministic truncation: before truncating, entries are sorted by (timestamp, sig bytes) so the cap is reproducible across calls for the same queue state.
- Filter location: the deployer pubkey filter is applied in BlockAPI::list_pending_deploys (after reading both pools), keeping the Casper::list_pending_deploys trait method filter-free.
- Read-only nodes: when Casper is not initialised (with_casper() -> None), the call returns an empty snapshot, not an error — matching the trait's default and the doc-comment.
- Best-effort snapshot: the queue may mutate on the next propose; the call is a point-in-time read (open question from the issue, preserved as documented behaviour).
Changes
Proto (models/src/main/protobuf/):
- DeployServiceCommon.proto — added uint32 totalAvailable = 2 to PendingDeploysResponsePayload.
- DeployServiceV1.proto — getPendingDeploys rpc and PendingDeploysResponse message (pre-staged skeleton).
Casper engine + API (casper/src/rust/):
- api/pending_deploys.rs (new) — PENDING_DEPLOYS_MAX_RESULTS constant, PendingDeploysSnapshot type.
- api/mod.rs — module registration.
- casper.rs — Casper::list_pending_deploys trait method with empty default (pre-staged skeleton).
- engine/multi_parent_casper/block_admission.rs — admit_list_pending_deploys: reads deploy_storage (is_rejected=false) and rejected_deploy_buffer (is_rejected=true).
- engine/multi_parent_casper/dispatch.rs — thin trait delegate.
- api/block_api.rs — BlockAPI::list_pending_deploys: filter by deployer, sort, cap, report total_available.
gRPC (node/src/rust/api/deploy_grpc_service_v1.rs):
- get_pending_deploys handler — converts Signed<DeployData> to PendingDeployInfo via DeployData::to_proto.
HTTP (node/src/rust/):
- api/web_api.rs — WebApi::get_pending_deploys trait method, PendingDeployJson/PendingDeploysJson types, WebApiImpl with hex encoding of deployer/sig.
- web/web_api_routes.rs — GET /pending-deploys route, DeployerQuery, pending_deploys_handler with OpenAPI annotation.
- web/web_api_docs.rs — registered in PublicApi and AdminApi.
Tests
- casper/tests/api/pending_deploys_test.rs (7 tests): no-casper returns empty (not error), fresh deploy → is_rejected=false, rejected buffer → is_rejected=true, both pools in one snapshot, deployer filter, cap truncation with total_available, empty helper.
- node/src/rust/web/web_api_routes.rs (2 tests): returns all when no deployer, filters by deployer.
Verification
cargo build -p node                                          # proto rebuild + compile
cargo test -p casper --test mod api::pending_deploys_test    # 7/7
cargo test -p casper --test mod api::deploy_finalization_status_test  # 8/8 (regression)
cargo test -p node --lib                                     # 136/136
cargo clippy -p casper -p node --all-targets                 # 0 warnings
cargo fmt --all -- --check                                   # clean


spreston8 (changes requested):
Reviewed at `5607d0a3c` against `dev`. Suggestions below.

**High**

1. Worth deriving the snapshot from the predicate in `fresh_local_deploy_stats` (`block_creator.rs:1977-1985`) rather than from `read_all()`. A deploy already in a block, one whose `valid_after_block_number` is ahead of the tip, and a time-expired one are each returned as pending locally — the pool holds included deploys until the lifecycle register terminalizes the sig (`block_admission.rs:189`).
2. Worth deduping the two pools. A signature can sit in both (`block_creator.rs:526`) and is then emitted twice, `isRejected` false and true, double-counted in `totalAvailable`; the predicate above dedupes in its first clause.

**Medium**

3. Consider returning an error when `with_casper()` is `None`, as the other sixteen `BlockAPI` methods do (`block_api.rs:1751`). That branch is a bootstrapping node rather than a read-only one, so an empty snapshot reads as "nothing pending".
4. Worth documenting the queue as node-local. No `CasperMessage` variant carries a deploy and read-only nodes reject `doDeploy` (`block_api.rs:580`), so an observer always answers empty — and `api-reference.md:42-62` directs read traffic there. The indexer case in #270 would be better served by a bulk form of `deployFinalizationStatus`, which is DAG-derived and therefore consistent across nodes.
5. Worth adding an `/api/pending-deploys` section and a `getPendingDeploys` row to `docs/node/api-reference.md`. All 22 routes registered on `dev` have one.
6. Consider asserting the `(timestamp, sig)` order in `cap_truncates_and_reports_total_available`, and routing the handler tests through `pending_deploys_handler`. `StubWebApi` applies the deployer filter itself, so that test still holds if the filter is removed.

**Low**

7. `casper.rs:198` names `NoopCasper`, which does not exist. `web_api.rs:181` omits `expirationTimestamp` when unset, while the description shows it as `null`.

semyon-dev:
@spreston8 could you review it again?

spreston8 (approved):
Reviewed at `169d61066` against `dev`. Approving.

Pushed to this branch rather than requested: a `dev` merge, and a narrower predicate for the reporting path with two tests.

**That second one corrects my own earlier guidance, so it is worth explaining rather than just listing.** The previous review pointed at `stored_deploy_is_pending_for_snapshot` as the node's definition of pending. It is not quite — it answers whether the proposer may put a deploy in the *next* block, which is a stronger question than whether the deploy has landed. The two differ on one clause: a deploy whose `valid_after_block_number` is ahead of the tip is not proposable, but it is submitted, queued, and will land once the chain reaches its window. Filtering on it reported such a deploy as absent, which is the failure the original review was trying to remove, in a new place. `stored_deploy_is_queued` keeps the three exclusions that mean will-never-land or already-did, and drops the future clause.

It also now gates the recovery backlog, which was pushed unfiltered. A deploy whose window closed while it sat in the buffer can never land, and the buffer is only purged when a proposal runs, so a node that is not proposing would have reported it indefinitely.

PR #296, opened 2026-08-19
Rebuild mergeable data the recompute was silently failing to restore

spreston8:
A merge reaching a block whose mergeable data GC has collected calls `ensure_scope_mergeable_present`, which recomputes it. That recompute returned success having persisted nothing whenever the replay cache answered first — exactly the case for a block the node proposed itself, since the cache is keyed by sender and written only by the propose paths. The caller then reported the entry still absent, and that error became a slashable `InvalidTransaction` against the validated block's proposer.

Hit twice in CI on different architectures: run 32241186290 (amd64-docker, validator2) and run 32253774260 (arm64-subprocess, validator3).

- Honor the replay-cache shortcut only when the entry for its post-state is present; otherwise fall through to the full replay, which persists. The guard mirrors one already written into the `StateHashCache` branch, which has never executed since that cache is capped at 0 entries.
- `ensure_mergeable_entry` discarded the hash the replay returned, so a replay reproducing a different post-state stored the entry where nobody reads it and reported only "still absent". Those two cases are now distinct.
- Remove the `StateHashCache`, dead since the repo's first commit. Its key is the pre-state alone, which does not determine a post-state, so a hit was only safe after a mergeable lookup re-derived the identity the key omitted. The replay cache already provides the same optimization keyed by `(pre_state, sender, seq, payload_hash)`.
- Pin system-integration to main's #122 merge, which polls post-recovery FT in both convergence recovery tests.

Two pins run the production components end to end: the real `collect_garbage` collects the block, the real `ensure_mergeable_entry` rebuilds it.

Deliberately not addressed: GC still collects data a lagging merge needs, because its retention is measured from this node's live floor while a merge is anchored on the floor carried by the block being validated. This change makes that condition recoverable rather than fatal; bounding the window is a separate change.


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Verdict:** :bulb: Feedback Provided (50.0000% agreement)

**Reviewed by:**

---

### Summary

This is a well-motivated correctness fix: the replay-cache shortcut in `replay_compute_state` previously returned success without persisting the mergeable entry, causing `ensure_mergeable_entry` (invoked when GC has collected a lagging block's mergeable data) to report the entry still absent and escalate to a slashable InvalidTransaction against an innocent proposer. The fix mirrors the guard that already existed in the dead `StateHashCache` branch — honoring the cache hit only when the mergeable entry for its post-state is present, else falling through to the persisting full replay — and additionally distinguishes replay-determinism failures (computed post-state differs from declared) from storage failures. Removing the never-enabled `StateHashCache` (capped at 0 entries, keyed unsoundly by pre-state alone) is a clean simplification. The two new tests exercise the production `collect_garbage` and `ensure_mergeable_entry` end to end, directly reproducing the CI failure. Remaining issues are minor: a stale doc comment now mislabels `replay_lock`, a slightly overclaiming error message, a small store read added to the cache-hit hot path, and the carried-over pattern of rigging a replay runtime that is immediately dropped. Safe to merge after addressing the doc-comment nit.

The functional change appears sound: replay-cache shortcuts are now used only when the corresponding mergeable entry exists, otherwise execution falls through to the persisting full-replay path. Removing the disabled and incorrectly keyed StateHashCache also simplifies the logic, and the returned post-state is now checked explicitly during reconstruction. No security or material performance issues are evident. Only test documentation cleanup and focused coverage of the new mismatch branch are recommended.

Solid, well-targeted fix for a real slashable false positive: the replay-cache shortcut no longer reports success without restoring mergeable data, post-state mismatch is distinguished from storage failure, and the never-enabled StateHashCache is removed cleanly. The new tests exercise both the own-block cache-hit shape and the production GC + ensure path end to end. Remaining notes are naming/clarity and a minor orphan-key edge on determinism failure; nothing blocking merge.

The code changes appear to be well-implemented and address the identified issues effectively. The removal of the `StateHashCache` and the adjustments to the replay cache logic seem appropriate. Minor style issues were noted, specifically unnecessary debug logs that should be removed for cleaner code. Overall, the changes look good and are safe to merge.

---

### Issues Found

<details>
<summary><b>:yellow_circle: Minor: Stale doc comment left attached to the wrong field</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 215)

The field `state_hash_cache` was removed but its doc comment `/// Optional state hash cache for skipping known replays` was left in place. It now attaches to `replay_lock: Arc<ReplayLock>`, documenting that field as a state hash cache, which is wrong and will confuse readers and doc tooling.

**Suggestion:**
```
Delete the leftover doc comment or replace it with an accurate description of `replay_lock`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary debug log</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 572)

The debug log `[CACHE] Stored state hash mapping` is no longer relevant after the removal of `StateHashCache`. Consider removing it to clean up the code.

**Suggestion:**
```
Remove the debug log at line 572.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary debug log</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 682)

The debug log `[CACHE] Stored state hash mapping` is no longer relevant after the removal of `StateHashCache`. Consider removing it to clean up the code.

**Suggestion:**
```
Remove the debug log at line 682.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Every replay-cache hit now performs a mergeable-store read plus bincode encode</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 741)

The guard adds a `bincode` serialization and a `mergeable_store.contains_key` lookup on the hot cache-hit path of `replay_compute_state`, including the common validation case where the entry is always present. This is almost certainly negligible relative to a full replay, but it converts a pure in-memory shortcut into one that touches the KV store on every hit.

**Suggestion:**
```
Acceptable as-is given correctness requirements; if profiling ever shows this store read on the hot path, consider a small in-memory presence bitset invalidated by GC.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Determinism mismatch error implies entry may already be stored under the wrong key</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 1173)

When `computed_post_state != block.body.state.post_state_hash`, the error says the mergeable entry was stored under the computed key. Full replay persists before this check, so a non-deterministic or corrupt block can leave an orphan mergeable entry under the computed post-state. That is better than the previous silent mis-report, and failing closed is correct, but operators may want GC/orphan cleanup awareness.

**Suggestion:**
```
Acceptable as-is for this fix. If orphan keys matter in practice, consider deleting the computed-key entry on mismatch, or only persisting after the declared post-state check succeeds in the ensure/recompute path.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Error message may misstate where the entry landed</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 1191)

The new post-state mismatch error in `ensure_mergeable_entry` asserts 'the mergeable entry was stored under the computed key'. That is only guaranteed on the full-replay path. If the replay-cache shortcut fired (its guard only checks that an entry exists under `entry.post_state`, not that it matches the block's declared post-state), nothing was newly stored — the entry merely pre-existed under the computed key. The distinction matters for anyone debugging a determinism failure from this message.

**Suggestion:**
```
Soften the wording to 'the mergeable entry exists (or was stored) under the computed key', or track whether the return came from the cache shortcut vs. full replay and say which.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Regression test name and comments describe the pre-fix behavior</b> (Reported by: openai)</summary>

**File:** `casper/tests/util/rholang/runtime_manager_test.rs` (line 3648)

The test `gc_collects_mergeable_data_that_the_recompute_cannot_rebuild` now asserts that recomputation successfully rebuilds the collected entry. Its name and comments such as “the later recompute a no-op” therefore contradict the behavior being tested and could mislead future maintainers.

**Suggestion:**
```
Rename the test to something like `gc_collected_mergeable_data_is_rebuilt_by_recompute` and update the comments to say that this shape previously made recomputation a no-op.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Cache-hit path still rigs a replay runtime only to return a hash</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 731)

On a ReplayCache hit with mergeable present, the code still spawns a replay runtime and rigs the event log before returning `entry.post_state`. That matches prior ReplayCache behavior and is outside the bug fix, but for callers such as `ensure_mergeable_entry` that only consume the hash it is pure work. Worth a follow-up only if this path is hot under recovery load.

**Suggestion:**
```
Optional follow-up: split a 'post-state only' fast path from callers that need a rigged runtime, or document why rigging remains required for all `replay_compute_state` consumers.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Rigged replay runtime is dropped immediately after the cache-hit shortcut</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 767)

On a cache hit with the entry present, the code spawns a replay runtime, rigs it with the cached event log, and then returns — the runtime appears to be dropped without the rigged events being consumed. This behavior is carried over verbatim from the pre-existing cache-hit path, so it is not a regression, but if the rig has no observable side effect on shared state it is wasted work; if it does, that side effect deserves a comment.

**Suggestion:**
```
Either document why rigging the runtime before returning is necessary (e.g., a shared/pooled runtime side effect) or remove the spawn/rig from the shortcut path in a follow-up.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add focused coverage for the post-state mismatch branch</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/util/rholang/runtime_manager.rs` (line 1189)

The new distinction between a missing mergeable entry and replay producing a different post-state is important consensus-facing behavior, but the added tests only exercise successful reconstruction. Without focused coverage, a future change could accidentally remove the comparison or regress the diagnostic while still passing the reconstruction tests.

**Suggestion:**
```
Add a test whose block declares a different post-state, then assert that `ensure_mergeable_entry` returns the replay-determinism error and does not report the generic still-absent storage error.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: GC regression test constructs DAG internals by hand</b> (Reported by: anthropic)</summary>

**File:** `casper/tests/util/rholang/runtime_manager_test.rs` (line 3639)

The `gc_collects_mergeable_data_that_the_recompute_cannot_rebuild` test builds a `KeyValueDagRepresentation` by populating ~10 internal maps/fields directly. This is a valuable end-to-end reproduction of the CI failure using the production `collect_garbage` and `ensure_mergeable_entry`, but it will silently break (or worse, silently stop testing what it intends to) whenever the DAG representation's internal invariants change.

**Suggestion:**
```
Consider extracting a shared test-support builder for synthetic linear DAGs, or asserting the constructed DAG passes whatever invariant checks the storage layer exposes, so future field additions fail loudly.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test name describes the bug, not the expected post-fix behavior</b> (Reported by: xai)</summary>

**File:** `casper/tests/util/rholang/runtime_manager_test.rs` (line 3648)

`gc_collects_mergeable_data_that_the_recompute_cannot_rebuild` reads as if the recompute is still expected to fail. After this fix the test asserts the opposite: GC deletes the entry and `ensure_mergeable_entry` successfully rebuilds it. Future readers may misread a passing test as documenting unrepaired behavior.

**Suggestion:**
```
Rename to something like `gc_collects_mergeable_data_and_recompute_rebuilds_it` (or `..._via_full_replay_when_replay_cache_hits`), and keep the bug narrative in the doc comment.
```

</details>
---

<details>

This is a well-motivated correctness fix: the replay-cache shortcut in `replay_compute_state` previously returned success without persisting the mergeable entry, causing `ensure_mergeable_entry` (invoked when GC has collected a lagging block's mergeable data) to report the entry still absent and escalate to a slashable InvalidTransaction against an innocent proposer. The fix mirrors the guard that already existed in the dead `StateHashCache` branch — honoring the cache hit only when the mergeable entry for its post-state is present, else falling through to the persisting full replay — and additionally distinguishes replay-determinism failures (computed post-state differs from declared) from storage failures. Removing the never-enabled `StateHashCache` (capped at 0 entries, keyed unsoundly by pre-state alone) is a clean simplification. The two new tests exercise the production `collect_garbage` and `ensure_mergeable_entry` end to end, directly reproducing the CI failure. Remaining issues are minor: a stale doc comment now mislabels `replay_lock`, a slightly overclaiming error message, a small store read added to the cache-hit hot path, and the carried-over pattern of rigging a replay runtime that is immediately dropped. Safe to merge after addressing the doc-comment nit.

The code changes appear to be well-implemented and address the identified issues effectively. The removal of the `StateHashCache` and the adjustments to the replay cache logic seem appropriate. Minor style issues were noted, specifically unnecessary debug logs that should be removed for cleaner code. Overall, the changes look good and are safe to merge.

#### openai (gpt-5.6-sol)

The functional change appears sound: replay-cache shortcuts are now used only when the corresponding mergeable entry exists, otherwise execution falls through to the persisting full-replay path. Removing the disabled and incorrectly keyed StateHashCache also simplifies the logic, and the returned post-state is now checked explicitly during reconstruction. No security or material performance issues are evident. Only test documentation cleanup and focused coverage of the new mismatch branch are recommended.

#### xai (grok-4.5)

Solid, well-targeted fix for a real slashable false positive: the replay-cache shortcut no longer reports success without restoring mergeable data, post-state mismatch is distinguished from storage failure, and the never-enabled StateHashCache is removed cleanly. The new tests exercise both the own-block cache-hit shape and the production GC + ensure path end to end. Remaining notes are naming/clarity and a minor orphan-key edge on determinism failure; nothing blocking merge.
</details>

---

jeffrey-l-turner (approved):

Issue #294, opened 2026-08-19
Same-key contention can starve a rejected deploy through repeated losing merges until it expires


jeffrey-l-turner:
**Affected:** rejected-deploy recovery and merge conflict resolution — deploy liveness under sustained same-key contention
**Severity:** medium-high — a valid deploy can terminally fail (`Expired`) with no error surfaced, while consensus stays healthy

## Observed

`test_user_contract_concurrency.py::test_concurrent_delete_and_set_same_key` failed in the Heavy Pipeline (run 32284516371, job `Integration Tests (amd64-subprocess, ucc)`, 2026-08-19, PR #279 at `97596b577` — Rust content byte-identical to dev `421774a3b`). Three concurrent deploys performed delete-and-set on the same contract key; one never reached a coherent terminal verdict. The harness's re-homing-aware classification:

> `304402207fb1a5fe@boot no terminal verdict -> merge-starved (state=Expired): rejected 2x (20 losing merge(s) in this shard's logs) and never re-landed before the validity window closed`

Forensics show consensus was healthy throughout: all 5 nodes agreed on the LFB (block 380, `a153059dfe`), the chain advanced continuously, and no forbidden log signatures appeared. 10 of 11 suite tests passed.

## Suspected mechanism

This is a liveness/fairness defect, not a safety defect. The deploy was conflict-rejected in a merge, re-proposed by the recovery pipeline, and rejected again against competing same-key deploys until its validity window closed. `conflict_set_merger::fold_rejection` resolves same-key conflicts deterministically (the lex-larger sig loses), so a deploy that keeps meeting the same contender can lose every merge deterministically — the recovery pipeline re-proposes it faithfully, but nothing prevents starvation before expiry.

## Distinguish from adjacent issues

- #42 tracks re-inclusion window bloat (explicitly not unsafe) — this issue is about the deploy never landing.
- #102 tracks silent effect loss from finalized noncanonical blocks — here the loss is visible as `Expired`.
- #287/#288 track merge-outcome incorrectness/nondeterminism — here every individual merge outcome is arguably correct; the composition starves.

## Direction

An anti-starvation measure in recovery or conflict resolution, for example: retry-aware priority for previously rejected deploys, inclusion ordering that serializes same-key contenders, or rejection tie-breaking that accounts for prior losses. A deterministic reproduction should encode the losing-every-merge shape before a fix.

spreston8:
[...]

Issue #206, opened 2026-08-06
Formal verification: make exhaustive TLA+ coverage green on a schedule (profile, right-size, disposition the unbounded configurations)


jeffrey-l-turner:
## Context

PR #204 restored the nightly TLA+ gate to green with real multi-validator coverage: 10 fast-tier configurations complete in ~4.3 minutes, including the new `MC_EquivocationDetectorEager_3v2s` (first-ever completed detector verification at 3 validators) and `MC_EquivocationDetector_liveness_2v` (liveness at 2 validators). The exhaustive tier (`RUN_EXHAUSTIVE_TLA=1`, manual dispatch only) remains **red by design**: its three configurations exceed the 45-minute per-configuration cap and are documented as unbounded references.

The maintainer's stated goal is that **all of this eventually succeeds as a regular part of CI** — exhaustive coverage included. This issue tracks the formal verification work required to get there.

Evidence baseline (2026-08-05):
- Exhaustive dispatch [run 31027278093](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/31027278093): 10 fast configurations OK; 3 exhaustive configurations `TIMEOUT` at the full 45m cap; roll-up correctly reports `3 cap timeout(s), 0 violation-or-error(s)`
- Nightly-tier dispatch [run 31052570576](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/31052570576): TLA+ job green
- No configuration has **ever** produced a violation — every red in this system's history has been resource cost, not a counterexample

## The three configurations that cannot complete under the cap

| Configuration | What it checks | Why it can't complete | Prognosis |
| --- | --- | --- | --- |
| `MC_EquivocationDetector_safety` (2v×2s×2b) | Full invariant sweep, safety only | Runtime — docs say "many hours"; **never actually measured** | Likely green with a measured cap on a scheduled run |
| `MC_EquivocationDetectorEager_3v` (3v×3s×2b) | Full-depth 3-validator Eager model; liveness folded into `Inv_LivenessAsSafety` | Pure state-space cost: extrapolating from `_3v2s` (5.7M distinct states / ~2 min), the 3-seqnum space is plausibly hundreds of millions of states — hours of runtime and likely **more memory than a 7GB hosted runner** | Possible on a larger runner with a measured cap; needs profiling |
[...]

Issue #71, opened 2026-06-05
Finalized state can regress under multi-parent merge


spreston8:
Finalization in our Casper guarantees that a **block** is a permanent ancestor of the chain. It does **not** guarantee that the block's **state effects** survive. Under multi-parent merge the two come apart: a later block can keep a finalized block as an ancestor while its merge **overwrites that block's state with a concurrent sibling's value**. So a finalized effect can be silently undone by a block that finalizes after it.

It bites hardest on the PoS **bonds map**, which lives in a single whole-value cell. Two validators bonding concurrently contend on that one cell; the merge keeps one write and drops the other; a **finalized bond is reverted** and the validator set never converges. But the defect is general to any deploy effect on a contended single-value cell.

## What it looks like

Validators produce continuously. A block's **justifications** are the latest message it had seen from each validator; its **parents** are the blocks whose post-states it merges. The two used to differ: before `#288`, fork choice (GHOST) selected a narrow subset of the justifications as parents — the heaviest tip plus a few — and orphaned the rest. **`#288` removed that filter, so a block's parents are now simply all of its justifications** (one block per bonded validator). A block's **pre-state** is the merge of those parents' post-states; its **post-state** is what the pre-state becomes once the block runs its own deploys. So the merge that reconciles a contended cell *is* a block's pre-state computation.

The defect, shown generally as a contended single-value cell `X` (in PoS, `X` is the bonds map and `a`/`b` are two validators bonding concurrently):

<img width="1400" height="1400" alt="Image" src="https://github.com/user-attachments/assets/10567a10-1980-406d-812c-3f0353852008" />

`#4` and `#5` write `X` concurrently — both off the same value at their common ancestor `#3`. `#4`'s write (`X=a`) finalizes, so `#4` becomes the LFB. `#6` then merges `#4` and `#5`, but its base is the **LCA `#3`, which sits below the LFB** — so `#6` re-derives `X` from `#3`, reconciles the two writes, and keeps only one (`X=b`). When `#6` finalizes, the finalized `X` goes `a → b`: a finalized effect is reverted. Block-finality held; state-finality did not.

[...]

PR #299, opened 2026-08-20
fix(casper): rank prior on-DAG rejections above content order in merge adjudication

jeffrey-l-turner:
## Summary

This PR mitigates #294 through phase-1 loss-aware merge adjudication. Sub-issue #317 tracks the partial mitigation delivered by #299 and #312.

Authenticated rejection history outranks content order during merge adjudication. The policy gives a repeatedly rejected deploy priority inside an adjudicable merge.

This PR does not guarantee termination under fixed-proposer main-parent base bias. PR #312 adds merged-frontier packaging and rotating-proposer evidence.

## Changes

- Add the non-identity `prior_rejections` field to `DeployChainIndex`.
- Count kept records across the merge scope and applicable main-parent lineage.
- Return `BlockNotHeld` when required history is unavailable.
- Stamp actual and late chains before conflict resolution.
- Apply loss priority at all three merge adjudication sites.
- Stream rejection counting and use DAG metadata for lineage traversal.
- Add missing-history and fixed-lead adversarial tests.
- Add direct nonzero-loss tests for `get_optimal_rejection`.
- Add an active three-validator neutral-base integration test.
- Preserve the ignored fixed-proposer expected-RED sentinel.
- Stabilize the long PoS contract test and its pre-push timeout.

## Review context and disposition

The merge gate incorporates these reviews:

- [Multi-Agent Code Review](https://github.com/F1R3FLY-io/f1r3node-rust/pull/299#issuecomment-5357673318)
- [spreston8 review](https://github.com/F1R3FLY-io/f1r3node-rust/pull/299#pullrequestreview-4997605837)

| Review concern | Disposition | Merge effect |
| --- | --- | --- |
| Missing lineage can produce node-local counts. | Resolved in `efd103afe`. Missing bodies return `BlockNotHeld`. | Closed. |
| Late chains did not receive rejection-history stamps. | Resolved in `efd103afe`. Actual and late chains receive equal stamping. | Closed. |
| Counting materialized all records and fetched lineage bodies twice. | Resolved in `efd103afe`. Counting streams records, and the ancestry walk uses DAG metadata. | The benchmark remains required. |
| Priority can dominate cost and users can influence rejection history. | The strict loss-first rule is ratified for phase 1. Continued loss farming remains a documented risk. | Accepted partial-mitigation risk. |
| One historical loss can select an economically worse result. | Ratified. Cost decides only when prior-rejection counts are equal. | Accepted consensus rule. |
| The priority owner and chain aggregation were unspecified. | Each deploy signature owns its count. A dependency chain uses the maximum member count. | Ratified. The follow-up branch update applies `max`. |
| User Contract Concurrency evidence is absent. | User Contract Concurrency is waived for PR #299. The disabled suite does not fail on starvation. | A separate change must enable the job and add the expiry assertion. |
| The active test lacked a neutral-base phase-1 matchup. | Resolved in `eac5665a0` by `three_validator_neutral_base_applies_prior_loss_priority`. | Closed. |
| Direct nonzero-loss tests were absent. | Resolved in `eac5665a0`. Tests cover loss-over-cost, equal-loss fallback, and multi-branch loss totals. | Closed. |
| Rejection-count determinism lacked an explicit invariant. | The call site states the invariant. Missing-history tests enforce refusal instead of partial derivation. | Closed. |
| The ignored fixed-proposer test remains RED. | Preserved as the main-parent base-bias sentinel. | #294 remains open.  Sub-issue closes: https://github.com/F1R3FLY-io/f1r3node-rust/issues/317#issue-5219688702|

## Ratified consensus policy

1. Kept, authenticated DAG rejection records supply priority.
2. Each deploy signature owns its prior-rejection count.
3. Dependency-chain priority uses the maximum member count.
4. Prior-rejection count strictly outranks cost.
5. Cost and deterministic content order decide equal-count cases.
6. Fork choice stays independent of deploy content.
7. Missing required history returns `BlockNotHeld`.

A fixed priority cap is not part of phase 1. Saturation can restore deterministic starvation after all contenders reach the cap.

## Ratified scope and release policy

- PR #299 can merge as a partial mitigation while #294 remains open.
- Issue #294 requires deterministic termination for every valid schedule.
- PR #299 and PR #312 do not require atomic merging.
- Sub-issue #317 closes only after PR #299 and PR #312 merge.
- User Contract Concurrency is waived as a PR #299 merge gate.
- The follow-up User Contract Concurrency gate must detect contention expiry and pass three consecutive runs.

## Correct by Construction scope

These production artifacts require mandatory Correct by Construction scope:

- `casper/src/rust/merging/conflict_set_merger.rs`
- `casper/src/rust/merging/dag_merger.rs`
- `casper/src/rust/merging/deploy_chain_index.rs`
- `casper/src/rust/util/rholang/interpreter_util.rs`

The claims cover deterministic count derivation, unavailable-history refusal, total ordering, non-identity priority, and equal chain stamping.

`interpreter_util.rs` already has `cbc=mandatory`. PR #299 records a temporary waiver. PR #311 adds the remaining attributes and formal discharge evidence.

Tests provide implementation evidence. Tests do not prove Rust conformance.

## Performance budget

The rejection-history scan must stay `O(B + R)`. `B` is the unique block count, and `R` is the rejection-record count.

The implementation must load each unique block body no more than once. The benchmark uses these limits:

- Floor distance: 256 blocks.
- Visible scope: 512 blocks.
- p95 latency regression against `dev`: at most 10 percent.
- Peak-memory regression against `dev`: at most 10 percent.

The benchmark is a release gate. Node-local elapsed time must never control block validity or consensus admission.

## Remaining merge gates

- Push the follow-up that applies maximum chain aggregation, block deduplication, and the explicit equal-loss tie-break.
- Record the performance measurement against `dev`.
- Complete required GitHub checks on the final head.
- Run the final multi-review and record each disposition.
- Request a current-head review from `spreston8`.

## Verification

- `cargo test -p casper --lib`: 314 passed on the ratified local follow-up.
- `cargo test -p casper --test mod three_validator_neutral_base_applies_prior_loss_priority -- --nocapture`: passed.
- `cargo fmt --all -- --check`: passed.
- `git diff --check`: passed.

## Remaining Casper issue scope

| Issue | Status after this stack |
| --- | --- |
| #294 same-key starvation | Mitigated for selected schedules. Fixed-proposer base bias remains. |
| #71 finalized state regression | Formal requirements exist. The production repair remains pending. |
| #104 number-channel corruption | Formal isolation models exist. A production resolution is not established. |
| #42 rejected-buffer re-inclusion | The frontier gate does not resolve the re-inclusion window. |
| #24 merge pipeline cost | The ratified benchmark measurement remains required. |
| #206 exhaustive formal coverage | Open. Most new models are not PR-gated. |

## Stack

Stack ancestry: `#311 -> #312 -> #299 -> dev`

Merge order: PR #299, PR #312, PR #311.

This PR targets `dev` and merges first. PR #312 adds phase-2 recovery next. PR #311 delivers formal-verification scope last.


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| xai | `grok-4.5` | Feedback Provided | 0.78 |

---

### Summary

This PR adds loss-aware conflict adjudication to Casper merging: DeployChainIndex gains a non-identity prior_rejections count derived from kept on-DAG rejection records, and all three adjudication sites (keep-one, rejection-option selection, greedy claim order) rank prior losses above the content-deterministic ordering to fix the same-key starvation-to-expiry defect (#294). The implementation is careful and well-tested at the unit level, with unusually thorough documentation of the design rationale, ratified principles, and the still-open base-bias facet (the e2e spec is intentionally ignored/RED). Two concerns warrant human review before merge: (1) the shipped mechanism itself introduces a user-influenceable adjudication input whose griefing surface (manufactured losses, especially if rejected deploys are uncharged) is analyzed in the docs only for the unshipped C1 option, and (2) the count-derivation path silently tolerates missing blocks (walk break, empty-record default), which undermines the 'identical counts on every validator' claim and risks InvalidRejectedDeploy divergence between nodes with differing block availability. Smaller items include unstamped late-set chains, double block fetches on the hot path, and the changed economic semantics of 'optimal' rejection.

The loss-aware ordering is implemented consistently at the individual adjudication sites, but the production derivation and propagation of its consensus-sensitive input need further review. In particular, silently treating unavailable blocks as empty history can make results depend on local storage state, and only the actual-chain collection is stamped despite late chains also being supplied to conflict resolution. The allocation pattern and helper API can also be tightened, and an active end-to-end proposer/validator test is needed before relying on this as a consensus-safe change.

Phase-1 loss-aware adjudication is thoughtfully designed: prior_rejections is non-identity state derived from on-DAG kept rejection records, and the three merge sites generally prefer fewer losses when choosing what to reject (or higher losses when choosing what to keep). Unit tests cover keep-one preference, equal-loss content-order fallback, sustained contention, and scope/lineage count provenance; the ignored e2e test correctly documents the remaining main-parent base-bias facet. Before merge, confirm late_seq chains are stamped the same way as actual_set (or document why late paths cannot adjudicate on losses), tighten equal-loss keep-one ordering to explicit DeployChainIndex cmp, and add direct get_optimal_rejection tests for non-zero losses so all three sites stay regression-proof.

The code changes look generally good with minor issues related to function naming, documentation, and potential confusion in function signatures. These issues are not critical but could improve the maintainability and clarity of the code.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Only actual_set_vec is stamped; late-set chains always carry prior_rejections = 0</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 944)

`merge` stamps `actual_set_vec`, but the adjacent `late_set_vec` is left with the constructor default of zero. Both actual and late sequences are passed into `resolve_conflicts`, whose rejection options and ordering use `prior_rejections`. If a late chain participates in a conflict branch or dependency ordering, its recorded history is ignored and the result can differ solely because the chain was classified as late. This also makes the generic `prior_losses` callback appear fully wired while some production inputs retain zero.

---

`stamp_prior_rejections(&mut actual_set_vec, prior_rejection_counts)` stamps the actual set only. The late set (`late_seq`) passed into `resolve_conflicts` retains the default `prior_rejections: 0`. If late chains ever participate in rejection-option scoring or the greedy claim order alongside actual chains, they compete with an artificially zero loss count, which subtly biases adjudication against them. If late chains are always auto-rejected regardless, this is harmless but should be stated.

---

stamp_prior_rejections is applied to actual_set_vec but not to late_set / late_seq chains. resolve_conflicts receives late_seq and ranks via |chain| chain.prior_rejections. If late chains can enter keep-one, optimal-rejection, or unavailable-consume adjudication, they will always appear as zero-loss contenders even when on-DAG kept rejection records exist for their sigs. That reintroduces content-order starvation for the late path and can diverge from the intended loss-aware policy.

**Suggestion:**
```
Either stamp late_seq as well, or add a comment/assertion documenting that late chains never reach a loss-aware comparison site.
```

</details>

<details>
<summary><b>:orange_circle: Major: Prior-rejection priority is user-influenceable and may be farmable</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 553)

Loss-aware adjudication elevates deploys based on prior_rejection_counts, which are ultimately derived from user-submitted deploy history. An adversary who can cheaply accumulate kept rejection records (e.g., by deliberately racing their own conflicting deploys as siblings so one side loses each merge) gains adjudication priority over honest contenders on the same key — inverting the starvation this PR fixes onto honest users. Critically, if rejected deploys are not charged (they never execute at the merged state), manufacturing losses may be free. The CONSENSUS_PHILOSOPHY.md doc acknowledges a griefing vector for option C1 (fork-choice/main-parent influence) but does not analyze the griefing surface of the shipped phase-1 mechanism itself. P4 ('fork choice stays deploy-content-blind') is preserved, but P1 introduces deploy-history influence over merge outcomes, which deserves explicit adversarial analysis.

**Suggestion:**
```
Document and bound the adversarial surface of phase 1 explicitly: confirm whether rejected deploys incur cost, consider capping the effective priority delta (e.g., saturating count or requiring records from distinct proposers), and add an adversarial test where an attacker manufactures losses against a victim key.
```

</details>

<details>
<summary><b>:orange_circle: Major: Silent tolerance of missing blocks can produce divergent counts across validators</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1688)

The base-lineage ancestry walk breaks silently when `block_store.get(&cursor)?` returns `None`, and `records_of` maps a missing block to an empty record list (`.unwrap_or_default()`). If two nodes have different block availability (e.g., pruning, partially synced store, or a store bug), they will derive different visible-block sets and therefore different prior_rejection_counts — which feed adjudication and change the rejection set. Since validators reject blocks whose rejection set differs (`InvalidRejectedDeploy`), a silently truncated walk on one node is a consensus-split / block-rejection hazard rather than a graceful degradation. The whole design claim ('every validator derives identical counts from on-DAG data') rests on every node resolving the identical block set, so missing blocks should be a hard error, not an empty default.

**Suggestion:**
```
Return an error (or at minimum log at ERROR with the missing hash) when a block inside the settled window or the visible set cannot be fetched, instead of breaking the walk or defaulting to empty records.
```

</details>

<details>
<summary><b>:orange_circle: Major: Missing blocks are silently treated as having no rejection history</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1698)

The priority calculation is consensus-sensitive, but both ancestry traversal and record loading tolerate missing block bodies. The ancestry walk breaks when `block_store.get(&cursor)?` returns `None`, while `records_of` converts a missing visible block to an empty rejection list with `unwrap_or_default()`. Nodes with different local block availability or pruning state can therefore derive different prior-rejection counts from the same declared parents and compute different rejection sets. This contradicts the claim that validators deterministically derive identical counts from on-DAG data.

**Suggestion:**
```
Return a validation/storage error whenever a block in the required visible set or ancestry window is unavailable. Do not interpret missing consensus data as an empty rejection list. Add a test showing that an unavailable visible or lineage block fails computation rather than changing adjudication.
```

</details>

<details>
<summary><b>:orange_circle: Major: scope_prior_rejection_counts materializes all records before counting</b> (Reported by: anthropic, openai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 577)

`scope_prior_rejection_counts` extends a `Vec<RejectedDeploy>` with every record across all visible blocks and then iterates it once in `prior_rejection_counts`. Records can be counted incrementally per block, avoiding cloning/accumulating the full record set (each record carries a 32-byte carrier and a sig Bytes).

---

`scope_prior_rejection_counts` gathers all rejection records from all visible blocks into one `Vec` and only then counts them. Merge computation can cover a full lifespan window, so this creates memory proportional to all rejection records and delays counting. The ancestry walk also loads blocks once to discover parents and then loads them again through `records_of`.

**Suggestion:**
```
Accumulate counts while each block's records are loaded, avoiding `all_records`. Where practical, reuse blocks loaded during ancestry traversal or provide an iterator/store operation that exposes both parent metadata and rejection records.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Potential confusion in function signature</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 335)

The function `get_optimal_rejection` now takes an additional parameter `losses_f`, which might cause confusion for developers not familiar with the recent changes. Consider renaming the function or adding a clear comment to explain the new parameter.

**Suggestion:**
```
Rename the function to `get_optimal_rejection_with_losses` or add a comment explaining the purpose of the `losses_f` parameter.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: No direct unit coverage for loss-first optimal rejection</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 833)

get_optimal_rejection now ranks total prior losses before cost, and the PR claims all three adjudication sites are loss-aware. Existing conflict_set_merger tests only pass |_branch| 0u64 for losses_f, so they never exercise the new zeroth criterion. Regressions in option-selection polarity (min losses of the rejected set vs max) would not be caught here; only keep-one/unavailable paths are covered in dag_merger tests.

**Suggestion:**
```
Add get_optimal_rejection cases: (1) higher-loss branch is kept when cost would reject it; (2) equal losses fall back to cost/signature order; (3) multi-branch option sums losses correctly.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Loss criterion can now select an economically worse rejection option</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 836)

The zeroth sort criterion in `get_optimal_rejection` (fewest prior losses rejected) strictly outranks total cost. A merge may therefore reject a strictly higher-cost branch set to protect a previously-lost chain, reducing the economic value retained by the merge. This is the intended trade-off, but it changes the economic semantics of 'optimal rejection' and could interact with fee/reward accounting expectations elsewhere.

**Suggestion:**
```
Confirm no downstream invariant assumes the rejection set is cost-minimal, and consider renaming or re-documenting `get_optimal_rejection` to reflect that optimality is now (losses, cost)-lexicographic.
```

</details>

<details>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 530)

The function `prior_rejection_counts` uses snake_case, while other functions in the same module use camelCase. Consider renaming the function to follow the consistent naming convention.

**Suggestion:**
```
Rename the function to `priorRejectionCounts` to match the camelCase convention used in the module.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Counting helper double-counts duplicate block identifiers</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 570)

`scope_prior_rejection_counts` documents a visible block set but accepts any iterator and does not deduplicate block hashes. The production caller currently constructs a `HashSet`, but other callers can pass the same block more than once and inflate consensus-sensitive priority counts; the unit test itself uses a `Vec`, demonstrating that uniqueness is not enforced by the API.

**Suggestion:**
```
Accept a set type or deduplicate block hashes inside the helper. This makes the consensus invariant local to the function instead of relying on every caller.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Extra main-chain walk plus full rejection-record load per merge</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1671)

prior_rejection_counts builds a visible set from scope ∪ base_lineage, then walks main-parent ancestry to settled_walk_bound while block_store.get-ing every block and cloning all rejected_deploys. This is correct for the retry-gate provenance note but adds another linear ancestry walk and record materialization on the merge hot path, on top of existing settled-sig / lineage work.

**Suggestion:**
```
If profiles show merge latency growth, reuse any ancestry/rejection data already gathered for sig_settled_in_base / base_lineage_blocks, or stream-count per block without retaining full Vec<RejectedDeploy>.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Blocks on the base lineage are fetched twice per merge</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1681)

The ancestry walk calls `block_store.get(&cursor)` for each block to read its number and parent, then `scope_prior_rejection_counts` fetches every one of those blocks again via `records_of` just to read `body.rejected_deploys`. For a deep settled window this doubles block-store reads (full block bodies) on every `compute_parents_post_state` invocation, which is on the hot proposal/validation path.

**Suggestion:**
```
Accumulate rejected_deploys records during the ancestry walk itself (the block is already in hand), and only fetch the remaining scope/base-lineage blocks in the counting pass. Consider caching counts keyed by (anchor, window edge) since the same scope is recomputed by every validator.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Production consensus wiring has no active end-to-end test</b> (Reported by: openai)</summary>

**File:** `casper/tests/batch2/loss_priority_spec.rs` (line 93)

The only proposer/validator test that exercises production count assembly and rejection-set validation is ignored, while the active tests directly stamp chains or test record aggregation in isolation. Consequently, regressions in visible-block selection, lineage traversal, merge parameter wiring, or validator reproduction will not fail CI. This is especially risky for a consensus-affecting change.

**Suggestion:**
```
Add a non-ignored end-to-end test for a neutral-base matchup where loss-aware adjudication is expected to apply. The known main-parent-base-bias scenario can remain ignored separately as a phase-two sentinel.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Missing documentation for new principles</b> (Reported by: bedrock)</summary>

**File:** `docs/casper/CONSENSUS_PHILOSOPHY.md` (line 150)

The new principles P1-P6 are introduced but not documented in the `docs/Glossary.md` file. Consider adding these principles to the glossary for better reference.

**Suggestion:**
```
Add the new principles P1-P6 to the `docs/Glossary.md` file with brief descriptions.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Bare `&|_| 0` prior-loss hook in the wrapper merge path</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 774)

The convenience `merge` wrapper hardcodes `&|_| 0` for prior losses, meaning any caller of this path silently gets pure content-order adjudication. The comment explains why, but a named constant/closure (e.g., `no_prior_losses`) would make future call sites less likely to copy the zero hook where real counts exist.

**Suggestion:**
```
Introduce a named helper like `fn no_prior_losses<R>(_: &R) -> u64 { 0 }` and use it at this and test call sites.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Keep-one tie-break still depends on collection order</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 512)

sort_by_key((!pinned, Reverse(prior_rejections))) is stable, so equal-loss ties preserve the pre-sort order of prod.iter(). If prod is a HashSet-backed collection, that order is not content order. The new test overfilled_splitter_equal_prior_losses_keep_content_order may pass only when insertion/iteration order happens to match DeployChainIndex ordering. Pre-existing risk, but loss-awareness makes the tie path more important to pin down.

**Suggestion:**
```
After the priority key, explicitly .then_with(|| a.cmp(b)) on DeployChainIndex (as dependency_ordered_branch_items already does) instead of relying on stable sort + iteration order.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Clear helpers and glossary alignment</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 550)

prior_rejection_counts / scope_prior_rejection_counts / stamp_prior_rejections are well-named, duplicate records are excluded consistently, and prior_rejections is correctly omitted from DeployChainIndex identity. Glossary and CONSENSUS_PHILOSOPHY changes match the implementation principles (P1/P2).

**Suggestion:**
```
None required for merge; keep phase-2 ladder status as Draft until maintainer ratification.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: E2E racing-shape test intentionally ignored</b> (Reported by: xai)</summary>

**File:** `casper/tests/batch2/loss_priority_spec.rs` (line 99)

repeatedly_rejected_deploy_gains_priority_and_lands is #[ignore] pending the main-parent base-bias decision. That matches the TDD plan and CONSENSUS_PHILOSOPHY ladder, but phase-1 merge alone does not close the issue’s full observed shape. Fine to ship if reviewers accept phase-1 scope; not a code defect.

**Suggestion:**
```
Call out in the PR merge checklist that B4 remains blocked and B6/B7 are not in this change, so #294 should stay open or be tracked as partial.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Committed permanently-RED #[ignore]d end-to-end test</b> (Reported by: anthropic)</summary>

**File:** `casper/tests/batch2/loss_priority_spec.rs` (line 130)

The e2e spec is ignored 'RED by design' pending the phase-2 base-bias decision. This is well-documented (ignore message points to the TDD plan), but ignored tests tend to rot silently — the harness APIs it uses (create_block_unsafe, process_block, play_exploratory_deploy) can drift without CI noticing.

**Suggestion:**
```
Consider a tracking issue reference in the #[ignore] string and/or a compile-only guard so the spec at least keeps building, plus a target date for the phase-2 decision.
```

</details>
---

<details>

This PR adds loss-aware conflict adjudication to Casper merging: DeployChainIndex gains a non-identity prior_rejections count derived from kept on-DAG rejection records, and all three adjudication sites (keep-one, rejection-option selection, greedy claim order) rank prior losses above the content-deterministic ordering to fix the same-key starvation-to-expiry defect (#294). The implementation is careful and well-tested at the unit level, with unusually thorough documentation of the design rationale, ratified principles, and the still-open base-bias facet (the e2e spec is intentionally ignored/RED). Two concerns warrant human review before merge: (1) the shipped mechanism itself introduces a user-influenceable adjudication input whose griefing surface (manufactured losses, especially if rejected deploys are uncharged) is analyzed in the docs only for the unshipped C1 option, and (2) the count-derivation path silently tolerates missing blocks (walk break, empty-record default), which undermines the 'identical counts on every validator' claim and risks InvalidRejectedDeploy divergence between nodes with differing block availability. Smaller items include unstamped late-set chains, double block fetches on the hot path, and the changed economic semantics of 'optimal' rejection.

The code changes look generally good with minor issues related to function naming, documentation, and potential confusion in function signatures. These issues are not critical but could improve the maintainability and clarity of the code.

#### openai (gpt-5.6-sol)

The loss-aware ordering is implemented consistently at the individual adjudication sites, but the production derivation and propagation of its consensus-sensitive input need further review. In particular, silently treating unavailable blocks as empty history can make results depend on local storage state, and only the actual-chain collection is stamped despite late chains also being supplied to conflict resolution. The allocation pattern and helper API can also be tightened, and an active end-to-end proposer/validator test is needed before relying on this as a consensus-safe change.

#### xai (grok-4.5)

Phase-1 loss-aware adjudication is thoughtfully designed: prior_rejections is non-identity state derived from on-DAG kept rejection records, and the three merge sites generally prefer fewer losses when choosing what to reject (or higher losses when choosing what to keep). Unit tests cover keep-one preference, equal-loss content-order fallback, sustained contention, and scope/lineage count provenance; the ignored e2e test correctly documents the remaining main-parent base-bias facet. Before merge, confirm late_seq chains are stamped the same way as actual_set (or document why late paths cannot adjudicate on losses), tighten equal-loss keep-one ordering to explicit DeployChainIndex cmp, and add direct get_optimal_rejection tests for non-zero losses so all three sites stay regression-proof.
</details>

---

spreston8 (commented):
Reviewed at `1dcda5b3f` against `dev`. Suggestions below. Items 1–3 read as pre-merge; 4–6 are improvements. The first two are about what feeds `prior_rejections` rather than the mechanism consuming it — in one case the value is partly node-local, in the other partly self-selected.

**High**

1. Worth making the prior-rejection walk refuse an incomplete lineage rather than truncate. `interpreter_util.rs` breaks the walk on a missing block and defaults `rejected_deploys` to empty, so a node missing bodies derives lower counts, picks a different keep-one winner, and lands on `InvalidRejectedDeploy` — the failure mode P2 names, and one `block_status.rs:273` treats as non-slashable precisely because it can reflect local state. `effect_in_state_of_above` walks the same lineage to the same bound and errors instead ("refusing to judge membership from an incomplete lineage"), so matching it costs nothing that is not already being paid. Worth weighing that a restored node is permanently truncated and the clique oracle needs metadata only — this would be the first consensus read to require bodies. The trade is worth naming: refusing turns a silent divergence into a deferral, so validation gains a way to fail-to-decide on availability that it does not have today.

2. Consider composing losses with cost rather than ranking strictly above it. A rejected chain's diff never reaches `compute_merged_state`, and the buffer re-admits the deploy, so a loss costs the deployer no phlo — the counter is free to increment. Under lexicographic ordering any positive count beats any cost gap, and a freshly-submitted rival always starts at zero, so contending cheaply on a hot key and losing is a way to outrank it. Bounded starvation needs only that priority eventually rises far enough, and P5 prefers the weakest sufficient remedy; a saturating term gets the same liveness property without that. Worth settling before merge rather than after: adjudication order is a consensus rule, so moving from dominating to bounded later is a lockstep upgrade, while changing it now is a comparator edit. Worth rating Phase 1's adversarial surface in the comparison table too — it currently covers all six alternatives but not what shipped.

3. Worth gating merge on ucc being green and repeatably so. It is the only suite that exercises same-key contention at shard level, which is what this changes, and its last failure was a concurrent-proposal race — so one green run would not distinguish "fixed" from "did not reproduce this time". Three consecutive passes is the bar the stall campaign used. Two things stand in the way, neither belonging to this PR: `ucc_tests` still carries `if: false` (with the launcher count and the `ci.yml` slot gate as its stated unit), and its last real run surfaced a `rank_forkchoices` divergence that is fixed only in unmerged work (the thing I mentioned in standup today). Worth noting separately that green alone would not demonstrate the fix — the suite logs contention losses as STARVATION-RECORD rather than failing, so it can pass with the starvation intact. Making it an acceptance criterion for this needs that assertion changed on the system-integration side.

**Medium**

4. Worth adding a neutral-base variant of `loss_priority_spec`: three validators, proposer's main parent holding neither contender. Both chains are live in that shape, so the mechanism is decisive and facet 1 gets end-to-end evidence without waiting on the base-bias decision or on ucc. The existing spec should stay ignored — its shape needs facet 2, and P6 makes that rejection mandatory rather than orderable.

5. Worth pinning the determinism the counts rest on. `scope_prior_rejection_counts` is node-identical only if the visible-block set and the base-lineage walk are identical on every validator of the same block; the counting itself is order-free, so that assumption carries the whole property and nothing currently tests it. A test asserting two derivations over the same block agree — or a stated invariant at the call site — would make a later change to scope construction fail loudly rather than quietly skew adjudication.

**Low**

6. The panel's Critical does not reach the production path — `late_blocks` is empty whenever a scope is provided, and `compute_parents_post_state` always passes one. Worth a comment at the `stamp_prior_rejections` call site so that stays true.

jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.86 |

---

### Summary

This is a carefully constructed and unusually well-documented consensus-policy change: prior on-DAG rejection counts now strictly outrank content ordering at all three merge adjudication sites, with the count kept out of chain identity, streamed record counting, BlockNotHeld refusal on missing history, and equal stamping of actual and late chains. Test coverage is strong (direct nonzero-loss unit tests, adversarial fixed-lead bounds, missing-history refusal, and a three-validator neutral-base integration test), and the accepted risks (loss farming, non-termination under fixed-proposer base bias) are explicitly documented. The main substantive concerns are: (1) the lineage walk in interpreter_util.rs may not distinguish 'reached genesis' from 'missing DAG metadata', which could silently violate the count-determinism invariant the PR itself ratifies; (2) SUM aggregation at the rejection-option level is inconsistent with the ratified MAX chain-level rule and creates a coalition-farming lever; and (3) pinned chains strictly dominate loss priority in keep-one, preserving a starvation channel that should be explicitly tied to the phase-2 remedy ladder. Per-merge full-body loads across the visible window remain the dominant performance cost pending the required benchmark. The code is functional and safe to merge with these clarifications addressed or documented.

The loss-aware merge changes appear internally coherent: rejection history is derived from authenticated held blocks, missing bodies fail closed, both actual and late chains are stamped, and chain priority uses the documented maximum member count. No clear security or consensus-correctness defect is evident from the diff. The main actionable concerns are repeated metric calculation in a potentially large rejection-option sort and documentation that overstates the liveness guarantee or conflicts with the stated phase-2 status.

This is a careful phase-1 consensus change: prior kept rejection history is stamped onto DeployChainIndex without affecting identity, missing visible history correctly becomes BlockNotHeld, and loss priority is applied at keep-one, claim-order, and optimal-rejection sites with solid unit and neutral-base integration coverage. The main residual design risk is get_optimal_rejection’s sum aggregation over multi-branch options, which can still prefer rejecting one high-loss chain over several low-loss ones; that should be explicitly ratified or tightened before relying on phase 1 for complex conflict graphs. Remaining notes are contract clarity (sig vs deploy_id), scan/benchmark cost, and CI timeout retry hygiene—none look like merge-blocking security defects given the documented accepted farming/base-bias limits.

The PR introduces significant changes to the merge adjudication process by incorporating prior on-DAG rejections. The changes are well-documented and tested, but additional test cases and minor improvements to comments and glossary terms are recommended to ensure clarity and correctness.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential logic error in `get_optimal_rejection`</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 335)

The function `get_optimal_rejection` now considers both cost and prior losses. Ensure that the logic correctly handles cases where prior losses are equal but costs differ, and vice versa.

**Suggestion:**
```
Add additional test cases to cover scenarios where prior losses and costs are varied to ensure the function behaves as expected.
```

</details>

<details>
<summary><b>:orange_circle: Major: Rejection-option loss aggregation uses sum, which can still drop a high-loss chain</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 833)

get_optimal_rejection minimizes the sum of prior losses over the entire rejected branch set. That is correct for simple 1-vs-1 keep-one cases and is covered by tests, but in multi-way rejection options a single high-loss chain (e.g. losses=5) can still be rejected when the alternative rejects several low-loss chains whose summed losses are smaller (e.g. 2+2=4). Phase-1 policy ratifies loss-before-cost and max-within-chain, but does not clearly specify cross-branch aggregation. This can reintroduce bounded starvation in multi-conflict graphs even though pairwise keep-one is loss-aware.

**Suggestion:**
```
Document sum-as-total-rejected-unfairness as the ratified rule, or change option comparison to a loss-aware lexicographic policy (e.g. compare multiset of per-chain losses, or compare max then sum) so a strictly higher-loss chain is never rejected when a lower-loss alternative exists. Add a multi-way unit test that locks the chosen rule.
```

</details>

<details>
<summary><b>:orange_circle: Major: Lineage walk may terminate silently on missing DAG metadata instead of returning BlockNotHeld</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1698)

The main-parent ancestry walk uses `cursor = s.dag.main_parent(&hash)` and stops when it returns `None`. If `main_parent` returns `None` both for genesis AND for a block whose metadata this node does not hold, a node with a truncated DAG view would silently derive counts from a shorter lineage window rather than refusing with `BlockNotHeld`. This would violate the stated invariant that 'two nodes with different block availability must not derive different counts from the same parents' — the exact determinism property the PR claims to enforce. Body availability is checked in `records_of`, but metadata availability during the walk is not distinguished from reaching genesis.

**Suggestion:**
```
Distinguish the genesis/window-edge terminal case from missing metadata. If `main_parent` can return `None` for a non-genesis block, return `CasperError::BlockNotHeld` when the walk terminates before reaching the `settled_walk_bound` for a reason other than genesis. Add a test covering truncated DAG metadata (not just missing bodies).
```

</details>

<details>
<summary><b>:orange_circle: Major: Retry gate matches on error-message string prefix</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/tests/genesis/contracts/pos_spec.rs` (line 110)

The pos_spec retry triggers on `message.starts_with("Timeout of ") && message.contains("phase 'eval-test-source'")`. String-matching error text is brittle — any rewording of the interpreter timeout message silently disables the retry, and the retry itself can mask a genuine performance regression rather than parallel-load flakiness.

---

pos_spec now retries once on InterpreterError::BugFoundError timeout during phase eval-test-source. That stabilizes pre-push (with the 900s hook default) but can mask genuine slowdowns or intermittent interpreter issues as a quiet retry.

**Suggestion:**
```
Log/metric the retry path in CI, or bound retries behind an explicit env flag so local pre-push stays resilient while CI still surfaces repeated timeouts.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Sort comparator repeatedly recalculates rejection metrics</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 833)

Every comparison performed by `sort_by` traverses both rejection options to recalculate prior-loss totals and, when those are equal, cost totals. Since rejection-option generation can already produce many options, this changes sorting to approximately O(O log O * S), where O is the option count and S is the average number of branch items. It also invokes the supplied closures an unspecified number of times. This is particularly relevant because the PR designates rejection-history performance as a merge gate.

**Suggestion:**
```
Precompute each option's loss total, cost total, and deterministic ordering key once, store those alongside the option, and sort the annotated values. Use checked or explicitly defined saturating arithmetic while calculating totals.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Loss aggregation semantics differ between adjudication sites: SUM at rejection-option level vs MAX at chain level</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 840)

The ratified policy states 'dependency-chain priority uses the maximum member count' (implemented via `max` in `stamp_prior_rejections`), but `get_optimal_rejection` sums losses across all branches in a rejection option. An option rejecting one chain with 3 prior losses (sum=3) is preferred for rejection over an option rejecting two chains with 2 losses each (sum=4), so the single high-loss veteran loses to a coalition of moderately-lossy chains. This asymmetry is a potential farming lever: many cheap low-loss chains can collectively outweigh a legitimately starved deploy. The behavior is exercised by `optimal_rejection_sums_losses_across_branches`, so it appears intentional, but the SUM-vs-MAX inconsistency is not documented in the ratified policy table.

**Suggestion:**
```
Either document the option-level SUM semantics explicitly in CONSENSUS_PHILOSOPHY.md and the glossary, or consider using max-per-option to keep aggregation semantics uniform with the chain-level rule.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improve comment clarity</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 483)

The comment explaining the sorting logic could be more concise and clear.

**Suggestion:**
```
Rephrase the comment to clearly state the purpose of the sorting logic and how it affects the merge process.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Pinned status strictly outranks prior_rejections in keep-one, reintroducing a starvation channel</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 505)

In `split_overfilled_single_value_cells`, the sort key is `(!pinned, Reverse(prior_rejections))`, so a pinned chain with zero losses always beats an unpinned chain with arbitrarily many recorded losses. If a proposer's main parent consistently commits the contender's effect (the exact main-parent base-bias shape from issue #294 phase 2), the starved chain gains losses forever but can never outrank the pinned rival. This is presumably intentional (pinning is a correctness constraint, and the base-bias facet is deferred to PR #312), but the interaction is not called out in the comment or the philosophy document's adversarial-surface section.

**Suggestion:**
```
Document that pinning dominates loss priority and that this preserves the phase-2 starvation facet, so the residual exposure is traceable to the remedy ladder.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Counts are keyed by rejection sig; stamping looks up deploy_id</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 547)

scope/prior counters key HashMap entries by RejectedDeploy.sig, while stamp_prior_rejections reads counts.get(&d.deploy_id). This is only correct if deploy_id is always the deploy signature bytes. That is likely true in this codebase, but it is an implicit cross-type contract: a future change that makes deploy_id a hash or other identifier would silently zero every prior_rejections stamp and disable the mitigation without failing compilation.

**Suggestion:**
```
Assert or type-alias the equality (e.g. debug_assert key conventions, or map through a single DeploySig newtype) and add one test that uses production deploy-id construction rather than hand-built Bytes keys.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Lineage walk inserts after main_parent fetch; bound membership is easy to misread</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1671)

The ancestry loop fetches main_parent before inserting the current hash, and breaks before insert when number < settled_walk_bound. Behavior looks deterministic and consistent with the comments, but the control flow is subtle: readers may miss that the bound block itself is excluded and that scope_anchor is only added when above the bound. Any future edit that inserts before the bound check, or walks parents differently, would silently change consensus-visible counts.

**Suggestion:**
```
Extract a small helper (e.g. main_parent_window(anchor, bound)) with an explicit unit test for: anchor below bound, anchor equal bound, and long chain with duplicates already present in visible/base sets.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Full block bodies loaded per merge across the entire visible window with no caching</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1690)

`scope_prior_rejection_counts` loads every visible block's full body via `block_store.get(hash)` just to read `body.rejected_deploys`, and this runs on every call to `compute_parents_post_state` that performs a merge. With the documented 512-block visible scope plus a 256-block lineage window, this is hundreds of body deserializations per merge, repeated for every proposal and validation with no cross-merge memoization. The PR notes the benchmark remains required, but the O(B) body loads per merge (not per block lifetime) is the dominant new cost.

**Suggestion:**
```
Consider a per-block memoized map of sig→kept-record counts (block hash → counts), or a dedicated rejected-deploys index in the block store, so repeated merges over overlapping windows amortize body loads.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Update glossary terms</b> (Reported by: bedrock)</summary>

**File:** `docs/Glossary.md` (line 269)

The glossary has been updated with new terms related to the PR. Ensure all terms are correctly defined and referenced.

**Suggestion:**
```
Review the glossary to ensure all terms are consistent with the code changes and provide clear definitions.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Glossary overstates the starvation guarantee</b> (Reported by: openai)</summary>

**File:** `docs/Glossary.md` (line 300)

The new definition says that every loss raises priority "so starvation stays bounded." That is broader than the implementation and contradicts both the PR description and the updated Casper glossary, which limit the benefit to an adjudicable matchup and explicitly retain fixed-main-parent base-bias starvation as an unresolved case.

**Suggestion:**
```
Match `docs/casper/GLOSSARY.md`, for example: "Every loss raises the loser's priority in an adjudicable matchup." Avoid claiming a general starvation bound until phase 2 provides that guarantee.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Merged-frontier packaging status is inconsistent</b> (Reported by: openai)</summary>

**File:** `docs/Glossary.md` (line 370)

This glossary labels merged-frontier retry packaging as "proposed," while the PR summary and `docs/casper/CONSENSUS_PHILOSOPHY.md` state that PR #312 implemented it, with only its liveness guarantee pending ratification. These conflicting statuses make it unclear whether the mechanism exists or is merely planned.

**Suggestion:**
```
Use a consistent status such as "implemented by PR #312; liveness guarantee pending ratification," if that reflects the current branch.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Public merge wrapper hard-codes zero prior losses</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/merging/conflict_set_merger.rs` (line 774)

The context-free merge wrapper passes |_| 0 for prior_losses. The comment says there is no block context, which is fine if this path is test/tooling only, but it is an easy footgun if a production caller later reuses merge() expecting loss-aware behavior.

---

`conflict_set_merger::merge` passes `&|_| 0` for prior losses, so any caller of this wrapper gets the pre-#294 pure content-order behavior. The comment explains why, but if this wrapper is ever reachable from a consensus-relevant path alongside `dag_merger::merge`, the two paths would adjudicate the same conflict set differently.

**Suggestion:**
```
Rename or doc-comment the wrapper as loss-agnostic/legacy, or thread an explicit PriorLossPolicy enum so call sites cannot forget stamping.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: prior_rejection_counts clones each record</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 556)

`prior_rejection_counts` takes `&'a RejectedDeploy` and calls `.cloned()` before passing to `count_kept_records`, which only needs `record.duplicate` and `record.sig`. The clone is cheap for `Bytes` (refcount) but unnecessary.

**Suggestion:**
```
Have `count_kept_records` accept `&RejectedDeploy` (cloning only the sig on insert) or restructure to avoid the whole-record clone.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Full block body load per visible hash for rejection records only</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1695)

scope_prior_rejection_counts loads each block body via block_store.get just to read body.rejected_deploys. Dedup prevents repeat loads and the philosophy doc gates a 256/512-block benchmark, so this is acceptable if the benchmark remains required. Still, merge latency will track body fetch cost on large windows.

**Suggestion:**
```
Keep the release-gate benchmark mandatory as documented. If it regresses, consider a records-only store accessor so merge adjudication does not pull full bodies.
```

</details>
---

<details>

This is a carefully constructed and unusually well-documented consensus-policy change: prior on-DAG rejection counts now strictly outrank content ordering at all three merge adjudication sites, with the count kept out of chain identity, streamed record counting, BlockNotHeld refusal on missing history, and equal stamping of actual and late chains. Test coverage is strong (direct nonzero-loss unit tests, adversarial fixed-lead bounds, missing-history refusal, and a three-validator neutral-base integration test), and the accepted risks (loss farming, non-termination under fixed-proposer base bias) are explicitly documented. The main substantive concerns are: (1) the lineage walk in interpreter_util.rs may not distinguish 'reached genesis' from 'missing DAG metadata', which could silently violate the count-determinism invariant the PR itself ratifies; (2) SUM aggregation at the rejection-option level is inconsistent with the ratified MAX chain-level rule and creates a coalition-farming lever; and (3) pinned chains strictly dominate loss priority in keep-one, preserving a starvation channel that should be explicitly tied to the phase-2 remedy ladder. Per-merge full-body loads across the visible window remain the dominant performance cost pending the required benchmark. The code is functional and safe to merge with these clarifications addressed or documented.

The PR introduces significant changes to the merge adjudication process by incorporating prior on-DAG rejections. The changes are well-documented and tested, but additional test cases and minor improvements to comments and glossary terms are recommended to ensure clarity and correctness.

#### openai (gpt-5.6-sol)

The loss-aware merge changes appear internally coherent: rejection history is derived from authenticated held blocks, missing bodies fail closed, both actual and late chains are stamped, and chain priority uses the documented maximum member count. No clear security or consensus-correctness defect is evident from the diff. The main actionable concerns are repeated metric calculation in a potentially large rejection-option sort and documentation that overstates the liveness guarantee or conflicts with the stated phase-2 status.

#### xai (grok-4.5)

This is a careful phase-1 consensus change: prior kept rejection history is stamped onto DeployChainIndex without affecting identity, missing visible history correctly becomes BlockNotHeld, and loss priority is applied at keep-one, claim-order, and optimal-rejection sites with solid unit and neutral-base integration coverage. The main residual design risk is get_optimal_rejection’s sum aggregation over multi-branch options, which can still prefer rejecting one high-loss chain over several low-loss ones; that should be explicitly ratified or tightened before relying on phase 1 for complex conflict graphs. Remaining notes are contract clarity (sig vs deploy_id), scan/benchmark cost, and CI timeout retry hygiene—none look like merge-blocking security defects given the documented accepted farming/base-bias limits.
</details>

---

jeffrey-l-turner:
## Multi-Agent Review: resolution of the four major findings

The fixes are in the working tree and land with the next push to this branch.

| # | Finding | Reporter | Resolution |
|---|---------|----------|------------|
| 1 | Lineage walk may stop silently on missing DAG metadata (`interpreter_util.rs:1698`) | anthropic | **Fixed.** Verified that `block_number_map` and `main_parent_map` are written together from one metadata record, so `main_parent` returns `None` only for a held block with no parent (genesis). The walk did not shorten silently, but a metadata gap surfaced as a generic `KvStoreError`. The walk now uses `block_number(..)` and returns `CasperError::BlockNotHeld` for the gap, with a comment that states the genesis-vs-gap distinction. |
| 2 | `get_optimal_rejection` aggregates prior losses by SUM while the ratified chain rule is MAX (`conflict_set_merger.rs:833`, also anthropic minor at `:840`) | xai, anthropic | **Fixed.** New `LossProfile { max, sum }`. Branch losses fold max-then-sum over members. Option selection ranks by highest member count, then count total, then cost, then the existing size and order tie-breaks. A coalition of low-loss chains can no longer outweigh one chain with a higher loss. Numeric keys are computed once per option (closes the openai minor note on repeated metric calculation). Tests: `optimal_rejection_keeps_highest_loss_chain_over_low_loss_coalition` replaces `optimal_rejection_sums_losses_across_branches` with the flipped expectation, plus `optimal_rejection_equal_max_losses_fall_back_to_loss_sum` and `branch_losses_takes_max_and_sum_over_members`. Decision-record row added to `CONSENSUS_PHILOSOPHY.md`, marked **pending ratification**. |
| 3 | Add tests for equal-loss/different-cost and equal-cost/different-loss (`conflict_set_merger.rs:335`) | bedrock | **No change needed.** `optimal_rejection_keeps_higher_loss_branch_despite_cost` and `optimal_rejection_equal_nonzero_losses_fall_back_to_cost_and_order` already cover both cases. |
| 4 | Retry gate matches the timeout error by message prefix (`pos_spec.rs:110`) | anthropic, xai | **Fixed.** `rho_spec.rs` now owns `timeout_error(..)` and a typed `timeout_phase(&InterpreterError) -> Option<&str>` with `EVAL_TEST_SOURCE_PHASE`. Encoder and decoder live together with a round-trip test, so a message change cannot disable the retry silently. `pos_spec` uses the classifier, and `RHO_SPEC_NO_RETRY=1` disables the retry for runs that must surface a slowdown. |

Verification: `cargo build -p casper --tests`, `cargo clippy -p casper --tests` (0 warnings), `cargo fmt --check` clean. `conflict_set_merger` 18/18, `dag_merger` 25/25, `timeout_phase_tests` 2/2, and `three_validator_neutral_base_applies_prior_loss_priority` pass.

## CI failure on `edc92470a`

`Heavy Pipeline / Integration Tests (amd64-subprocess)` failed on one test: `test_web_api.py::test_get_blocks@shared`, a deploy that stayed `Pending` with `rejection_count=18` (three validators, six consecutive losing merges). The other three integration variants passed.

This failure is inherited from `dev`, not introduced here:

- `dev` run [32544160782](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/32544160782) (`8b895792f`): `test_get_blocks@shared`, `rejection_count=24`
- `dev` run [32540926176](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/32540926176) (`39edf1855`): `test_last_finalized_block@shared`, `rejection_count=21`

The last green CI on this branch was `6a80ce6e3`, before the `dev` merge. The production diff since then (`edc92470a`) is behavior-neutral for this path: `all_chains` is pre-sorted so the `then_with` tie-break matches the old stable sort, and sum-to-max is identical for single-deploy chains.

The symptom is the fixed-main-parent base-bias facet that #294 phase 2 (#317) defers: a deploy loses every merge to the same base and loss priority cannot rescue it. This PR's phase-1 scope does not claim to close that facet.

PR #301, opened 2026-08-20
The `global_lock` write-side cost and the `/api/status` instrumentation blind spot

9Neechan:
Fixes two of the three suggested next steps from #203 — the `global_lock` write-side cost and the `/api/status` instrumentation blind spot. Does **not** address the dominant cause identified in #203 (the `explore-deploy` DAG-merge recompute's CPU cost) — It was closed by PR #211.

**1. Skip the `propagate_ft_to_finalized_blocks()` scan when it can't change anything**

`propagate_ft_to_finalized_blocks()` runs after every finalization and does an unbounded scan over _all_ finalized blocks in the chain's history while holding `global_lock` exclusively — a cost that grows with chain length and queues out every reader (external pollers and the node's own block-processing/finalization pipeline alike), per #203's root-cause section.

Added `ft_lower_bound: Arc<AtomicU32>` (float bits) tracking the lowest FT value seen among directly/indirectly finalized blocks. If the incoming `ft_value` is already `<=` that bound, the scan can't raise anything, so it's skipped before the exclusive lock is even taken (`propagate_ft.skipped` counter). Otherwise the scan runs as before, now timed (`propagate_ft.scan.time`, `propagate_ft.scan.blocks`) so the actual cost is visible instead of inferred.

The bound is lowered whenever a block finalizes with a `ft_value` below it, and raised to the propagated value once a scan completes — so it always tracks the true minimum without itself requiring a scan.

**2. Fix the `/api/status` slow-path instrumentation blind spot**

#203 found that `STATUS_SLOW_THRESHOLD`'s timer stopped _before_ the `BlockAPI::last_finalized_block()` call — the one step in `status()` that touches `global_lock` — so the "Web API status assembly is slow" warning could never fire for the actual slow path (confirmed: zero occurrences across all three test runs in #203, despite `status` p95 hitting 3.2s under `high` load).

Moved the `last_finalized_block()` call above the threshold check and added `lfb_elapsed` to the warning fields, so the log line now names the step it's actually blind to.

Refs #203

### DIFF
--- a/block-storage/Cargo.toml
+++ b/block-storage/Cargo.toml
@@ -43,21 +43,14 @@ proptest = { workspace = true }
 # See block-storage/tests/loom_equivocations_tracker.rs.
 loom = "=0.7.2"
 
-# P2-16: integration tests in `tests/block_dag_storage_test.rs` call the
-# gated `equivocation_records` / `insert_equivocation_record` /
-# `update_equivocation_record` helpers. They require the
-# `test-internals` feature; `cargo test --workspace --all-features`
-# already enables it, but the per-target `required-features` makes the
-# dependency explicit and prevents `cargo test` (default features only)
-# from confusingly omitting these tests.
-[[test]]
-name = "block_dag_storage_test"
-required-features = ["test-internals"]
-
-[[test]]
-name = "mod"
-required-features = ["test-internals"]
-
+# P2-16: these targets used to carry `required-features = ["test-internals"]`
+# because two equivocation tests called the gated lock-bypass helpers. A crate
+# cannot dev-depend on itself, so that feature could only be reached by passing
+# `--features` on the command line — and a target whose required features are
+# unmet is SKIPPED, not failed, so CI (`cargo nextest run -p block-storage`,
+# no `--features`) silently ran none of these tests. The two tests now go
+# through `access_equivocations_tracker`, the same locked path production uses,
+# so nothing here needs the feature and every test is visible by default.
 [[test]]
 name = "loom_equivocations_tracker"
 
--- a/block-storage/src/rust/dag/block_dag_key_value_storage.rs
+++ b/block-storage/src/rust/dag/block_dag_key_value_storage.rs
@@ -35,7 +35,7 @@
 // See block-storage/src/main/scala/coop/rchain/blockstorage/dag/BlockDagKeyValueStorage.scala
 
 use std::collections::{BTreeSet, HashMap, HashSet, VecDeque};
-use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
 use std::sync::Arc;
 
 use models::rust::block_hash::{self, BlockHash, BlockHashSerde};
@@ -719,6 +719,7 @@ pub struct BlockDagKeyValueStorage {
     /// Monotonically increasing counter incremented on every successful block insert.
     /// Used by caches to detect when the DAG has changed.
     pub(crate) dag_generation: Arc<AtomicU64>,
+    pub(crate) ft_lower_bound: Arc<AtomicU32>,
 }
 
 impl BlockDagKeyValueStorage {
@@ -779,6 +780,7 @@ impl BlockDagKeyValueStorage {
             lifecycle: Arc::new(PlRwLock::new(lifecycle_tables)),
             latest_messages_index: latest_messages_db,
             dag_generation: Arc::new(AtomicU64::new(0)),
+            ft_lower_bound: Arc::new(AtomicU32::new(0.0f32.to_bits())),
         })
     }
 
@@ -847,6 +849,7 @@ impl BlockDagKeyValueStorage {
             equivocation_tracker_index,
             lifecycle: Arc::new(PlRwLock::new(DeployLifecycleTables::in_memory())),
             dag_generation,
+            ft_lower_bound: Arc::new(AtomicU32::new(0.0f32.to_bits())),
         }
     }
 
@@ -1273,7 +1276,17 @@ impl BlockDagKeyValueStorage {
                     directly_finalized_hash.clone(),
                     indirectly_finalized,
                     ft_value,
-                )
+                )?;
+                drop(block_metadata_index_guard);
+
+                // These blocks enter at `ft_value`, possibly below what earlier
+                // rounds propagated — the bound has to follow them down.
+                let _ = self.ft_lower_bound.fetch_update(
+                    Ordering::Relaxed,
+                    Ordering::Relaxed,
+                    |bits| (ft_value < f32::from_bits(bits)).then(|| ft_value.to_bits()),
+                );
+                Ok(())
             };
 
         let mut effect_applied: HashSet<BlockHash> = HashSet::new();
@@ -1336,12 +1349,32 @@ impl BlockDagKeyValueStorage {
         // P2-12: mutates `block_metadata_index`; exclusive lock.
         let _lock_guard = self.global_lock.write();
 
+        // Nothing is below the bound and this scan only raises, so it would
+        // rewrite nothing. Skipping keeps `global_lock` off an O(finalized) walk.
+        if ft_value <= f32::from_bits(self.ft_lower_bound.load(Ordering::Relaxed)) {
+            metrics::counter!("propagate_ft.skipped", "source" => "f1r3fly.casper.block-dag")
+                .increment(1);
+            return Ok(());
+        }
+
         // Update ALL finalized blocks with lower FT, not just ancestors of the
         // current LFB. In a multi-parent DAG, finalized blocks on orphaned
         // branches are not reachable via the ancestor chain of the new LFB.
+        let scan_started = std::time::Instant::now();
         let mut block_metadata_index_guard = self.block_metadata_index.write();
         let finalized_hashes = block_metadata_index_guard.finalized_block_hashes();
-        block_metadata_index_guard.update_ft_if_higher(finalized_hashes, ft_value)
+        let scanned = finalized_hashes.len();
+        block_metadata_index_guard.update_ft_if_higher(finalized_hashes, ft_value)?;
+        drop(block_metadata_index_guard);
+        metrics::histogram!("propagate_ft.scan.time", "source" => "f1r3fly.casper.block-dag")
+            .record(scan_started.elapsed().as_secs_f64());
+        metrics::histogram!("propagate_ft.scan.blocks", "source" => "f1r3fly.casper.block-dag")
+            .record(scanned as f64);
+
+        // Every finalized block is now at or above `ft_value`.
+        self.ft_lower_bound
+            .store(ft_value.to_bits(), Ordering::Relaxed);
+        Ok(())
     }
 }
 
--- a/block-storage/src/rust/dag/block_dag_key_value_storage.rs
+++ b/block-storage/src/rust/dag/block_dag_key_value_storage.rs
@@ -719,6 +719,21 @@ pub struct BlockDagKeyValueStorage {
     /// Monotonically increasing counter incremented on every successful block insert.
     /// Used by caches to detect when the DAG has changed.
     pub(crate) dag_generation: Arc<AtomicU64>,
+    /// Lower bound on `fault_tolerance_value` across `finalized_block_set`,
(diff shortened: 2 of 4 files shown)


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR delivers two focused improvements: an atomic f32-bits lower bound that lets propagate_ft_to_finalized_blocks skip the O(finalized) rewrite scan when it cannot raise anything (plus scan timing/size metrics), and moving the last_finalized_block call inside the /api/status slow-path timer so the warning can actually name the global_lock-bound step. The web_api change is straightforward and correct. The block-dag change is well-tested for the sequential case (the included test correctly exercises the tricky lower-then-raise ordering), but has a few correctness risks: the unconditional post-scan store can race with a concurrent lowering fetch_update unless both are provably serialized under global_lock; initializing the bound to 0.0 is unsafe if negative FT values are possible (NEG_INFINITY would be strictly safer); the bound-lowering runs after a fallible update and could be skipped on partial failure; and the skip check actually runs after acquiring the exclusive lock, contradicting the PR description. None of these break the happy path, but the race and initialization issues could silently leave blocks with stale FT values, which is exactly the class of bug this optimization must never introduce — they deserve a response before merge.

The status instrumentation change looks correct, and the sequential lower-bound test covers an important case. However, the FT optimization has a major race because the propagated bound is stored after releasing the metadata lock, allowing a concurrent lower update to be lost and causing incorrect future skips. Additionally, the skip check remains behind the exclusive global lock, so it does not deliver the advertised reduction in reader queueing. These issues should be addressed before merging.

The code changes look good and address the identified issues. There are minor style improvements and a suggestion to add documentation to the test for better clarity.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Initializing ft_lower_bound to 0.0 is unsafe if fault-tolerance values can be negative</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 783)

The bound is a claimed lower bound on FT across all finalized blocks, and the safety property is: 'if bound is too LOW, we only do an unnecessary scan; if bound is too HIGH, we skip a scan that was needed.' Casper fault-tolerance values conventionally range over [-1, 1]. If any finalized block ever carries a negative FT (or `record_directly_finalized` is ever called with a negative `ft_value`), initializing the bound to `0.0` on startup overstates it: a subsequent propagate with a negative-but-higher `ft_value` (e.g. -0.2 vs a stored -0.5) satisfies `ft_value <= 0.0` and is skipped, leaving those blocks un-raised. Persisted state from a prior run makes this reachable across restarts even if the in-process lowering path would normally track it.

**Suggestion:**
```
Initialize to `f32::NEG_INFINITY.to_bits()` (always safe — worst case is one extra full scan that then establishes the true bound), or compute the true minimum from storage at startup.
```

</details>

<details>
<summary><b>:orange_circle: Major: Fast-path check still waits for the exclusive global lock</b> (Reported by: openai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1352)

The `global_lock.write()` call occurs before the lower-bound check. Therefore skipped propagations still acquire and wait for the exclusive lock, continuing to queue all readers. This contradicts the PR's stated goal of skipping before the exclusive lock is taken and leaves a substantial part of the reported write-side contention intact, although it does avoid the finalized-history scan.

**Suggestion:**
```
Place a concurrency-safe fast-path check before taking `global_lock`. A simple move is not sufficient unless bound changes are coordinated correctly; first serialize bound updates with metadata mutations, then design the fast path so concurrent lower finalizations cannot produce a false skip. Optionally recheck after acquiring the lock for the scan path.
```

</details>

<details>
<summary><b>:orange_circle: Major: Bound update can overwrite a concurrently lowered value</b> (Reported by: openai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1370)

The metadata write guard is dropped before `ft_lower_bound.store(ft_value.to_bits(), ...)`. In that gap, another finalization can acquire the metadata lock, add a block with a lower FT, and lower the atomic bound. The unconditional store then overwrites that lower value with the propagated higher value. A later propagation may consequently skip even though the newly finalized block still needs its FT raised. For example, a scan at 0.5 can release the metadata lock, a block can finalize at 0.4 and lower the bound, and then the scan can store 0.5; a subsequent propagation at 0.45 will incorrectly skip.

**Suggestion:**
```
Update `ft_lower_bound` before releasing `block_metadata_index_guard`, and likewise lower the bound while the finalization metadata guard is still held. This serializes metadata changes and bound transitions under the same lock. Add a concurrent finalization/propagation regression test, since the current test only covers sequential rounds.
```

</details>

<details>
<summary><b>:orange_circle: Major: Unconditional store of ft_lower_bound can lose a concurrent lower value</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1375)

After the scan completes, `self.ft_lower_bound.store(ft_value.to_bits(), Relaxed)` unconditionally raises the bound. The lowering path in the finalization effect uses `fetch_update` after `drop(block_metadata_index_guard)`, and it is not evident from the diff that both paths are serialized under the same `global_lock` acquisition (propagate acquires `global_lock.write()` itself, so the finalize effect presumably runs under a separate acquisition). If a concurrent finalization inserts a block at a lower ft_value (and lowers the bound) after the scan takes its snapshot of `finalized_block_hashes()` but before the `store`, the store overwrites the lowered bound. That block is then below the recorded bound, and a future `propagate_ft_to_finalized_blocks` with a value between the block's FT and the (stale-high) bound will be incorrectly skipped, leaving the block's FT permanently under-propagated.

---

If `ft_value` is ever NaN, the skip comparison `ft_value <= bound` is false, so the scan runs, and then `NaN.to_bits()` is stored as the bound. All subsequent `x <= NaN` comparisons are false, so every future scan runs — fail-safe but silently defeats the optimization forever. Additionally, distinct NaN bit patterns and -0.0/+0.0 bit-level asymmetries are lurking footguns when round-tripping floats through `AtomicU32`.

**Suggestion:**
```
Guard with `debug_assert!(ft_value.is_finite())` or skip the store when `!ft_value.is_finite()`.
```

</details>

<details>
<summary><b>:orange_circle: Major: Slow-status warning threshold semantics changed implicitly</b> (Reported by: anthropic, bedrock)</summary>

**File:** `node/src/rust/api/web_api.rs` (line 353)

Moving `last_finalized_block()` above the threshold check means `total_elapsed` now includes the LFB call, so the warning can fire for endpoints that were previously under threshold. This is the intended fix, but the change also shifts what `STATUS_SLOW_THRESHOLD` measures without any adjustment or comment on the constant itself; anyone tuning the threshold based on historical data should be aware the baseline moved.

---

The semicolon after the closing brace of the `collect()` call is unnecessary and can be removed for better readability.

**Suggestion:**
```
Add a brief note near `STATUS_SLOW_THRESHOLD` that the measured window now includes the LFB lookup, and consider whether the threshold value needs re-tuning.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary semicolon</b> (Reported by: bedrock)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1273)

The semicolon after the `?>` is unnecessary and can be removed for better readability.

**Suggestion:**
```
Remove the semicolon after the `?>`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Bound is not lowered if update_finalized partially succeeds then errors</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1279)

The effect closure now does `update_finalized(...)?` before running the `fetch_update` that lowers the bound. If the underlying update partially applies (some blocks admitted at `ft_value`) and then returns an error, the `?` short-circuits and the bound is never lowered, leaving admitted blocks below the recorded bound — the same skipped-scan hazard as the race above. Whether this is reachable depends on the atomicity guarantees of `update_finalized`, which aren't visible in this diff.

**Suggestion:**
```
Lower the bound before calling `update_finalized` (lowering too early is always safe — worst case is one unnecessary scan), i.e. run the `fetch_update` first, then propagate the result.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Skip check happens after acquiring global_lock, contradicting the PR description</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1352)

The PR description states the scan is 'skipped before the exclusive lock is even taken', but in the code `let _lock_guard = self.global_lock.write();` executes before the skip check. Writers still queue behind every reader to perform the skip. Note that checking under the lock is actually the safer choice given the atomic's Relaxed ordering (a pre-lock read could observe a stale-high bound), so the code is arguably right and the description is wrong — but the discrepancy should be resolved one way or the other, and the in-code comment ('Skipping keeps global_lock off an O(finalized) walk') should make clear the lock is still briefly acquired.

**Suggestion:**
```
Either update the PR description/comment to reflect that the lock is taken briefly, or move the check before the lock with an analysis of why a stale read is safe (it isn't, without stronger ordering or lock coverage of all writers).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: ft_lower_bound field lacks a doc comment</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 722)

The adjacent `dag_generation` field has a doc comment explaining its purpose and encoding; `ft_lower_bound` has none, even though it has a non-obvious encoding (f32 bits stored in an AtomicU32) and a subtle invariant (lower bound over FT of all finalized blocks; lowering is always safe, raising only after a completed scan).

**Suggestion:**
```
Add a doc comment describing the f32-bits encoding, the invariant, and which lock (if any) writers must hold.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add test documentation</b> (Reported by: bedrock)</summary>

**File:** `block-storage/tests/block_dag_storage_test.rs` (line 1234)

Add a brief description of what the test `a_lower_finalization_round_does_not_block_a_later_raise` is verifying.

**Suggestion:**
```
Add a comment above the test describing its purpose and expected behavior.
```

</details>
---

<details>

This PR delivers two focused improvements: an atomic f32-bits lower bound that lets propagate_ft_to_finalized_blocks skip the O(finalized) rewrite scan when it cannot raise anything (plus scan timing/size metrics), and moving the last_finalized_block call inside the /api/status slow-path timer so the warning can actually name the global_lock-bound step. The web_api change is straightforward and correct. The block-dag change is well-tested for the sequential case (the included test correctly exercises the tricky lower-then-raise ordering), but has a few correctness risks: the unconditional post-scan store can race with a concurrent lowering fetch_update unless both are provably serialized under global_lock; initializing the bound to 0.0 is unsafe if negative FT values are possible (NEG_INFINITY would be strictly safer); the bound-lowering runs after a fallible update and could be skipped on partial failure; and the skip check actually runs after acquiring the exclusive lock, contradicting the PR description. None of these break the happy path, but the race and initialization issues could silently leave blocks with stale FT values, which is exactly the class of bug this optimization must never introduce — they deserve a response before merge.

The code changes look good and address the identified issues. There are minor style improvements and a suggestion to add documentation to the test for better clarity.

#### openai (gpt-5.6-sol)

The status instrumentation change looks correct, and the sequential lower-bound test covers an important case. However, the FT optimization has a major race because the propagated bound is stored after releasing the metadata lock, allowing a concurrent lower update to be lost and causing incorrect future skips. Additionally, the skip check remains behind the exclusive global lock, so it does not deliver the advertised reduction in reader queueing. These issues should be addressed before merging.

#### xai (grok-4.5)

</details>

---

spreston8 (approved):
Reviewed at `31ac5c218` against `dev`. Approving.

Pushed to this branch rather than requested: the bound now lowers before `record_finalized` rather than after; the two equivocation tests moved onto `access_equivocations_tracker`; `required-features` dropped from both `[[test]]` targets, so the 23 tests in them run under a plain `cargo test -p block-storage` and therefore in CI; `update_equivocation_record` deleted once that removed its last caller; and a doc comment on `ft_lower_bound`.

The Major findings in the multi-agent review above do not hold. Both writes to the bound happen under the same `global_lock` acquisition, and `0.0` is the exact minimum rather than an overstatement, since every writer of `fault_tolerance_value` only raises.

PR #125, opened 2026-07-15
fv: port merge-recovery FV campaign from #114 (Phase A artifacts)

jeffrey-l-turner:
- Port formal/{rocq,tlaplus,z3,sage,wolfram} models and proofs for finalized-floor, fork-choice, merge-algebra, deploy-lifecycle; extend the existing slashing campaign
- Port docs/theory dossiers, specifications, glossaries, and diagrams
- Port local-only verification gate scripts (scripts/check-*-ALL.sh)
- Add docs/validation/fv-campaign-gap-analysis.md: PR lineage findings, defect matrix (#114 fixes vs #122 ports), and Phase A/B/C plan
- Ignore .localtools and editor backup files

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/125"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786734049&installation_id=145946300&pr_number=125&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F125&signature=7adb4136391490ef5b7cb167f4c0b61d5b787ae397f3de97383b2e1bcc72b7fc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>/codesmith</code> with what you need. Autofix is disabled.</sup>

PR #305, opened 2026-08-20
chore(ci): pin SI verdict-sweep branch and re-enable ucc

spreston8:
**Draft, not for merge as-is.** The pin points at a PR branch head, not a main SHA.

**Pin** — `SYSTEM_INTEGRATION_REF` -> `86a5fa10`, the head of F1R3FLY-io/system-integration#125 (shared-deadline verdict resolution). The point is to run ucc against that resolver before #125 merges. Once it does, repin to the resulting main SHA — that repin is a required follow-up before this can merge.

**ucc re-enabled**, as the one unit its own note described:
- `ucc_tests`: drop `if: false`
- amd64 ephemeral launcher: 2 -> 4 (2 integration legs + 2 ucc legs; arm64 stays 2)
- `ci.yml` amd64 slot gate: 2 -> 4, so the aggregator expects the new job count

**Why now.** Four ucc failures across both providers in one day — `test_guarded_rmw_conflict_no_double_apply`, `test_concurrent_delete_and_set_same_key`, `test_set_cell_concurrent_distinct_elements_union`, `test_single_cell_map_concurrent_adds_all_resolve` — all reported "no coherent terminal verdict". In every one, all five nodes had written terminal `Expired` within 100-185 ms of each other, and `boot` (named in all four) was consistently among the last to decide. Those shards were finality-healthy throughout: floor gap 1, floor pins 11.6-12.6s, all validators advancing. That is the serial-polling race #125 fixes, not a consensus fault.

**Caveat worth watching in this run.** ucc was disabled for RSS-guardian kills, and it is the heaviest memory consumer in the pipeline. Raising the launcher to 4 amd64 restores the VM-per-job ratio, but whether the memory pressure is resolved is exactly what this draft is meant to find out.

### DIFF
--- a/.github/oci-validation.env
+++ b/.github/oci-validation.env
@@ -16,4 +16,4 @@
 # Bump both in the same commit: `just repin-system-integration <sha>`
 # (scripts/repin-system-integration.sh) rewrites every pin site together and
 # runs the invariants checker.
-SYSTEM_INTEGRATION_REF=56884ab896a8d37d7fa7cba8eed2a0ffef9c1b6a
+SYSTEM_INTEGRATION_REF=86a5fa108d4996d690eef71d28889859c9cdfd4d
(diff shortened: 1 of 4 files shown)


spreston8:
Superseded by #318 — its CI-enable commit carries this PR's job (SI pin + parallel ucc leg re-enable with the coupled launcher count and slot gate) plus eight audited deselect drops.

Issue #268, opened 2026-08-14
mntd-calibration at high cap sometimes stalls finalization to exactly zero with the node fully alive, distinct from the OOM crash-loop in #244


9Neechan:
**Affected:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (`prefer_deploy_support_main_parent` — confirmed proximate cause under flood load, mitigated here, see "Mitigation applied, not confirmed sufficient" below); `casper/src/rust/finality/finalizer.rs` (`Finalizer::run`, specifically the `Finalizer::cannot_be_orphaned` >50%-agreement gate); `casper::rust::blocks::proposer::block_creator` (backpressure/fallback logic, "Ordinary user deploy fallback enabled" log line); `mntd-calibration` experiment (`experiment/mntd`); `mid` profile (2/8 CPU, 4Gi/16Gi RAM, 100Gi disk)

**Severity:** medium, **not fixed** — a distinct failure mode from [#244](https://github.com/F1R3FLY-io/f1r3node-rust/issues/244)'s OOM crash-loop, which restarts the pod. Here the node never dies — block production and gossip continue — but finalization output is exactly zero for the entire measurement window. Root-caused to `prefer_deploy_support_main_parent` under flood load; a sig-count ceiling on the branches it's allowed to promote (`MAX_PROMOTABLE_BACKLOG_SIGS`) was applied and eliminated the stall on the specific flood-load caps it was tested against, but a 2026-08-24 `smoke` run (light load, 4 validators, no flood) hit the same `Finalizer::cannot_be_orphaned` gate (`finalizer.rs:85-90`, "non-positive stake totals: total=300000, agreeing=0") and stalled finalization permanently — see "Mitigation applied, not confirmed sufficient" below. The underlying liveness bug is still open. Note: a *different*, unrelated bug ([#194](https://github.com/F1R3FLY-io/f1r3node-rust/issues/194)) causes an identically-shaped zero-finalization stall at extreme cap (≥102) — see the note at the end of that section.

## The problem

Running `mntd-calibration` (multi-validator flood-propose load) at various caps, finalization sometimes stops completely — `FIN_B=0`, `TPS=0.00` — for the whole measurement window, while the node stays fully alive: `restart_count=0`, no crash, `min_peers(>=2)` checks pass throughout, blocks keep getting proposed and gossiped.

[...]

Issue #306, opened 2026-08-20
LFS-synced joiners fail block admission: weight_from_validator_by_dag reads a main parent below the restore horizon (sixth walk for #303)


jeffrey-l-turner:
**Affected:** LFS-synced joiners and observers — block admission and the fork-choice weight lookup
**Severity:** high — blocks the 60h stability soak gate on master; late-joining nodes fail to add blocks near their sync anchor

## Observed

The 60h stability soak against master `358cc85e3` (PR #302) failed its full integration preflight twice on 2026-08-20:

- run 32390713819 — OCI runner lost (infrastructure), automatic in-window restart fired
- run 32403395848 — the restart reached a genuine verdict: **7 failed, 6 errors of 124**

The dominant forbidden-log storm is one signature, thousands of entries on `joiner5`, `joiner7`, `observer1`, `observer2`:

```
KvStore error: Key not found: weight_from_validator_by_dag: main-parent metadata missing from index (sync/prune window): <hash>
```

Affected tests: `test_validator_lifecycle` (joiner7: "received but not added yet" after 135s), `test_trim_state`, `test_websocket::test_block_events`, `test_readonly_catchup_bounded` (503 on `explore-deploy`), and `test_observer_lfs_sync` (observer LFB drift never converges). The same storm killed the scheduled dev integration soak on every night since at least 2026-08-13 against pre-promotion master, so the promotion did not clear it.

## Mechanism

`casper/src/rust/util/proto_util.rs`, `weight_from_validator_by_dag` (lines 162-195): the function reads the validator weight from the block's **main parent** metadata. On an LFS-restored node, the main parent of blocks near the restore horizon is settled history held as a hash only, so `dag.lookup(parent_hash)` returns `None` and the function raises `KvStoreError::KeyNotFound` with a string message. Block processing treats that as a hard failure, the block is never added, and admission of everything above it stalls.

Two existing remedies do not catch this site:

- The `BlockNotHeld` collapse (`b62532b59`) converts only `KvStoreError::MissingBlock { hash }` in `From<KvStoreError> for CasperError`. This site raises the string `KeyNotFound` variant, so the deferral never fires.
[...]

PR #303, opened 2026-08-20
fix: truncation-horizon campaign — five walks learn the restore horizon

spreston8:
Every walk that assumes full history must learn the restore horizon: an LFS-restored node holds a window, and references below it are settled history held as hashes only. Five defects were the five walks that didn't know that — all fixed red-first, live-verified across validator-lifecycle runs 15-22 (three consecutive full passes).

- `bd37ede7` finalized-ancestry marking terminates at the horizon (`held_ancestors`)
- `a45ec57d` + `e5b15f9e` newly-bonded slots seed the network-uniform genesis (write-once register, refuse-not-invent)
- `849f4d1c` the restore ships the genesis block itself, verified by claimed hash and content re-hash, inserted Approved
- `fda7b49b` a propose needing an unheld block requests it from peers instead of backing off (`ProposeFailure::MissingBlock`, non-bug heartbeat)
- `20054e78` the deploy-lifecycle register abstains at the horizon instead of erroring block admission (typed `BlockNotHeld`; no invented Expired over an unreadable segment)
- `2b00702f` block-pipeline outcomes are typed, not stringly-typed errors
- `b7e28128` CI: pin system-integration to the race-proofed lifecycle harness (F1R3FLY-io/system-integration#124 branch) and seat `test_validator_lifecycle.py` — re-pin to the merged main SHA once this PR's CI proves it green

Co-Authored-By: Claude <noreply@anthropic.com>

PR #318, opened 2026-08-22
fix(consensus): finality divergence and stall recovery

spreston8:
Supersedes #303 (folded wholesale — its commits and authorship are in this branch's history) and #305 (the CI pin + ucc re-enable land here). Consensus-affecting: fork choice, the clique-oracle walk, and the slash-evidence taxonomy change together and require atomic activation (all validators run this build together — the A9 posture; no mixed-version window exists for an unreleased node).

### Diff anatomy

|---|---:|---|---|
| Production (src + proto) | 28 | +2,254 −912 | The consensus changes below; deletions are the retired tip-ranking machinery, leader-only heartbeat lanes, and `prefer_certified_main_parent` with its in-module spec |
| Tests | 43 | +2,897 −359 | Red-first specs for every fix + taxonomy pin flips |
| Replay fixtures | 2 | +6,243 | `stall_fixtures/{i1,i5}.json` — machine-distilled from two CI shards' logs (senders, heights, justification maps, true stakes); consumed by one spec; skim, don't read |
| Docs | 17 | +329 −230 | Spec/verification/operational docs rewritten to the shipped semantics (one commit, `b7f9317b6`) |
| Config | 1 | +27 −12 | `defaults.conf`: recovery cadence 3s/3s, frontier chase 20 — pinned by an embedded-defaults test |
| CI | 4 | +40 −28 | system-integration pin → the audited harness; nine earned deselects dropped + the ucc leg re-enabled (see "Re-enabled integration coverage") |

### What changed

- **Fork choice**: true heaviest-subtree descent replaces flat tip ranking — the ucc-i6 finality-divergence root, where a hash-tie head flip reverted a 200-of-300 certificate with zero equivocations. The spec (R-GHOST) always demanded this; the old code violated it at any fork of depth ≥ 2.
- **Clique oracle**: the disagreement walk is two-sided by height — below the target, only a rival prefix vetoes; the target's own ancestry is ignorance, not disagreement. Scope pinned before the fix: 37 of 777 replayed verdicts flip, all in i5's withdrawn-committee era; 0 of i1's 264 rival vetoes move.
- **Slash evidence narrows to the equivocation class** (19 → 2 variants): evidence demands a fault every honest node attributes identically from the signed block alone; run 32588262605 showed view-relative verdicts diverging across honest nodes and burning FT −18.55 of honest stake. Demoted verdicts still reject blocks — only evidence minting narrows; per-variant re-promotion path documented. Plus: no slash issuance from a view lagging its own finalized frontier (restored-joiner clock skew).
- **Finalizer + restore horizon**: `held_ancestors` as the one canonical unheld-ancestry walk; the finalizer holds its cycle (`AbsenceHold`) instead of erroring at the retention edge; `ContainmentHold` streaks escalate via a `DivergenceMonitor` (one ERROR + metric; proposing not halted); the restore ships genesis itself, verified twice against a write-once register; newly-bonded LM slots seed network-uniform genesis; the lifecycle register abstains at the horizon; settled-history admission gains a fifth conjunct and never touches latest messages.
- **Heartbeat**: stale-LFB recovery open to every bonded validator, paced by time alone (leader-only lanes deleted — every observed CI stall had all lanes shut while one leader proposed alone); backpressure exemption re-keyed temporal so a stall bounds churn to the recovery cadence, never to zero.
- **Validation pipeline hygiene**: typed pipeline outcomes (routine drops off the error channel); hard validation errors bounded (cap 32 + quarantine + loud purge); propose fetches an unheld block instead of backing off; the fork-choice weight walk raises typed `MissingBlock` so a restored node's below-horizon parent defers the block instead of hard-failing admission (#306's storm signature).
- **Negative-FTT regime made explicit**: under `ftt < 0` "finalized" is bare majority agreement per snapshot, so incompatible same-height certificates are an expected transient — the live finalizer now holds that cycle quietly (typed `IncompatibleFinalizedFork` → `IncompatibilityHold`) while the θ ≥ 0 safety alarm stays loud and byte-identical; startup warns when a node is configured with a negative threshold.

Nine tests come off the CI deselect list (each earned its seat: audited and
where needed hardened in system-integration #126, then validated in the local
full-suite mirror and the green CI runs), plus the ucc leg returns as its own
parallel job:

- `test_validator_lifecycle` — the full join/bond/withdraw arc (the branch's primary live gauntlet)
- `test_bonding_validators` (incl. Phase C)
- `test_joiner_self_proposes_at_epoch_boundary`
- `test_observer_lfs_sync`
- `test_contract_lifecycle`
- `test_consensus_safety::test_epoch_transition_under_heartbeat`
- `test_active_validator_cap`
- `test_asymmetric_bonds::test_finalization_asymmetric_bonds`
- `test_finality_stall_recovery` (the cadence-and-width bound from the revised amplification claim)
- ucc leg: `test_user_contract_concurrency` — all 11 contention shapes, dedicated job per arch/provider

Still deselected (unchanged): `test_load`, `test_shard_degradation`,
`test_network_converges_after_slow_deploy`, `test_cold_start_readiness`,
`test_concurrent_bridge_locks` (owned by a separate node-side PR),
`test_slow_peer_notification`.

### Suggested reading order

1. `block_status.rs::is_slashable` + the `slashing-specification.md` amendment (smallest, self-contained)
2. `estimator.rs::rank_forkchoices` + `tests/fork_choice/heaviest_subtree_descent.rs`
3. `clique_oracle.rs::never_eventually_see_disagreement` + `tests/finalized_floor/oracle_stall_replay_spec.rs`
4. `finality/floor.rs` (`FloorOfView`) + `finalization_runner.rs`
5. `heartbeat_proposer.rs::decide_lanes` + `docs/claims/heartbeat-proposal-amplification-bound.md`
6. `block_dag_key_value_storage.rs` (held_ancestors, genesis register, `SettledHistory`) + `block_processor.rs`
7. Restore path (`initializing.rs`, `running.rs`, proto), then CI pins and docs

Two notes that preempt commit-by-commit confusion: `864c39ee4` contributes tests only (two branches independently fixed the same sweep bug; the fold kept one fix, both test suites); the settled-history door itself predates this branch — it gains one conjunct and a consensus-neutral insert mode here.

### Validation

- CI full green ×3 on the exact merge tree: run 32691266446 attempts 1+2, run 32759382579 (all ~1h05)
- Two full CI simulations green on CI-shaped VMs before the real runs
- Local combined mirror (full suite + bonding + lifecycle + ucc, one invocation): 118 passed / 0 failed / 1h48m, forbidden-log scanners active on every shard, zero trips
- Workspace gate green (`pos_spec` = standing full-suite load flake, solo-green)

### Known follow-ups (deferred deliberately)

- Formal-layer re-derivation: `Rank.v` / `GuardBridge.v` pipeline theorems / `InvalidBlock.v` taxonomy model cite pre-branch code (marked historical in the dossiers)
- Demoted-verdict re-promotion once individual checks are shown admission-order-free
- `AbsenceHold` observability (streak counter + active fetch)


jeffrey-l-turner:
> Supersedes #303 (folded wholesale — its commits and authorship are in this branch's history) and #305 (the CI pin + ucc re-enable land here).

Does this address?: https://github.com/F1R3FLY-io/f1r3node-rust/issues/268#issue-5151928531

spreston8:
> Does this address?: https://github.com/F1R3FLY-io/f1r3node-rust/issues/268#issue-5151928531

@jeffrey-l-turner Structurally, yes — with one caveat. Every mechanism #268 names is gone from this branch: `prefer_deploy_support_main_parent` (the confirmed proximate cause), the `MAX_PROMOTABLE_BACKLOG_SIGS` mitigation, and `finalizer.rs::cannot_be_orphaned` all have zero occurrences here. The main parent is now the pure GHOST head from a true heaviest-subtree descent — exactly the no-override configuration #268's own A/B measured at 0/14 stalls. The promotion's legitimate purpose (rejected-deploy re-inclusion pinning) is covered differently: deploy finalization is spine-independent (`verdict_convergence_spec::a_deploy_finalizes_from_a_carrier_the_spine_never_holds`). The branch also adds recovery for the general self-referential-chain fragmentation: an open, interval-paced stale-recovery lane plus a one-shot multi-parent convergence proposal, and a DivergenceMonitor so a persistent stall is an ERROR + metric instead of silence.

The caveat: the 2026-08-24 recurrence in #268 was captured on an image that still contains `finalizer.rs` (cited at :85-90), i.e. the pre-rewrite architecture — and nobody has run `mntd-calibration`/smoke against this branch's build. I'd keep #268 open until the same cap grid reruns on an image built from this branch (or dev after merge); if the stall reproduces there, it's a genuinely new mechanism worth a fresh capture.

spreston8:
> > Supersedes #303 (folded wholesale — its commits and authorship are in this branch's history) and #305 (the CI pin + ucc re-enable land here).
> 
> 
> 
> Does this address?: https://github.com/F1R3FLY-io/f1r3node-rust/issues/268#issue-5151928531

Structurally, yes — with one caveat. Every mechanism #268 names is gone from this branch: `prefer_deploy_support_main_parent` (the confirmed proximate cause), the `MAX_PROMOTABLE_BACKLOG_SIGS` mitigation, and `finalizer.rs::cannot_be_orphaned` all have zero occurrences here. The main parent is now the pure GHOST head from a true heaviest-subtree descent — exactly the no-override configuration #268's own A/B measured at 0/14 stalls. The promotion's legitimate purpose (rejected-deploy re-inclusion pinning) is covered differently: deploy finalization is spine-independent (`verdict_convergence_spec::a_deploy_finalizes_from_a_carrier_the_spine_never_holds`). The branch also adds recovery for the general self-referential-chain fragmentation: an open, interval-paced stale-recovery lane plus a one-shot multi-parent convergence proposal, and a DivergenceMonitor so a persistent stall is an ERROR + metric instead of silence.

The caveat: the 2026-08-24 recurrence in #268 was captured on an image that still contains `finalizer.rs` (cited at :85-90), i.e. the pre-rewrite architecture — and nobody has run `mntd-calibration`/smoke against this branch's build. I'd keep #268 open until the same cap grid reruns on an image built from this branch (or dev after merge); if the stall reproduces there, it's a genuinely new mechanism worth a fresh capture.

spreston8:
> > > Supersedes #303 (folded wholesale — its commits and authorship are in this branch's history) and #305 (the CI pin + ucc re-enable land here).
> > 
> > 
> > Does this address?: [#268 (comment)](https://github.com/F1R3FLY-io/f1r3node-rust/issues/268#issue-5151928531)
> 
> Structurally, yes — with one caveat. Every mechanism #268 names is gone from this branch: `prefer_deploy_support_main_parent` (the confirmed proximate cause), the `MAX_PROMOTABLE_BACKLOG_SIGS` mitigation, and `finalizer.rs::cannot_be_orphaned` all have zero occurrences here. The main parent is now the pure GHOST head from a true heaviest-subtree descent — exactly the no-override configuration #268's own A/B measured at 0/14 stalls. The promotion's legitimate purpose (rejected-deploy re-inclusion pinning) is covered differently: deploy finalization is spine-independent (`verdict_convergence_spec::a_deploy_finalizes_from_a_carrier_the_spine_never_holds`). The branch also adds recovery for the general self-referential-chain fragmentation: an open, interval-paced stale-recovery lane plus a one-shot multi-parent convergence proposal, and a DivergenceMonitor so a persistent stall is an ERROR + metric instead of silence.
> 
> The caveat: the 2026-08-24 recurrence in #268 was captured on an image that still contains `finalizer.rs` (cited at :85-90), i.e. the pre-rewrite architecture — and nobody has run `mntd-calibration`/smoke against this branch's build. I'd keep #268 open until the same cap grid reruns on an image built from this branch (or dev after merge); if the stall reproduces there, it's a genuinely new mechanism worth a fresh capture.
> 

Correction to my earlier reply, after reading #268's comment thread: @9Neechan reproduced the stall on an unmodified dev build (`aebf358ef`) that already has `prefer_deploy_support_main_parent` deleted — cap68, `FIN_B=0`, no OOM, zero promotion events. So the "0/14 with the function disabled" result was specific to the old architecture, and removal alone does not close #268 under `mntd-calibration` flood. The mechanism that survives is view fragmentation (76% self-parent chains under flood).

What stands from my earlier comment: this branch replaces the flat tip-ranking `aebf358ef` still runs with a true heaviest-subtree descent (untested against that workload), deletes both content-driven promotions (P4-aligned), and adds the convergence one-shot + open recovery cadence — but none of that is demonstrated against the flood workload, so #268 should be treated as open regardless of this PR, and the rerun I suggested is the discriminating test either way.

jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This is a large, consensus-critical PR (fork choice, clique oracle, slash-evidence taxonomy, finalizer horizon handling, heartbeat recovery) with unusually strong test discipline — red-first specs, truth-table pins, and fixture-driven replays accompany every visible change. The visible engineering is careful, but three areas need human judgment before merge: (1) the write-once genesis-hash register is recorded from a peer response with no visible independent verification, making first-responder poisoning permanent; (2) `held_ancestors` in the finalization sweep silently masks genuine missing-block corruption on non-truncated nodes; and (3) the wholesale demotion of 17 slash verdicts removes economic deterrence for several faults that are provable from the signed block alone (InvalidBlockHash, DeployNotSigned). The fifth settled-admission conjunct may also re-wedge restored nodes for blocks authored by since-unbonded validators. Note this review covers only the first ~15% of a 98-file diff — the fork-choice descent, clique-oracle walk, heartbeat, and proto changes were truncated and are unreviewed here; given they are the core consensus changes, a full human review of those files is mandatory.

The shown portion contains substantial, well-tested recovery work, but two trust-boundary assumptions require correction before merge: the restore path can permanently record an unverified genesis identity, and the finalization walk treats any missing parent as proven settled ancestry without an explicit restore-horizon discriminator. There are also consistency and observability issues around non-atomic genesis installation and prematurely clearing divergence state. Because this is a consensus-affecting change and the supplied diff is truncated, the remaining files should receive focused review after these issues are addressed.

This is a large, consensus-critical change set (fork choice, clique-oracle walk, slash taxonomy, finalizer/restore horizon, heartbeat recovery, validation hygiene) that correctly aims at real CI failures (finality divergence, honest-stake burns, restored-node LFB freeze, validation storm). Visible production code is carefully commented and backed by strong targeted tests (held_ancestors horizon, SettledHistory LM isolation, genesis write-once + content verify, DivergenceMonitor streaks, five-condition settled admission, validation-error cap). It still needs expert human review before merge: the slashable-set collapse is an economic-security policy decision that may be broader than the incident requires; held_ancestors can mask non-horizon missing parents if reused broadly; settled admission trusts a hard-coded bonded flag; and divergence remains operator-visible-only while proposing continues. Approve only after consensus owners sign the taxonomy/fork-choice/oracle semantics and the SI pin/soak defaults under the hotter recovery cadence.

The code changes introduce significant modifications to the consensus mechanism, including fork choice, clique oracle, and finality divergence handling. These changes require careful review to ensure they do not introduce new vulnerabilities or logic errors. The new InsertMode::SettledHistory and the narrowing of slash evidence are particularly critical and should be tested extensively.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Fifth conjunct may re-wedge settled-history admission for senders absent from latest messages</b> (Reported by: anthropic, openai)</summary>

**File:** `casper/src/rust/blocks/block_processor.rs` (line 1406)

The code separately loads settled_admissions, checks it against the budget, inserts the block, and then increments the counter. If block processing can run concurrently, multiple callers can pass the check at the boundary and exceed the storage budget. This weakens the documented security property that attacker-triggered settled-history growth is strictly bounded.

---

`seq_below_senders_latest` evaluates to `false` when the sender has no latest message (`None => false`) or when the latest message hash fails lookup. On a truncated node, historic settled blocks are frequently authored by validators that have since unbonded and therefore have no latest-message slot — every such block now fails the fifth conjunct and takes the judged/deferral path. Since settled-history admission was introduced precisely to break restore-era deferral wedging, this conjunct may reintroduce the stall for exactly the blocks it was meant to admit. The truth-table test covers the boolean, but no test in the visible diff covers the 'sender has no latest message but is genuinely settled history' scenario end-to-end.

**Suggestion:**
```
For senders with no latest-message slot, fall back to a height-based check (block height strictly below the anchor's justification frontier) instead of unconditionally failing, or add an integration-level test demonstrating restored nodes still converge when settled history includes unbonded historic senders.
```

</details>

<details>
<summary><b>:red_circle: Critical: Write-once genesis register can be permanently poisoned by the first restore peer</b> (Reported by: anthropic, openai)</summary>

**File:** `casper/src/rust/engine/initializing.rs` (line 1095)

During restore, `record_genesis_hash(response.genesis_hash.clone())` is called for any non-empty genesis_hash in the floor-cache response. `record_genesis_hash` is write-once and errors on any different value forever after. If the exchange peer is malicious (or the response arrives from an unauthenticated/compromised channel), the first responder permanently fixes a wrong genesis hash on this node — it then poisons every newly-bonded latest-message seed (which the comment says must be network-uniform), and `insert` will hard-fail or, worse, seed a divergent sentinel that reads as equivocation network-wide. The diff shows no verification of `response.genesis_hash` against the approved block, the bonded set, or a quorum of peers before it is written; the block-body path is verified (re-hash) but the hash-only path is trusted outright.

---

The restore path calls record_genesis_hash(response.genesis_hash) before requiring a genesis block or validating its contents. Because the register is write-once, a malformed or malicious response containing an arbitrary non-empty hash can permanently pin the node to the wrong genesis identity. An absent or invalid genesis block only makes genesis_landed false, after which restore still returns success; a later correct response then fails because the register refuses overwrite. Moreover, validating a supplied block against a hash learned from the same response establishes only self-consistency, not that this is the shard's canonical genesis, unless the surrounding exchange independently authenticates that hash.

**Suggestion:**
```
Authenticate the genesis hash against trusted restore metadata, the approved anchor, or a quorum-backed proof before persisting it. At minimum, require the block to be present and successfully re-hash to the supplied hash before calling record_genesis_hash, and do not complete restore successfully when a first-time genesis identity cannot be verified. Add a test where the first response supplies an invalid hash/block and a later response supplies the correct genesis.
```

</details>

<details>
<summary><b>:red_circle: Critical: Finality divergence is detect-only; proposing continues on a frozen read surface</b> (Reported by: bedrock, xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 72)

DivergenceMonitor escalates after 10 rising ContainmentHold refusals on one pinned LFB (ERROR + metric) but deliberately does not halt proposing. That matches the incident analysis, yet operators and downstream consumers can keep reading a frozen LFB while the node still emits blocks. Without runbook automation (alert route, auto-resync, API surface exposing diverged()), this can look like a healthy validator while its finalized state is permanently wrong.

---

The DivergenceMonitor should be thoroughly tested to ensure it correctly detects and handles finality divergences.

**Suggestion:**
```
Expose diverged() on admin/status APIs and alert rules in the same PR/docs set. Confirm soak/CI asserts the metric fires on the ucc-i6 shape and that status makes the frozen finalized surface unmistakable.
```

</details>

<details>
<summary><b>:orange_circle: Major: InsertMode::SettledHistory</b> (Reported by: bedrock)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 83)

The new InsertMode::SettledHistory should be thoroughly tested to ensure it behaves as expected, especially in scenarios involving truncated DAGs and restore operations.

</details>

<details>
<summary><b>:orange_circle: Major: held_ancestors silently treats any missing parent as settled horizon</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 674)

The finalization marking walk now terminates on unheld parents instead of erroring. That fixes restored-node LFB freeze, but the same helper cannot distinguish 'below restore truncation' from 'transient MissingBlock / corruption / partial DAG'. If record_directly_finalized is ever invoked while live parents are merely not yet fetched, ancestry can be marked finalized without those parents ever being examined or marked, leaving an incomplete finalized set relative to peers who hold the full chain.

**Suggestion:**
```
Constrain call sites so held_ancestors is only used when the node is known-truncated (restore/LFS path) or after an explicit horizon check. Alternatively pass a horizon hash/height and only suppress MissingBlock for parents at-or-below that floor; surface MissingBlock above it.
```

</details>

<details>
<summary><b>:orange_circle: Major: All missing parents are treated as below the restore horizon</b> (Reported by: openai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 684)

held_ancestors stops at every unheld parent without proving that the edge crosses a trusted restore boundary. The method is then used by record_directly_finalized globally, including for ordinary DAGs. The added test even constructs a Normal block with an arbitrary missing parent and treats it as settled. A missing live merge parent, incomplete dependency, or malformed admitted block is therefore indistinguishable from genuinely truncated finalized ancestry. This can silently finalize held descendants while omitting ancestry that was never established as settled, and can leave that parent unmarked if it arrives later.

**Suggestion:**
```
Gate horizon termination on explicit restore metadata or a persisted settled-boundary proof. On a non-truncated node, or when the parent is not proven below the anchor, preserve the existing MissingBlock behavior. Tests should cover both a proven sub-horizon parent that terminates successfully and an unheld live parent that still defers finalization.
```

</details>

<details>
<summary><b>:orange_circle: Major: `held_ancestors` in the finalization sweep silently masks genuine missing-block corruption on non-truncated nodes</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1425)

`record_directly_finalized` now uses `held_ancestors` unconditionally. On a truncated (restored) node this is correct — an unheld parent is the horizon. But on a fully-synced ceremony node a missing parent is data loss/corruption, and the walk now silently terminates there instead of surfacing an error: the subtree below the gap never receives finalized marks, `is_finalized` returns false for genuinely finalized ancestry, and finalized-effects callbacks are silently skipped for those blocks. The old behavior (error) was wrong for restored nodes but at least loud for corrupted full nodes; the new behavior is silent in both cases.

**Suggestion:**
```
Gate the lenient walk on the node actually being truncated (e.g., presence of a restore anchor / learned genesis without a held height-0 block), or at minimum emit a WARN + metric when a parent is referenced-but-unheld during the finalization sweep on a node that completed a full sync.
```

</details>

<details>
<summary><b>:orange_circle: Major: Slash Evidence Narrowing</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/block_status.rs` (line 237)

The narrowing of slash evidence to the equivocation class should be carefully reviewed to ensure it does not introduce new edge cases or vulnerabilities.

</details>

<details>
<summary><b>:orange_circle: Major: Slashable set narrowed from 19 variants to 2 — economic security tradeoff</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/block_status.rs` (line 237)

Only AdmissibleEquivocation and IgnorableEquivocation remain slashable. Many faults that are arguably objective from the signed block alone (e.g. InvalidBlockHash, DeployNotSigned, InvalidSequenceNumber, InvalidParents) no longer mint evidence. The motivation (view-relative honest-stake burns in run 32588262605) is sound for JustificationRegression / UnauthorizedSlashDeploy / neglected-* families, but the demotion is blanket. Under A9 atomic activation this permanently changes attacker economics: invalid blocks are still rejected, but there is no stake penalty for a large class of Byzantine payloads.

**Suggestion:**
```
Before merge, explicitly ratify per-variant which demotions are temporary vs permanent. Prefer keeping purely syntactic, admission-order-free checks (hash/signature/deploy-sign) slashable now, and only demote the proven view-relative family. Track re-promotion with tests that pin 'objective from bytes alone' so they cannot silently stay demoted.
```

</details>

<details>
<summary><b>:orange_circle: Major: Mass demotion of slashable verdicts (19 → 2) removes economic deterrence for several objectively provable faults</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/block_status.rs` (line 250)

The rationale (evidence must be attributable identically from the signed block alone) is sound for view-relative verdicts like JustificationRegression and NeglectedEquivocation, but several demoted variants ARE provable from the signed block alone: `InvalidBlockHash` (content does not hash to the claimed hash), `DeployNotSigned` (deploy signature invalid), `InvalidSequenceNumber` and `InvalidBlockNumber` (checkable against the block's own cited justifications/parents when those are included), `InvalidShardId`. Demoting these makes producing malformed-but-signed blocks economically free, weakening the spam/DoS cost model. The comment acknowledges per-variant re-promotion is possible, but shipping all demotions at once as an atomic consensus change means the deterrence gap is live until then.

**Suggestion:**
```
Keep the provable-from-the-block-alone subset (at minimum InvalidBlockHash and DeployNotSigned) slashable in this PR, or document the concrete interim mitigation (rate limits / peer scoring) that replaces the economic deterrent for these variants.
```

</details>

<details>
<summary><b>:orange_circle: Major: try_admit_settled hard-codes solicited_by_bonded=true</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/block_processor.rs` (line 1403)

admit_as_settled's third conjunct is documented as 'citation from a bonded sender', but try_admit_settled always passes true. Safety then depends entirely on callers only invoking this after bonded solicitation bookkeeping. Any future call-site mistake turns settled-history into an unbonded storage/admission path (bounded by budget, but still skips validation).

**Suggestion:**
```
Thread the actual bonded-solicitation boolean from settled_solicitations membership into admit_as_settled instead of a literal true. Add a negative test where an untracked hash cannot take the settled path even if height/budget/seq match.
```

</details>

<details>
<summary><b>:orange_circle: Major: Genesis block store and DAG updates are not atomic</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/engine/initializing.rs` (line 186)

receive_shipped_genesis writes the block body to block_store before inserting it into the DAG. If DAG insertion fails, the function returns an error but leaves the block in the block store. This contradicts the stated refusal invariant that an unsuccessful shipped genesis leaves no trace and can make subsequent retries observe a partially installed genesis.

**Suggestion:**
```
Validate all insertion preconditions before writing either store, or perform the block-store and DAG changes in a transaction with rollback. Add a failure-injection test for DAG insertion after put_block_message and assert that neither store retains the block.
```

</details>

<details>
<summary><b>:orange_circle: Major: Aggressive recovery cadence defaults (15s/12s → 3s/3s) combined with all-validator heartbeat may amplify traffic at scale</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/casper_conf.rs` (line 338)

Stale-LFB recovery is now open to every bonded validator (leader-only lanes deleted) and the default cadence drops to 3s for both self-propose cooldown and stale-recovery interval, with frontier chase raised 0 → 20. During a stall, this means every bonded validator can propose/chase every 3 seconds. The PR asserts backpressure bounds churn to the recovery cadence, but on large validator sets N proposals per 3s window is materially more block and gossip traffic than the previous regime; the visible diff shows CI validation on small shards only.

---

Stale-LFB recovery is open to every bonded validator and paced only by time, with much hotter defaults. This is the stall fix, but it can multiply propose/fetch churn and bandwidth on large validator sets or slow IO nodes, especially with frontier_chase_max_lag=20.

**Suggestion:**
```
Keep the embedded-defaults pin, and confirm soak/merge-recovery jobs measure propose rate, fetch QPS, and CPU under the new cadence on CI-shaped topology before calling the defaults final.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: `genesis_hash()` fallback picks the min-hash height-0 block without uniqueness check</b> (Reported by: anthropic)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 891)

When the register is empty, `genesis_hash()` derives genesis as `height_map.get(&0).min()`. If more than one height-0 block ever exists in the map (e.g., an invalid block admitted at height 0, or a bug elsewhere), the function silently picks the lexicographically smallest, potentially returning a non-genesis hash that then seeds newly-bonded latest-message slots — precisely the network-uniformity violation this PR fixes. Also note `genesis_hash()` does not take `global_lock`, so it can interleave with a concurrent `record_genesis_hash`.

**Suggestion:**
```
Assert/log when `height_map[0]` has more than one entry, and consider taking the global read lock for a consistent read of register + height map.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Validation Failure Handling</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/blocks/block_processor.rs` (line 114)

The validation failure handling logic should be reviewed for clarity and consistency with existing error handling patterns.

</details>

<details>
<summary><b>:yellow_circle: Minor: Validation-error quarantine reuses missing-dependency quarantine duration</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/blocks/block_processor.rs` (line 1218)

note_validation_failure stamps validation_error_quarantine_until with missing_dependency_quarantine_ms(). The failure modes differ (hard Err loop vs dependency wait); coupling them means tuning dependency fetch backoff also changes hard-fail purge damping and vice versa.

**Suggestion:**
```
Give validation errors their own env default (e.g. F1R3_VALIDATION_ERROR_QUARANTINE_MS) or document the intentional shared knob next to both constants.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Quarantine timing uses wall clock; clock steps can void or extend quarantines</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/block_processor.rs` (line 1246)

`note_validation_failure`, `is_validation_failure_quarantined`, and the sweep all use `SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or(0)`. A backwards clock step (NTP correction — explicitly relevant given this PR's own restored-joiner clock-skew fix in the slashing path) can make active quarantines appear far in the future (blocking harvest for hours) or immediately expired (restarting the hot loop the mechanism exists to bound). Additionally, `unwrap_or(0)` on a pre-epoch clock silently disables quarantine entirely.

**Suggestion:**
```
Use a monotonic clock (`Instant`) for quarantine bookkeeping, matching the intent of pacing rather than absolute time.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: TOCTOU window in `sweep_orphaned_validation_error_attempts`</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/blocks/block_processor.rs` (line 1330)

The sweep collects orphaned hashes under one lock acquisition, releases it, then re-acquires the lock to remove them. A block can re-enter the buffer (fresh peer delivery) between the two acquisitions and have its just-restarted attempt counter incorrectly removed. Effect is benign (the counter restarts from zero, extending retries slightly), but the same double-acquire pattern is repeated from the missing-dependency sweep and could be done atomically.

**Suggestion:**
```
Hold the attempts lock across the classify-and-remove pass, or use `retain` with the buffer-membership check inline.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: DivergenceMonitor uses Mutex::expect — poison becomes a finalizer panic</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 130)

on_advance/on_containment_hold call state.lock().expect(...). A poisoned mutex (prior panic while holding the lock) will panic the finalization path rather than degrade to 'hold this cycle'.

**Suggestion:**
```
Match other casper locks: map poison to a logged hold/reset, or use a non-poisoning concurrency primitive if panics in this path must never wedge finalization.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Divergence state is cleared before finalization advancement succeeds</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 425)

divergence_monitor.on_advance() is called as soon as floor_of_view returns Advance, before the new LFB is actually recorded and its effects complete. If the subsequent update fails, the LFB remains pinned but the refusal streak and diverged flag have already been cleared. Repeated failures can therefore suppress or delay the intended divergence alarm.

**Suggestion:**
```
Call on_advance only after update_last_finalized_block and all required persistence/effect work succeed. Merely deriving an advancing candidate should not clear an alarm tied to the persisted LFB.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: SYSTEM_INTEGRATION_REF still documents branch SHAs pending main land</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 140)

Pin history correctly records d94fffd as PR #126 branch-with-main-merged and repeats 're-pin to merged main SHA once #126 lands'. Merging the node PR while SI is still a moving branch pin risks CI drift if that branch is force-pushed or superseded.

**Suggestion:**
```
Land/merge the SI pin to an immutable main SHA in the same release train, or fail CI if the pin is not an ancestor of system-integration main.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Strong red-first tests and comments match the consensus risk</b> (Reported by: xai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 83)

SettledHistory LM isolation, truncated-window finalization, learned-genesis bonding placeholder, DivergenceMonitor threshold tests, and five-condition settled admission tests are the right shape for this change set. Style is verbose but appropriate for A9 consensus code.

**Suggestion:**
```
No change required; keep the taxonomy/fork-choice/oracle specs equally pin-oriented in the portions of the diff not shown here.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Defaulted no-op trait methods on `Casper`/`MultiParentCasper` risk silent misbehavior in future impls</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/casper.rs` (line 92)

`request_block_from_peers` defaults to `Ok(())` and `genesis_block_hash` defaults to `Ok(None)` to spare effect mocks. A future production implementation that forgets to override gets a silently non-functional block fetch (propose retries forever without ever requesting the gap) or a permanently `None` genesis. Defaulting for test convenience trades a compile-time forcing function for a runtime footgun.

**Suggestion:**
```
Move the defaults to the mock implementations (or a `MockCasperExt` helper trait) and keep the production trait methods required.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Shipped genesis verification path looks correct</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/engine/initializing.rs` (line 154)

receive_shipped_genesis checks claimed hash == learned register and content re-hash == register, refuses forged bodies without leaving DAG/store traces, and tests assert LFB stays on the anchor. Write-once genesis_hash_index similarly fails closed on conflicting values. No issue — call out as a positive control for restore integrity.

**Suggestion:**
```
Optional: metric/counter on refused shipped genesis to make peer equivocation on the restore channel visible in ops dashboards.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: DIVERGENCE_ESCALATION_REFUSALS constant lacks a stated relationship to the finalizer cycle period</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 82)

The escalation threshold (10 rising refusals) is justified as topology-independent, but its real-time meaning depends entirely on the finalizer cycle cadence, which this PR also changes (recovery cadence 3s). At 3s cycles, escalation fires ~30s into a divergence; the cited incident ran 14.5 silent minutes. The comment argues against configurability but never states the assumed cycle period the '10' was calibrated against, making future cadence changes silently re-tune the detection latency.

**Suggestion:**
```
Document the assumed finalizer cadence next to the constant, or express the threshold in wall-clock terms derived from the configured cycle.
```

</details>
---

<details>

This is a large, consensus-critical PR (fork choice, clique oracle, slash-evidence taxonomy, finalizer horizon handling, heartbeat recovery) with unusually strong test discipline — red-first specs, truth-table pins, and fixture-driven replays accompany every visible change. The visible engineering is careful, but three areas need human judgment before merge: (1) the write-once genesis-hash register is recorded from a peer response with no visible independent verification, making first-responder poisoning permanent; (2) `held_ancestors` in the finalization sweep silently masks genuine missing-block corruption on non-truncated nodes; and (3) the wholesale demotion of 17 slash verdicts removes economic deterrence for several faults that are provable from the signed block alone (InvalidBlockHash, DeployNotSigned). The fifth settled-admission conjunct may also re-wedge restored nodes for blocks authored by since-unbonded validators. Note this review covers only the first ~15% of a 98-file diff — the fork-choice descent, clique-oracle walk, heartbeat, and proto changes were truncated and are unreviewed here; given they are the core consensus changes, a full human review of those files is mandatory.

The code changes introduce significant modifications to the consensus mechanism, including fork choice, clique oracle, and finality divergence handling. These changes require careful review to ensure they do not introduce new vulnerabilities or logic errors. The new InsertMode::SettledHistory and the narrowing of slash evidence are particularly critical and should be tested extensively.

#### openai (gpt-5.6-sol)

The shown portion contains substantial, well-tested recovery work, but two trust-boundary assumptions require correction before merge: the restore path can permanently record an unverified genesis identity, and the finalization walk treats any missing parent as proven settled ancestry without an explicit restore-horizon discriminator. There are also consistency and observability issues around non-atomic genesis installation and prematurely clearing divergence state. Because this is a consensus-affecting change and the supplied diff is truncated, the remaining files should receive focused review after these issues are addressed.

#### xai (grok-4.5)

This is a large, consensus-critical change set (fork choice, clique-oracle walk, slash taxonomy, finalizer/restore horizon, heartbeat recovery, validation hygiene) that correctly aims at real CI failures (finality divergence, honest-stake burns, restored-node LFB freeze, validation storm). Visible production code is carefully commented and backed by strong targeted tests (held_ancestors horizon, SettledHistory LM isolation, genesis write-once + content verify, DivergenceMonitor streaks, five-condition settled admission, validation-error cap). It still needs expert human review before merge: the slashable-set collapse is an economic-security policy decision that may be broader than the incident requires; held_ancestors can mask non-horizon missing parents if reused broadly; settled admission trusts a hard-coded bonded flag; and divergence remains operator-visible-only while proposing continues. Approve only after consensus owners sign the taxonomy/fork-choice/oracle semantics and the SI pin/soak defaults under the hotter recovery cadence.
</details>

---

spreston8:
> ## Multi-Agent Code Review
> **Verdict:** 🔍 Needs Review (100.0000% agreement)
> 
> **Reviewed by:**
> 
> Provider	Model	Verdict	Confidence
> anthropic	`claude-fable-5`	Needs Review	0.6
> bedrock	`us.amazon.nova-pro-v1:0`	Needs Review	0.9
> openai	`gpt-5.6-sol`	Needs Review	0.91
> xai	`grok-4.5`	Needs Review	0.78
> ### Summary
> This is a large, consensus-critical PR (fork choice, clique oracle, slash-evidence taxonomy, finalizer horizon handling, heartbeat recovery) with unusually strong test discipline — red-first specs, truth-table pins, and fixture-driven replays accompany every visible change. The visible engineering is careful, but three areas need human judgment before merge: (1) the write-once genesis-hash register is recorded from a peer response with no visible independent verification, making first-responder poisoning permanent; (2) `held_ancestors` in the finalization sweep silently masks genuine missing-block corruption on non-truncated nodes; and (3) the wholesale demotion of 17 slash verdicts removes economic deterrence for several faults that are provable from the signed block alone (InvalidBlockHash, DeployNotSigned). The fifth settled-admission conjunct may also re-wedge restored nodes for blocks authored by since-unbonded validators. Note this review covers only the first ~15% of a 98-file diff — the fork-choice descent, clique-oracle walk, heartbeat, and proto changes were truncated and are unreviewed here; given they are the core consensus changes, a full human review of those files is mandatory.
> 
> The shown portion contains substantial, well-tested recovery work, but two trust-boundary assumptions require correction before merge: the restore path can permanently record an unverified genesis identity, and the finalization walk treats any missing parent as proven settled ancestry without an explicit restore-horizon discriminator. There are also consistency and observability issues around non-atomic genesis installation and prematurely clearing divergence state. Because this is a consensus-affecting change and the supplied diff is truncated, the remaining files should receive focused review after these issues are addressed.
> 
> This is a large, consensus-critical change set (fork choice, clique-oracle walk, slash taxonomy, finalizer/restore horizon, heartbeat recovery, validation hygiene) that correctly aims at real CI failures (finality divergence, honest-stake burns, restored-node LFB freeze, validation storm). Visible production code is carefully commented and backed by strong targeted tests (held_ancestors horizon, SettledHistory LM isolation, genesis write-once + content verify, DivergenceMonitor streaks, five-condition settled admission, validation-error cap). It still needs expert human review before merge: the slashable-set collapse is an economic-security policy decision that may be broader than the incident requires; held_ancestors can mask non-horizon missing parents if reused broadly; settled admission trusts a hard-coded bonded flag; and divergence remains operator-visible-only while proposing continues. Approve only after consensus owners sign the taxonomy/fork-choice/oracle semantics and the SI pin/soak defaults under the hotter recovery cadence.
> 
> The code changes introduce significant modifications to the consensus mechanism, including fork choice, clique oracle, and finality divergence handling. These changes require careful review to ensure they do not introduce new vulnerabilities or logic errors. The new InsertMode::SettledHistory and the narrowing of slash evidence are particularly critical and should be tested extensively.
> 
> ### Issues Found
> 🔴 Critical: Fifth conjunct may re-wedge settled-history admission for senders absent from latest messages (Reported by: anthropic, openai)
> **File:** `casper/src/rust/blocks/block_processor.rs` (line 1406)
> 
> The code separately loads settled_admissions, checks it against the budget, inserts the block, and then increments the counter. If block processing can run concurrently, multiple callers can pass the check at the boundary and exceed the storage budget. This weakens the documented security property that attacker-triggered settled-history growth is strictly bounded.
> 
> `seq_below_senders_latest` evaluates to `false` when the sender has no latest message (`None => false`) or when the latest message hash fails lookup. On a truncated node, historic settled blocks are frequently authored by validators that have since unbonded and therefore have no latest-message slot — every such block now fails the fifth conjunct and takes the judged/deferral path. Since settled-history admission was introduced precisely to break restore-era deferral wedging, this conjunct may reintroduce the stall for exactly the blocks it was meant to admit. The truth-table test covers the boolean, but no test in the visible diff covers the 'sender has no latest message but is genuinely settled history' scenario end-to-end.
> 
> **Suggestion:**
> 
> ```
> For senders with no latest-message slot, fall back to a height-based check (block height strictly below the anchor's justification frontier) instead of unconditionally failing, or add an integration-level test demonstrating restored nodes still converge when settled history includes unbonded historic senders.
> ```
> 
> 🔴 Critical: Write-once genesis register can be permanently poisoned by the first restore peer (Reported by: anthropic, openai)
> **File:** `casper/src/rust/engine/initializing.rs` (line 1095)
> 
> During restore, `record_genesis_hash(response.genesis_hash.clone())` is called for any non-empty genesis_hash in the floor-cache response. `record_genesis_hash` is write-once and errors on any different value forever after. If the exchange peer is malicious (or the response arrives from an unauthenticated/compromised channel), the first responder permanently fixes a wrong genesis hash on this node — it then poisons every newly-bonded latest-message seed (which the comment says must be network-uniform), and `insert` will hard-fail or, worse, seed a divergent sentinel that reads as equivocation network-wide. The diff shows no verification of `response.genesis_hash` against the approved block, the bonded set, or a quorum of peers before it is written; the block-body path is verified (re-hash) but the hash-only path is trusted outright.
> 
> The restore path calls record_genesis_hash(response.genesis_hash) before requiring a genesis block or validating its contents. Because the register is write-once, a malformed or malicious response containing an arbitrary non-empty hash can permanently pin the node to the wrong genesis identity. An absent or invalid genesis block only makes genesis_landed false, after which restore still returns success; a later correct response then fails because the register refuses overwrite. Moreover, validating a supplied block against a hash learned from the same response establishes only self-consistency, not that this is the shard's canonical genesis, unless the surrounding exchange independently authenticates that hash.
> 
> **Suggestion:**
> 
> ```
> Authenticate the genesis hash against trusted restore metadata, the approved anchor, or a quorum-backed proof before persisting it. At minimum, require the block to be present and successfully re-hash to the supplied hash before calling record_genesis_hash, and do not complete restore successfully when a first-time genesis identity cannot be verified. Add a test where the first response supplies an invalid hash/block and a later response supplies the correct genesis.
> ```
> 
> 🔴 Critical: Finality divergence is detect-only; proposing continues on a frozen read surface (Reported by: bedrock, xai)
> **File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 72)
> 
> DivergenceMonitor escalates after 10 rising ContainmentHold refusals on one pinned LFB (ERROR + metric) but deliberately does not halt proposing. That matches the incident analysis, yet operators and downstream consumers can keep reading a frozen LFB while the node still emits blocks. Without runbook automation (alert route, auto-resync, API surface exposing diverged()), this can look like a healthy validator while its finalized state is permanently wrong.
> 
> The DivergenceMonitor should be thoroughly tested to ensure it correctly detects and handles finality divergences.
> 
> **Suggestion:**
> 
> ```
> Expose diverged() on admin/status APIs and alert rules in the same PR/docs set. Confirm soak/CI asserts the metric fires on the ucc-i6 shape and that status makes the frozen finalized surface unmistakable.
> ```
> 
> 🟠 Major: InsertMode::SettledHistory (Reported by: bedrock)
> **File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 83)
> 
> The new InsertMode::SettledHistory should be thoroughly tested to ensure it behaves as expected, especially in scenarios involving truncated DAGs and restore operations.
> 
> 🟠 Major: held_ancestors silently treats any missing parent as settled horizon (Reported by: xai)
> **File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 674)
> 
> The finalization marking walk now terminates on unheld parents instead of erroring. That fixes restored-node LFB freeze, but the same helper cannot distinguish 'below restore truncation' from 'transient MissingBlock / corruption / partial DAG'. If record_directly_finalized is ever invoked while live parents are merely not yet fetched, ancestry can be marked finalized without those parents ever being examined or marked, leaving an incomplete finalized set relative to peers who hold the full chain.
> 
> **Suggestion:**
> 
> ```
> Constrain call sites so held_ancestors is only used when the node is known-truncated (restore/LFS path) or after an explicit horizon check. Alternatively pass a horizon hash/height and only suppress MissingBlock for parents at-or-below that floor; surface MissingBlock above it.
> ```
> 
> 🟠 Major: All missing parents are treated as below the restore horizon (Reported by: openai)
> **File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 684)
> 
> held_ancestors stops at every unheld parent without proving that the edge crosses a trusted restore boundary. The method is then used by record_directly_finalized globally, including for ordinary DAGs. The added test even constructs a Normal block with an arbitrary missing parent and treats it as settled. A missing live merge parent, incomplete dependency, or malformed admitted block is therefore indistinguishable from genuinely truncated finalized ancestry. This can silently finalize held descendants while omitting ancestry that was never established as settled, and can leave that parent unmarked if it arrives later.
> 
> **Suggestion:**
> 
> ```
> Gate horizon termination on explicit restore metadata or a persisted settled-boundary proof. On a non-truncated node, or when the parent is not proven below the anchor, preserve the existing MissingBlock behavior. Tests should cover both a proven sub-horizon parent that terminates successfully and an unheld live parent that still defers finalization.
> ```
> 
> 🟠 Major: `held_ancestors` in the finalization sweep silently masks genuine missing-block corruption on non-truncated nodes (Reported by: anthropic)
> **File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1425)
> 
> `record_directly_finalized` now uses `held_ancestors` unconditionally. On a truncated (restored) node this is correct — an unheld parent is the horizon. But on a fully-synced ceremony node a missing parent is data loss/corruption, and the walk now silently terminates there instead of surfacing an error: the subtree below the gap never receives finalized marks, `is_finalized` returns false for genuinely finalized ancestry, and finalized-effects callbacks are silently skipped for those blocks. The old behavior (error) was wrong for restored nodes but at least loud for corrupted full nodes; the new behavior is silent in both cases.
> 
> **Suggestion:**
> 
> ```
> Gate the lenient walk on the node actually being truncated (e.g., presence of a restore anchor / learned genesis without a held height-0 block), or at minimum emit a WARN + metric when a parent is referenced-but-unheld during the finalization sweep on a node that completed a full sync.
> ```
> 
> 🟠 Major: Slash Evidence Narrowing (Reported by: bedrock)
> **File:** `casper/src/rust/block_status.rs` (line 237)
> 
> The narrowing of slash evidence to the equivocation class should be carefully reviewed to ensure it does not introduce new edge cases or vulnerabilities.
> 
> 🟠 Major: Slashable set narrowed from 19 variants to 2 — economic security tradeoff (Reported by: xai)
> **File:** `casper/src/rust/block_status.rs` (line 237)
> 
> Only AdmissibleEquivocation and IgnorableEquivocation remain slashable. Many faults that are arguably objective from the signed block alone (e.g. InvalidBlockHash, DeployNotSigned, InvalidSequenceNumber, InvalidParents) no longer mint evidence. The motivation (view-relative honest-stake burns in run 32588262605) is sound for JustificationRegression / UnauthorizedSlashDeploy / neglected-* families, but the demotion is blanket. Under A9 atomic activation this permanently changes attacker economics: invalid blocks are still rejected, but there is no stake penalty for a large class of Byzantine payloads.
> 
> **Suggestion:**
> 
> ```
> Before merge, explicitly ratify per-variant which demotions are temporary vs permanent. Prefer keeping purely syntactic, admission-order-free checks (hash/signature/deploy-sign) slashable now, and only demote the proven view-relative family. Track re-promotion with tests that pin 'objective from bytes alone' so they cannot silently stay demoted.
> ```
> 
> 🟠 Major: Mass demotion of slashable verdicts (19 → 2) removes economic deterrence for several objectively provable faults (Reported by: anthropic)
> **File:** `casper/src/rust/block_status.rs` (line 250)
> 
> The rationale (evidence must be attributable identically from the signed block alone) is sound for view-relative verdicts like JustificationRegression and NeglectedEquivocation, but several demoted variants ARE provable from the signed block alone: `InvalidBlockHash` (content does not hash to the claimed hash), `DeployNotSigned` (deploy signature invalid), `InvalidSequenceNumber` and `InvalidBlockNumber` (checkable against the block's own cited justifications/parents when those are included), `InvalidShardId`. Demoting these makes producing malformed-but-signed blocks economically free, weakening the spam/DoS cost model. The comment acknowledges per-variant re-promotion is possible, but shipping all demotions at once as an atomic consensus change means the deterrence gap is live until then.
> 
> **Suggestion:**
> 
> ```
> Keep the provable-from-the-block-alone subset (at minimum InvalidBlockHash and DeployNotSigned) slashable in this PR, or document the concrete interim mitigation (rate limits / peer scoring) that replaces the economic deterrent for these variants.
> ```
> 
> 🟠 Major: try_admit_settled hard-codes solicited_by_bonded=true (Reported by: xai)
> **File:** `casper/src/rust/blocks/block_processor.rs` (line 1403)
> 
> admit_as_settled's third conjunct is documented as 'citation from a bonded sender', but try_admit_settled always passes true. Safety then depends entirely on callers only invoking this after bonded solicitation bookkeeping. Any future call-site mistake turns settled-history into an unbonded storage/admission path (bounded by budget, but still skips validation).
> 
> **Suggestion:**
> 
> ```
> Thread the actual bonded-solicitation boolean from settled_solicitations membership into admit_as_settled instead of a literal true. Add a negative test where an untracked hash cannot take the settled path even if height/budget/seq match.
> ```
> 
> 🟠 Major: Genesis block store and DAG updates are not atomic (Reported by: openai)
> **File:** `casper/src/rust/engine/initializing.rs` (line 186)
> 
> receive_shipped_genesis writes the block body to block_store before inserting it into the DAG. If DAG insertion fails, the function returns an error but leaves the block in the block store. This contradicts the stated refusal invariant that an unsuccessful shipped genesis leaves no trace and can make subsequent retries observe a partially installed genesis.
> 
> **Suggestion:**
> 
> ```
> Validate all insertion preconditions before writing either store, or perform the block-store and DAG changes in a transaction with rollback. Add a failure-injection test for DAG insertion after put_block_message and assert that neither store retains the block.
> ```
> 
> 🟠 Major: Aggressive recovery cadence defaults (15s/12s → 3s/3s) combined with all-validator heartbeat may amplify traffic at scale (Reported by: anthropic, xai)
> **File:** `casper/src/rust/casper_conf.rs` (line 338)
> 
> Stale-LFB recovery is now open to every bonded validator (leader-only lanes deleted) and the default cadence drops to 3s for both self-propose cooldown and stale-recovery interval, with frontier chase raised 0 → 20. During a stall, this means every bonded validator can propose/chase every 3 seconds. The PR asserts backpressure bounds churn to the recovery cadence, but on large validator sets N proposals per 3s window is materially more block and gossip traffic than the previous regime; the visible diff shows CI validation on small shards only.
> 
> Stale-LFB recovery is open to every bonded validator and paced only by time, with much hotter defaults. This is the stall fix, but it can multiply propose/fetch churn and bandwidth on large validator sets or slow IO nodes, especially with frontier_chase_max_lag=20.
> 
> **Suggestion:**
> 
> ```
> Keep the embedded-defaults pin, and confirm soak/merge-recovery jobs measure propose rate, fetch QPS, and CPU under the new cadence on CI-shaped topology before calling the defaults final.
> ```
> 
> 🟡 Minor: `genesis_hash()` fallback picks the min-hash height-0 block without uniqueness check (Reported by: anthropic)
> **File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 891)
> 
> When the register is empty, `genesis_hash()` derives genesis as `height_map.get(&0).min()`. If more than one height-0 block ever exists in the map (e.g., an invalid block admitted at height 0, or a bug elsewhere), the function silently picks the lexicographically smallest, potentially returning a non-genesis hash that then seeds newly-bonded latest-message slots — precisely the network-uniformity violation this PR fixes. Also note `genesis_hash()` does not take `global_lock`, so it can interleave with a concurrent `record_genesis_hash`.
> 
> **Suggestion:**
> 
> ```
> Assert/log when `height_map[0]` has more than one entry, and consider taking the global read lock for a consistent read of register + height map.
> ```
> 
> 🟡 Minor: Validation Failure Handling (Reported by: bedrock)
> **File:** `casper/src/rust/blocks/block_processor.rs` (line 114)
> 
> The validation failure handling logic should be reviewed for clarity and consistency with existing error handling patterns.
> 
> 🟡 Minor: Validation-error quarantine reuses missing-dependency quarantine duration (Reported by: xai)
> **File:** `casper/src/rust/blocks/block_processor.rs` (line 1218)
> 
> note_validation_failure stamps validation_error_quarantine_until with missing_dependency_quarantine_ms(). The failure modes differ (hard Err loop vs dependency wait); coupling them means tuning dependency fetch backoff also changes hard-fail purge damping and vice versa.
> 
> **Suggestion:**
> 
> ```
> Give validation errors their own env default (e.g. F1R3_VALIDATION_ERROR_QUARANTINE_MS) or document the intentional shared knob next to both constants.
> ```
> 
> 🟡 Minor: Quarantine timing uses wall clock; clock steps can void or extend quarantines (Reported by: anthropic)
> **File:** `casper/src/rust/blocks/block_processor.rs` (line 1246)
> 
> `note_validation_failure`, `is_validation_failure_quarantined`, and the sweep all use `SystemTime::now().duration_since(UNIX_EPOCH).unwrap_or(0)`. A backwards clock step (NTP correction — explicitly relevant given this PR's own restored-joiner clock-skew fix in the slashing path) can make active quarantines appear far in the future (blocking harvest for hours) or immediately expired (restarting the hot loop the mechanism exists to bound). Additionally, `unwrap_or(0)` on a pre-epoch clock silently disables quarantine entirely.
> 
> **Suggestion:**
> 
> ```
> Use a monotonic clock (`Instant`) for quarantine bookkeeping, matching the intent of pacing rather than absolute time.
> ```
> 
> 🟡 Minor: TOCTOU window in `sweep_orphaned_validation_error_attempts` (Reported by: anthropic)
> **File:** `casper/src/rust/blocks/block_processor.rs` (line 1330)
> 
> The sweep collects orphaned hashes under one lock acquisition, releases it, then re-acquires the lock to remove them. A block can re-enter the buffer (fresh peer delivery) between the two acquisitions and have its just-restarted attempt counter incorrectly removed. Effect is benign (the counter restarts from zero, extending retries slightly), but the same double-acquire pattern is repeated from the missing-dependency sweep and could be done atomically.
> 
> **Suggestion:**
> 
> ```
> Hold the attempts lock across the classify-and-remove pass, or use `retain` with the buffer-membership check inline.
> ```
> 
> 🟡 Minor: DivergenceMonitor uses Mutex::expect — poison becomes a finalizer panic (Reported by: xai)
> **File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 130)
> 
> on_advance/on_containment_hold call state.lock().expect(...). A poisoned mutex (prior panic while holding the lock) will panic the finalization path rather than degrade to 'hold this cycle'.
> 
> **Suggestion:**
> 
> ```
> Match other casper locks: map poison to a logged hold/reset, or use a non-poisoning concurrency primitive if panics in this path must never wedge finalization.
> ```
> 
> 🟡 Minor: Divergence state is cleared before finalization advancement succeeds (Reported by: openai)
> **File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 425)
> 
> divergence_monitor.on_advance() is called as soon as floor_of_view returns Advance, before the new LFB is actually recorded and its effects complete. If the subsequent update fails, the LFB remains pinned but the refusal streak and diverged flag have already been cleared. Repeated failures can therefore suppress or delay the intended divergence alarm.
> 
> **Suggestion:**
> 
> ```
> Call on_advance only after update_last_finalized_block and all required persistence/effect work succeed. Merely deriving an advancing candidate should not clear an alarm tied to the persisted LFB.
> ```
> 
> ⚪ Suggestion: SYSTEM_INTEGRATION_REF still documents branch SHAs pending main land (Reported by: xai)
> **File:** `.github/workflows/merge-recovery-soak.yml` (line 140)
> 
> Pin history correctly records d94fffd as PR #126 branch-with-main-merged and repeats 're-pin to merged main SHA once #126 lands'. Merging the node PR while SI is still a moving branch pin risks CI drift if that branch is force-pushed or superseded.
> 
> **Suggestion:**
> 
> ```
> Land/merge the SI pin to an immutable main SHA in the same release train, or fail CI if the pin is not an ancestor of system-integration main.
> ```
> 
> ⚪ Suggestion: Strong red-first tests and comments match the consensus risk (Reported by: xai)
> **File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 83)
> 
> SettledHistory LM isolation, truncated-window finalization, learned-genesis bonding placeholder, DivergenceMonitor threshold tests, and five-condition settled admission tests are the right shape for this change set. Style is verbose but appropriate for A9 consensus code.
> 
> **Suggestion:**
> 
> ```
> No change required; keep the taxonomy/fork-choice/oracle specs equally pin-oriented in the portions of the diff not shown here.
> ```
> 
> ⚪ Suggestion: Defaulted no-op trait methods on `Casper`/`MultiParentCasper` risk silent misbehavior in future impls (Reported by: anthropic)
> **File:** `casper/src/rust/casper.rs` (line 92)
> 
> `request_block_from_peers` defaults to `Ok(())` and `genesis_block_hash` defaults to `Ok(None)` to spare effect mocks. A future production implementation that forgets to override gets a silently non-functional block fetch (propose retries forever without ever requesting the gap) or a permanently `None` genesis. Defaulting for test convenience trades a compile-time forcing function for a runtime footgun.
> 
> **Suggestion:**
> 
> ```
> Move the defaults to the mock implementations (or a `MockCasperExt` helper trait) and keep the production trait methods required.
> ```
> 
> ⚪ Suggestion: Shipped genesis verification path looks correct (Reported by: xai)
> **File:** `casper/src/rust/engine/initializing.rs` (line 154)
> 
> receive_shipped_genesis checks claimed hash == learned register and content re-hash == register, refuses forged bodies without leaving DAG/store traces, and tests assert LFB stays on the anchor. Write-once genesis_hash_index similarly fails closed on conflicting values. No issue — call out as a positive control for restore integrity.
> 
> **Suggestion:**
> 
> ```
> Optional: metric/counter on refused shipped genesis to make peer equivocation on the restore channel visible in ops dashboards.
> ```
> 
> ⚪ Suggestion: DIVERGENCE_ESCALATION_REFUSALS constant lacks a stated relationship to the finalizer cycle period (Reported by: anthropic)
> **File:** `casper/src/rust/engine/multi_parent_casper/finalization_runner.rs` (line 82)
> 
> The escalation threshold (10 rising refusals) is justified as topology-independent, but its real-time meaning depends entirely on the finalizer cycle cadence, which this PR also changes (recovery cadence 3s). At 3s cycles, escalation fires ~30s into a divergence; the cited incident ran 14.5 silent minutes. The comment argues against configurability but never states the assumed cycle period the '10' was calibrated against, making future cadence changes silently re-tune the detection latency.
> 
> **Suggestion:**
> 
> ```
> Document the assumed finalizer cadence next to the constant, or express the threshold in wall-clock terms derived from the configured cycle.
> ```
> 
> ---
> View individual reviewer assessments
> #### anthropic (claude-fable-5)
> This is a large, consensus-critical PR (fork choice, clique oracle, slash-evidence taxonomy, finalizer horizon handling, heartbeat recovery) with unusually strong test discipline — red-first specs, truth-table pins, and fixture-driven replays accompany every visible change. The visible engineering is careful, but three areas need human judgment before merge: (1) the write-once genesis-hash register is recorded from a peer response with no visible independent verification, making first-responder poisoning permanent; (2) `held_ancestors` in the finalization sweep silently masks genuine missing-block corruption on non-truncated nodes; and (3) the wholesale demotion of 17 slash verdicts removes economic deterrence for several faults that are provable from the signed block alone (InvalidBlockHash, DeployNotSigned). The fifth settled-admission conjunct may also re-wedge restored nodes for blocks authored by since-unbonded validators. Note this review covers only the first ~15% of a 98-file diff — the fork-choice descent, clique-oracle walk, heartbeat, and proto changes were truncated and are unreviewed here; given they are the core consensus changes, a full human review of those files is mandatory.
> 
> #### bedrock (us.amazon.nova-pro-v1:0)
> The code changes introduce significant modifications to the consensus mechanism, including fork choice, clique oracle, and finality divergence handling. These changes require careful review to ensure they do not introduce new vulnerabilities or logic errors. The new InsertMode::SettledHistory and the narrowing of slash evidence are particularly critical and should be tested extensively.
> 
> #### openai (gpt-5.6-sol)
> The shown portion contains substantial, well-tested recovery work, but two trust-boundary assumptions require correction before merge: the restore path can permanently record an unverified genesis identity, and the finalization walk treats any missing parent as proven settled ancestry without an explicit restore-horizon discriminator. There are also consistency and observability issues around non-atomic genesis installation and prematurely clearing divergence state. Because this is a consensus-affecting change and the supplied diff is truncated, the remaining files should receive focused review after these issues are addressed.
> 
> #### xai (grok-4.5)
> This is a large, consensus-critical change set (fork choice, clique-oracle walk, slash taxonomy, finalizer/restore horizon, heartbeat recovery, validation hygiene) that correctly aims at real CI failures (finality divergence, honest-stake burns, restored-node LFB freeze, validation storm). Visible production code is carefully commented and backed by strong targeted tests (held_ancestors horizon, SettledHistory LM isolation, genesis write-once + content verify, DivergenceMonitor streaks, five-condition settled admission, validation-error cap). It still needs expert human review before merge: the slashable-set collapse is an economic-security policy decision that may be broader than the incident requires; held_ancestors can mask non-horizon missing parents if reused broadly; settled admission trusts a hard-coded bonded flag; and divergence remains operator-visible-only while proposing continues. Approve only after consensus owners sign the taxonomy/fork-choice/oracle semantics and the SI pin/soak defaults under the hotter recovery cadence.
> 
> _Generated by Multi-Agent Review System_

Triage of all 23 findings, each verified against the code rather than the diff excerpt. Note the bot reviewed ~15% of the diff by its own account; the truncated 85% (fork-choice descent, oracle walk, heartbeat, proto) received a full manual consensus review earlier — the reviewer outline and doc-currency ledger from that pass back the PR body's reading order.

**Fixed in `bc3ba16a2` (review hardenings):**
- Validation-error quarantine now paces on a monotonic clock (`Instant`) — the wall-clock-step finding was correct, and the `unwrap_or(0)` hazard goes with it.
- The finalization sweep WARNs when it skips an unheld parent — routine while a restored node's sweep first crosses its horizon, alarm-grade on a full node. (The three `held_ancestors` findings share this root; the invariant that makes the lenient walk safe is that admission never inserts a block whose parents aren't held — only a restore can create that edge — but corruption visibility was genuinely lost, hence the WARN.)
- The divergence streak clears only after the adoption effect persists (the `on_advance` ordering finding).
- `DivergenceMonitor` recovers from mutex poison instead of panicking the finalizer.
- The settled-admission budget slot is reserved atomically (`fetch_update`, returned on insert failure), and the solicitation boolean is threaded into `admit_as_settled` instead of a literal `true` (the one-shot `take_settled_solicitation` gate at the top remains the enforcement point).
- Refused shipped genesis increments `restore.genesis.refused` (the suggested ops visibility).

**Fixed in `7b7fd0b91` (the strongest finding — thank you):** a sender with no latest-message slot now PASSES the fifth settled-admission conjunct. Deep settled history is routinely authored by since-unbonded validators; refusing them re-wedged exactly the restore gaps the door exists to close. Live material wearing a sub-anchor height always carries a live latest message, so the pollution guard is done entirely by the seq comparison; a bonded citer vouching for a no-slot author stays priced by the admission budget. Red-first: `settled_admission_admits_an_unbonded_historic_author`.

**Fixed in `70c605164`:** system-integration is now pinned to the immutable main merge SHA (`8b4da0f`, tree-identical to the previous branch pin — the CI green ×3 evidence carries over). This was the SI-pin suggestion.

**By design (no change):**
- *Genesis-register "poisoning"*: the floor-cache exchange IS the LFS bootstrap channel — the same peer the node already trusts for the approved anchor and the entire restored window. First-responder poisoning of the genesis hash is exactly first-responder poisoning of the anchor, the pre-existing trust-on-first-use assumption of LFS restore itself; the hash rides that channel deliberately (stated on the proto field). Quorum-authenticating genesis alone would not change the restore trust model. The block-body path's double verification plus the new refusal metric are the enforcement inside that model.
- *Divergence detect-only*: deliberate — the wedged node's proposals followed the estimator and stayed valid throughout the observed incident; halting would shrink the honest proposer set exactly when certification needs it. The harness treats the ERROR line as forbidden (alarm channel). Exposing `diverged()` on the status API is filed as a follow-up, not a merge gate.
- *Non-atomic genesis install*: the "leaves no trace" invariant is about the refusal path (verification fails ⇒ no writes), which holds. A DAG-insert failure after `put_block_message` leaves a content-addressed body with no DAG entry — inert, re-verified on the next attempt.
- *Shared quarantine knob*: deliberate and documented at both the code site and the env-var table in `docs/node/README.md`; the two loops share pacing intent.
- *Recovery cadence at scale*: agreed this is soak's question; the defaults are pinned by the embedded-defaults test and the merge-recovery soak measures the hot cadence on CI-shaped topology. Current validation is small-shard by the nature of an unreleased node.

**Held as-is, follow-up filed (the policy finding):** keeping the slashable set at the equivocation class — including demoting `InvalidBlockHash`/`DeployNotSigned` — is the deliberate conservative posture the spec amendment ratifies: an unreleased node whose evidence system just burned FT −18.55 of honest stake ships the provably-safe set first. "Provable from bytes alone" is the exact re-promotion bar the code comment documents, and those two variants are the first candidates (the attribution subtlety to discharge: whether the signature binds the claimed hash such that evidence is attributable without local state). Validity enforcement is unchanged; demotion costs economics only.

Minor items not listed are either fixed above or noted in code comments; the two content-free entries ("should be tested") are covered by the existing red-first suites the review itself calls out as strong.

Issue #341, opened 2026-08-25
Validator loses its bond at an epoch boundary before its own propose (joiner preflight failure, 4 consecutive runs)


jeffrey-l-turner:
## Symptom

`test_joiner_self_proposes_at_epoch_boundary` fails reproducibly in the soak integration preflight:

```
AssertionError: V4 unexpectedly dropped from bonds before V4's propose (latest LFB #7)
```

The bonds list at the assertion holds three validators; V4 is absent before its own propose at the epoch boundary.

## Occurrences

The same assertion failed in four consecutive preflight runs across different refs, so it is not stack-specific:

| Run | Ref | Date |
|---|---|---|
| 32668023403 | `master` (scheduled) | 2026-08-23 21:36Z |
| 32757906436 | stack top `470ddf812` | 2026-08-24 17:39Z |
| 32780706977 | `fix/key-contention-base-bias` `3f488aeb3` | 2026-08-24 21:40Z |
| 32808149007 | stack top `e31ecde69` (contains #318) | 2026-08-25 04:13Z |

LFB at failure varies (#7, #8), so the drop tracks the epoch transition, not a fixed height.

## New evidence from run 32808149007

The forbidden-log check captured WARN entries near the epoch boundary that did not appear before #318 merged:

```
[RecordingInvalidBlock] level=WARN "Recording invalid block 22349445cf... for Justification..."
[RecordingInvalidBlock] level=WARN "Recording invalid block 4d4a03c639... for Justification..."
```

A node records other validators' blocks as invalid on a justification ground around the same window in which V4 loses its bond. If V4's block is judged invalid on a justification technicality at the boundary and a slash or bond-removal follows, that would produce exactly this assertion.

## Suspected area

Epoch-transition bonding logic and justification validation around the boundary (`casper` epoch handling; possibly interacting with the #318 heartbeat/recovery cadence, which increases propose frequency at the boundary). Related but distinct from #306 (history availability): this failure persists in runs where the #306 storm is much reduced.

## Acceptance

- `test_joiner_self_proposes_at_epoch_boundary` passes in the soak integration preflight on three consecutive runs.
- Root cause documented: why a bonded validator's bond disappears before its first propose at an epoch boundary, and why blocks are recorded invalid for justification grounds in that window.

jeffrey-l-turner:
## Prior reproduction campaign — this is a known, long-hunted flake

[...]

PR #343, opened 2026-08-25
fix issue-341  validator-loses-bond-epoch-boundary

jeffrey-l-turner:
## Summary

Fixes the mechanism behind issue #341 (a freshly-bonded validator's bond vanishing around an epoch boundary) at the merge layer, with a deterministic reproduction, an end-shape regression guard, and a completed TDD plan documenting what is and is not covered.

## The mechanism (B1)

`partition_base_conflicts` protects the BASE lineage's committed content, and `sig_settled_in_base` protects effects already in the base state — but nothing protected a **finalized sibling**: under multi-parent finality a finalized block can sit off the base's main-parent chain with its effects absent from the base state. Its chains entered cost adjudication like any scope chain and could LOSE; recovery then purged the finalized effects ("finalized canonical wins"), vanishing settled state. Observed twice in production shapes: the bridge-admin registry flake and the #341 bond loss.

The fix mirrors the base protection: `dag_merger::merge` splits scope chains by carrier finalization; an ordinary chain conflicting with the finalized chains' combined event log is rejected deterministically (same record/buffer/recovery path as a base conflict); finalized chains re-join the merge set exempt from cost adjudication. Two mutually conflicting finalized chains (a finality-safety violation upstream) fall through to ordinary adjudication rather than wedging the merge.

`casper/tests/merging/finalized_protection_spec.rs` reproduces the bug deterministically: pre-fix, a finalized sibling's cheap deploy was rejected in favor of an expensive contender.

## The end-shape guard (B2)

`a_finalized_bond_survives_an_epoch_boundary_merge` (in the bonding spec) sets PoS `epoch_length: 4` — walking into the close-block epoch-change path the default test genesis deliberately avoids — finalizes a bond, then drives sibling blocks and multi-parent merges across two epoch boundaries and asserts the bond survives.

**Honesty note:** this test never went RED, and a control run against the pre-fix tree also passes — the deterministic boundary shape alone does not reproduce the production loss (consistent with the 19-variant SI reproduction campaign concluding the heartbeat propose race is required). The soak preflight (3s heartbeat cadence) remains the arbiter for the production symptom; this PR closes the mechanism-level hole and adds the guards.

## TDD plan

`docs/tdd-plans/epoch-boundary-bond-loss-2026-08-25T07-30-00Z.md` records all three cycles, including the B3 deviation (JustificationRegression is sequence-number-based; bond content is not an input) and discovered follow-ups: window-rule exemption for finalized carriers, and the rejection-expansion interaction.

## Verification

- `finalized_protection_spec` RED pre-fix, GREEN post-fix.
- Full casper release suite: 314 lib + 904 integration tests pass.
- Pre-fix control run of the B2 guard documented in the plan.

Refs #341

## Review remediation (multi-agent review, 2026-08-25)

The review found 2 critical and 3 major issues; all are addressed:

| Finding | Resolution |
|---|---|
| **Merge outcome depended on node-local finality state** (`dag.finalized_blocks_set` is view-relative; proposer and validator could compute different rejection sets) | Classifier redesigned. The settled test is now a caller-supplied `sig_settled_in_floor` probe — the same settled definition as the `assert_no_settled_rejection` tripwire, evaluated over the merging block's **settled floors** (derived from its frozen justification snapshot), so replay answers identically. Carrier height alone was also rejected as a classifier: it cannot discriminate a finalized sibling from a silent validator's dead below-floor sibling, and collided with `batch2/merge_window_spec` (which stays green). |
| **Window rule ran before the protection** (a settled carrier's closed-window chain could still be rejected) | The settled partition is hoisted before `split_window_closed_chains`; settled chains are exempt. New spec test: `a_settled_carriers_closed_window_deploy_is_not_window_rejected`, with an ordinary closed-window chain asserted still window-rejected. |
| **Rejection expansion could still reject settled content** | Settled chains (the pinned set) are exempt from lineage rejection-expansion. |
| **Mutually conflicting settled chains adjudicated silently** | Settled-versus-settled and settled-versus-base conflicts now log a `finality-safety violation` error for evidence before falling back to ordinary adjudication (liveness over wedging); the caller-side tripwire hard-fails the merge if settled effects are in fact dropped. |
| (bonus) | The settled set now populates `pinned`, so the over-filled single-value-cell keep-one prefers settled writers. |

Verification after remediation: both `finalized_protection_spec` tests green, `merge_window_spec` green, full casper release suite green (314 lib + 906 integration).


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

This PR adds a merge-layer protection so that deploy chains whose carrier block is FINALIZED are exempt from cost adjudication and win deterministically against conflicting scope chains, mirroring the existing base-lineage protection. The implementation is clean, deterministic (partition + combined event log + re-join + sort), and comes with a strong mechanism-level test (B1) that went properly RED, plus an end-to-end epoch-boundary test and an unusually candid TDD plan. However, the plan's own notes document at least two remaining paths where finalized content can still be rejected (the deploy-window rule running before the partition, and post-fold-in lineage rejection expansion), the interaction with the pre-existing base-conflict partition is untested for finalized-vs-base conflicts, and the end-to-end test never reproduced the actual #341 failure — the production repro appears to require a heartbeat race not covered here. The change is safe and an improvement, but a maintainer familiar with the merge/finality invariants should weigh in on whether the acknowledged gaps must be closed before this can be considered the fix for #341, hence needs_review rather than approval.

The change addresses the observed cost-adjudication mechanism and adds useful regression coverage, but it introduces a consensus-sensitive dependency on mutable node-local finality state and does not fully enforce its stated finalized-content invariant. Finalized chains can still be rejected by window/dependency processing or by conflicts with other finalized chains. These semantics need architectural review and broader invariant tests before merge.

The merge-layer fix correctly mirrors base-content protection for finalized carriers: split by finalized_blocks_set, deterministically reject scope conflicts against the combined finalized event log, re-join finalized chains, and fold those rejects into the normal rejection path. The dedicated unit test is the real safety net and matches the #341/bridge-admin mechanism; the epoch-boundary bonding test is valuable but non-regressing on its own. Before treating #341 as fully closed, address (or consciously defer with tests) the two residual reject paths called out in the TDD notes—window rejection before this partition, and post-fold rejection expansion onto finalized-carrier dependents—since either can still admit loss of settled state.

The code changes address the issue of validators losing bonds at epoch boundaries by ensuring that finalized content is never rejected during merge adjudication. However, there are some areas for improvement, such as handling conflicts between two finalized chains and reducing code duplication in test helpers. Overall, the changes are functional and safe to merge with minor adjustments.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Finalized chains remain rejectable outside cost adjudication</b> (Reported by: bedrock, openai, xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1182)

The TDD plan itself records that the deploy-window rule runs before this finalized-carrier partition. A chain whose carrier is already finalized can still be window-rejected and never reach this protection path. That weakens the stated invariant ('finalized carrier content is committed and must never be rejected') and could still drop settled effects (including bonds) under timing/window edge cases.

---

The code handles conflicts between finalized chains and ordinary chains, but it does not explicitly handle conflicts between two finalized chains. This could lead to issues if two finalized chains conflict with each other.

---

The new partition occurs only after earlier rejection processing and protects finalized chains only from the cost adjudication addressed here. As the accompanying TDD document notes, the deploy-window rule runs before this partition, while lineage-based rejection expansion runs afterward. Consequently, a finalized chain can still be rejected because its window is closed or because it depends on another rejected chain. That contradicts the stated invariant that finalized content must never be rejected and may leave alternate paths for the same settled-state loss.

**Suggestion:**
```
Apply the same finalized-carrier exemption (or run an equivalent partition) before window-based rejection, or explicitly filter window rejects so chains with source_block_hash in finalized_blocks_set are retained. Add a focused unit test that places a finalized-carrier chain in a closed window and asserts it is not rejected.
```

</details>

<details>
<summary><b>:red_circle: Critical: Post-fold-in rejection expansion can still reject finalized content</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1667)

Finalized conflicts are folded into resolved.rejected with base conflicts, after which lineage/dependency rejection expansion runs. A non-finalized loser can cause dependent chains to be rejected even when those dependents are carried by finalized blocks. That is another path to 'finalized canonical wins' purges and state vanishing, adjacent to the #341 failure mode.

---

The plan notes that rejection-expansion over block lineage runs AFTER the fold-in of finalized_conflicting: a chain that DEPENDS on a rejected chain is rejected even if its own carrier is finalized. This is a second acknowledged path by which finalized (committed) content can be rejected by the merge, contradicting the stated invariant. It is deferred without a covering test.

**Suggestion:**
```
Add a behavior/test for a finalized chain depending on a rejected chain, and either exempt finalized carriers from lineage expansion or document why the transitive rejection is safe.
```

</details>

<details>
<summary><b>:orange_circle: Major: Conflicting finalized chains can still be rejected</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1192)

All finalized chains are reinserted into `actual_seq_all`, so mutually conflicting finalized chains proceed through ordinary adjudication and one may be rejected. This directly violates the new committed-content invariant. Treating a finality-safety violation as ordinary competition also hides evidence of a serious upstream consistency failure and can produce a state that silently discards finalized effects.

**Suggestion:**
```
Detect conflicts among finalized chains explicitly. Return a dedicated finality-safety error or quarantine the merge rather than cost-adjudicating finalized content. If liveness fallback is an intentional protocol rule, it should be specified precisely and covered by a test demonstrating the required deterministic behavior.
```

</details>

<details>
<summary><b>:orange_circle: Major: Deploy-window rejection runs before finalized-carrier protection</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1196)

As acknowledged in the TDD plan's own 'discovered' notes, the deploy-window rule executes BEFORE the new finalized partition. A chain whose carrier block is FINALIZED but whose deploy window has closed can still be window-rejected, violating the very invariant this PR pins ('content whose carrier block is FINALIZED must never be rejected'). This leaves an open path for the #341 bond-loss shape.

---

The new finalized partition operates on `actual_seq_all` after the existing base-conflict partition has already removed `base_conflicting` chains. If a finalized-carrier chain conflicts with the base's committed content, it appears it would be rejected as an ordinary base conflict before finalized protection ever sees it — rejecting finalized effects via the base path. The code comment argues two finalized chains conflicting is an upstream finality violation, but base-vs-finalized-sibling conflicts are exactly the multi-parent scenario this fix targets and this ordering is not tested.

**Suggestion:**
```
Either exempt finalized-carrier chains from the window rule (the plan's candidate B4), or move the finalized partition ahead of the window rule and add a regression test for a finalized carrier with a closed window.
```

</details>

<details>
<summary><b>:orange_circle: Major: Merge result now depends on node-local finalization state</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1200)

Classifying chains through `dag.finalized_blocks_set` makes merge adjudication depend on the node's current DAG/finality view rather than solely on immutable data committed by the block being merged. A carrier may be considered finalized on one node but not another, or may become finalized between initial execution and later replay. Those executions can then choose different conflict winners and derive different rejected sets or post-state hashes for the same merge inputs, creating a consensus and validation divergence risk.

**Suggestion:**
```
Derive protected carriers from finality evidence or a finalized frontier committed to the merge context/block, or otherwise prove and document that `finalized_blocks_set` is identical for every validator replaying this merge. Add a test that executes the same merge against DAG views with different later finality knowledge and verifies identical output.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Quadratic accumulation when combining finalized event logs</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1204)

`EventLogIndex::combine(&finalized_log, &chain.event_log_index)` builds a fresh combined index on every iteration, copying the accumulated log each time. With many finalized-carrier chains this is O(n²) in total event-log size on the hot merge path.

**Suggestion:**
```
If EventLogIndex supports an in-place/extend-style merge, use it; otherwise consider a fold that reuses the accumulator's internal sets rather than reallocating per chain.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Mutually conflicting finalized chains silently fall through to cost adjudication</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1233)

When two finalized chains conflict with each other, both are re-joined to the merge set and one can still lose ordinary adjudication — i.e., finalized content is rejected. The code comment calls this a deliberate liveness fallback for an upstream finality-safety violation, but there is no runtime signal when it occurs.

**Suggestion:**
```
Detect pairwise conflicts among `finalized_committed` and emit a `tracing::warn!` (or error metric) so an actual finality-safety violation is observable in production rather than silently adjudicated.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: B2 end-to-end test never went RED — it does not reproduce #341</b> (Reported by: anthropic)</summary>

**File:** `casper/tests/batch1/multi_parent_casper_bonding_spec.rs` (line 128)

Per the TDD plan, `a_finalized_bond_survives_an_epoch_boundary_merge` also passes against the pre-fix tree ('DEVIATION: never went RED'). It is a characterization guard, not a reproduction of the bug, and the plan itself concludes the production failure needs an additional ingredient (heartbeat propose race / LFS-restore gaps). The PR may therefore not actually close #341 as titled; only the mechanism-level unit spec (B1) went red.

**Suggestion:**
```
Make it explicit in the PR description that #341's production repro (the ~33%-failing heartbeat-race variant) is not covered by this change, and keep the issue open or track the remaining race in a follow-up before claiming the fix.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: B2 end-to-end test does not regress the fix</b> (Reported by: xai)</summary>

**File:** `casper/tests/batch1/multi_parent_casper_bonding_spec.rs` (line 128)

Per the TDD plan, a_finalized_bond_survives_an_epoch_boundary_merge passes on the pre-fix tree as well. It is a useful characterization guard for the bonds-map shape, but it does not prove the merge-layer fix or the production heartbeat/LFS race path.

**Suggestion:**
```
Keep B1 as the merge-mechanism gate (CI-required). Label B2 clearly as characterization-only, and track a preflight/heartbeat-race repro if #341 production failure needs ingredients beyond deterministic epoch-boundary merges.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Duplicated code in test helpers</b> (Reported by: bedrock)</summary>

**File:** `casper/tests/merging/finalized_protection_spec.rs` (line 246)

The functions `produce_on`, `consume_on`, and `chain_from` are duplicated between `base_protection_spec.rs` and `finalized_protection_spec.rs`.

**Suggestion:**
```
Extract these functions into a shared test helper module to avoid duplication.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Module declaration breaks alphabetical ordering</b> (Reported by: anthropic)</summary>

**File:** `casper/tests/merging/mod.rs` (line 2)

`finalized_protection_spec` is inserted between `base_protection_spec` and `conflict_detection_perf_spec`, breaking the existing alphabetical ordering of module declarations.

**Suggestion:**
```
Move `mod finalized_protection_spec;` after `mod conflict_detection_perf_spec;`.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Missing documentation for deferred refactor</b> (Reported by: bedrock)</summary>

**File:** `docs/tdd-plans/epoch-boundary-bond-loss-2026-08-25T07-30-00Z.md` (line 1)

The TDD plan mentions a deferred refactor to extract shared test helpers but does not provide detailed documentation on how this should be implemented.

**Suggestion:**
```
Add a section in the documentation detailing the planned refactor and the steps to implement it.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Fully-qualified EventLogIndex paths reduce readability</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1205)

The repeated `rspace_plus_plus::rspace::merger::event_log_index::EventLogIndex::...` paths inside the loop make the new block harder to read than necessary.

**Suggestion:**
```
Add a `use rspace_plus_plus::rspace::merger::event_log_index::EventLogIndex;` at module scope (or a local alias) and reference `EventLogIndex::empty()` / `EventLogIndex::combine` directly.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: partition_base_conflicts reused for non-base committed content</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1214)

Reusing partition_base_conflicts for finalized-carrier logs is fine mechanically, but the name implies base-only semantics and may confuse future merge changes.

**Suggestion:**
```
Rename to a neutral helper (e.g. partition_committed_conflicts) or add a brief comment at the call site that 'base' here means any committed event log, not only the merge base lineage.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Consider info/warn level for finalized-conflict rejections</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1218)

Rejections of chains conflicting with finalized content are consensus-significant events, but they are logged at `debug`, which is typically disabled in production — exactly where #341 manifested.

**Suggestion:**
```
Log at `info` (or `warn`) so operators can correlate finalized-protection rejections with bond/state anomalies without enabling debug logging.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Redundant identity rebinding of actual_seq_all</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/merging/dag_merger.rs` (line 1225)

After extend/sort, `let actual_seq_all = actual_seq_all;` only freezes mutability via shadowing. It is correct but noisy and easy to misread as meaningful transformation.

**Suggestion:**
```
Drop the rebind and use a non-mut binding from the partition result, e.g. build into a new immutable Vec, or call `actual_seq_all.shrink_to_fit()` only if needed; otherwise just leave the mut vec for subsequent use.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Duplicated merge test helpers</b> (Reported by: xai)</summary>

**File:** `casper/tests/merging/finalized_protection_spec.rs` (line 1)

produce_on/consume_on/chain_from builders duplicate base_protection_spec.rs helpers, increasing drift risk between the two protection specs.

**Suggestion:**
```
Extract shared helpers into something like tests/merging/support.rs as noted in the TDD deferred_refactor list.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test helpers duplicated from base_protection_spec</b> (Reported by: anthropic)</summary>

**File:** `casper/tests/merging/finalized_protection_spec.rs` (line 39)

`produce_on`, `consume_on`, `chain_consuming_on`, `chain_producing_on`, and `chain_from` duplicate helpers in base_protection_spec.rs, as the plan itself notes under 'deferred_refactor'. Divergence between the copies will make future invariant changes error-prone.

**Suggestion:**
```
Extract a shared `tests/merging/test_helpers.rs` module now rather than deferring — the extraction is mechanical and low-risk.
```

</details>
---

<details>

This PR adds a merge-layer protection so that deploy chains whose carrier block is FINALIZED are exempt from cost adjudication and win deterministically against conflicting scope chains, mirroring the existing base-lineage protection. The implementation is clean, deterministic (partition + combined event log + re-join + sort), and comes with a strong mechanism-level test (B1) that went properly RED, plus an end-to-end epoch-boundary test and an unusually candid TDD plan. However, the plan's own notes document at least two remaining paths where finalized content can still be rejected (the deploy-window rule running before the partition, and post-fold-in lineage rejection expansion), the interaction with the pre-existing base-conflict partition is untested for finalized-vs-base conflicts, and the end-to-end test never reproduced the actual #341 failure — the production repro appears to require a heartbeat race not covered here. The change is safe and an improvement, but a maintainer familiar with the merge/finality invariants should weigh in on whether the acknowledged gaps must be closed before this can be considered the fix for #341, hence needs_review rather than approval.

The code changes address the issue of validators losing bonds at epoch boundaries by ensuring that finalized content is never rejected during merge adjudication. However, there are some areas for improvement, such as handling conflicts between two finalized chains and reducing code duplication in test helpers. Overall, the changes are functional and safe to merge with minor adjustments.

#### openai (gpt-5.6-sol)

The change addresses the observed cost-adjudication mechanism and adds useful regression coverage, but it introduces a consensus-sensitive dependency on mutable node-local finality state and does not fully enforce its stated finalized-content invariant. Finalized chains can still be rejected by window/dependency processing or by conflicts with other finalized chains. These semantics need architectural review and broader invariant tests before merge.

#### xai (grok-4.5)

The merge-layer fix correctly mirrors base-content protection for finalized carriers: split by finalized_blocks_set, deterministically reject scope conflicts against the combined finalized event log, re-join finalized chains, and fold those rejects into the normal rejection path. The dedicated unit test is the real safety net and matches the #341/bridge-admin mechanism; the epoch-boundary bonding test is valuable but non-regressing on its own. Before treating #341 as fully closed, address (or consciously defer with tests) the two residual reject paths called out in the TDD notes—window rejection before this partition, and post-fold rejection expansion onto finalized-carrier dependents—since either can still admit loss of settled state.
</details>

---

jltatbeach (approved):

PR #345, opened 2026-08-25
test(rspace): isolate event_log/produce_counter contention cost (#50)

### LINKED ISSUE #50: Intra-deploy par-operator parallelism does not deliver speedup — investigation
> Migrated from [F1R3FLY-io/f1r3node#514](https://github.com/F1R3FLY-io/f1r3node/issues/514) as part of the deprecation of that repository. Original author: @elmiringos. Comments on the legacy issue are not mirrored here — see the linked legacy issue for the full thread.

---

### What did you expect to see?

The implementation already looks like it's aiming for parallel execution of par-branches:

- The reducer spawns a separate tokio task per branch (`rholang/src/rust/interpreter/reduce.rs:259`).
- The runtime is multi-threaded with one worker per CPU core (`node/src/main.rs:52`, `TOKIO_WORKER_THREADS=32` on a 32-core pod — verified at runtime).

Given that setup, a deploy with `N` par-branches doing identical work would reasonably be expected to scale across cores:

| par-branches | per-branch work | expected execution time | expected CPU peak |
|---|---|---|---|
| 1  | ~2.5 s on one core | ~2.5 s             | ~1 core |
| 4  | ~2.5 s each        | ~2.5 s             | ~4 cores |
| 32 | ~2.5 s each        | ~2.5 s (slowest branch wins) | toward 32 cores (saturated) |

As Kevin clarified, `|` is declarative — by the language spec it doesn't *require* concurrent execution, and the implementation is free to choose any strategy. But the runtime is already configured for parallelism end-to-end, so a 32-branch deploy on a 32-core pod should at least *attempt* to use those cores.

What we measured doesn't match this — see below.

### What did you see instead?

Propose time **grows linearly with total work**, with super-linear overhead at high `N`. CPU peak **stays at 1.5–4.8 cores out of 32** regardless of `N`. Replay times grow much faster than propose and exceed practical timeouts above `N ≥ 8`.

Run `rholang-par-08-05-26-091918`, profile `max` (32 CPU pod limit):

| forks | total iterations | **propose ms** | s/fork (`propose/N/1000`) | CPU peak | CPU % limit | replay PHASE | replay/propose |
|---|---|---|---|---|---|---|---|
| 1  | 5 000   | **2 560**   | 2.560  | n/a     | n/a  | 17 267 ms      | 6.7×  |
| 2  | 10 000  | **3 897**   | 1.948  | n/a     | n/a  | 233 521 ms     | 60×   |
| 4  | 20 000  | **8 472**   | 2.118  | n/a     | n/a  | 678 996 ms     | 80×   |
| 8  | 40 000  | **26 467**  | 3.308  | 3 482m  | 11 % | timed out 900s | >34×  |
| 16 | 80 000  | **92 816**  | 5.801  | 3 528m  | 11 % | timed out 900s | (~5h est) |
| 32 | 160 000 | **438 824** | 13.713 | 4 829m  | 15 % | timed out 900s | (~days est) |

`forks=1 → forks=32` multiplies total work by exactly 32×, but multiplies propose wall-clock by **171×**. A truly parallel reducer would have left wall-clock flat (~2.6 s); a perfectly sequential one would multiply by exactly 32× (=~82 s). 

CPU "n/a" rows = propose window was shorter than the 15s kubelet sampling tick; CPU was used, just not measured.

### Steps to reproduce the bug
**1. The smart contract.** A single signed deploy whose rholang body is generated programmatically — `N` par-composed `busy!(...)` invocations joined by an N-receive `for`. Each branch performs an identical 5 000-round recursive countdown through rspace. Concrete expansion for `forks=4`:

```rholang
new done, stdout(`rho:io:stdout`) in {
  new busy in {
    contract busy(@iters, @id, done) = {
      new loop in {
        contract loop(@n) = {
          if (n <= 0) { done!(id) } else { loop!(n - 1) }
        } |
        loop!(iters)
      }
    } |
    busy!(5000, 0, *done) | busy!(5000, 1, *done) | busy!(5000, 2, *done) | busy!(5000, 3, *done)
  } |
  for (@_ <- done; @_ <- done; @_ <- done; @_ <- done) {
    stdout!(("ParTerm done", 4))
  }
}
```

Each branch uses its **own private `loop` channel** (created via `new loop in {…}` inside `contract busy`), so the only shared rspace state across branches is the implicit tuplespace itself.

**2. Sweep and measurement loop.** For each `N ∈ {1, 2, 4, 8, 16, 32}` we provision a fresh 3-validator chain (one proposer + two followers, single shard, `max-user-deploys-per-block = 1`, heartbeat disabled), submit one signed deploy with the contract above, call `propose` on the proposer, wait for the block to advance on the proposer, then wait for the follower to replay it. Teardown the namespace, then next `N`. Total runtime ≈ 90 minutes for the full sweep on a 32-core pod.

**3. What we recorded per step.** Numbers come from two independent sources, both sampled live during each step:

- **Propose wall-clock** — measured client-side as `time.Since(proposeStart)` around the `propose` RPC against the proposer. Cross-checked against the node-side log line `Propose timing: mode=sync, propose_core_ms=…, total_ms=…` (within ~200 ms RPC overhead).
- **CPU peak** on the proposer pod — `usageNanoCores` from the kubelet's `/stats/summary` API, polled every 15 s; we keep the max sample observed during the propose window. Converted to millicores (`/1_000_000`).
- **Replay PHASE / EVAL** on the follower — Prometheus histograms `block.replay.phase.user-deploys.time` and `block.replay.deploy.evaluate.time` scraped from `:40403/metrics` before and after each step.
- **Contract completion confirmation** — `("ParTerm done", N)` appears in validator1's stdout log only after the N-receive `for` fires, i.e. after all N branches have signalled completion. Verified present for all 6 steps.

**4. How we concluded what we concluded.** Verdict is derived from propose wall-clock scaling, not replay PHASE (replay PHASE includes rig-replay overhead that is orthogonal to reducer concurrency). If the reducer were honouring `|` as concurrent execution we would see propose wall-clock stay roughly flat as `N` grows and CPU peak rise toward 32 cores. Instead propose wall-clock grows ~linearly with `N` (171× growth at `N = 32` for 32× more work) and CPU peak never exceeds 4.8 of 32 cores — across **all** rows of the sweep, in a consistent pattern.

### What version of F1R3FLY Node are you using?
rust-node:0.4.14

### What is your environment?
**Node**:
- Tokio multi-thread runtime, `TOKIO_WORKER_THREADS=32` (verified at runtime — env override in `pkg/manifest/statefulset.go:182`)
- Casper config: heartbeat disabled (`F1R3_HEARTBEAT_ENABLED=false`), `max-user-deploys-per-block = 1`, per-deploy phlo limit `100_000_000`
- 3 validators (one proposer + two followers) on a single shard, genesis-only, no dynamic bonding
- Bootstrap + 3 validators + 1 observer, all in one k3s namespace per fork-step
- Replay measured on validator2 (follower); propose measured on validator1

**Host**:
- k3s 1.x, containerd, single-node cluster
- Validator pod resource envelope (`max` profile):
  - `limits.cpu: 32`, `requests.cpu: 8`
  - `limits.memory: 64Gi`, `requests.memory: 16Gi`
  - PVC: 800 Gi (local-path)
  - cgroup `cpu.max`: `3200000 100000` (32-core CFS quota, verified)

9Neechan:
## Summary

Adds one diagnostic test that isolates the wall-clock cost of `event_log`/`produce_counter` (`RSpace`'s two remaining `std::sync::Mutex`es) from everything else on the `produce()` path. No production code changes — this test does not fix #50, it pins down which lock the residual serialization from #72's "out of scope" note actually lives in.

## Motivation

#72 fixed six layers of serialization for intra-deploy par-parallelism (#50, #43) and got `rholang-par` from SEQUENTIAL to PARTIALLY PARALLEL (423.7s → 52.9s at forks=32). Its own "out of scope" section names `event_log`/`produce_counter` as the prime suspect for the remaining ~2.5 idle cores, but that was never directly measured — only inferred from the flame graph at the time.

Re-running `bench_par_branches` (the rspace-only rung of #72's own benchmark ladder) against current `dev` HEAD confirmed the residual is still there: **0.88–0.96× speedup at forks=32 against an ideal of 32×**, even with a unique channel per operation — which rules out the striped per-channel lock (#72's fix #4) as the explanation, since it can't contend when every op touches a different channel. That still leaves two candidates open: `event_log`/`produce_counter` themselves, or something else on `produce()`'s path (`HotStore`, `get_store()`'s read lock, tokio scheduling).

## Changes

`rspace++/src/rspace/rspace.rs`: adds `event_log_and_produce_counter_isolated_cost_at_rholang_par_scale`, alongside the two existing `event_log` contention tests. It calls the private `log_produce()` directly — exactly what every `produce()` call does to these two locks (`event_log.push()`, then for non-persistent produces `produce_counter.insert()`) — with no `produce_lock()`, no `HotStore`, no `get_store()` read lock, no matcher. Same scale as `bench_par_branches`: 32 branches × 5,000 ops, sequential baseline vs. concurrent tokio tasks.

## Non-goals

Does not fix `event_log`/`produce_counter`. Follow-up fix direction (sharding `produce_counter` like `HotStore` was sharded, lock-free `event_log`) is scoped as separate work — `event_log` in particular feeds the rig/replay mechanism's event ordering, so its replacement needs its own PR and validation against replay determinism.

## Testing

```
cargo test --release -p rspace_plus_plus --lib rspace::rspace::tests -- --nocapture
```

All 3 tests in the module pass (`test result: ok. 3 passed; 0 failed`).

## Related

#50, #72, #43

### DIFF
--- a/rspace++/src/rspace/rspace.rs
+++ b/rspace++/src/rspace/rspace.rs
@@ -1374,7 +1374,8 @@ mod tests {
 
     // Disambiguates two candidate explanations for the near-zero wall-clock
     // speedup measured end-to-end at rholang-par scale (bench_par_branches:
-    // 0.88-0.96x at 32 branches, ideal 32x — see issue #50 follow-up):
+    // 0.88-0.96x at 32 branches against a worker-thread-bounded ideal — see
+    // issue #50 follow-up):
     // (a) event_log/produce_counter themselves cost that much, or
     // (b) something else in produce()'s path (HotStore, the striped
     //     per-channel lock, tokio scheduling) is the real cost and these
@@ -1425,13 +1426,24 @@ mod tests {
         }
         let par_ms = t_par.elapsed().as_millis().max(1);
 
+        // log_produce() is synchronous and the branch loops never await, so
+        // each task holds its worker for the whole loop: at most num_workers
+        // (further capped by the host's cores) branches run at once. That, not
+        // PAR_BRANCHES, is the achievable ideal for the efficiency metric.
+        let num_workers = tokio::runtime::Handle::current().metrics().num_workers();
+        let ideal = PAR_BRANCHES.min(num_workers).min(
+            std::thread::available_parallelism()
+                .map(|n| n.get())
+                .unwrap_or(usize::MAX),
+        );
         let speedup = seq_ms as f64 / par_ms as f64;
-        let efficiency = speedup / PAR_BRANCHES as f64 * 100.0;
+        let efficiency = speedup / ideal as f64 * 100.0;
 
         eprintln!(
             "event_log+produce_counter isolated cost: branches={PAR_BRANCHES} \
              ops_per_branch={OPS_PER_BRANCH} total_ops={} sequential={seq_ms}ms \
-             concurrent={par_ms}ms speedup={:.2}x (ideal {PAR_BRANCHES}x) efficiency={:.1}%",
+             concurrent={par_ms}ms speedup={:.2}x (ideal {ideal}x: {PAR_BRANCHES} branches on \
+             {num_workers} workers) efficiency={:.1}%",
             PAR_BRANCHES * OPS_PER_BRANCH,
             speedup,
             efficiency,
--- a/rspace++/src/rspace/rspace.rs
+++ b/rspace++/src/rspace/rspace.rs
@@ -1371,4 +1371,75 @@ mod tests {
             PAR_BRANCHES * OPS_PER_BRANCH,
         );
     }
+
+    // Disambiguates two candidate explanations for the near-zero wall-clock
+    // speedup measured end-to-end at rholang-par scale (bench_par_branches:
+    // 0.88-0.96x at 32 branches, ideal 32x — see issue #50 follow-up):
+    // (a) event_log/produce_counter themselves cost that much, or
+    // (b) something else in produce()'s path (HotStore, the striped
+    //     per-channel lock, tokio scheduling) is the real cost and these
+    //     two locks are not the story despite being std::sync::Mutex.
+    //
+    // Calls log_produce() directly — the exact call log_produce() itself
+    // takes (event_log.push then, for non-persist, produce_counter.insert)
+    // with no produce_lock(), no HotStore, no get_store() RwLock read, no
+    // matcher involved. Isolates these two locks from every other cost
+    // produce() incurs, at the same branch/op counts as bench_par_branches.
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn event_log_and_produce_counter_isolated_cost_at_rholang_par_scale() {
+        const PAR_BRANCHES: usize = 32;
+        const OPS_PER_BRANCH: usize = 5000;
+
+        // Sequential baseline: identical total ops, one after another.
+        let seq_rspace = make_rspace().await;
+        let t_seq = Instant::now();
+        for b in 0..PAR_BRANCHES {
+            for i in 0..OPS_PER_BRANCH {
+                let channel = format!("seq_{}_{}", b, i);
+                let data = "datum".to_string();
+                let produce_ref = Produce::create(&channel, &data, false);
+                seq_rspace.log_produce(&produce_ref, &channel, &data, false);
+            }
+        }
+        let seq_ms = t_seq.elapsed().as_millis().max(1);
+
+        // Concurrent: PAR_BRANCHES tokio tasks, each on its own channel set,
+        // all sharing one RSpace and therefore one event_log/produce_counter.
+        let par_rspace = Arc::new(make_rspace().await);
+        let t_par = Instant::now();
+        let handles: Vec<_> = (0..PAR_BRANCHES)
+            .map(|b| {
+                let s = par_rspace.clone();
+                tokio::spawn(async move {
+                    for i in 0..OPS_PER_BRANCH {
+                        let channel = format!("par_{}_{}", b, i);
+                        let data = "datum".to_string();
+                        let produce_ref = Produce::create(&channel, &data, false);
+                        s.log_produce(&produce_ref, &channel, &data, false);
+                    }
+                })
+            })
+            .collect();
+        for h in handles {
+            h.await.unwrap();
+        }
+        let par_ms = t_par.elapsed().as_millis().max(1);
+
+        let speedup = seq_ms as f64 / par_ms as f64;
+        let efficiency = speedup / PAR_BRANCHES as f64 * 100.0;
+
+        eprintln!(
+            "event_log+produce_counter isolated cost: branches={PAR_BRANCHES} \
+             ops_per_branch={OPS_PER_BRANCH} total_ops={} sequential={seq_ms}ms \
+             concurrent={par_ms}ms speedup={:.2}x (ideal {PAR_BRANCHES}x) efficiency={:.1}%",
+            PAR_BRANCHES * OPS_PER_BRANCH,
+            speedup,
+            efficiency,
+        );
+
+        // No assertion: this test is diagnostic, not a regression gate. It
+        // reports the isolated cost of these two locks so it can be compared
+        // against bench_par_branches' end-to-end number for the same
+        // branch/op counts (see issue #50 investigation).
+    }
 }


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

---

### Summary

The test is safe and does isolate log_produce from the rest of produce(), but its headline efficiency calculation is misleading because only eight worker threads can execute the 32 non-yielding tasks. It also adds a relatively heavy, assertion-free benchmark to the normal test suite. Correcting the concurrency baseline and making the diagnostic opt-in would make the measurements more actionable.

This is a well-motivated, production-safe diagnostic-only test with clear issue linkage and no security or production-logic risk. The main actionable concern is interpretive: reporting ideal 32× efficiency while pinning the runtime to 8 worker threads can make the isolated lock cost look worse than the threading model allows, which weakens the very disambiguation the PR aims for—align thread count with branch count or correct the ideal/efficiency baseline. Secondary nits are the always-pass nature of the test (consider #[ignore] or a bench), possible alloc noise in the timed loop, and small comment clarifications; none of those should block merge if the methodology fix (or explicit justification vs existing benches) is addressed.

The code change adds a new diagnostic test to isolate the cost of `event_log` and `produce_counter` contention. The test is well-structured and follows the existing testing conventions. There are no security, logic, or performance issues identified. The only suggestion is to add more detailed comments to enhance maintainability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Reported 32-way ideal is impossible with eight worker threads</b> (Reported by: openai, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1389)

The concurrent workload consists entirely of synchronous work inside spawned futures and never reaches an await point. Each task therefore occupies one Tokio worker until its loop completes, so at most eight tasks can execute simultaneously with worker_threads = 8. Calculating efficiency against PAR_BRANCHES = 32 and describing 32x as the ideal produces a misleading diagnostic metric; even without lock contention, this test cannot demonstrate 32x speedup.

---

The test is annotated with `worker_threads = 8` but reports `ideal {PAR_BRANCHES}x` (32×) and computes `efficiency = speedup / 32`. At most ~8 tasks can run truly in parallel on this runtime, so a perfectly scalable critical section would still top out near 8× (~25% “efficiency”). That makes the diagnostic easy to misread when comparing to `bench_par_branches` or when deciding whether these two Mutexes explain residual idle cores. If the sibling benches use the same 8-thread runtime the relative comparison can still be valid, but the absolute “ideal 32×” framing is methodologically wrong for this process.

**Suggestion:**
```
Either set `worker_threads = 32` (or omit a low cap) to match PAR_BRANCHES, or define ideal as `min(PAR_BRANCHES, worker_threads)` and print both the task count and the runtime thread count in the eprintln so speedup/efficiency cannot be over-interpreted.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Expensive diagnostic benchmark runs as an ordinary assertion-free test</b> (Reported by: openai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1390)

This test performs 320,000 total operations across two RSpace instances, retains large event logs and produce-counter contents, and provides no pass/fail regression signal. Running it in every normal library test invocation adds CPU, allocation, and memory cost while timing results—especially in debug builds and shared CI—are not stable correctness checks.

**Suggestion:**
```
Mark the diagnostic test with #[ignore] and document the command for running it explicitly, or move it into a benchmark target. If it remains a regular test, add a stable assertion that provides regression value.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Diagnostic test never fails (no assertion / benchmark gate)</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1435)

The test explicitly has no assertion and only prints timings. That matches the PR non-goal, but it will always show green in CI, so regressions in lock cost will not be caught and the test adds wall-clock cost on every `cargo test` run of this module without enforcing a property. It is also closer to a micro-benchmark than a unit test (sync Mutex work inside `tokio::spawn` async tasks).

**Suggestion:**
```
Keep it if the team wants an always-on probe, but consider `#[ignore]` (run via `--ignored` in the #50 investigation) or moving it next to `bench_par_branches` as a criterion/bench. If it stays as a test, document in the module that green ≠ “locks are fine,” only that the probe ran.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add more detailed comments</b> (Reported by: bedrock)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1375)

Consider adding more detailed comments to explain the purpose and expected outcomes of the new test. This will help future maintainers understand the context and significance of the test.

**Suggestion:**
```
Add comments explaining the expected speedup and efficiency values, and how they relate to the overall performance goals.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Comment wording is slightly confusing</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1382)

“Calls log_produce() directly — the exact call log_produce() itself takes” reads circular. From the PR text, the intent is that this is the same lock sequence every `produce()` uses via `log_produce` (push + optional counter insert), not a description of what `log_produce` calls on itself.

**Suggestion:**
```
Rephrase to something like: “Calls log_produce() directly — the same event_log/produce_counter sequence produce() uses — without produce_lock(), HotStore, get_store(), or matcher.”
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Hot-loop allocations are fine for parity but dominate if locks are cheap</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1396)

Each op does `format!(...)`, `"datum".to_string()`, and `Produce::create(...)` on both sequential and concurrent paths. Relative speedup remains comparable, but if the isolated locks are very cheap, allocation/formatting noise can flatten the measured speedup and understate lock contention (or overstate “something else” when compared carelessly to end-to-end benches that share similar alloc patterns).

**Suggestion:**
```
Optional: precompute `Vec` of `(channel, data, Produce)` per branch (or reuse one channel string pattern like the existing contention tests) so the timed region is closer to pure `log_produce` lock work. Not required if strict parity with `bench_par_branches` setup is the goal—say that explicitly in the comment.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Blocking std::sync::Mutex inside async tasks is intentional but worth one line</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1410)

Spawning async tasks that call a sync path taking `std::sync::Mutex` will block tokio worker threads. That likely mirrors production `produce()` behavior and is what you want to measure, but readers may think the test should use `spawn_blocking` or `tokio::sync::Mutex`.

**Suggestion:**
```
Add one comment that blocking here is deliberate parity with the produce path (and part of the cost under study), so future cleanups do not “fix” it into spawn_blocking and change the measurement.
```

</details>
---

<details>

The code change adds a new diagnostic test to isolate the cost of `event_log` and `produce_counter` contention. The test is well-structured and follows the existing testing conventions. There are no security, logic, or performance issues identified. The only suggestion is to add more detailed comments to enhance maintainability.

#### openai (gpt-5.6-sol)

The test is safe and does isolate log_produce from the rest of produce(), but its headline efficiency calculation is misleading because only eight worker threads can execute the 32 non-yielding tasks. It also adds a relatively heavy, assertion-free benchmark to the normal test suite. Correcting the concurrency baseline and making the diagnostic opt-in would make the measurements more actionable.

#### xai (grok-4.5)

This is a well-motivated, production-safe diagnostic-only test with clear issue linkage and no security or production-logic risk. The main actionable concern is interpretive: reporting ideal 32× efficiency while pinning the runtime to 8 worker threads can make the isolated lock cost look worse than the threading model allows, which weakens the very disambiguation the PR aims for—align thread count with branch count or correct the ideal/efficiency baseline. Secondary nits are the always-pass nature of the test (consider #[ignore] or a bench), possible alloc noise in the timed loop, and small comment clarifications; none of those should block merge if the methodology fix (or explicit justification vs existing benches) is addressed.
</details>

---

jeffrey-l-turner (approved):

jeffrey-l-turner:
Maintainer-directed remediation of the critical review finding ("32-way ideal is impossible with eight worker threads"): commit cfaf06bae changes the diagnostic in `event_log_and_produce_counter_isolated_cost_at_rholang_par_scale` to compute its efficiency ideal as min(PAR_BRANCHES, tokio worker count, available_parallelism) instead of dividing by PAR_BRANCHES, and to print the basis (`ideal 8x: 32 branches on 8 workers`). The workload is unchanged — it deliberately mirrors `bench_par_branches` — and the corrected output still supports the PR's conclusion (0.61x speedup against the honest 8x ideal).

PR #346, opened 2026-08-25
fix(rspace): shard produce_counter to remove global lock (#50)

9Neechan:
## Summary

Shards `RSpace`'s `produce_counter` from one global `std::sync::Mutex<BTreeMap>` into 256 independent stripes, removing it as a source of serialization in the intra-deploy par-parallelism chain tracked in #50. This is one of two parts — `event_log` (the other lock #345's isolation test proved sufficient, together with this one, to explain the residual bottleneck) is deliberately left for a separate PR, see Non-goals.

## Motivation

#345 isolated the residual serialization left after #72 to `event_log`/`produce_counter` specifically: calling `log_produce()` directly, bypassing `HotStore`/the striped per-channel lock, showed a speedup of 0.68× against an ideal of 32× — at or below the full `bench_par_branches` number (0.88–0.96×), proving these two locks are sufficient to explain the entire residual. Starting with `produce_counter` as the simpler, lower-risk case — unlike `event_log`, it carries no ordering dependency for replay/rig.

## Changes

`rspace++/src/rspace/rspace.rs`:

- `produce_counter` field: `Arc<std::sync::Mutex<BTreeMap<Produce, i32>>>` → `Arc<Vec<std::sync::Mutex<BTreeMap<Produce, i32>>>>`, 256 stripes (`striped_locks::NUM_LOCK_STRIPES`), the same channel-hash-based sharding scheme already used for `phase_a_locks`/`phase_b_locks`.
- New helpers: `produce_counter_shard()`, `take_produce_counter()`, `reset_produce_counter()`, `restore_produce_counter()`.
- The hot path (`log_produce`, `produce_counters`) now only ever locks the one shard its key hashes to.
- The five checkpoint-boundary call sites (`create_checkpoint`, `reset`, `create_soft_checkpoint`, `take_event_log`, `revert_to_soft_checkpoint`) switched to the new helpers; per-shard `mem::take` preserves the same no-data-loss guarantee the single global lock had.

## Non-goals

Does not fix `event_log` — deliberately deferred. `event_log`'s insertion order feeds the rig/replay mechanism, so a lock-free or sharded replacement needs its own design and validation against replay determinism, not bundled into this smaller, lower-risk change.

Also unaddressed: striped per-channel lock contention visible in the testbed's `Lock%` figures (a different lock, not yet investigated), and the replay-side `Mutex<ReplayRSpace>` + rig-matching bottleneck — a separate, more severe problem.

## Related

#50, #72, #345

### DIFF
--- a/rspace++/src/rspace/rspace.rs
+++ b/rspace++/src/rspace/rspace.rs
@@ -60,13 +60,14 @@ pub struct RSpace<C, P, A, K> {
     event_log: Arc<std::sync::Mutex<Log>>,
     // Striped like phase_a/phase_b_locks below: NUM_LOCK_STRIPES independent
     // shards keyed by channel_hash(produce) % NUM_LOCK_STRIPES, instead of
-    // one global std::sync::Mutex<BTreeMap>. Per-shard mem::take keeps the
-    // same no-data-loss guarantee the single-lock version had (a concurrent
-    // produce() racing a checkpoint either lands in the pre-drain snapshot
-    // or the post-drain empty shard, never both/neither), while the hot path
-    // (log_produce/produce_counters) only ever contends the one shard its
-    // key hashes to. See take_produce_counter/reset_produce_counter/
-    // restore_produce_counter below for the checkpoint-boundary helpers.
+    // one global std::sync::Mutex<BTreeMap>. The hot path (log_produce/
+    // produce_counters) only ever contends the one shard its key hashes to.
+    // Checkpoint boundaries (take/reset/restore_produce_counter below) lock
+    // every shard for the whole operation, in the same fixed index order
+    // every time, restoring the single global lock's point-in-time
+    // linearization for this field specifically — not across it and
+    // event_log, which remain two separate locks taken sequentially, same
+    // as before this change (pre-existing, out of scope here).
     produce_counter: Arc<Vec<std::sync::Mutex<BTreeMap<Produce, i32>>>>,
     matcher: Arc<Box<dyn Match<P, A, K>>>,
     // Fixed-size striped locks replace the growing DashMap<u64, Mutex>.
@@ -545,6 +546,9 @@ where
         Ok((history_repo, hot_store))
     }
 
+    // Single-shard lookup for the hot path (log_produce): a given key always
+    // hashes to the same shard, so per-key correctness holds regardless of
+    // what any other shard is doing concurrently.
     fn produce_counter_shard(
         &self,
         produce_ref: &Produce,
@@ -553,49 +557,81 @@ where
         &self.produce_counter[idx]
     }
 
+    // Groups by shard first so a slice with repeats or same-shard keys locks
+    // each needed shard only once, instead of once per produce_ref.
     fn produce_counters(&self, produce_refs: &[Produce]) -> BTreeMap<Produce, i32> {
-        produce_refs
+        let mut by_shard: HashMap<usize, Vec<&Produce>> = HashMap::new();
+        for p in produce_refs {
+            let idx = (striped_locks::channel_hash(p) as usize) % self.produce_counter.len();
+            by_shard.entry(idx).or_default().push(p);
+        }
+
+        let mut result = BTreeMap::new();
+        for (idx, refs) in by_shard {
+            let guard = self.produce_counter[idx]
+                .lock()
+                .expect("produce counter shard lock");
+            for p in refs {
+                result.insert(p.clone(), guard.get(p).copied().unwrap_or(0));
+            }
+        }
+        result
+    }
+
+    // Locks every shard, in the same fixed index order every helper below
+    // uses, before touching any of them, and holds all guards for the whole
+    // operation. This is the equivalent of the former single global lock's
+    // linearization point, restricted to this field: any produce() racing
+    // take/reset/restore blocks on whichever shard it hashes to until the
+    // full boundary operation finishes on every shard, so the result is a
+    // consistent point-in-time snapshot across shards, not just per-key.
+    //
+    // (This does not make produce_counter atomic *with* event_log — those
+    // remain two separate locks taken sequentially, same as before this
+    // change; that cross-field ordering gap is pre-existing, not introduced
+    // here, and is out of scope until event_log itself is addressed.)
+    fn all_produce_counter_shards(&self) -> Vec<std::sync::MutexGuard<'_, BTreeMap<Produce, i32>>> {
+        self.produce_counter
             .iter()
-            .cloned()
-            .map(|p| {
-                let count = self
-                    .produce_counter_shard(&p)
-                    .lock()
-                    .expect("produce counter shard lock")
-                    .get(&p)
-                    .copied()
-                    .unwrap_or(0);
-                (p, count)
-            })
+            .map(|shard| shard.lock().expect("produce counter shard lock"))
             .collect()
     }
 
-    // Drains all shards into one combined map and leaves every shard empty.
-    // Per-shard mem::take is atomic under that shard's own lock, so a
-    // produce() racing this call either lands in the returned snapshot or
-    // in the now-empty shard afterwards — never both, never neither.
+    // Drains all shards into one combined map and leaves every shard empty,
+    // atomically across the whole field (see all_produce_counter_shards).
     fn take_produce_counter(&self) -> BTreeMap<Produce, i32> {
+        let mut guards = self.all_produce_counter_shards();
         let mut combined = BTreeMap::new();
-        for shard in self.produce_counter.iter() {
-            let mut guard = shard.lock().expect("produce counter shard lock");
-            combined.extend(std::mem::take(&mut *guard));
+        for guard in guards.iter_mut() {
+            combined.extend(std::mem::take(&mut **guard));
         }
         combined
     }
 
+    // Empties every shard atomically across the whole field.
     fn reset_produce_counter(&self) {
-        for shard in self.produce_counter.iter() {
-            *shard.lock().expect("produce counter shard lock") = BTreeMap::new();
+        let mut guards = self.all_produce_counter_shards();
+        for guard in guards.iter_mut() {
+            **guard = BTreeMap::new();
         }
     }
 
+    // Partitions the incoming map by shard before taking any locks, then
+    // installs each partition with one assignment per shard under all
+    // guards held together — avoids the reset-then-insert window a
+    // concurrent produce() could otherwise land in between.
     fn restore_produce_counter(&self, map: BTreeMap<Produce, i32>) {
-        self.reset_produce_counter();
+        let mut partitioned: Vec<BTreeMap<Produce, i32>> = (0..self.produce_counter.len())
+            .map(|_| BTreeMap::new())
+            .collect();
         for (k, v) in map {
-            self.produce_counter_shard(&k)
-                .lock()
-                .expect("produce counter shard lock")
-                .insert(k, v);
+            let idx = (striped_locks::channel_hash(&k) as usize) % self.produce_counter.len();
+            partitioned[idx].insert(k, v);
+        }
+
+        let mut guards = self.all_produce_counter_shards();
+        for (guard, part) in guards.iter_mut().zip(partitioned) {
+            **guard = part;
         }
     }
 
--- a/rspace++/src/rspace/rspace.rs
+++ b/rspace++/src/rspace/rspace.rs
@@ -58,7 +58,16 @@ pub struct RSpace<C, P, A, K> {
     pub store: Arc<std::sync::RwLock<Arc<Box<dyn HotStore<C, P, A, K>>>>>,
     installs: Arc<std::sync::Mutex<HashMap<Vec<C>, Install<P, K>>>>,
     event_log: Arc<std::sync::Mutex<Log>>,
-    produce_counter: Arc<std::sync::Mutex<BTreeMap<Produce, i32>>>,
+    // Striped like phase_a/phase_b_locks below: NUM_LOCK_STRIPES independent
+    // shards keyed by channel_hash(produce) % NUM_LOCK_STRIPES, instead of
+    // one global std::sync::Mutex<BTreeMap>. Per-shard mem::take keeps the
+    // same no-data-loss guarantee the single-lock version had (a concurrent
+    // produce() racing a checkpoint either lands in the pre-drain snapshot
+    // or the post-drain empty shard, never both/neither), while the hot path
+    // (log_produce/produce_counters) only ever contends the one shard its
+    // key hashes to. See take_produce_counter/reset_produce_counter/
+    // restore_produce_counter below for the checkpoint-boundary helpers.
+    produce_counter: Arc<Vec<std::sync::Mutex<BTreeMap<Produce, i32>>>>,
     matcher: Arc<Box<dyn Match<P, A, K>>>,
     // Fixed-size striped locks replace the growing DashMap<u64, Mutex>.
     // See striped_locks.rs for the stripe count and hashing/lock scheme.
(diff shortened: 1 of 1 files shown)


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

| xai | `grok-4.5` | Feedback Provided | 0.78 |

---

### Summary

This is a well-scoped, well-motivated change that shards produce_counter into 256 stripes using the existing channel-hash scheme, and the hot path (log_produce / produce_counters) correctly always resolves the same key to the same shard, so per-key correctness is preserved. The main caveat is a semantic weakening the field comment glosses over: the single global lock gave point-in-time-consistent whole-map snapshots at checkpoint boundaries, whereas per-shard draining (take_produce_counter) and the reset-then-insert restore (restore_produce_counter) do not — this is safe only if checkpoint/revert operations never race with in-flight produces, an invariant that should be documented or asserted rather than assumed. No security concerns; remaining points are minor performance (per-key shard locking in produce_counters) and style/documentation polish. Safe to merge once the concurrency assumptions at checkpoint boundaries are made explicit.

The hot-path sharding is reasonable and should reduce contention, but the checkpoint-boundary helpers do not preserve the atomic semantics of the former single mutex. Since locks are released shard by shard, concurrent producers can see partial resets or restores and produce snapshots that cannot correspond to any single point in time. Boundary operations should lock all stripes in a stable order for their full duration before this change is merged.

Solid, low-risk sharding of produce_counter along the same stripe scheme as the phase locks: hot-path log_produce only contends one shard, helpers cover checkpoint boundaries, and init matches NUM_LOCK_STRIPES. No security concerns. The main follow-up is semantic: multi-shard take/reset/restore are no longer one atomic map swap, so SoftCheckpoint counter state can tear relative to concurrent produce and to the still-global event_log; that is fine only if checkpoint/revert runs quiescently—worth documenting or fixing with all-shard lock ordering / per-shard replace. Minor cleanups: partition-then-replace in restore, optional batching in produce_counters. Safe to merge once the quiescence/atomicity expectation is explicitly accepted.

The code change effectively shards the `produce_counter` to improve concurrency, but there are some logic and style issues that need addressing. Specifically, ensure consistent locking order to avoid deadlocks, standardize the use of `Arc::new`, and add documentation for new elements.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential deadlock in `produce_counters` method</b> (Reported by: bedrock)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 533)

The `produce_counters` method locks each shard individually, which could lead to a deadlock if two threads attempt to access the same shard in a different order. Consider using a consistent locking order to avoid this.

**Suggestion:**
```
Ensure that shards are always locked in a consistent order, perhaps by sorting the `produce_refs` before locking.
```

</details>

<details>
<summary><b>:orange_circle: Major: Checkpoint helpers are no longer globally atomic across counter shards</b> (Reported by: openai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 564)

The previous single mutex gave take, reset, and restore operations one linearization point for the complete produce-counter map. The new helpers lock and release each shard independently, allowing concurrent producers to observe a partially completed boundary operation. For example, reset_produce_counter can clear shard A, allow an increment in A to survive, then allow a later increment in shard B to be erased when B is cleared. Likewise, restore_produce_counter resets all shards and inserts entries individually, so an increment occurring after its shard was reset can still be overwritten by a subsequent restore insertion. take_produce_counter does not lose individual entries, but it can classify an earlier update in an already-drained shard as post-checkpoint while classifying a later update in a not-yet-drained shard as pre-checkpoint. These outcomes were impossible under the former global counter lock and can make checkpoint, reset, and rollback state inconsistent under the concurrent checkpoint races explicitly discussed in the change.

**Suggestion:**
```
Keep the hot path sharded, but have boundary helpers acquire all shard mutexes in a fixed order and retain every guard until the complete take, reset, or restore operation finishes. Then drain, clear, or redistribute entries while all guards are held. This restores a single global linearization point only at checkpoint boundaries without reintroducing the global lock on log_produce.
```

</details>

<details>
<summary><b>:orange_circle: Major: take_produce_counter is not a point-in-time consistent snapshot across shards</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 573)

The old single global Mutex guaranteed that mem::take produced an atomic, point-in-time snapshot of the *entire* counter map. The new take_produce_counter locks and drains shards one at a time, so a concurrent produce() racing the checkpoint can land in an already-drained shard while a related counter in a not-yet-drained shard is captured. Per-key the 'in snapshot XOR in post-drain shard' guarantee holds (as the comment says), but cross-key consistency within a single checkpoint snapshot is weaker than before. If checkpoint creation is guaranteed to happen at a quiescent point (no in-flight produces), this is harmless — but that invariant should be stated explicitly, because the field comment currently claims the 'same no-data-loss guarantee the single-lock version had', which overstates the equivalence.

**Suggestion:**
```
Document (or assert) the quiescence invariant at checkpoint boundaries, or note explicitly in the comment that cross-shard snapshot consistency is weaker than the single-lock version and why that is acceptable for produce_counter semantics.
```

</details>

<details>
<summary><b>:orange_circle: Major: Multi-shard take/reset is not a point-in-time snapshot</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 575)

take_produce_counter and reset_produce_counter lock shards one-by-one. Unlike the old single Mutex, a concurrent log_produce can interleave mid-drain so the combined map (or the cleared state) is torn across shards. The per-key 'never both/neither' claim holds, but the SoftCheckpoint { event_log, produce_counter } pair can now diverge further: e.g. event_log is taken under one global lock first, then counters are drained shard-by-shard, so a produce may appear in the counter snapshot without its IoEvent (or the reverse, depending on timing). If create_soft_checkpoint / revert can run concurrently with produce (intra-deploy par), this weakens checkpoint consistency versus the previous atomic map swap. If those boundaries are always externally quiescent, this is benign—but that invariant is not enforced or documented at the call sites.

**Suggestion:**
```
Document the quiescence assumption at create_soft_checkpoint/revert_to_soft_checkpoint/create_checkpoint, or lock all shards (in index order to avoid deadlock) for the duration of take/reset/restore so the counter map remains a single atomic snapshot. At minimum, take event_log and all counter shards under a defined lock order if paired consistency matters for rig/replay.
```

</details>

<details>
<summary><b>:orange_circle: Major: produce_counters locks a shard per key with no batching</b> (Reported by: anthropic, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 553)

Each produce_ref acquires/releases its shard lock independently. Callers passing many refs that hash to the same stripe pay repeated lock overhead; the old path locked once for the whole read. Unlikely hot relative to log_produce, but easy to tighten.

---

produce_counters iterates produce_refs and acquires the shard lock once per key. When multiple keys hash to the same shard (or the same key appears repeatedly), the same mutex is locked and unlocked repeatedly. For small ref slices this is negligible, but it's easy to batch.

**Suggestion:**
```
Group keys by shard index, lock each needed shard once, and batch the lookups into the result map.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent use of `Arc::new`</b> (Reported by: bedrock)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 427)

The `Arc::new` is used inconsistently. In some places, it wraps the `Mutex::new`, while in others, it directly wraps the collection. Standardize the usage for clarity and consistency.

**Suggestion:**
```
Always wrap the `Mutex::new` with `Arc::new` to maintain consistency.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: restore_produce_counter is two-phase and more racy than before</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 589)

Old code atomically replaced the whole BTreeMap under one lock. restore_produce_counter first clears every shard, then re-inserts key-by-key. Concurrent produce() can land in a cleared shard and later be overwritten by restore, or survive in a shard if that key is absent from the checkpoint map—widening the race window versus a per-shard mem::replace of a pre-partitioned map.

**Suggestion:**
```
Partition the incoming map into NUM_LOCK_STRIPES local BTreeMaps first (no locks), then for each shard index do `*shard.lock() = partitioned[i]` (single replace per shard). That removes the empty intermediate state and matches take's mem::take style.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: restore_produce_counter has a window between reset and re-insert</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 590)

restore_produce_counter first clears all shards, then re-inserts entries one shard-lock at a time. Under the old single lock, restoration (`*guard = checkpoint.produce_counter`) was one atomic assignment. Now a concurrent produce() interleaved between reset_produce_counter() and the insert loop could increment a key that is subsequently overwritten by insert(), or add a key not present in the checkpoint map that then survives the revert. This only matters if revert_to_soft_checkpoint can race with in-flight produces; if the runtime guarantees exclusive access during revert, it's fine, but that assumption is currently implicit.

**Suggestion:**
```
Add a comment stating that revert_to_soft_checkpoint must not run concurrently with produce/consume, or restructure to install the restored state per-shard atomically (partition the checkpoint map by shard index first, then swap each shard's contents in a single lock acquisition).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Long line / missing rustfmt on produce_counter_shard signature</b> (Reported by: anthropic, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 548)

The signature `fn produce_counter_shard(&self, produce_ref: &Produce) -> &std::sync::Mutex<BTreeMap<Produce, i32>>` exceeds typical rustfmt width and repeats the fully qualified mutex/map type that already appears in the struct field. A type alias would reduce duplication and keep the field, helpers, and constructor in sync if the type ever changes.

---

produce_counter_shard duplicates the hash-modulo indexing pattern used for phase_a/phase_b locks. Fine for clarity, but if striped_locks already exposes a stripe index helper, reusing it would keep hashing policy in one place (avoids drift if NUM_LOCK_STRIPES or channel_hash usage changes).

**Suggestion:**
```
Introduce `type ProduceCounterShard = std::sync::Mutex<BTreeMap<Produce, i32>>;` (or similar) and run rustfmt.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add documentation for new fields and methods</b> (Reported by: bedrock)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 58)

The new `produce_counter` field and associated methods lack documentation. Adding comments will help other developers understand the purpose and usage of these changes.

**Suggestion:**
```
Add inline comments or docstrings to explain the new field and methods.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: reset_produce_counter and restore_produce_counter lack doc comments</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 583)

take_produce_counter has a good explanatory comment about its atomicity properties, but reset_produce_counter and restore_produce_counter — which have their own (weaker) concurrency semantics per the issues above — have none.

**Suggestion:**
```
Add brief comments describing when each helper is called and what concurrency assumptions it relies on.
```

</details>
---

<details>

This is a well-scoped, well-motivated change that shards produce_counter into 256 stripes using the existing channel-hash scheme, and the hot path (log_produce / produce_counters) correctly always resolves the same key to the same shard, so per-key correctness is preserved. The main caveat is a semantic weakening the field comment glosses over: the single global lock gave point-in-time-consistent whole-map snapshots at checkpoint boundaries, whereas per-shard draining (take_produce_counter) and the reset-then-insert restore (restore_produce_counter) do not — this is safe only if checkpoint/revert operations never race with in-flight produces, an invariant that should be documented or asserted rather than assumed. No security concerns; remaining points are minor performance (per-key shard locking in produce_counters) and style/documentation polish. Safe to merge once the concurrency assumptions at checkpoint boundaries are made explicit.

The code change effectively shards the `produce_counter` to improve concurrency, but there are some logic and style issues that need addressing. Specifically, ensure consistent locking order to avoid deadlocks, standardize the use of `Arc::new`, and add documentation for new elements.

#### openai (gpt-5.6-sol)

The hot-path sharding is reasonable and should reduce contention, but the checkpoint-boundary helpers do not preserve the atomic semantics of the former single mutex. Since locks are released shard by shard, concurrent producers can see partial resets or restores and produce snapshots that cannot correspond to any single point in time. Boundary operations should lock all stripes in a stable order for their full duration before this change is merged.

#### xai (grok-4.5)

Solid, low-risk sharding of produce_counter along the same stripe scheme as the phase locks: hot-path log_produce only contends one shard, helpers cover checkpoint boundaries, and init matches NUM_LOCK_STRIPES. No security concerns. The main follow-up is semantic: multi-shard take/reset/restore are no longer one atomic map swap, so SoftCheckpoint counter state can tear relative to concurrent produce and to the still-global event_log; that is fine only if checkpoint/revert runs quiescently—worth documenting or fixing with all-shard lock ordering / per-shard replace. Minor cleanups: partition-then-replace in restore, optional batching in produce_counters. Safe to merge once the quiescence/atomicity expectation is explicitly accepted.
</details>

---

9Neechan:
Addressed the review. Summary of what changed and what I'm disputing:

## Fixed

**Checkpoint helpers not atomic across shards** (openai's "Checkpoint helpers are no longer globally atomic", anthropic's "take_produce_counter is not a point-in-time consistent snapshot", xai's "Multi-shard take/reset is not a point-in-time snapshot", and the two related Minor findings on `restore_produce_counter`'s reset-then-insert window) — all one underlying issue, now fixed. `take_produce_counter`/`reset_produce_counter`/`restore_produce_counter` now go through a new `all_produce_counter_shards()` that locks every shard, in the same fixed index order every time, and holds all guards for the whole operation — restoring the old single lock's point-in-time linearization for this field. `restore_produce_counter` also now partitions the incoming map by shard *before* taking any locks, then installs each partition with one assignment per shard (no more reset-then-insert window). Locking 256 shards costs nothing meaningful since these are checkpoint-boundary calls, not hot path.

One caveat carried over explicitly in the updated field comment: this makes `produce_counter` atomic *with itself* across all 256 shards, but not *with* `event_log` — those remain two separate locks taken sequentially, exactly as they were before this PR (`create_soft_checkpoint` already did two independent `.lock()` calls for `event_log` then `produce_counter`, not one atomic pair, in the original single-mutex code too). That specific cross-field gap is pre-existing, not introduced or worsened here, and is in scope for the `event_log` PR, not this one.

**`produce_counters` locking a shard per key with no batching** (anthropic, xai) — now groups `produce_refs` by shard index first, locks each needed shard once, and reads all keys for that shard under one guard.

**Missing doc comments on `reset_produce_counter`/`restore_produce_counter`/`produce_counter_shard`** (bedrock, anthropic) — added.

## Disputed, no change

**"Potential deadlock in `produce_counters`" (bedrock, Major)** — this doesn't hold up. The function (before or after the batching fix) never acquires more than one shard lock at a time — it locks a shard, reads, drops the guard, moves to the next. A lock-ordering deadlock requires two threads each holding one lock while waiting on a lock the other holds; that requires holding ≥2 locks simultaneously, which this code never does. (The *new* helpers do hold multiple locks at once, but always in the same fixed index order across every call site, which is exactly the standard way to make multi-lock acquisition deadlock-free — not a source of one.)

**"Inconsistent use of `Arc::new`" (bedrock, Minor)** — comparing `produce_counter: Arc<Vec<Mutex<...>>>` (no per-element `Arc`) against `phase_a_locks: Arc<Vec<Arc<Mutex<()>>>>` (per-element `Arc`). This is intentional, not inconsistent: `phase_a_locks` uses `tokio::sync::Mutex` and needs an owned `Arc<Mutex<()>>` per stripe because `striped_locks::acquire_locks` calls `.clone().lock_owned().await`, producing an `OwnedMutexGuard` that gets held across an `.await` point — that requires an owned `Arc`, not a borrow. `produce_counter` uses plain `std::sync::Mutex`, is never held across an `.await` (`log_produce`/the checkpoint helpers are synchronous), and is only ever accessed through `&self.produce_counter[idx]`, so there's no reason to wrap each element in its own `Arc`.

**Type alias for the shard `Mutex<BTreeMap<Produce, i32>>` type** — the long-line issue itself is already fixed by `cargo fmt` in an earlier commit on this branch. The type now appears in exactly two places (the field and `produce_counter_shard`'s return type); didn't add the alias given how little duplication that is.

All 307 tests in `rspace_plus_plus` still pass, `cargo fmt --all -- --check` is clean.

jltatbeach (approved):

PR #355, opened 2026-08-27
fix(rspace): replace store's RwLock with lock-free ArcSwap

9Neechan:
## What this fixes

`get_store()` (`self.store.read().expect(...).clone()`) is called at least 3x per `produce()`/`consume()` (`produce_lock()`, `locked_produce()`'s `get_joins`, `store_data()`) and never contends a writer on that hot path — writes to `store` only happen at checkpoint/spawn boundaries, replacing the whole pointer wholesale. `std::sync::RwLock` still does atomic RMW on shared reader-count state on every read-lock/unlock, and in isolation this measured genuine **negative** scaling: concurrent 173ms vs sequential 16ms for the same call count — 10x *slower* under concurrency, with zero writer activity.

`store`'s type changes from `Arc<std::sync::RwLock<Arc<Box<dyn HotStore<C, P, A, K>>>>>` to `Arc<arc_swap::ArcSwap<Box<dyn HotStore<C, P, A, K>>>>`. `ArcSwap` is lock-free and matches the actual access pattern exactly (read-mostly, rare wholesale-pointer swaps): `load_full()` is an atomic pointer read, `store()` an atomic pointer write, no reader-count bookkeeping. The two write sites (`revert_to_soft_checkpoint`, `create_new_hot_store`) both already replaced the pointer wholesale, so the swap is a straight drop-in.

## What improved

- The isolated `get_store()` read-lock benchmark went from **10x slower under concurrency (0.09x-0.04x speedup, i.e. negative scaling)** to consistently faster-or-flat, never worse, than sequential.
- New regression test `get_store_read_lock_isolated_cost_at_rholang_par_scale` guards against this regressing back to lock-based contention.

## What this does *not* fix (found and ruled out during the same investigation)

Follow-up to #346 (`produce_counter` fix). While isolating `get_store()`'s cost, this branch also investigated and **ruled out** `event_log` as needing a similar fix — `event_log_isolated_cost_at_rholang_par_scale` (kept as a regression test) confirms it parallelizes fine on its own. It further found that `bench_par_branches`, the benchmark this whole investigation line is anchored on, gives every operation a brand-new never-reused channel, which forces every `get_joins()` call through a cold history-store read path the real `rholang-par` contract essentially never takes (it reuses one channel per branch). That means `bench_par_branches`'s current "no speedup" reading may not represent the real remaining bottleneck — this needs a corrected benchmark before further rspace-layer fixes are pursued. Full writeup: `asi-chain-testbed/issues/02-intra-deploy-par-mutex-rspace.md`, "Correction (2026-08-27)".

This `get_store()` fix stands on its own regardless of that open question — it fixes a genuine bug independent of which benchmark is used to measure it.

Part of the [#50](https://github.com/F1R3FLY-io/f1r3node-rust/issues/50) investigation.


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Verdict:** :bulb: Feedback Provided (80.0000% agreement)

**Reviewed by:**

| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.85 |
| xai | `grok-4.5` | Feedback Provided | 0.85 |

---

### Summary

This is a sound, well-motivated micro-optimization: replacing `RwLock<Arc<Box<dyn HotStore>>>` with `ArcSwap<Box<dyn HotStore>>` exactly matches the read-mostly/rare-wholesale-swap access pattern, and both write sites already replaced the pointer wholesale, making the swap semantically safe (readers could already observe a stale store under the old code since the Arc was cloned out of the lock). The main concerns are around the supporting material rather than the core change: the advertised 'regression tests' contain no assertions and thus guard nothing while adding significant test runtime; the doc comment overstates `load_full()`'s cheapness (it still does a contended atomic refcount increment — `load()` would avoid that where a Guard suffices); a test comment describes the now-deleted implementation; and the Cargo.lock bundles unrelated windows-sys/socket2 bumps. None of these block merge, but the assertion-free tests and stale comments should be addressed.

The ArcSwap conversion itself appears semantically consistent with the previous get_store behavior: callers already cloned the Arc and released the RwLock immediately, so replacing the pointer atomically does not newly allow operations to retain an old store. No security or clear production correctness issue is apparent. The main concerns are that the claimed regression test has no assertion, its comments and output still describe the removed RwLock, diagnostic benchmarks are included as ordinary tests, and the lockfile includes unrelated dependency upgrades.

Replacing the store's RwLock with arc_swap::ArcSwap is a sound fit for the stated access pattern (read-mostly, rare wholesale pointer replacement) and preserves prior semantics where callers already cloned an Arc and dropped the lock before using the HotStore. Write sites correctly use store.store(...), get_store() correctly uses load_full(), and removing lock poisoning on this path is a small robustness win. The main gap is that the new 'regression' test does not assert any performance bound despite the PR claiming it guards against lock-based contention, and its comments/labels still describe the old RwLock implementation. Address the test/docs mismatch; otherwise this is a solid, mergeable performance fix.

The core change — replacing the hot-path RwLock around the hot store with arc_swap::ArcSwap — is correct, well-motivated, and matches the actual read-mostly/rare-swap access pattern; the two write sites were already wholesale pointer replacements, making this a clean drop-in. Main concerns: the 'regression test' contains no assertion and therefore guards nothing (the PR description overstates this), the store swap now publishes before the event_log reset which should be sanity-checked against checkpoint consistency assumptions, and the Cargo.lock diff includes a broad set of unrelated dependency upgrades that should be separated. None of these block merging, but the missing assertion and the lockfile noise should be addressed.

The code change effectively replaces the `RwLock` with `ArcSwap` for the `store` field, which should improve performance by reducing contention under concurrent reads. The change is well-documented and includes regression tests to ensure the fix remains effective. Minor style improvements could be made to the comments for better readability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Claimed regression test does not assert anything</b> (Reported by: openrouter, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1635)

The PR description claims `get_store_read_lock_isolated_cost_at_rholang_par_scale` 'guards against this regressing back to lock-based contention', but the test explicitly states 'No assertion: diagnostic only'. A test that only prints timing via eprintln! will pass silently regardless of performance. If the intent is a regression guard, it needs at least a lenient assertion (e.g. par_ms <= seq_ms * some_factor, or speedup >= 1.0) — possibly gated behind an env var or #[ignore] to avoid CI flakiness. Same applies to `event_log_isolated_cost_at_rholang_par_scale`.

---

The PR description says get_store_read_lock_isolated_cost_at_rholang_par_scale 'guards against this regressing back to lock-based contention', but the test explicitly ends with 'No assertion: diagnostic only' and only eprintln!s timing. It will never fail CI on a performance regression, so it does not protect the fix.

**Suggestion:**
```
Add a bounded assertion on concurrent vs sequential behavior (e.g. speedup >= 1.0, or par_ms <= seq_ms * some tolerance), or reword the PR/test docs to call it a diagnostic benchmark rather than a regression guard. Prefer a stable threshold that won't flake on busy hosts.
```

</details>

<details>
<summary><b>:orange_circle: Major: "Regression test" contains no assertions and cannot guard against regression</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1665)

The PR description claims `get_store_read_lock_isolated_cost_at_rholang_par_scale` "guards against this regressing back to lock-based contention", but the test explicitly ends with `// No assertion: diagnostic only`. It only prints timing via eprintln! and will pass regardless of any performance regression. The same applies to `event_log_isolated_cost_at_rholang_par_scale`. These are benchmarks masquerading as tests; they provide zero regression protection and add ~320,000 iterations of work to every test run.

**Suggestion:**
```
Either add a loose assertion (e.g., `assert!(speedup > 0.5, ...)` to catch catastrophic negative scaling while tolerating CI noise), move these to a Criterion benchmark under `benches/`, or update the PR description to stop calling them regression tests. If kept as tests, consider gating them behind `#[ignore]` so CI time isn't spent on assertion-free diagnostics.
```

</details>

<details>
<summary><b>:orange_circle: Major: load_full() still performs a contended atomic RMW on the Arc refcount</b> (Reported by: anthropic, openrouter, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 98)

`ArcSwap::load_full()` is documented as slower than `load()`: it must increment the inner Arc's strong count, which is an atomic RMW on a single shared cache line — the same class of cross-core contention the PR aims to eliminate from RwLock's reader count. The measured improvement is real (one contended RMW instead of two, and no lock-release RMW), but the doc comment's claim that `load_full()` is "an atomic pointer read ... no reader-count bookkeeping" is inaccurate; the refcount bump is exactly reader-count bookkeeping on shared state.

---

`load_full()` performs an atomic load plus a refcount increment/decrement per call. `ArcSwap::load()` returns a guard that avoids the refcount bump and is cheaper for short-lived borrows. Since `get_store()` is called 3x per produce/consume, using `load()` and returning/dereferencing the guard where lifetimes permit would squeeze a bit more out of the change. Not a blocker — `load_full()` is already strictly better than the RwLock path — but worth noting since the PR is performance-motivated.

---

load_full() matches the previous 'clone the Arc under the lock, then drop the lock' semantics and is the right API for returning Arc<Box<dyn HotStore...>>. The broader cost may still be repeated get_store() calls per produce/consume (PR notes ≥3x), each bumping refcounts.

**Suggestion:**
```
Consider `self.store.load()` with a returned `Guard`, or measure; if callers need an owned Arc for storage in results, `load_full()` is correct as-is.
```

</details>

<details>
<summary><b>:orange_circle: Major: Subtle semantics change: writers no longer wait for in-flight readers (pre-existing, but now structural)</b> (Reported by: anthropic, openrouter, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 261)

In `revert_to_soft_checkpoint` and `create_new_hot_store`, the old code held the store write-lock while updating; with ArcSwap, `self.store.store(...)` publishes the new hot store immediately, before the subsequent `event_log` reset (`revert_to_soft_checkpoint` line ~265). A concurrent reader can now observe the new store while the event log still reflects the pre-revert state. This window likely existed before too (readers never held the store lock across event_log access), so this is probably not a new bug — but the publish ordering is now more explicit and should be confirmed against checkpoint semantics.

---

With the old RwLock, `store.write()` at least serialized against threads currently inside `get_store()`'s lock acquisition. With ArcSwap, `store()` never waits: any task that has already loaded the old store Arc continues operating on the stale HotStore after `revert_to_soft_checkpoint`/`create_new_hot_store` swaps the pointer. In practice this was already true because `get_store()` cloned the Arc and released the read lock immediately, so this is not a new bug — but the change removes even the small synchronization window that existed, and correctness now relies entirely on the checkpoint/spawn-boundary invariant that no produce/consume is in flight during a swap.

---

store remains pub but changed from Arc<RwLock<Arc<Box<dyn HotStore...>>>> to Arc<ArcSwap<Box<dyn HotStore...>>>. In-crate call sites in this diff are updated; any external code using .store.read()/write() would fail to compile. Semantics of swap vs locked replace remain 'wholesale pointer replace', which matches checkpoint/spawn usage.

**Suggestion:**
```
Document the invariant explicitly at both `store()` call sites (revert_to_soft_checkpoint, create_new_hot_store): callers must guarantee no concurrent produce/consume operations are in flight when the hot store is replaced.
```

</details>

<details>
<summary><b>:orange_circle: Major: Unrelated transitive dependency bumps bundled into the PR</b> (Reported by: anthropic, openrouter)</summary>

**File:** `Cargo.lock` (line 1065)

Beyond adding arc-swap, the lockfile bumps windows-sys (0.48/0.52/0.59 → 0.61.2) and socket2 (0.5.10 → 0.6.5) across several unrelated crates (hyper-util, quinn, rustix, tempfile, etc.). These are not required by the arc-swap addition and widen the change's blast radius, making bisection harder if a dependency bump causes issues.

---

The Cargo.lock changes include upgrades of `windows-sys` to 0.61.2, `socket2` to 0.6.5, and related changes across mio, hyper-util, iroh, rustix, tempfile, etc. These are unrelated to adding `arc-swap` and appear to be the result of a broad `cargo update`. Mixing unrelated dependency upgrades into a targeted perf fix increases review surface and bisection risk if one of those upgrades introduces a regression.

**Suggestion:**
```
Regenerate the lockfile with a minimal update (`cargo update -p arc-swap` or `cargo add` without a full `cargo update`) so the PR only contains the dependency it actually needs, or split the dependency bumps into a separate chore PR.
```

</details>

<details>
<summary><b>:orange_circle: Major: Single-line function body deviates from surrounding formatting</b> (Reported by: anthropic, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 98)

`pub fn get_store(&self) -> Arc<Box<dyn HotStore<C, P, A, K>>> { self.store.load_full() }` is collapsed onto one line, unlike every other method in this impl block. Standard rustfmt would expand this.

---

get_store was collapsed to a single line while the struct field gained a large inline rationale block. Functionally fine, but inconsistent with typical multi-line method formatting and makes the struct definition harder to scan.

**Suggestion:**
```
Keep a short field comment (read-mostly, rare wholesale swap → ArcSwap) and move the benchmark/issue narrative next to the diagnostic test or a module-level note. Format get_store as a normal multi-line method.
```

</details>

<details>
<summary><b>:orange_circle: Major: Test comment references the old RwLock implementation that this PR removes</b> (Reported by: anthropic, xai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1621)

After the ArcSwap migration, the test comment still says get_store() is self.store.read().expect(...).clone() / a std::sync::RwLock read lock, and the eprintln label still says 'get_store() RwLock read isolated cost'. That is misleading for future readers debugging #50.

---

The doc comment on `get_store_read_lock_isolated_cost_at_rholang_par_scale` says "get_store() (rspace.rs:87-89) is `self.store.read().expect(...).clone()` — a std::sync::RwLock read lock", describing the pre-PR implementation (and a hardcoded line number that is already wrong). After this PR merges, the comment is stale on arrival.

**Suggestion:**
```
Rewrite the comment in past tense ("was previously a RwLock read lock...") and drop the hardcoded line numbers, framing the test as measuring the current lock-free implementation against that historical baseline.
```

</details>

<details>

**File:** `rspace++/src/rspace/rspace.rs` (line 55)

The comment explaining the change could be formatted for better readability.

**Suggestion:**
```
Reformat the comment for better readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Diagnostic microbenchmarks run as ordinary tests</b> (Reported by: openai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1558)

Both added tests perform timing experiments with many allocations, mutex operations, and spawned tasks but make no correctness assertions. Running diagnostic benchmarks as regular tokio tests increases normal test-suite cost while producing results that are sensitive to CI load and build profile.

**Suggestion:**
```
Move these measurements to Criterion or the repository's benchmark infrastructure. If retaining them in the test module is useful, consider marking purely diagnostic tests with #[ignore] and documenting how to run them.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: New test documentation still describes the removed RwLock</b> (Reported by: openai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1594)

The comment says get_store() uses self.store.read() and explains that the test isolates an RwLock read, while the implementation now uses ArcSwap::load_full(). The emitted diagnostic at line 1640 also labels the measurement as an RwLock read. This makes the test misleading after the change.

**Suggestion:**
```
Update the comment and diagnostic text to describe ArcSwap loads and frame the test as validating the replacement implementation. Remove obsolete call-count claims if they are not kept synchronized with the production paths.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Performance regression test has no assertion</b> (Reported by: openai)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1647)

The new get_store_read_lock_isolated_cost_at_rholang_par_scale test only prints timing information and therefore cannot guard against reintroducing the RwLock contention described by the PR. It will pass regardless of whether concurrent access becomes substantially slower than sequential access.

**Suggestion:**
```
Either add a deliberately tolerant regression assertion based on repeated samples, move this measurement into a benchmark suite, or rename and mark it ignored if it is intended to be diagnostic only. Avoid a strict wall-clock assertion that would be flaky in CI.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Lockfile contains unrelated transitive dependency upgrades</b> (Reported by: openai)</summary>

**File:** `Cargo.lock` (line 1053)

Adding arc-swap, whose shown dependency is only rustversion, also upgrades socket2 and several windows-sys dependency selections used by unrelated networking and platform crates. These changes broaden the PR's effective scope and make it harder to attribute platform or networking regressions.

**Suggestion:**
```
Regenerate the lockfile while preserving unrelated locked versions where possible, or explain why these transitive upgrades are required and ensure affected platform/network builds are covered by CI.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: history_repository still uses the same RwLock<Arc<...>> pattern</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 55)

`history_repository` retains the identical `Arc<RwLock<Arc<Box<dyn ...>>>>` shape that this PR identifies as pathological for read-mostly/rare-wholesale-swap access. If it is read on any hot path, it will exhibit the same negative scaling.

**Suggestion:**
```
Evaluate whether `history_repository` has the same access pattern and, if so, migrate it to ArcSwap in a follow-up (or note in the field comment why it was intentionally left as-is).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test comment is stale relative to the new implementation</b> (Reported by: openrouter)</summary>

**File:** `rspace++/src/rspace/rspace.rs` (line 1605)

The comment above `get_store_read_lock_isolated_cost_at_rholang_par_scale` still describes get_store() as `self.store.read().expect(...).clone()` — a std::sync::RwLock read lock — but this PR changes it to `load_full()`. The comment describes the pre-fix state; after merge it reads as documenting the current implementation, which is now incorrect.

**Suggestion:**
```
Update the comment to describe the benchmark as a historical-isolation test for the (now-removed) RwLock read path, noting it now measures the ArcSwap load path as a regression sentinel.
```

</details>
---

<details>

This is a sound, well-motivated micro-optimization: replacing `RwLock<Arc<Box<dyn HotStore>>>` with `ArcSwap<Box<dyn HotStore>>` exactly matches the read-mostly/rare-wholesale-swap access pattern, and both write sites already replaced the pointer wholesale, making the swap semantically safe (readers could already observe a stale store under the old code since the Arc was cloned out of the lock). The main concerns are around the supporting material rather than the core change: the advertised 'regression tests' contain no assertions and thus guard nothing while adding significant test runtime; the doc comment overstates `load_full()`'s cheapness (it still does a contended atomic refcount increment — `load()` would avoid that where a Guard suffices); a test comment describes the now-deleted implementation; and the Cargo.lock bundles unrelated windows-sys/socket2 bumps. None of these block merge, but the assertion-free tests and stale comments should be addressed.

The code change effectively replaces the `RwLock` with `ArcSwap` for the `store` field, which should improve performance by reducing contention under concurrent reads. The change is well-documented and includes regression tests to ensure the fix remains effective. Minor style improvements could be made to the comments for better readability.

#### openai (gpt-5.6-sol)

The ArcSwap conversion itself appears semantically consistent with the previous get_store behavior: callers already cloned the Arc and released the RwLock immediately, so replacing the pointer atomically does not newly allow operations to retain an old store. No security or clear production correctness issue is apparent. The main concerns are that the claimed regression test has no assertion, its comments and output still describe the removed RwLock, diagnostic benchmarks are included as ordinary tests, and the lockfile includes unrelated dependency upgrades.

The core change — replacing the hot-path RwLock around the hot store with arc_swap::ArcSwap — is correct, well-motivated, and matches the actual read-mostly/rare-swap access pattern; the two write sites were already wholesale pointer replacements, making this a clean drop-in. Main concerns: the 'regression test' contains no assertion and therefore guards nothing (the PR description overstates this), the store swap now publishes before the event_log reset which should be sanity-checked against checkpoint consistency assumptions, and the Cargo.lock diff includes a broad set of unrelated dependency upgrades that should be separated. None of these block merging, but the missing assertion and the lockfile noise should be addressed.

#### xai (grok-4.5)

Replacing the store's RwLock with arc_swap::ArcSwap is a sound fit for the stated access pattern (read-mostly, rare wholesale pointer replacement) and preserves prior semantics where callers already cloned an Arc and dropped the lock before using the HotStore. Write sites correctly use store.store(...), get_store() correctly uses load_full(), and removing lock poisoning on this path is a small robustness win. The main gap is that the new 'regression' test does not assert any performance bound despite the PR claiming it guards against lock-based contention, and its comments/labels still describe the old RwLock implementation. Address the test/docs mismatch; otherwise this is a solid, mergeable performance fix.
</details>

---

9Neechan:
Addressed the review (`d41d9c7e6`, `c62a67925`).

## Addressed

- No assertions in the "regression" tests → added `assert!(par_ms <= seq_ms.saturating_mul(5), ...)` to both `event_log_isolated_cost_at_rholang_par_scale` and `get_store_read_lock_isolated_cost_at_rholang_par_scale`.
- Stale RwLock comments/labels describing the removed implementation → deleted, not reworded.
- `load_full()` doc comment overstated cheapness ("no reader-count bookkeeping") → corrected: it does one atomic refcount RMW, cheaper than RwLock's read-lock-plus-unlock pair but not free.
- Unrelated `windows-sys`/`socket2` bumps in `Cargo.lock` → reverted; diff against `dev` is now just the `arc-swap` entry (20 lines), verified against `cargo test -p rspace_plus_plus --lib`.
- `store`/`event_log` swap ordering across checkpoint/spawn → invariant (no in-flight produce/consume during a swap) documented explicitly at both `store.store()` call sites.
- `history_repository`'s identical `RwLock<Arc<...>>` shape → noted directly in the field comment why it's out of scope: its only reader is the rare `spawn()`, not the hot path.

## Not changed

- `get_store()`'s single-line body → `cargo fmt -p rspace_plus_plus -- --check` is clean on it; not a project style violation.
- Move the two diagnostic tests to Criterion/`#[ignore]` → matches the existing pattern already merged in this file (#346/#345); the added assertions above cover the actual complaint (they "guard nothing") without a structural rewrite.

jeffrey-l-turner (approved):

PR #356, opened 2026-08-27
chore(rspace): add channel-reuse variant of bench_par_branches

9Neechan:
## Why

`bench_par_branches` gives every `produce()` call a brand-new, never-reused channel. That means every `get_joins()` call misses `HotStore`'s in-memory `joins` cache and falls through to a real history-store read (`hot-store.get_joins.history_fill` fires on 100% of calls in that benchmark — confirmed via its existing instrumentation). The real `rholang-par` contract this whole investigation is about does not do this: each branch's `loop` channel is created once (`new loop in { contract loop(@n) = ... }`) and reused for every recursive `loop!(n-1)` within that branch, so the real cache-miss rate is ~N/(N×iters), not 100%.

This means `bench_par_branches`'s long-standing "no speedup, 0.88–0.96×" reading — the number PR #72's own internal rspace-layer attribution and this entire follow-up investigation ([#50](https://github.com/F1R3FLY-io/f1r3node-rust/issues/50)) are anchored on — may be measuring an artifact of the benchmark's own design, not the real workload's rspace-layer behavior.

## What this adds

`bench_par_branches_channel_reuse`: identical harness to `bench_par_branches` (same produce()-only workload, same CLI), except each of the `N` branches gets **one** channel reused for all its `ops_per_branch` calls, isolating that single variable.

## Result

At 32×5,000 (the exact branch/op count the real `rholang-par` contract uses):

| | `bench_par_branches` (unique channel/op) | `bench_par_branches_channel_reuse` (one channel/branch) |
|---|---|---|
| sequential | 637ms | 908ms |
| concurrent | 698ms | 245ms |
| **speedup** | **0.91×** (no real parallelism) | **3.5–3.7×** (reproducible across 4 runs) |

Against this laptop's realistic ceiling (~8 worker threads, not 32), 3.5× is ~44% efficiency — real, if partial, parallelism, consistent with PR #72's own "PARTIALLY PARALLEL" verdict. This supports the hypothesis above: the channel-uniqueness design is very likely responsible for most or all of `bench_par_branches`'s "no speedup" reading.

## A second finding surfaced along the way (not yet investigated further)

At larger scale (32×50,000) this new benchmark's own speedup drops sharply (1.87×). Traced to `HotStore::put_datum`'s `existing.insert(0, d)` ([`hot_store.rs:646`](https://github.com/F1R3FLY-io/f1r3node-rust/blob/dev/rspace%2B%2B/src/rspace/hot_store.rs#L646)) — an O(n) front-insert on the per-channel data `Vec`. This accumulates unboundedly *in this benchmark* because it's produce()-only with no `consume()` to drain the channel — unlike the real contract, whose persistent `contract loop(@n)` consumes every produce immediately, keeping stored data near-zero. Flagging this as a separate candidate finding (same `Vec::insert(0, …)` O(n) pattern PR #72's item #1 fixed for `event_log`/`produce_counter`, apparently not applied to `HotStore`'s own per-channel data), not a contradiction of the primary 32×5,000 result above — that comparison is not affected since it uses the real contract's own op count, before this accumulation effect has room to dominate.

Part of the [#50](https://github.com/F1R3FLY-io/f1r3node-rust/issues/50) investigation, follow-up to #346 and #355.


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Reviewed by:**

| anthropic | `claude-fable-5` | Approve | 0.9 |
| openrouter | `moonshotai/kimi-k3` | Abstain | 0.5 |

---

### Summary

This PR adds a single new standalone benchmark binary that is a well-documented, minimal variant of the existing `bench_par_branches`, changing exactly one variable (channel reuse per branch instead of a unique channel per op) to test a specific, well-articulated hypothesis about a benchmark-design artifact skewing prior parallelism measurements. The change touches no production code, has no security surface, and the extensive header comment plus PR description clearly explain the motivation, methodology, and results. The only findings are minor robustness/methodology suggestions (millisecond-resolution timing risking div-by-zero at tiny scales, silent fallback on malformed CLI args, no warm-up phase, and a possible rustfmt deviation), none of which affect the validity of the headline 32×5,000 comparison. Safe to merge as-is.

The new benchmark correctly isolates channel reuse and has no apparent security or substantive concurrency defects. Its main weakness is edge-case handling in benchmark reporting: zero or sufficiently small workloads can yield NaN or infinite speedup because durations are truncated to whole milliseconds and inputs are not constrained. Addressing input validation and duration precision would make the benchmark output reliable.

Solid, well-motivated benchmark twin that correctly isolates per-branch channel reuse versus unique-channel-per-op. No security concerns; logic matches the stated harness (sequential one-channel-per-branch vs concurrent tokio tasks on a shared RSpace). Main actionable nits are millisecond timer / zero-branch edge cases for speedup math, aligning or documenting the 500 vs 5000 default, and calling out produce-only datum accumulation so large-N runs are not misread. Safe to merge after minor hardening of the reporting path if desired.

The code change introduces a new benchmark variant that reuses channels, which helps in isolating the effect of channel reuse on performance. The code is functional and safe, but there are minor style and performance improvements that could be made, along with additional documentation to enhance clarity.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential division by zero / infinity in speedup calculation</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 106)

Both elapsed durations are converted with `as_millis()`, which truncates sub-millisecond runs to zero. Consequently, small workloads can report infinite or NaN speedup. The CLI also accepts `branches=0` or `ops_per_branch=0`, producing zero-duration workloads and, for zero branches, an additional division by zero in the efficiency calculation. These results are misleading for a benchmark even though floating-point division does not panic.

---

If `par_ms` is 0 (which can happen with very small branch/op counts, since `as_millis()` truncates sub-millisecond durations), `seq_ms as f64 / par_ms as f64` yields `inf` or `NaN`. The default parameters make this unlikely, but a user passing e.g. `1 1` on the CLI would produce a misleading result.

---

par_ms is taken from as_millis(). On a fast machine or with small ops_per_branch, elapsed can be 0ms, so `seq_ms as f64 / par_ms as f64` becomes inf/NaN. The same risk exists for branches=0 when computing efficiency (`speedup / branches as f64`).

**Suggestion:**
```
Reject zero values for both arguments and calculate using untruncated durations, for example `let seq_secs = t_seq.elapsed().as_secs_f64()` and the corresponding parallel value. Consider retaining the full `Duration` for display and explicitly handling a zero measured duration.
```

</details>

<details>
<summary><b>:orange_circle: Major: No warm-up run or repetition; single-shot timing may be noisy</b> (Reported by: anthropic, bedrock)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 73)

Cloning strings in a loop can be inefficient. Consider using string references or a more efficient data structure if possible.

---

The sequential pass runs first on a cold process (allocator warm-up, code paths not yet JIT'd by branch predictors, tokio runtime spin-up), then the concurrent pass runs on a warmed process. This asymmetry can slightly favor the concurrent measurement. The PR description mentions reproducing results across 4 manual runs, which mitigates this, but the harness itself does one measurement per mode.

**Suggestion:**
```
Consider an optional warm-up iteration and/or reporting min/median across a few in-process repetitions. Only worth doing if you want the number to stand alone without manual re-runs — and any change should be mirrored in `bench_par_branches` to keep the A/B comparison valid.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary `unwrap()` calls</b> (Reported by: bedrock)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 114)

The use of `unwrap()` can lead to panics if the operation fails. It would be better to handle these cases more gracefully, perhaps by logging an error and exiting cleanly.

**Suggestion:**
```
Consider using `expect` with a meaningful error message or proper error handling with `match` statements.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Document produce-only accumulation caveat in the binary header</b> (Reported by: bedrock, xai)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 1)

The PR correctly notes that at large ops_per_branch, HotStore::put_datum’s insert(0, …) dominates because nothing consumes. Future readers running 32×50_000 may think channel-reuse parallelism “disappeared” without that context.

---

While the code is well-commented, adding more detailed comments on the logic within the loops and async functions could improve readability and maintainability.

**Suggestion:**
```
Add a short comment that this bench is produce-only (datums accumulate per channel), so large ops_per_branch is not comparable to the real contract’s consume-balanced loop, and point at the put_datum O(n) finding.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Single-line function body may not match rustfmt defaults</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 42)

`fn get(&self, _: &Wildcard, a: &String) -> Option<String> { Some(a.clone()) }` is written on one line; default rustfmt would expand the body onto separate lines. If the repo runs `cargo fmt --check` in CI this could fail; if this style intentionally mirrors `bench_par_branches`, it's fine.

**Suggestion:**
```
Run `cargo fmt` on the new file to match project formatting conventions.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Invalid CLI arguments silently fall back to defaults</b> (Reported by: anthropic)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 55)

`.and_then(|s| s.parse().ok()).unwrap_or(32)` means a typo like `bench_par_branches_channel_reuse 32x 5000` silently runs with the default value instead of erroring. For a benchmark whose entire purpose is comparing exact branch/op counts against `bench_par_branches`, a silent fallback could invalidate a comparison without the user noticing. (This presumably mirrors the original benchmark for consistency, which is a reasonable tradeoff.)

**Suggestion:**
```
Consider `expect`-ing on parse failure (or printing a warning) so a malformed argument aborts loudly rather than silently benchmarking the wrong configuration. If changed, apply the same change to `bench_par_branches` to keep the harnesses identical.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Default ops_per_branch differs from the headline comparison point</b> (Reported by: xai)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 59)

Default is 500, while the PR’s primary claim and the real rholang-par shape use 5,000 (32×5,000). Easy to mis-run and compare against the wrong baseline when invoking with no args.

**Suggestion:**
```
Either default to 5000 to match the investigation, or print a one-line hint that the issue #50 comparison used 32 5000.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Total operation count can overflow</b> (Reported by: openai)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 69)

`branches * ops_per_branch` uses unchecked `usize` multiplication. Very large command-line values will panic in overflow-checked builds or wrap in other builds, causing the displayed total to be incorrect before the likely resource exhaustion.

**Suggestion:**
```
Use `checked_mul` and report an invalid argument when the requested total cannot be represented.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Hot-loop String allocations are fine but worth keeping parity-noted</b> (Reported by: xai)</summary>

**File:** `rspace++/src/bin/bench_par_branches_channel_reuse.rs` (line 94)

channel.clone() and "datum".to_string() run every produce. Acceptable if bench_par_branches does the same (apples-to-apples). If the original later switches to static/&str-friendly APIs, this twin should stay in lockstep.

**Suggestion:**
```
No change required for merge; optionally bind datum once outside the inner loop (let datum = "datum".to_string();) if the produce API can take a clone from a single source—only if the sibling bench does likewise.
```

</details>
---

<details>

This PR adds a single new standalone benchmark binary that is a well-documented, minimal variant of the existing `bench_par_branches`, changing exactly one variable (channel reuse per branch instead of a unique channel per op) to test a specific, well-articulated hypothesis about a benchmark-design artifact skewing prior parallelism measurements. The change touches no production code, has no security surface, and the extensive header comment plus PR description clearly explain the motivation, methodology, and results. The only findings are minor robustness/methodology suggestions (millisecond-resolution timing risking div-by-zero at tiny scales, silent fallback on malformed CLI args, no warm-up phase, and a possible rustfmt deviation), none of which affect the validity of the headline 32×5,000 comparison. Safe to merge as-is.

The code change introduces a new benchmark variant that reuses channels, which helps in isolating the effect of channel reuse on performance. The code is functional and safe, but there are minor style and performance improvements that could be made, along with additional documentation to enhance clarity.

#### openai (gpt-5.6-sol)

The new benchmark correctly isolates channel reuse and has no apparent security or substantive concurrency defects. Its main weakness is edge-case handling in benchmark reporting: zero or sufficiently small workloads can yield NaN or infinite speedup because durations are truncated to whole milliseconds and inputs are not constrained. Addressing input validation and duration precision would make the benchmark output reliable.

> **ABSTAIN**: {
    "verdict": "approve",
    "confidence": 0.85,
    "issues": [
        {
            "severity": "minor",
            "category": "logic",
            "file": "rspace++/src/bin/bench_par_branches_channel_reuse.rs",
            "line": 104,
            "title": "Potential division by zero in speedup calculation",
            "description": "If par_ms is 0 (possible with very small branch/op counts on a fast machine, since as_millis() truncates), the speedup becomes inf/NaN. Not a real concern at the intended scale (32x5000), but a guard or using as_micros() would make the tool more robust for small smoke-test runs.",
            "suggestion": "Use t_par.elapsed().as_micros() (and same for seq) for the ratio computation, or clamp par_ms to a minimum of 1."
        },
        {
            "severity": "suggestion",
            "category": "performance",
            "file": "rspace++/src/bin/bench_par_branches_channel_reuse.rs",
            "line": 74,
            "title": "Per-op channel.clone() and datum allocation included in measured time",
            "description": "Each produce() call clones the channel String and constructs a new \"datum\".to_string(). These allocations are inside the timed region and are identical in both sequential and concurrent paths, so the comparison is fair — but they add constant overhead that slightly dilutes the measured rspace-layer effect. Since the goal is isolating rspace behavior, hoisting a pre-built datum clone (still required by the API) is unavoidable, but it's worth a comment noting this overhead is part of the measurement.",
            "suggestion": "Optionally add a brief comment noting the per-op String allocations are inside the timed region and equal across both paths."
        },
        {
            "severity": "suggestion",
            "category": "documentation",
            "file": "rspace++/src/bin/bench_par_branches_channel_reuse.rs",
            "line": 113,
            "title": "Efficiency metric measured against branch count, not available cores",
            "description": "The printed 'efficiency' is speedup/branches, which understates efficiency when branches (32) exceeds actual worker threads (~8 on the author's

#### xai (grok-4.5)

Solid, well-motivated benchmark twin that correctly isolates per-branch channel reuse versus unique-channel-per-op. No security concerns; logic matches the stated harness (sequential one-channel-per-branch vs concurrent tokio tasks on a shared RSpace). Main actionable nits are millisecond timer / zero-branch edge cases for speedup math, aligning or documenting the 500 vs 5000 default, and calling out produce-only datum accumulation so large-N runs are not misread. Safe to merge after minor hardening of the reporting path if desired.
</details>

---

9Neechan:
Addressed the review (`db6a1f0e9`).

## Addressed
- `as_millis()` truncation → NaN/inf speedup at tiny scales → reject `branches=0`/`ops_per_branch=0` up front; speedup now computed from `Duration::as_secs_f64()` instead of truncated ms.
- `branches * ops_per_branch` unchecked overflow → `checked_mul`, exits with a message instead of wrapping/panicking mid-print.
- Default `ops_per_branch` (500) didn't match the issue #50 headline comparison (32×5000) → default changed to 5000.
- Missing produce-only accumulation caveat → added to the header comment (points at `put_datum`'s O(n) insert for large `ops_per_branch` runs).

## Not changed
- No warm-up/repetition — would diverge from `bench_par_branches`'s identical single-shot methodology and invalidate the A/B comparison this file exists for.
- `unwrap()` on `produce()`/task-join — intentional for a benchmark binary (fail loud on unexpected error, not swallow it into bogus timing).
- Single-line `get()` impl — `cargo fmt -p rspace_plus_plus -- --check` is clean on it; mirrors `bench_par_branches`'s identical line.
- Silent fallback on unparseable CLI args — mirrors `bench_par_branches` for harness-identical behavior; changing one without the other breaks the pairing.
- Per-op `String` allocations — identical in both sequential and concurrent paths, so the comparison stays fair; reviewer's own note says no change needed.

jeffrey-l-turner (approved):

Issue #24, opened 2026-04-17
perf: block replay pipeline bottleneck limits sustained deploy throughput


jeffrey-l-turner:
> **Migrated from F1R3FLY-io/f1r3node#474** — originally filed 2026-04-08 by @spreston8.
> This issue was moved as part of the f1r3node → f1r3node-rust deprecation (EPOCH-005).
> Original: https://github.com/F1R3FLY-io/f1r3node/issues/474

## Summary

Load testing on a 3-validator shard (FTT=0.1, heartbeat 2s) shows sustained deploy throughput tops out at ~5 deploys/sec before finalization degrades. The bottleneck is the block replay pipeline — each peer block takes 500-1500ms to validate, even for trivial deploys.

## Current Performance (after PR #473 optimizations)

| Deploy Rate | Inclusion p50 | Finalization p50 | LFB Rate |
|-------------|---------------|------------------|----------|
| 1/sec | 1.7s | 4.7s | 40.3 blk/min |
| 5/sec | 25.7s | 49.7s | 9.4 blk/min |
| 10/sec | 58.9s | 95.3s | 3.8 blk/min |

## Detailed Merger Breakdown (high load, avg per merge)

| Step | Time | Notes |
|------|------|-------|
| branches (depends O(D²)) | 349ms | Pre-computed derived sets (was 731ms) |
| conflicts_map (O(B²)) | 860ms | Cached branch data (was 2809ms) |
| rejection_options | 0ms | |
| channel_reads (storage) | 1ms | |
| combine_changes | 1ms | |
| compute_trie_actions | ~2ms | |
| apply_trie_actions | ~18ms | |

The conflicts_map (860ms) and branches (349ms) remain the dominant costs. Both are O(N²) pairwise comparisons — the caching in PR #473 reduced per-call cost but the quadratic iteration count remains.

## Replay Breakdown (high load, avg per block)

| Step | Time |
|------|------|
| user_deploys replay | 3962ms |
| system_deploys | 114ms |
| checkpoint | 44ms |
| reset | 3ms |

The `user_deploys` replay (3962ms) for trivial `@N!(N)` deploys indicates the cost is in runtime spawn and state management, not Rholang execution.

## Root Cause

Feedback loop: more load → more blocks → larger unfinalized DAG scope → slower conflict detection (O(N²)) → slower replay → finalization falls behind → more blocks accumulate.

## Already Optimized (PR #473)

- [x] Removed adaptive deploy cap
[...]

PR #362, opened 2026-08-27
perf(casper): specify and batch the settled-effect probes (one walk per merge)

jeffrey-l-turner:
Specifies AND remediates the settled-effect probe cost behind the sustained finalization lag.

Addresses #24 (root-cause attribution: https://github.com/F1R3FLY-io/f1r3node-rust/issues/24#issuecomment-5445499814). The probe semantics preserved here are the settled-content protection introduced for #341.

## Why

Soak run [33099406770](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33099406770) (master `af1e5209d`, first run with the parents-post-state sub-stage instrumentation) attributed the sustained finalization lag: **92% of the 2.16s average `merge_call` cost is the per-sig settled probes** (`sig_settled_in_base` / `sig_settled_in_floor`) — ~30 probes per merge, each an O(depth) lineage walk with a full block-body load per step. Walk depth is `tip − (floor − deploy_lifespan)`, so lag deepens the walk and the walk deepens the lag. Because the probes shape consensus content (rejected-deploy records checked via `InvalidRejectedDeploy`), the semantics were pinned formally before the optimization.

## Specification (commits 1–2)

- **`docs/claims/settled-effect-probe-equivalence.md`** — CLAIM-FINALITY-001: the reference semantics of `effect_in_state_of` (lineage stepping rules, `BlockNotHeld` deferral, probe-site bindings), claims C1–C4, seam premises, the model↔Rust mapping, and the discharge plan.
- **`formal/rocq/finalized_floor/theories/SettledEffectProbe.v`** — mechanizes the walk algebra: `walk_collect_equiv` (batching), `walk_segment_composition` (splitting), `walk_memo_false_stable` (memoization), `walk_true_stable` (TRUE stability). Zero `Admitted`, no axioms; registered in `_CoqProject`; the `check-finalized-floor-ALL.sh` gate asserts all four alongside the domain capstones (14 → 18 headline results, verified 18/18 under Rocq 9.1.1).

## Implementation (commit `c9aff7732`)

- **`deploy_lifecycle.rs`** — `settled_sigs_of_lineage`: the batched form of the reference walk (one walk, every sig), backed by a content-addressed per-block `LineageStep` cache (hard cap 16,384) so repeat merges answer without deserializing block bodies. `effect_in_state_of` is untouched as the specification oracle.
- **`interpreter_util.rs`** — both probe closures build their applied-sig sets lazily on the first probe (base: one walk of the main parent's lineage; floors: union over settled floors) and answer by membership. Bounds are byte-identical to the previous closures; a merge that never probes never walks.
- **Telemetry** — new `settled-index.build.time` / `settled-index.blocks` histograms time the batched walk; the existing `settled-probe.wrapper` counters keep running and now measure membership lookups, giving the before/after evidence on the next soak preflight.

## Claim discharge status

1. Reference walk kept as oracle — **done**.
2. Generative equivalence test (12 seeded lineages × failed/non-failed deploys, `applied_from_scope`, decoy sigs on non-base parents, absent sigs, three bounds, cache-hit second pass) plus a fail-closed test pinning the one documented divergence (a gapped segment refuses with typed `BlockNotHeld` where the per-sig walk could answer above the gap) — **done**.
3. CbC evidence recorded for both touched `cbc=mandatory` artifacts, citing the claim — **done**.
4. `InvalidRejectedDeploy` equality end-to-end — **open**: rides the casper suites now (1,259/1,259 green in release) and the next soak preflight for the sustained-load run.

## Review remediations

**First review (spec-only diff):**
- **Major — gate omitted `walk_true_stable`**: fixed; the gate asserts all four probe theorems, and the claim doc says "four probe theorems".
- **Major — unreadable single-line pass message**: fixed; the gate's pass output is a short verdict line plus a five-bullet per-family breakdown.
- **Major — PR described as WIP**: fixed; this description replaces the WIP stub.

**Second review (spec + implementation diff), remediated in `1e166ce3a`:**
- **Critical + 2 majors — cache keyed by hash alone (test isolation, availability semantics)**: fixed. A lineage-step cache hit is revalidated against the caller's store via the new `KeyValueBlockStore::contains_key` (raw key check, no decompress/decode), so the walk stays a function of the supplied store and `BlockNotHeld` semantics survive the cache.
- **Major — floor probe lost the per-floor short-circuit**: fixed. `FloorSettledProbe` scans floors in the reference loop's order with lazily built per-floor sets; a floor after the answering one is never read. Two new tests pin equivalence and gap behavior.
- **Major — entry-count cap did not enforce the memory bound**: fixed. The cache budget is byte-tracked from measured sig bytes; clear-rebuild thrash is unreachable because the floor-distance merge backstop bounds walk depth far below the budget (now documented).
- **Major — acceptance criterion open**: intentionally open. Discharge item 4 (`InvalidRejectedDeploy` equality under sustained load) requires the soak preflight run; code cannot close it.
- **Minors**: separate floor-index build metrics; the gate's axiom-free count now derives from its own assertion list instead of a hardcoded number; claim-doc references use function names, not line numbers; the availability seam premise records the per-floor short-circuit; evidence records reference the landed commit; the Rocq module doc names all four results; RefCell sync-invariant and first-probe-spike attribution are commented. bedrock's two unspecified style notes carried no actionable content.

**CI note**: `cargo deny` was red on every branch after `chacha20 0.10.1` was yanked upstream (the scheduled run was last green 2026-08-24); this PR carries the two-line lock bump to 0.10.2.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -852,9 +852,9 @@ checksum = "f079e83a288787bcd14a6aea84cee5c87a67c5a3e660c30f557a3d24761b3527"
 
 [[package]]
 name = "chacha20"
-version = "0.10.1"
+version = "0.10.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d524456ba66e72eb8b115ff89e01e497f8e6d11d78b70b1aa13c0fbd97540a81"
+checksum = "65c35e4b699c7e15ccbe7ee35c005e4fc0a278d22238a2857e6ce2dadeda1b06"
 dependencies = [
  "cfg-if",
  "cpufeatures 0.3.0",
--- a/block-storage/src/rust/key_value_block_store.rs
+++ b/block-storage/src/rust/key_value_block_store.rs
@@ -245,6 +245,19 @@ impl KeyValueBlockStore {
         }
     }
 
+    /// Key-existence check against the underlying store, skipping the
+    /// decompression and protobuf decode that `contains` (via `get`) pays.
+    /// The cheap availability probe for callers revalidating cached
+    /// per-block facts against THIS store.
+    pub fn contains_key(&self, block_hash: &BlockHash) -> Result<bool, KvStoreError> {
+        Ok(self
+            .store
+            .contains(&vec![block_hash.to_vec()])?
+            .first()
+            .copied()
+            .unwrap_or(false))
+    }
+
     fn error_approved_block(cause: String) -> String {
         format!("Approved block decoding error. Cause: {}", cause)
     }
--- a/casper/src/rust/finality/deploy_lifecycle.rs
+++ b/casper/src/rust/finality/deploy_lifecycle.rs
@@ -161,22 +161,47 @@ struct LineageStep {
     sigs: std::sync::Arc<HashSet<Bytes>>,
 }
 
-/// Hard cap for the per-block lineage-step cache. At roughly a kilobyte
-/// per fat block the full cache stays under ~16MB; on overflow the cache
-/// is cleared rather than evicted piecewise — entries are pure functions
-/// of immutable bodies, so losing them costs only a re-read.
-const LINEAGE_STEP_CACHE_MAX: usize = 16_384;
+/// Byte budget for the per-block lineage-step cache, tracked from each
+/// entry's measured sig bytes (an entry-count cap understates fat blocks:
+/// 128 sigs × ~70B is ~9KB for one entry). On overflow the cache is
+/// cleared rather than evicted piecewise — entries are pure functions of
+/// immutable bodies, so losing them costs only a re-read. Repeated
+/// clear-rebuild thrash would need one walk's entries to approach the
+/// budget, and walk depth is bounded far below it by the deterministic
+/// floor-distance merge backstop (`merge_scope_backstop_exceeded`).
+const LINEAGE_STEP_CACHE_MAX_BYTES: usize = 32 * 1024 * 1024;
 
-fn lineage_step_cache(
-) -> &'static parking_lot::Mutex<HashMap<BlockHash, std::sync::Arc<LineageStep>>> {
-    static CACHE: std::sync::OnceLock<
-        parking_lot::Mutex<HashMap<BlockHash, std::sync::Arc<LineageStep>>>,
-    > = std::sync::OnceLock::new();
-    CACHE.get_or_init(|| parking_lot::Mutex::new(HashMap::new()))
+/// Per-entry overhead estimate added to the measured sig bytes: map slot,
+/// `Arc` headers, hash key.
+const LINEAGE_STEP_ENTRY_OVERHEAD_BYTES: usize = 128;
+
+#[derive(Default)]
+struct LineageStepCache {
+    map: HashMap<BlockHash, std::sync::Arc<LineageStep>>,
+    approx_bytes: usize,
+}
+
+impl LineageStepCache {
+    fn insert(&mut self, hash: BlockHash, step: std::sync::Arc<LineageStep>) {
+        let entry_bytes =
+            LINEAGE_STEP_ENTRY_OVERHEAD_BYTES + step.sigs.iter().map(Bytes::len).sum::<usize>();
+        if self.approx_bytes + entry_bytes > LINEAGE_STEP_CACHE_MAX_BYTES {
+            self.map.clear();
+            self.approx_bytes = 0;
+        }
+        self.approx_bytes += entry_bytes;
+        self.map.insert(hash, step);
+    }
+}
+
+fn lineage_step_cache() -> &'static parking_lot::Mutex<LineageStepCache> {
+    static CACHE: std::sync::OnceLock<parking_lot::Mutex<LineageStepCache>> =
+        std::sync::OnceLock::new();
+    CACHE.get_or_init(|| parking_lot::Mutex::new(LineageStepCache::default()))
 }
 
 #[cfg(test)]
-fn lineage_step_cache_len() -> usize { lineage_step_cache().lock().len() }
+fn lineage_step_cache_bytes() -> usize { lineage_step_cache().lock().approx_bytes }
 
 /// One walk, every sig: the applied-sig union of `block_hash`'s state
 /// lineage down to `min_height` — the batched form of
@@ -204,10 +229,24 @@ pub(crate) fn settled_sigs_of_lineage(
     let mut walked = 0usize;
     let mut cur = block_hash.clone();
     loop {
-        let cached = lineage_step_cache().lock().get(&cur).cloned();
+        // The cache is process-global and keyed by hash alone, while this
+        // function's answer must be a function of the SUPPLIED store: a hit
+        // is revalidated with a raw key-existence check (no decompression,
+        // no decode) so a block this store does not hold is `BlockNotHeld`
+        // exactly as on the cold path — availability semantics survive the
+        // cache, and one process serving several stores (tests) cannot
+        // cross-answer.
+        let cached = lineage_step_cache().lock().map.get(&cur).cloned();
+        let cached = match cached {
+            Some(step) if block_store.contains_key(&cur)? => Some(step),
+            _ => None,
+        };
         let step = match cached {
             Some(step) => step,
             None => {
+                // Miss path: two short lock acquisitions (lookup above,
+                // insert below) are deliberate — the store read between
+                // them is I/O and must not run under the lock.
                 let Some(block) = block_store.get(&cur)? else {
                     return Err(CasperError::BlockNotHeld(cur));
                 };
@@ -240,11 +279,9 @@ pub(crate) fn settled_sigs_of_lineage(
                     base,
                     sigs: std::sync::Arc::new(sigs),
                 });
-                let mut cache = lineage_step_cache().lock();
-                if cache.len() >= LINEAGE_STEP_CACHE_MAX {
-                    cache.clear();
-                }
-                cache.insert(cur.clone(), step.clone());
+                lineage_step_cache()
+                    .lock()
+                    .insert(cur.clone(), step.clone());
                 step
             }
         };
@@ -257,6 +294,58 @@ pub(crate) fn settled_sigs_of_lineage(
             None => return Ok((settled, walked)),
             Some(base) => cur = base.clone(),
         }
+    }
+}
+
+/// Batched multi-floor settled probe with the reference loop's per-floor
+/// short-circuit. The reference form checks floors IN ORDER and returns
+/// TRUE at the first floor whose lineage holds the sig; floors after the
+/// answering one are never read. This probe builds each floor's applied-sig
+/// set lazily via [`settled_sigs_of_lineage`] the first time the in-order
+/// scan reaches that floor, so an unavailable LATER floor cannot poison a
+/// probe an earlier floor answers — the error surface matches the
+/// reference at floor granularity (within one floor's segment the
+/// fail-closed strengthening of the batched walk applies).
+pub(crate) struct FloorSettledProbe {
+    /// `(floor hash, min_height)` in the reference scan order.
+    floors: Vec<(BlockHash, i64)>,
+    /// Lazily built per-floor applied-sig sets, index-aligned with
+    /// `floors`.
+    sets: Vec<Option<HashSet<Bytes>>>,
+    /// Blocks walked across every set built so far, for the caller's
+    /// metrics.
+    pub(crate) total_walked: usize,
+}
+
+impl FloorSettledProbe {
+    pub(crate) fn new(floors: Vec<(BlockHash, i64)>) -> Self {
+        let sets = floors.iter().map(|_| None).collect();
+        Self {
+            floors,
+            sets,
+            total_walked: 0,
+        }
+    }
+
+    /// TRUE iff some floor's lineage (down to that floor's bound) holds
+    /// the sig, scanning floors in order and stopping at the first hit.
+    pub(crate) fn settled(
+        &mut self,
+        block_store: &KeyValueBlockStore,
+        sig: &Bytes,
+    ) -> Result<bool, CasperError> {
+        for i in 0..self.floors.len() {
+            if self.sets[i].is_none() {
+                let (hash, min_height) = &self.floors[i];
+                let (sigs, walked) = settled_sigs_of_lineage(block_store, hash, *min_height)?;
+                self.total_walked += walked;
+                self.sets[i] = Some(sigs);
+            }
+            if self.sets[i].as_ref().expect("just built").contains(sig) {
+                return Ok(true);
+            }
+        }
+        Ok(false)
     }
 }
 
@@ -1259,8 +1348,85 @@ mod tests {
             }
         }
         assert!(
-            lineage_step_cache_len() <= LINEAGE_STEP_CACHE_MAX,
-            "lineage-step cache must respect its hard cap"
+            lineage_step_cache_bytes() <= LINEAGE_STEP_CACHE_MAX_BYTES,
+            "lineage-step cache must respect its byte budget"
+        );
+    }
+
+    /// The multi-floor probe answers exactly as the reference per-floor
+    /// loop (first floor whose lineage holds the sig wins), across sigs
+    /// held by the first floor, only a later floor, or none.
+    #[tokio::test]
+    async fn floor_probe_matches_the_reference_floor_loop() {
+        let store = store().await;
+        let genesis = block_at(0, vec![], 300);
+        let (sig_a, pd_a) = processed(301, false);
+        let mut floor1 = block_at(1, vec![genesis.block_hash.clone()], 301);
+        floor1.body.deploys = vec![pd_a];
+        let (sig_b, pd_b) = processed(302, false);
+        let mut mid = block_at(1, vec![genesis.block_hash.clone()], 302);
+        mid.body.deploys = vec![pd_b];
+        let floor2 = block_at(2, vec![mid.block_hash.clone()], 303);
+        for b in [&genesis, &floor1, &mid, &floor2] {
+            store.put_block_message(b).expect("store block");
+        }
+        let floors = vec![
+            (floor1.block_hash.clone(), 0),
+            (floor2.block_hash.clone(), 0),
+        ];
+        let sig_absent = Bytes::from_static(b"absent-floor-sig");
+        let mut probe = FloorSettledProbe::new(floors.clone());
+        for sig in [&sig_a, &sig_b, &sig_absent] {
+            let reference = floors
+                .iter()
+                .map(|(hash, bound)| effect_in_state_of(&store, hash, sig, *bound))
+                .try_fold(false, |acc, r| r.map(|hit| acc || hit))
+                .expect("reference loop");
+            assert_eq!(
+                reference,
+                probe.settled(&store, sig).expect("probe"),
+                "floor probe diverged from the reference loop for sig {}",
+                hex::encode(&sig[..8.min(sig.len())]),
+            );
+        }
+    }
+
+    /// The probe keeps the reference loop's short-circuit: a sig the FIRST
+    /// floor holds answers TRUE without reading the second floor at all,
+    /// so a gap in the later floor's lineage cannot poison that probe. A
+    /// sig no floor holds must still reach the gap and refuse, exactly as
+    /// the reference loop does.
+    #[tokio::test]
+    async fn floor_probe_short_circuit_skips_unavailable_later_floors() {
+        let store = store().await;
+        let genesis = block_at(0, vec![], 310);
+        let (sig_a, pd_a) = processed(311, false);
+        let mut floor1 = block_at(1, vec![genesis.block_hash.clone()], 311);
+        floor1.body.deploys = vec![pd_a];
+        let absent = block_at(1, vec![], 312);
+        let floor2 = block_at(2, vec![absent.block_hash.clone()], 313);
+        for b in [&genesis, &floor1, &floor2] {
+            store.put_block_message(b).expect("store block");
+        }
+        let floors = vec![
+            (floor1.block_hash.clone(), 0),
+            (floor2.block_hash.clone(), 0),
+        ];
+
+        let mut probe = FloorSettledProbe::new(floors.clone());
+        assert!(
+            probe.settled(&store, &sig_a).expect("first floor answers"),
+            "a first-floor hit must not read the gapped later floor"
+        );
+
+        let sig_unknown = Bytes::from_static(b"unknown-floor-sig");
+        let err = probe
+            .settled(&store, &sig_unknown)
+            .expect_err("an unanswered probe must reach the gap and refuse");
+        assert!(
+            matches!(err, CasperError::BlockNotHeld(ref h) if *h == absent.block_hash),
+            "the refusal must name the missing block typed; got: {}",
+            err
         );
     }
 
--- a/casper/src/rust/finality/deploy_lifecycle.rs
+++ b/casper/src/rust/finality/deploy_lifecycle.rs
@@ -149,6 +149,117 @@ fn effect_in_state_of_above(
     }
 }
 
+/// One cached lineage step: everything the batched walk needs from a block
+/// without re-reading its body. Content-addressed by block hash, so an
+/// entry can never go stale.
+struct LineageStep {
+    block_number: i64,
+    /// Next block on the state lineage; `None` means genesis (exhausted).
+    base: Option<BlockHash>,
+    /// The block's applied-sig facts: sigs of its non-failed `deploys`
+    /// entries plus its `applied_from_scope` list.
+    sigs: std::sync::Arc<HashSet<Bytes>>,
+}
+
+/// Hard cap for the per-block lineage-step cache. At roughly a kilobyte
+/// per fat block the full cache stays under ~16MB; on overflow the cache
+/// is cleared rather than evicted piecewise — entries are pure functions
+/// of immutable bodies, so losing them costs only a re-read.
+const LINEAGE_STEP_CACHE_MAX: usize = 16_384;
+
+fn lineage_step_cache(
+) -> &'static parking_lot::Mutex<HashMap<BlockHash, std::sync::Arc<LineageStep>>> {
+    static CACHE: std::sync::OnceLock<
+        parking_lot::Mutex<HashMap<BlockHash, std::sync::Arc<LineageStep>>>,
+    > = std::sync::OnceLock::new();
+    CACHE.get_or_init(|| parking_lot::Mutex::new(HashMap::new()))
+}
+
+#[cfg(test)]
+fn lineage_step_cache_len() -> usize { lineage_step_cache().lock().len() }
+
+/// One walk, every sig: the applied-sig union of `block_hash`'s state
+/// lineage down to `min_height` — the batched form of
+/// [`effect_in_state_of`] (CLAIM-FINALITY-001, C2:
+/// `docs/claims/settled-effect-probe-equivalence.md`). A caller holding
+/// the returned set answers any per-sig probe by membership, turning the
+/// merge's ~30 per-sig lineage walks into one walk plus O(1) lookups.
+/// Also returns the number of blocks walked, for the caller's metrics.
+///
+/// Stepping, bounds, and the non-failed/`applied_from_scope` fact kinds
+/// are exactly the reference walk's. One deliberate strengthening: this
+/// walk always covers the FULL segment, so an absent body (or a malformed
+/// multi-parent block without a recorded base) anywhere above the bound
+/// refuses the whole answer — even when a probed sig is applied above the
+/// gap and the per-sig reference walk would have answered TRUE without
+/// reaching it. Availability must not shape verdicts (the claim's
+/// availability-deferral seam premise); a deferral where the reference
+/// sometimes answered is the fail-closed direction.
+pub(crate) fn settled_sigs_of_lineage(
+    block_store: &KeyValueBlockStore,
+    block_hash: &BlockHash,
+    min_height: i64,
+) -> Result<(HashSet<Bytes>, usize), CasperError> {
+    let mut settled: HashSet<Bytes> = HashSet::new();
+    let mut walked = 0usize;
+    let mut cur = block_hash.clone();
+    loop {
+        let cached = lineage_step_cache().lock().get(&cur).cloned();
+        let step = match cached {
+            Some(step) => step,
+            None => {
+                let Some(block) = block_store.get(&cur)? else {
+                    return Err(CasperError::BlockNotHeld(cur));
+                };
+                let base = if !block.body.merge_base.is_empty() {
+                    Some(block.body.merge_base.clone())
+                } else {
+                    match block.header.parents_hash_list.as_slice() {
+                        [] => None,
+                        [parent] => Some(parent.clone()),
+                        _ => {
+                            return Err(CasperError::Other(format!(
+                                "settled_sigs_of_lineage: multi-parent block {} \
+                                 carries no recorded merge_base — refusing to \
+                                 guess its state lineage",
+                                hex::encode(&cur[..8.min(cur.len())]),
+                            )))
+                        }
+                    }
+                };
+                let mut sigs: HashSet<Bytes> = block
+                    .body
+                    .deploys
+                    .iter()
+                    .filter(|pd| !pd.is_failed)
+                    .map(|pd| pd.deploy.sig.clone())
+                    .collect();
+                sigs.extend(block.body.applied_from_scope.iter().cloned());
+                let step = std::sync::Arc::new(LineageStep {
+                    block_number: block.body.state.block_number,
+                    base,
+                    sigs: std::sync::Arc::new(sigs),
+                });
+                let mut cache = lineage_step_cache().lock();
+                if cache.len() >= LINEAGE_STEP_CACHE_MAX {
+                    cache.clear();
+                }
+                cache.insert(cur.clone(), step.clone());
+                step
+            }
+        };
+        if step.block_number < min_height {
+            return Ok((settled, walked));
+        }
+        walked += 1;
+        settled.extend(step.sigs.iter().cloned());
+        match &step.base {
+            None => return Ok((settled, walked)),
+            Some(base) => cur = base.clone(),
+        }
+    }
+}
+
 #[derive(Default)]
 struct Schedule {
     /// Sigs to re-evaluate once the max frozen lm-floor height reaches the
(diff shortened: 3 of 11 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :bulb: Feedback Provided (60.0000% agreement)

**Reviewed by:**

---

### Summary

This is a documentation-and-formalization-only change: a new CBC claim document pinning the reference semantics of the settled-effect probe, a small self-contained Coq theory mechanizing the walk algebra (batching equivalence, segment composition, FALSE-memo and TRUE stability), registration in _CoqProject, and an extension of the axiom-freedom gate from 14 to 17 headline results. No runtime Rust code is touched, so there is no security or performance surface. The Coq proofs are straightforward and correct: `walk_collect_equiv` by induction with `in_dec` case analysis, composition by induction on the first segment, and the two stability results as direct corollaries — with only a section-local decidable-equality hypothesis, so 'Closed under the global context' will hold. The gate arithmetic (14+3=17) is consistent with the three added Print Assumptions lines. Feedback is limited to maintainability concerns: the hardcoded count in the gate script should be derived rather than literal, the pass message has grown unreadable, a line-number citation in the claim doc will rot, and the memoization cache-key premise (bound-sensitivity) deserves explicit coverage in the mandated property test since it sits outside the mechanized model. Safe to merge after addressing the WIP marker.

The Rocq definitions and proofs are internally consistent and correctly establish batching, segment composition, false-result memoization, and true-result extension for the abstract segment model. The documentation also clearly identifies the important availability and walk-bound premises that are outside the model. No security or performance regressions are introduced. The only concern is that one claimed mechanized result is not independently covered by the axiom-free CI gate.

This change documents CLAIM-FINALITY-001 and adds a small, axiom-free Rocq walk algebra (batching ≡ per-sig walk, segment composition, false-memo and true stability) with project registration and an expanded finalized-floor gate. The proofs match the stated list/segment model and correctly leave BlockNotHeld, walk-bound soundness, and store terminal-stability as explicit seam premises. No security or runtime regressions are introduced. Remaining feedback is polish: gate walk_true_stable or document its exclusion, clarify list-vs-set abstraction and tip-first segment orientation for the upcoming batching PR, and tidy the WIP/CI message surface before merge.

This is a well-scoped documentation and formal-methods PR: the claim document clearly pins the reference semantics of the settled-effect probe and the obligations a batched/memoized optimization must preserve, the Rocq theory is small, correct, and genuinely axiom-free (the three proofs are straightforward inductions over list concatenation and are sound), and the gate script is extended consistently from 14 to 17 results. No security or correctness defects were found. The findings are limited to minor documentation concerns: the gate omits the fourth proven result (`walk_true_stable`), a hardcoded Rust line reference will drift, and the model's failed-deploy filtering assumption could be stated inline in the theory.

The code change introduces a well-documented and formally verified optimization for the settled-effect probe, addressing a significant performance bottleneck. The formal proofs and mechanization are thorough, and the documentation is clear and comprehensive. The change appears safe and beneficial, with minor suggestions for additional examples to enhance understanding.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: PR description marked WIP for a mechanized, gated claim</b> (Reported by: anthropic, xai)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 1)

Description is WIP while status in the claim is already `mechanized` and the gate is expanded. Fine for a docs/formal-only change, but reviewers may be unsure whether merge is intended before the remediation branch exists.

---

The claim's yaml frontmatter declares `status: mechanized` and the CI gate is updated to enforce 17/17, yet the PR description says WIP. If this is intentionally landing as final, the WIP marker should be removed to avoid confusion about whether the claim is authoritative.

**Suggestion:**
```
Clear WIP when the 17/17 gate is green on CI, and link the follow-up optimization PR in the discharge plan.
```

</details>

<details>
<summary><b>:orange_circle: Major: Axiom-free gate omits walk_true_stable</b> (Reported by: openai, xai)</summary>

**File:** `scripts/check-finalized-floor-ALL.sh` (line 107)

C3 and the claim doc list walk_true_stable as mechanized alongside the other walk theorems, and SettledEffectProbe.v proves it, but the gate only Print-Assumptions-checks walk_collect_equiv, walk_segment_composition, and walk_memo_false_stable (14→17). The corollary is tiny and follows from composition, so risk is low, but it can drift without CI pressure.

---

The claim and PR description identify walk_true_stable as one of the mechanized equivalence results, but the gate only runs Print Assumptions for the other three new results. Consequently, a future change could introduce an assumption into walk_true_stable without failing this check.

**Suggestion:**
```
Add `Print Assumptions walk_true_stable.` and update the expected count from 17 to 18. Alternatively, explicitly document that this corollary is intentionally excluded from the headline gate because it is derived directly from the checked walk_segment_composition result.
```

</details>

<details>
<summary><b>:orange_circle: Major: Pass message has become a single unreadable mega-line</b> (Reported by: anthropic, xai)</summary>

**File:** `scripts/check-finalized-floor-ALL.sh` (line 112)

The success message is now a ~600-character sentence chaining every headline result with '+' separators. Each new theory extends it further, making CI logs hard to scan and the diff hard to review. The new CLAIM-FINALITY-001 clause is appended to an already overloaded string.

---

The success string is a single growing catalogue of every capstone. It is easy to desync from the actual Print Assumptions list when adding theories, and hard to read in CI logs.

**Suggestion:**
```
Consider a shorter pass line plus a fixed list/array of theorem names driving both Print Assumptions and the expected count.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Doc claims three headline results are gated, but four are registered</b> (Reported by: openrouter)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 96)

The document says the gate 'asserts the three headline results', and the shell script checks `walk_collect_equiv`, `walk_segment_composition`, and `walk_memo_false_stable` — but the theory also proves `walk_true_stable` (relied upon by claim C3's 'TRUE answer survives any extension' and C4). The TRUE-stability corollary is proven but not included in the axiom-free gate, so a future regression introducing an axiom into `walk_true_stable` would not be caught by the gate.

**Suggestion:**
```
Add `Print Assumptions walk_true_stable` to the gate and bump the expected count from 17 to 18, or note explicitly why the corollary is excluded.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded source line reference will rot</b> (Reported by: anthropic)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 116)

The mechanization mapping table cites `deploy_lifecycle.rs:78` for the per-sig walk loop. Line-number references in prose documentation go stale on the first unrelated edit to that file, and there is no tooling here to detect the drift.

**Suggestion:**
```
Reference the function name only (`effect_in_state_of` in `deploy_lifecycle.rs`) or an anchor comment in the Rust source, rather than a raw line number.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded source line reference will go stale</b> (Reported by: openrouter)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 122)

The mechanization mapping cites `deploy_lifecycle.rs:78` for the per-sig walk. Line-number references in long-lived claim documents silently drift as the Rust source is edited; readers will later find the mapping pointing at the wrong code.

**Suggestion:**
```
Reference the function name only (e.g. `effect_in_state_of` in `deploy_lifecycle.rs`) or use a permalocked commit link so the citation remains accurate.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Brittle hardcoded headline count coupled to Print Assumptions lines</b> (Reported by: anthropic)</summary>

**File:** `scripts/check-finalized-floor-ALL.sh` (line 111)

The gate compares grep -c "Closed under the global context" against the literal string "17". Every time a headline result is added or removed, three places must be updated in lockstep (the Print Assumptions heredoc, the pass branch, and the fail branch). A mismatch would silently fail the gate or, worse, a copy-paste error that adds a Print Assumptions line without bumping the count would report failure even when all results are axiom-free. The arithmetic here is correct (14 + 3 = 17), but the pattern accumulates risk.

**Suggestion:**
```
Compute the expected count from the heredoc itself, e.g. expected=$(grep -c '^Print Assumptions' "$chk"), and compare n_closed against $expected so the two can never drift.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Memoization cache-key requirement stated only in prose (C4)</b> (Reported by: anthropic)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 100)

C4 correctly notes that cross-merge reuse is sound only if the cached segment is keyed on the lineage blocks it covered *and the bound it was computed under*. The Coq model deliberately elides the bound (segments are 'already truncated'), so bound-mismatch bugs — e.g. reusing a FALSE answer computed under a higher min_height for a probe with a lower bound — are outside what walk_memo_false_stable proves. This is disclosed as a seam premise, but it is the most likely real-world bug class in the planned remediation.

**Suggestion:**
```
In the discharge plan, explicitly require the property test to include cases where the same lineage is probed under two different min_height bounds, so the cache-key premise is exercised, not just documented.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Model uses lists; remediation text promises sets</b> (Reported by: xai)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 108)

The claim states batched probes answer by set membership, while collect is defined as concat (a list, possibly with duplicates). Membership equivalence is still valid, but implementers may assume the mechanization constrains dedup/hash-set behavior that was never modeled.

**Suggestion:**
```
One sentence clarifying that list In vs set membership is intentional extensional abstraction, and that production batching may use a set for complexity only.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add more detailed examples</b> (Reported by: bedrock)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 131)

Consider adding more detailed examples or use cases to illustrate the practical implications of the settled-effect probe and its optimizations.

**Suggestion:**
```
Include examples that show how the optimizations impact performance in real-world scenarios.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Stdlib namespace pins the theory to Rocq 9.x</b> (Reported by: anthropic)</summary>

**File:** `formal/rocq/finalized_floor/theories/SettledEffectProbe.v` (line 40)

`From Stdlib Require Import ...` is the Rocq 9 namespace and will not compile under Coq 8.x toolchains that use the `Coq` prefix. The PR states verification under Rocq 9.1.1 and the rest of the repo presumably follows suit, but worth confirming the CI runner and any developer setup docs pin the same version so this theory does not break older local installs.

**Suggestion:**
```
Confirm CI and contributor docs pin Rocq >= 9; no code change needed if so.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Model abstraction of failed deploys could be more explicit</b> (Reported by: openrouter)</summary>

**File:** `formal/rocq/finalized_floor/theories/SettledEffectProbe.v` (line 46)

The module header explains that `lineage_block` abstracts a block to 'the sigs whose effects it applies' (non-failed deploys plus `applied_from_scope`). This is a sound abstraction, but a reader validating the refinement against the Rust code must confirm that failed deploys are filtered before the model's `lineage_block` is constructed. The claim document's mapping table covers this, but a one-line note in the theory that the filtering happens at extraction time would strengthen the refinement argument.

**Suggestion:**
```
Add a comment noting that the extraction from `body.deploys` to `lineage_block` must drop entries with `is_failed = true`, so the soundness gap between model and code is visible in the theory itself.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Segment orientation only stated in comments</b> (Reported by: xai)</summary>

**File:** `formal/rocq/finalized_floor/theories/SettledEffectProbe.v` (line 54)

walk_segment_composition and the memo theorems are sound for either convention, but checked_below / cross-merge memoization in Rust depend on tip-first upper++lower layout matching s1 ++ s2. That convention is comment-only; a wrong stitch in the optimization PR would still type-check against these names.

**Suggestion:**
```
Name parameters upper/lower (or add a one-line remark next to the composition theorem) so the remediation maps s1=newer, s2=older without re-reading the claim.
```

</details>
---

<details>

This is a documentation-and-formalization-only change: a new CBC claim document pinning the reference semantics of the settled-effect probe, a small self-contained Coq theory mechanizing the walk algebra (batching equivalence, segment composition, FALSE-memo and TRUE stability), registration in _CoqProject, and an extension of the axiom-freedom gate from 14 to 17 headline results. No runtime Rust code is touched, so there is no security or performance surface. The Coq proofs are straightforward and correct: `walk_collect_equiv` by induction with `in_dec` case analysis, composition by induction on the first segment, and the two stability results as direct corollaries — with only a section-local decidable-equality hypothesis, so 'Closed under the global context' will hold. The gate arithmetic (14+3=17) is consistent with the three added Print Assumptions lines. Feedback is limited to maintainability concerns: the hardcoded count in the gate script should be derived rather than literal, the pass message has grown unreadable, a line-number citation in the claim doc will rot, and the memoization cache-key premise (bound-sensitivity) deserves explicit coverage in the mandated property test since it sits outside the mechanized model. Safe to merge after addressing the WIP marker.

The code change introduces a well-documented and formally verified optimization for the settled-effect probe, addressing a significant performance bottleneck. The formal proofs and mechanization are thorough, and the documentation is clear and comprehensive. The change appears safe and beneficial, with minor suggestions for additional examples to enhance understanding.

#### openai (gpt-5.6-sol)

The Rocq definitions and proofs are internally consistent and correctly establish batching, segment composition, false-result memoization, and true-result extension for the abstract segment model. The documentation also clearly identifies the important availability and walk-bound premises that are outside the model. No security or performance regressions are introduced. The only concern is that one claimed mechanized result is not independently covered by the axiom-free CI gate.

This is a well-scoped documentation and formal-methods PR: the claim document clearly pins the reference semantics of the settled-effect probe and the obligations a batched/memoized optimization must preserve, the Rocq theory is small, correct, and genuinely axiom-free (the three proofs are straightforward inductions over list concatenation and are sound), and the gate script is extended consistently from 14 to 17 results. No security or correctness defects were found. The findings are limited to minor documentation concerns: the gate omits the fourth proven result (`walk_true_stable`), a hardcoded Rust line reference will drift, and the model's failed-deploy filtering assumption could be stated inline in the theory.

#### xai (grok-4.5)

This change documents CLAIM-FINALITY-001 and adds a small, axiom-free Rocq walk algebra (batching ≡ per-sig walk, segment composition, false-memo and true stability) with project registration and an expanded finalized-floor gate. The proofs match the stated list/segment model and correctly leave BlockNotHeld, walk-bound soundness, and store terminal-stability as explicit seam premises. No security or runtime regressions are introduced. Remaining feedback is polish: gate walk_true_stable or document its exclusion, clarify list-vs-set abstraction and tip-first segment orientation for the upcoming batching PR, and tidy the WIP/CI message surface before merge.
</details>

---

jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :mag: Needs Review (40.0000% agreement)

**Reviewed by:**

| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.72 |
| xai | `grok-4.5` | Needs Review | 0.86 |

---

### Summary

A well-specified and well-tested optimization: the batched settled-sig walk with a content-addressed lineage-step cache is a sound remediation for the measured per-sig probe cost, the Coq mechanization is genuinely axiom-free and correctly proves the stated walk algebra, and the generative equivalence test plus the pinned fail-closed divergence give strong confidence in C2. The main concerns are around the process-global cache: it breaks test isolation (risking order-dependent flakiness in the fail-closed test), and it can answer for blocks no longer held in the store, which is a fail-open divergence from the claim's availability-deferral premise that the documentation does not cover. The floor-union probe also broadens the fail-closed scope beyond the single documented divergence, with a liveness cost under partial availability. None of these block merge in a non-pruning deployment, but the cache-vs-store divergence should be added to the claim's seam premises and the test isolation issue fixed before it bites.

The batching approach is well documented and its complete-segment set-membership equivalence is supported by tests and mechanization. However, the process-global cache changes per-store availability behavior and relies on a strong, unenforced content-addressing invariant. Its clear-all eviction policy can also produce a severe performance cliff once walk depth approaches the cap, which is especially concerning for a change intended to remediate lag-amplified lineage walks. These points should be addressed or explicitly validated before merging.

This is a careful, well-documented consensus-path performance fix: batching settled-effect probes into one lineage walk with a content-addressed step cache, keeping `effect_in_state_of` as oracle, and backing the algebra with axiom-free Coq theorems plus solid single-lineage equivalence tests. The main risk is semantic drift under partial availability—especially the floor probe now requiring a successful walk of every settled floor before any membership TRUE, which can turn former TRUE answers into `BlockNotHeld` when a later floor is gapped—plus the still-open `InvalidRejectedDeploy` end-to-end acceptance item. Treat as needs human/consensus review: resolve or explicitly specify the multi-floor short-circuit behavior, tighten evidence/commit pinning and metrics labeling, and avoid merging solely on suite green until rejected-record stability under the new fail-closed rules is demonstrated.

This is a well-scoped performance remediation with unusually strong assurance: the probe semantics are pinned by a Rocq mechanization (axiom-free, gate-asserted), the reference walk is preserved as an oracle, and the generative equivalence test plus the fail-closed gap test directly exercise the equivalence claim. The Rust implementation is straightforward and the semantics changes (full-segment coverage, all-floor union) are deliberately fail-closed and documented. The main points worth a follow-up are operational rather than correctness-critical: the global shared cache with clear-on-overflow can thrash near the cap during sustained lag, the floor-union form widens the availability-dependence of verdicts beyond the documented within-segment case, and the build telemetry only fires on cold paths. None of these block merge; they are candidates for the next soak-observation iteration.

The code change appears to be well-implemented and thoroughly tested. The performance optimization through batching and memoization is well-documented and mechanized. The changes are logically sound, and the tests cover a comprehensive set of scenarios. Minor style and documentation improvements could enhance readability and clarity.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Process-global lineage-step cache breaks test isolation and can make the fail-closed test order-dependent</b> (Reported by: anthropic, openrouter, xai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 170)

The `lineage_step_cache()` is a `OnceLock` static shared across the entire process, including all tests in the test binary (which cargo runs multi-threaded in one process). Tests create independent block stores but share the one cache. If any other test constructs a block whose hash collides with the `absent` block in `batched_walk_is_fail_closed_on_a_gapped_segment` (test blocks are built via `block_at(height, parents, seq)` with small, easily-reused seq values like 1 and 2) and stores/walks it first, the cache will hold a `LineageStep` for that hash and `settled_sigs_of_lineage` will answer instead of returning `BlockNotHeld`, flipping the assertion. Similarly, the `lineage_step_cache_len() <= LINEAGE_STEP_CACHE_MAX` assertion measures global state polluted by whatever tests ran before. Test outcomes now depend on execution order and parallelism.

---

`lineage_step_cache()` is a static `OnceLock<Mutex<HashMap<...>>>` with no namespace/reset outside tests. Parallel `cargo test` threads can interleave inserts/clears; a clear on cap (16_384) in one test can drop another test's entries mid-walk (re-read still correct) but the shared mutable global makes isolation and future “poisoned” entries harder to reason about. In production, the unbounded lifetime is fine for immutable content-addressed bodies, but the hard `clear()` on overflow can cause multi-thread stampeding re-deserialization after a single overflow.

---

lineage_step_cache() is a process-global OnceLock<Mutex<HashMap>>. It is keyed by BlockHash so entries are content-addressed and cannot go stale, which makes cross-store sharing sound in principle. However, in tests multiple store() instances share the same cache, so cache-hit behavior in one test is influenced by blocks inserted by prior tests in the same process (test execution order affects which pass actually exercises the cold path). Also, if the node ever runs multiple shards/stores in one process, the shared hard cap of 16,384 is consumed jointly, which may silently reduce cache effectiveness for each.

**Suggestion:**
```
Consider scoping the cache per KeyValueBlockStore (or making the cap/ownership explicit in a comment), and in the generative test either clear the cache at start or assert the second-pass hit path explicitly rather than relying on ambient cache state.
```

</details>

<details>
<summary><b>:orange_circle: Major: Global cache bypasses the supplied block store's availability semantics</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 176)

The static cache is keyed only by BlockHash, not by the KeyValueBlockStore from which the entry was read. Consequently, a call using store B can consume lineage data previously loaded from store A even when B does not hold that block. This conflicts with the documented rule that an absent body returns BlockNotHeld and can make the gapped-segment test order-dependent or flaky when tests use multiple stores. It is also only safe if every field cached here is cryptographically committed by BlockHash and every insertion path verifies that commitment; the tests mutate block bodies after constructing blocks, illustrating that this invariant is not enforced by the cache itself.

**Suggestion:**
```
Scope the cache to a block-store/database instance, include a stable store identity in the cache key, or make the cache an explicit store-owned layer. Add a test using two stores where a hash is cached from one but absent from the other. If cross-store reuse is intentional, validate that BlockHash commits merge_base, parents, block_number, deploy failure flags/signatures, and applied_from_scope before admitting an entry, and revise the BlockNotHeld specification accordingly.
```

</details>

<details>
<summary><b>:orange_circle: Major: Clear-all eviction can cause persistent cache thrashing for long lineages</b> (Reported by: openai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 195)

When the cache reaches 16,384 entries, the implementation clears every entry. If a settled walk is near or longer than that cap, rebuilding it can clear the hot tip portion and leave only the old tail in the cache. The next merge starts at the tip, misses again, and can repeat the same cycle. This creates a sharp performance cliff precisely when finalization lag makes lineage walks deep, undermining the purpose of this change and causing repeated full block deserialization.

**Suggestion:**
```
Use an eviction/admission policy that preserves the newest or highest hot lineage steps, such as height-aware eviction or a bounded cache designed to resist sequential-scan pollution. At minimum, benchmark and test repeated walks whose depth is just below, equal to, and above LINEAGE_STEP_CACHE_MAX.
```

</details>

<details>
<summary><b>:orange_circle: Major: Cache can answer for blocks no longer held in the store, weakening the availability-deferral premise</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 208)

The claim doc states 'an absent block body is BlockNotHeld (deferral). It is never an answer' and the batched walk is advertised as strictly fail-closed. However, once a block's `LineageStep` is cached, the walk answers from the cache without ever consulting `block_store`. If a block body is later pruned/removed from the store, the cached walk produces verdicts where the reference walk would return `BlockNotHeld`. This is the opposite (fail-open) direction from the documented strengthening: availability now shapes verdicts differently depending on whether a node happened to populate its cache before the body disappeared. Since these verdicts are consensus content checked via `InvalidRejectedDeploy`, two nodes with identical stores but different cache histories could diverge.

**Suggestion:**
```
Document this divergence explicitly in CLAIM-FINALITY-001 (the current seam premises do not cover it), or verify block presence (a cheap existence check) before trusting a cached step, or confirm and state that block bodies are never pruned in this deployment and add that as a seam premise.
```

</details>

<details>
<summary><b>:orange_circle: Major: Floor probe loses short-circuit; gaps on later floors can flip TRUE→error</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1790)

Previously `sig_settled_in_floor` returned Ok(true) on the first floor where `effect_in_state_of` was true and did not visit remaining floors. The batched path always walks every settled floor and unions the sets before membership. If an earlier floor would have answered TRUE but a later floor's lineage is gapped/malformed, the old closure returned TRUE and the new one returns `BlockNotHeld` (or Other). That is an extra fail-closed divergence beyond the documented per-segment gap case, and it can change deferred vs accepted merge outcomes under partial availability. The Coq `walk_segment_composition` justification covers union equivalence only when every floor segment is complete.

**Suggestion:**
```
Either (1) build per-floor sets and preserve existential short-circuit on membership (stop when any floor set contains sig, only failing if all visited floors are false and a later walk errors—or document/order floors deterministically), or (2) explicitly specify and test that requiring the full floor-union segment is intentional consensus behavior, including a generative test where floor[0] applies sig and floor[1] has a missing body.
```

</details>

<details>
<summary><b>:orange_circle: Major: Consensus acceptance criterion still open (InvalidRejectedDeploy)</b> (Reported by: xai)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 140)

Discharge plan item 4—byte-stable rejected-deploy records / `InvalidRejectedDeploy` equality with unfixed peers—is explicitly still open while this change alters probe availability behavior (full-segment refuse + possible full floor-union refuse). Suites being green does not by itself prove cross-version rejected-record identity under lag, gaps, or multi-floor snapshots. Merging a consensus-content probe rewrite without that gate is the highest product risk in the PR.

**Suggestion:**
```
Block merge on a targeted fixture or soak assertion that rejected-deploy sets match pre-batching on complete data, plus explicit cases for gapped base lineage and multi-floor partial availability. Record pass/fail under the claim before calling remediation complete.
```

</details>

<details>
<summary><b>:orange_circle: Major: Entry-count limit does not enforce the documented memory bound</b> (Reported by: openai, xai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 165)

The claimed approximately 16 MB bound assumes roughly one kilobyte per block, but each entry owns a HashSet plus cloned variable-length Bytes values and allocator/hash-table overhead. Fat blocks can therefore make the cache substantially larger than documented, and concurrent per-merge union sets add further memory pressure.

---

`LINEAGE_STEP_CACHE_MAX` (16_384) triggers `cache.clear()` rather than bounded eviction. Under sustained finalization lag many merges share prefixes; one overflow drops the entire working set and the next merges re-load bodies together. Correctness is preserved; latency jitter is not.

**Suggestion:**
```
If a global cache remains, use a simple clock/LRU or shard and evict a fraction; or key cache by tip segment id with a small per-merge map as the hot path.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Minor style issue</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 150)

Consider adding a blank line between the struct definition and the constant declaration for better readability.

**Suggestion:**
```
Add a blank line after the `LineageStep` struct definition.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Two mutex acquisitions per cache miss and clear-all eviction</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 214)

On a cache miss the mutex is taken once for the lookup and again for the insert; between the two, concurrent walkers can duplicate the block-body read and deserialization for the same hash. Additionally, on reaching the 16,384-entry cap the entire cache is cleared, so all concurrent merges simultaneously re-read the whole shared lineage (thundering herd on the block store). Both are correctness-safe (entries are content-addressed) but partially undercut the stated goal of eliminating repeated deserialization under sustained load.

**Suggestion:**
```
Use `entry()` under a single lock scope, or an LRU/random-eviction policy instead of clear-all so hot lineage prefixes survive the cap.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Cache overflow clears the entire map; near-cap thrash possible under sustained lag</b> (Reported by: openrouter)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 228)

On reaching LINEAGE_STEP_CACHE_MAX the whole cache is cleared. If the working set of recent lineage steps hovers just above the cap during sustained finalization lag (the exact condition this PR targets), every merge can clear and rebuild, reintroducing repeated block-body deserialization — the N+1-style cost being removed. A simple insertion-order or random eviction of a batch (e.g., drop oldest 25%) would degrade more gracefully.

**Suggestion:**
```
Either evict a fraction instead of clearing, or emit a metric/log on cache clear so the next soak can detect thrash at the cap.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Base and floor index builds share one histogram pair</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/metrics_constants.rs` (line 137)

Both `sig_settled_in_base` and `sig_settled_in_floor` record into `settled-index.build.time` / `settled-index.blocks`. Soak attribution of “one walk per merge” will mix two builders (and floor total_walked sums), weakening before/after evidence for issue #24.

**Suggestion:**
```
Add a `scope=base|floor` label, or separate metric names, so build time and blocks are distinguishable.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: RefCell-based lazy init is fragile against future async/parallel refactors</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1712)

The lazily-built sig sets use `RefCell` inside an async function, justified by a comment that the merge is synchronous on this thread. `RefCell` is !Sync and `borrow_mut` panics on reentrancy; if the merge is ever parallelized or the closures are held across an await, this becomes a runtime panic or a non-Send future compile break far from this comment. It also holds the `borrow_mut` across block-store I/O during the build.

**Suggestion:**
```
Consider `once_cell::unsync::OnceCell` (or `std::cell::OnceCell` with a fallible-init wrapper) to express 'build once' intent and eliminate the manual is_none/expect dance, or at minimum add a debug assertion documenting the single-threaded invariant.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Floor probe now fails on any floor's gap, even when another floor would have answered TRUE</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1800)

The previous `sig_settled_in_floor` iterated floors and returned TRUE on the first hit; an unavailable block on a *later* floor's lineage was never reached for a sig settled in an earlier floor. The batched version eagerly walks every settled floor to build the union, so a `BlockNotHeld` (or malformed multi-parent) anywhere in *any* floor's lineage now errors every probe. This is broader than the single-lineage gap divergence pinned by `batched_walk_is_fail_closed_on_a_gapped_segment` and has a liveness cost: merges that previously succeeded under partial availability will now defer. The claim doc only documents the single-segment gap case.

**Suggestion:**
```
Extend the claim doc's availability-deferral premise to cover the multi-floor union case, and consider a test pinning this divergence like the single-lineage one.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Floor-union build is all-or-nothing: one missing floor block poisons the probe for the whole merge</b> (Reported by: openrouter)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1807)

The previous per-sig loop short-circuited: effect_in_state_of on floor 1 returning TRUE answered the probe without touching other floors. The batched form walks every settled floor before answering anything, so a BlockNotHeld on any single floor now fails every floor probe (and defers the block) even when an earlier floor would have answered TRUE. This is the documented fail-closed strengthening and is arguably safe, but it widens the divergence beyond the within-segment gap case described in the claim (across-floor availability now also shapes verdicts). Under partial block availability this could increase deferrals.

**Suggestion:**
```
Document the across-floor strengthening explicitly in the claim doc alongside the within-segment one, and consider adding a test pinning behavior when a later floor is missing but an earlier floor settles the sig.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Floor index always walks every floor on first probe</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1810)

First floor probe materializes the union of all settled-floor lineages even when a single floor would decide membership. With many floors or deep `deploy_lifespan` windows this front-loads more block loads than the old short-circuiting OR, though later probes become O(1). Likely still a net win at ~30 probes/merge, but worst-case first-probe latency can regress.

**Suggestion:**
```
Telemetry-split base vs floor build histograms; if floor count is often >1, consider lazy per-floor sets with memoized negatives, or walk floors on demand until hit while caching completed floor sets.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: CbC evidence commit hash may not match landing commit</b> (Reported by: xai)</summary>

**File:** `docs/cbc-evidence/casper-src-rust-finality-deploy-lifecycle-rs.md` (line 6)

Evidence records commit `659966fbf` while the PR text cites implementation `c9aff7732`. If evidence is meant to pin the discharged tree, mismatched hashes weaken the audit trail for `cbc=mandatory` artifacts.

**Suggestion:**
```
Regenerate evidence against the final merge commit hash (or note working-tree OID policy explicitly in both evidence files).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Minor documentation issue</b> (Reported by: bedrock)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 1)

Consider adding a brief summary at the top of the document to provide an overview of the claim and its significance.

**Suggestion:**
```
Add a summary section at the beginning of the document.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded source line reference will rot</b> (Reported by: anthropic)</summary>

**File:** `docs/claims/settled-effect-probe-equivalence.md` (line 120)

The mechanization mapping table cites `deploy_lifecycle.rs:78`. Line numbers in a living file drift immediately; this PR itself inserts ~110 lines into that file, so the reference is likely already stale. (Acknowledged as a known minor in the PR description; recording it here.)

**Suggestion:**
```
Reference the symbol name only (`effect_in_state_of`) or an anchor comment rather than a line number.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Gate's hardcoded pass-count (18) couples the script to every future theorem addition</b> (Reported by: anthropic)</summary>

**File:** `scripts/check-finalized-floor-ALL.sh` (line 108)

The gate string-compares `n_closed == "18"`; every added `Print Assumptions` line requires editing the literal in two places (pass and fail branches), and a mismatch fails the gate even when all listed theorems are closed. (Also acknowledged in the PR description.)

**Suggestion:**
```
Derive the expected count from the number of `Print Assumptions` lines in the generated check file, e.g. `expected=$(grep -c '^Print Assumptions' "$chk")`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Equivalence test does not cover multi-floor union or probe closures</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 1276)

Generative coverage is strong for `settled_sigs_of_lineage` vs `effect_in_state_of` on single lineages (including decoys and cache second pass) and for single-segment fail-closed gaps. It does not exercise `sig_settled_in_floor`’s multi-floor union, lazy build in `interpreter_util`, or the TRUE-on-floor-A + gap-on-floor-B case.

**Suggestion:**
```
Add a focused test that builds two floors and asserts closure-level parity with the old OR-of-walks semantics (or the newly chosen fail-closed multi-floor rule).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: RefCell lazy cells inside async function rely on undocumented sync merge</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1708)

Closures capture `RefCell` and assume merge never re-enters or holds a borrow across `.await`. Current structure looks synchronous on the probe path, but this is brittle if `dag_merger::merge` later becomes async or callbacks nest.

**Suggestion:**
```
Keep a short invariant comment at the merge call site, or use `once_cell`-style outer `Option` filled before entering merge so closures only need shared immutable `&HashSet`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Index-build histogram only records on cold builds, skewing telemetry</b> (Reported by: openrouter)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1713)

PARENTS_POST_STATE_SETTLED_INDEX_BUILD_TIME_METRIC / BLOCKS are recorded only inside the cell.is_none() branch. On warm merges (cache hits or an already-built set) nothing is recorded, so the histograms will under-report frequency and conflate 'walked 0 extra blocks thanks to cache' with 'no build happened'. For before/after evidence on the soak preflight this distinction matters.

**Suggestion:**
```
Record the blocks histogram (possibly 0) on every probe that uses the index, or add a separate counter distinguishing cold builds from cache-served builds.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: First-probe latency spike is folded into the per-probe wrapper counter</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1720)

`settled_probe_time_ns` on the first probe now includes the full index build, so the per-probe wrapper metric mixes one large outlier with many O(1) lookups. The separate `settled-index.build.time` histogram captures the build, so the build time is effectively double-counted in the wrapper counter, slightly muddying the before/after comparison the PR relies on.

**Suggestion:**
```
Exclude the build duration from the wrapper counter (subtract build elapsed, or start the probe timer after the lazy build) so the wrapper cleanly measures membership-lookup cost.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Module doc says it proves three facts but four theorems are exported</b> (Reported by: openrouter)</summary>

**File:** `formal/rocq/finalized_floor/theories/SettledEffectProbe.v` (line 40)

The header comment enumerates walk_collect_equiv, walk_segment_composition, and walk_memo_false_stable (three facts); walk_true_stable is proved as a corollary and asserted in the gate but not mentioned in the header list. Trivially inconsistent documentation.

**Suggestion:**
```
Add walk_true_stable as a fourth item in the header enumeration.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Hardcoded count 18 couples the gate to the exact theorem list</b> (Reported by: openrouter)</summary>

**File:** `scripts/check-finalized-floor-ALL.sh` (line 117)

The n_closed == 18 assertion means any future added Print Assumptions line requires editing the magic number in two places (pass and fail branches). Already noted as a minor in the PR description; a computed expected count from the number of Print Assumptions lines would remove the coupling.

**Suggestion:**
```
Derive expected count from grep -c 'Print Assumptions' of the generated check file instead of a literal.
```

</details>
---

<details>

A well-specified and well-tested optimization: the batched settled-sig walk with a content-addressed lineage-step cache is a sound remediation for the measured per-sig probe cost, the Coq mechanization is genuinely axiom-free and correctly proves the stated walk algebra, and the generative equivalence test plus the pinned fail-closed divergence give strong confidence in C2. The main concerns are around the process-global cache: it breaks test isolation (risking order-dependent flakiness in the fail-closed test), and it can answer for blocks no longer held in the store, which is a fail-open divergence from the claim's availability-deferral premise that the documentation does not cover. The floor-union probe also broadens the fail-closed scope beyond the single documented divergence, with a liveness cost under partial availability. None of these block merge in a non-pruning deployment, but the cache-vs-store divergence should be added to the claim's seam premises and the test isolation issue fixed before it bites.

The code change appears to be well-implemented and thoroughly tested. The performance optimization through batching and memoization is well-documented and mechanized. The changes are logically sound, and the tests cover a comprehensive set of scenarios. Minor style and documentation improvements could enhance readability and clarity.

#### openai (gpt-5.6-sol)

The batching approach is well documented and its complete-segment set-membership equivalence is supported by tests and mechanization. However, the process-global cache changes per-store availability behavior and relies on a strong, unenforced content-addressing invariant. Its clear-all eviction policy can also produce a severe performance cliff once walk depth approaches the cap, which is especially concerning for a change intended to remediate lag-amplified lineage walks. These points should be addressed or explicitly validated before merging.

This is a well-scoped performance remediation with unusually strong assurance: the probe semantics are pinned by a Rocq mechanization (axiom-free, gate-asserted), the reference walk is preserved as an oracle, and the generative equivalence test plus the fail-closed gap test directly exercise the equivalence claim. The Rust implementation is straightforward and the semantics changes (full-segment coverage, all-floor union) are deliberately fail-closed and documented. The main points worth a follow-up are operational rather than correctness-critical: the global shared cache with clear-on-overflow can thrash near the cap during sustained lag, the floor-union form widens the availability-dependence of verdicts beyond the documented within-segment case, and the build telemetry only fires on cold paths. None of these block merge; they are candidates for the next soak-observation iteration.

#### xai (grok-4.5)

This is a careful, well-documented consensus-path performance fix: batching settled-effect probes into one lineage walk with a content-addressed step cache, keeping `effect_in_state_of` as oracle, and backing the algebra with axiom-free Coq theorems plus solid single-lineage equivalence tests. The main risk is semantic drift under partial availability—especially the floor probe now requiring a successful walk of every settled floor before any membership TRUE, which can turn former TRUE answers into `BlockNotHeld` when a later floor is gapped—plus the still-open `InvalidRejectedDeploy` end-to-end acceptance item. Treat as needs human/consensus review: resolve or explicitly specify the multi-floor short-circuit behavior, tighten evidence/commit pinning and metrics labeling, and avoid merging solely on suite green until rejected-record stability under the new fail-closed rules is demonstrated.
</details>

---

PR #364, opened 2026-08-28
fix(soak): keep a host OOM from taking the runner instead of the work…

### LINKED ISSUE #365: soak: host OOM takes the runner instead of the workload — VM self-terminates mid-window and the night loses its artifact
## Incident

Nightly soak run [33136185540](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33136185540) (2026-08-28, `dev` @ `dc625af22` — the first night with a green integration preflight): ~7h into the sustained phase the soak runner vanished. Soak segment 1 never concluded (null step conclusion), GitHub killed the job at lock expiry (12:26Z), the daily series lost the night's artifact, and the in-window restart consumed the remainder re-running the preflight.

## Root cause chain (OCI audit log + cloud-init source)

1. The OCI audit log shows `TerminateInstance` at 12:16:24Z issued by **the VM's own instance principal** (same IP that stamped the reaper-exemption tag at boot). Nothing external: no reaper run in the window (the 05:32Z reaper honored the `soak-deadline-epoch` exemption), no OCI-initiated event.
2. The only path to a mid-job self-terminate is the ephemeral runner's exit path: `run.sh` exits → `post_mortem` → `TerminateInstance`. Mid-soak that means the **wedge watch** (kills the listener when `Runner.Worker` is absent 600s) or outright runner-process death.
3. Worker death ~12:06Z + the 600s wedge window = 12:16Z — to the minute. The established fleet failure mode for worker death under a soak is the **kernel OOM killer choosing the runner process** (precedent: run 30713818751, where OOM shot pytest while node containers ran on — the reason the host guardian exists).
4. The host guardian (`SOAK_HOST_FREE_FLOOR_MB`, 3×5s samples, kills only node processes) never fired: the kernel acted inside or around its 15-second grace window, and its evidence (`post_mortem` dmesg/kern tails) went to the serial console of a VM that then deleted itself.

## Why this matters

- A guardian breach is a **recorded failure with an artifact**; a runner loss is a **vanished night**. The failure mode converts the recoverable case into the unrecoverable one.
- The RSS envelope (16.7–19.3GB) was measured on the merge-bottlenecked build. With #24's fix merged, sustained throughput actually sustains — the memory envelope under the un-throttled build is unmeasured and plausibly larger.

## Fix

PR #364 hardens the guardian: per-sample `oom_score_adj 1000` on all node processes/containers (the kernel now prefers the workload over the runner), a single-sample hard floor at half the configured floor, and durable `soak-health` freeform-tag last words that survive termination (proven readable off this incident's dead VM).

SI-side follow-ups (separate repo): `oom_score_adj -1000` for the runner processes in the ephemeral cloud-init, and a durable `post_mortem` destination instead of the console.

jeffrey-l-turner:
Hardens the soak host guardian so a host OOM can never again take the runner instead of the workload.

Fixes #365 (the incident record with the full audit-log forensics). Follows the #24 remediation: the first green-preflight night (run [33136185540](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33136185540)) ended with the kernel OOM killer choosing `Runner.Worker`, the wedge watch killing the listener, and the VM self-terminating through the ephemeral exit path — converting a recordable breach into a vanished night with no artifact.

## Changes (`scripts/run-merge-recovery-soak.sh`)

1. **The kernel now prefers the workload**: every 3rd guardian sample (15s) re-applies `oom_score_adj 1000` to all rnode processes and node containers. If the OOM killer beats the guardian again, it kills nodes (a recorded breach the segment reports) — never the runner (a vanished VM). A single once-per-guardian log line surfaces the case where the marking cannot be applied.
2. **Graduated floor**: a single sample below `floor/2` (clamped ≥1MB) fires immediately; the 3×5s grace applies only in the soft band.
3. **Durable last words**: in the warning band (`floor + SOAK_HOST_WARN_BAND_MB`, default 4GB) the guardian stamps a `soak-health` freeform tag (state, epoch, available MB) at most once a minute — backgrounded and on the healthy side of the breach checks so a slow tag write can never delay the emergency kill, with every remote call deadline-bounded (`timeout 15`). Tags outlive termination (proven by reading them off this incident's dead VM), and the existing post-mortem `instance.json` capture picks them up with zero workflow changes.

## Review remediations (multi-agent review, 5 providers)

- **Critical — `oci` calls without timeouts in the guardian path**: fixed; both CLI calls are wrapped in `timeout 15`, and the warning stamp additionally runs in the background.
- **Critical/major — silent `sudo -n` / `oom_score_adj` failures**: fixed; one log line per guardian lifetime reports when the workload could not be marked (per-sample logging would flood; silence hid that the runner was unprotected).
- **Major — health stamp could delay the emergency kill**: fixed; the stamp moved after the breach evaluation (healthy branch only) and is backgrounded. The breach-path stamp stays foreground deliberately — it runs after the kills, and durability outranks a bounded ≤30s exit delay there.
- **Critical — freeform-tag read-modify-write race**: accepted by design, now documented in the code: the tag API replaces the whole map, the write happens only while the host is already dying, and last words outrank a checkpoint signal in that state. Healthy runs never write tags.
- **Major — hard-floor edge for tiny configured floors**: fixed with a ≥1MB clamp.
- **Major — "potential command injection"**: declined; the interpolated values are `pgrep` PIDs and `docker ps -q` container IDs, both machine-generated.
- **Minors** — per-sample docker inspect cost: fixed by the 15s marking cadence; warn band made configurable (`SOAK_HOST_WARN_BAND_MB`); container-child PIDs beyond the init PID are noted as accepted scope (single-process node containers).

## Deliberately out of scope (SI repo hand-off)

`oom_score_adj -1000` for the runner processes in the ephemeral cloud-init, and a durable `post_mortem` destination — both belong to `system-integration`.


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :bulb: Feedback Provided (80.0000% agreement)

**Reviewed by:**

| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.82 |

---

### Summary

This is a well-reasoned hardening of the host memory guardian in a soak-test orchestration script, with an unusually good in-code post-mortem comment tying each change to the observed failure (run 33136185540). The three mechanisms — OOM-preferring the workload, a single-sample hard floor at half the configured floor, and durable freeform-tag stamping via instance principal — are logically sound and the failure handling (`|| true`, `return 0` guards, ocid validation, argv-passing into python) avoids the obvious injection and error-propagation pitfalls. Remaining concerns are operational rather than blocking: oom_score_adj is only applied to container init PIDs so pre-existing children keep their old score, the tag read-modify-write race the author acknowledges is still live during long warning periods, the oci CLI calls in the breach exit path are unbounded, and a silent sudo failure would quietly disable the very fix this PR delivers. All are addressable in follow-up; the change is safe to merge as-is for its CI-only scope.

The hard-floor and OOM-priority changes are directionally sound, but the new durable tagging path runs before emergency mitigation and can block the guardian at exactly the point it is intended to react immediately. The tag replacement also risks losing concurrent instance-tag updates. The kill path should be made independent of network operations, and tag writes should use optimistic concurrency control.

This is a focused, well-motivated hardening of the host OOM guardian: prefer killing workload via repeated oom_score_adj 1000, collapse grace on a half-floor hard trip, and leave durable soak-health freeform tags when the runner is about to disappear. The control flow for soft vs hard floor looks correct, warning stamps are rate-limited, and healthy runs avoid the tag RMW path as intended. Before relying on it in another long soak night, I would tighten the stamp path so OCI/python cannot block or allocate on the guardian hot path, and make oom_score_adj success observable rather than fully silent; neither issue is a merge-blocking security hole in this ephemeral runner context, but both undercut the failure mode the PR is fixing.

This is a well-reasoned, incident-driven hardening of the host OOM guardian: per-sample re-application of oom_score_adj to workload processes, a single-sample hard floor, and durable last-words tagging via instance principal are all sound mitigations for the observed runner-loss failure. The code is defensive (every external call is failure-tolerant, kill-before-marker ordering preserved, healthy runs never write tags) and safe to merge. Remaining feedback is minor: the pgrep pattern could over/under-match, a wedged docker daemon could stall the sampling loop, free_mb is not validated as numeric, and the warn-band/throttle constants are hard-coded rather than configurable. The residual tag read-modify-write race is acknowledged in the comments and acceptable.

The code change introduces several improvements to handle OOM situations more gracefully. However, there are potential security and logic issues that need to be addressed, such as command injection risks and inconsistent use of quotes. Additionally, the use of `python3` for JSON manipulation may not be portable across all environments.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: oci CLI calls have no explicit timeout in the breach exit path</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 767)

The IMDS curl uses `--max-time 5`, but the two `oci` CLI invocations (instance get, instance update) have no wrapper timeout. The breach path calls guardian_stamp_health_tag after the kills but before `exit 0`; a hung OCI API call (plausible under host memory pressure, since the python-based CLI needs to fork and allocate) delays guardian exit and thus delays the breach signal being observed. The CLI's internal defaults are long.

---

The comment correctly limits stamping to warning/breach to avoid the healthy-path race, but the RMW remains: get freeform-tags → merge soak-health → update --force. Any concurrent writer (soak-signal or operator tooling) between get and update can be clobbered by --force. On the vanishing-runner path this is still net-positive, but a warning-band stamp during an in-flight soak-signal update can erase signal state used by post-mortem.

---

The function reads the complete freeform-tags map, modifies it locally, and replaces the complete map without an ETag or other concurrency guard. A concurrent soak-signal or automation update between the get and update calls can therefore be silently lost. Limiting writes to warning and breach states reduces frequency but does not remove the race, and warning mode can repeat it every 60 seconds.

**Suggestion:**
```
Fetch and retain the instance ETag, pass it through an if-match condition on update, and refetch/retry on a conflict. If OCI provides an atomic mechanism for updating one tag, prefer that instead.
```

</details>

<details>
<summary><b>:red_circle: Critical: Silent sudo -n failure leaves the runner OOM-preferred with no observability</b> (Reported by: anthropic, bedrock)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 752)

Both write paths for oom_score_adj fall through to `|| true`. If `/proc/$pid/oom_score_adj` is not writable by the invoking user AND `sudo -n` is not passwordless on the host, the hardening this PR exists for silently never takes effect — the guardian believes it marked the workload, but the kernel will still score Runner.Worker normally. Since this is the fix for the vanished-runner failure mode, a silent no-op is worth surfacing.

---

The `guardian_stamp_health_tag` function uses `python3` to manipulate JSON, which may not be available on all systems. This could lead to unexpected behavior.

**Suggestion:**
```
Consider using a more portable solution for JSON manipulation, such as `jq` if available, to ensure consistency across different environments.
```

</details>

<details>
<summary><b>:red_circle: Critical: Freeform-tag read-modify-write race can clobber concurrent tag writers</b> (Reported by: anthropic, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 762)

Warning/breach stamping shells out to python3 and the OCI CLI (commonly a heavy Python entrypoint) while MemAvailable is already in the danger band. That can worsen pressure, fail spuriously, or push the host into the kernel OOM window the guardian is racing. The design goal (durable last words) is sound; the implementation competes with the workload for the resource being defended.

---

guardian_stamp_health_tag performs GET tags → merge → PUT full tag map with no ETag/if-match guard. If another actor (the soak-signal writer mentioned in the comment, or any external automation) updates freeform tags between the GET and the update, that write is silently lost. The comment acknowledges the race is kept 'out of the normal path', but in a degraded-memory night the warning band can persist for hours, stamping every 60s — exactly when the soak-signal tag is most likely to also be written.

---

guardian_stamp_health_tag runs curl + oci get + python3 + oci update synchronously on the guardian loop. Only curl has --max-time 5; the two oci invocations have no timeout. Under the exact failure mode this PR defends against (host memory collapse), a slow or hung OCI control-plane call can stall the loop for far longer than the 5s sample period, delaying or preventing the kill path. The breach path calls this after pkill/docker kill, which is better, but the warning-band call happens before the floor check and can still wedge sampling.

**Suggestion:**
```
Keep the stamp extremely cheap: pre-resolve instance OCID once at guardian start, build the freeform-tags JSON with pure shell/jq if available, or at least avoid cold-starting heavy CLIs every minute. Consider writing a local marker first and best-effort tagging in the background.
```

</details>

<details>
<summary><b>:orange_circle: Major: Potential command injection</b> (Reported by: bedrock)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 731)

The use of `pgrep` and `docker ps -q` followed by `xargs docker kill` could potentially be exploited for command injection if the patterns or names are not properly sanitized.

**Suggestion:**
```
Ensure that the patterns and names used in `pgrep` and `docker ps -q` are properly sanitized to prevent command injection.
```

</details>

<details>
<summary><b>:orange_circle: Major: Health-tag update can delay the emergency workload kill</b> (Reported by: openai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 792)

The warning tag is stamped before checking whether the sample is below the hard floor. On the first severe-memory sample, guardian_stamp_health_tag runs curl and two OCI CLI operations before the workload is killed. The OCI operations have no explicit timeout and may block under host or network pressure, recreating the window in which the kernel can OOM-kill Runner.Worker before the guardian acts. Even successful calls can add material delay to the intended single-sample hard-floor response.

**Suggestion:**
```
Evaluate the floor and perform the workload kill before any network or OCI operation. For a breach, write the local marker and then make the tag update best-effort with strict timeouts or asynchronously. Only perform a warning stamp after determining that the sample is not an immediate breach.
```

</details>

<details>
<summary><b>:orange_circle: Major: Strong incident-driven commentary; consider extracting helpers later</b> (Reported by: bedrock, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 726)

The incident narrative and three-hardening summary are excellent for future operators. The guardian block is now large and nested inside an anonymous subshell, which will be harder to test in isolation.

---

The script uses both single and double quotes inconsistently, which can lead to confusion and potential errors.

**Suggestion:**
```
Optional follow-up: move guardian_mark_workload_oom_preferred / guardian_stamp_health_tag to named functions at script level with a tiny unit-testable pure-shell core for floor decision logic.
```

</details>

<details>
<summary><b>:orange_circle: Major: oom_score_adj updates are best-effort and failures are fully silent</b> (Reported by: openrouter, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 748)

guardian_mark_workload_oom_preferred swallows all failures (no sudo rights, vanished PIDs, LSM denials, Docker inspect misses). If adj never sticks, the primary hardening for 'kernel OOM beats guardian' is ineffective while the rest of the guardian appears healthy. Re-apply-per-sample is the right idea for restarting nodes, but there is no once-per-run or periodic proof that at least one workload PID accepted adj 1000.

---

pgrep -f '/tmp/rnode' matches any process whose command line contains that substring, not just node binaries (e.g. an editor, tail, or log-shipper reading a /tmp/rnode path would also be marked with oom_score_adj 1000, and conversely nodes launched from a different path would be missed). Since the guardian kills with the same pattern on breach this is consistent with existing behavior, but the marking loop now makes over-match more consequential (unrelated processes become preferred OOM victims).

**Suggestion:**
```
Count successful writes per sample; occasionally log when zero successes occur while rnode PIDs/containers are present, so a permissions regression is visible in soak logs before the next vanished VM.
```

</details>

<details>
<summary><b>:orange_circle: Major: Hard floor edge behavior for small HOST_FREE_FLOOR_MB values</b> (Reported by: openrouter, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 780)

guardian_stamp_health_tag reads freeform-tags, merges locally, and writes the whole map back with --force. The comment acknowledges this race with the soak-signal tag and limits stamping to the warning/breach band, but in a genuinely pathological run (warning band for a long stretch, stamps every 60s) a concurrent soak-signal update could still be lost between the get and the update. A lost signal tag could confuse the orchestrator.

---

hard_floor_mb=$((HOST_FREE_FLOOR_MB / 2)) uses integer division. For very small configured floors (e.g. 1–3 MB, or odd values), the hard floor may be 0 or only 1 MB below the soft floor, so the 'single-sample fast plunge' path barely differs from the soft band. Unlikely in production soak configs, but the new semantics are not validated/documented for low floors.

**Suggestion:**
```
Acceptable as-is given the trade-off is documented, but consider re-reading tags immediately before update and retrying once if they changed, or note the residual risk explicitly in the header comment.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: docker ps/inspect on every 5s sample</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 749)

guardian_mark_workload_oom_preferred shells out to docker ps and then docker inspect per container every 5 seconds for the entire soak (potentially days). The cost is small, but each docker CLI call can block if the daemon is under memory pressure — which is exactly the condition under which this loop matters most. A hung docker inspect would stall the guardian's sampling loop and delay breach detection.

**Suggestion:**
```
Consider a short timeout around the docker calls (e.g. timeout 3 docker inspect ...) so a wedged daemon cannot freeze the guardian, or cache container PIDs and verify via /proc/<pid> existence instead.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: oom_score_adj only applied to container init PID, not children</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 757)

guardian_mark_workload_oom_preferred sets oom_score_adj on the PID returned by `docker inspect -f '{{.State.Pid}}'`, which is only the container's init process. Children forked BEFORE the adjustment do not inherit the new value (inheritance happens at fork time). If the node process inside the container forks workers early (e.g., JVM/child helpers), the kernel OOM killer scores those children with their original adj and may still choose a differently-scored victim. The 5s re-apply loop mitigates for restarts but never fixes pre-existing children.

**Suggestion:**
```
Enumerate all PIDs in the container's cgroup (e.g., `cat /sys/fs/cgroup/.../cgroup.procs` or `docker top $cid -eo pid`) and apply oom_score_adj to each, mirroring what the pgrep branch does for host processes.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Per-container docker inspect in a 5s loop</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 755)

guardian_mark_workload_oom_preferred runs `docker ps` plus one `docker inspect` per container every 5 seconds, plus a `sudo` fork per PID. For a multi-node soak this is several process spawns per second, contributing (marginally) load to a host being watched precisely because it is memory-starved.

**Suggestion:**
```
Batch: `docker inspect -f '{{.State.Pid}}' $(docker ps -q --filter 'name=rnode.')` in one call, and write all /proc paths in one `sudo -n tee -a`-style invocation or a single `sudo sh -c` loop.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Breach iteration can trigger two back-to-back OCI tag updates</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 800)

When free_mb drops below warn_floor_mb and the breach condition also fires in the same iteration, the guardian stamps a 'warning' tag (if 60s have elapsed since last_stamp) and then immediately stamps a 'breach' tag. Each stamp is a GET + PUT against the OCI API. Harmless in outcome (breach overwrites warning) but doubles API calls and widens the read-modify-write race window at the worst moment.

**Suggestion:**
```
Skip the warning stamp when the sample also satisfies the breach condition, or restructure so a single stamp per iteration carries the most severe state.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Excellent incident-driven header comment</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 730)

The block comment tying each hardening to the specific failure mode from run 33136185540 (with date, mechanism, and rationale for the tag-only-on-warning design) is exemplary and should be preserved as-is.

</details>

<details>
<summary><b>:white_circle: Suggestion: Broad pgrep -f pattern may match unintended processes</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 749)

pgrep -f '/tmp/rnode' matches any process whose full cmdline contains that string, not only the soak node binaries. That is likely acceptable on these ephemeral runners, but it is the same pattern used later for pkill -9, so over-matching would both prefer-OOM and hard-kill unexpected tasks.

**Suggestion:**
```
Prefer anchored/narrower matching (specific binary paths, pidfiles, or cgroup/docker labels already used via name=rnode.) and keep pgrep/pkill patterns identical by construction.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unescaped regex dot in pgrep/docker filter patterns</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 749)

`pgrep -f '/tmp/rnode'` and `--filter 'name=rnode.'` treat the pattern as a regex; the trailing dot matches any character and '/tmp/rnode' matches any cmdline containing that substring. This mirrors the pre-existing pkill/docker-kill patterns lower in the guardian, so behavior is at least consistent, but an over-broad match now means an unrelated process gets oom_score_adj=1000 rather than merely being killed on breach.

**Suggestion:**
```
Anchor or escape the patterns (e.g., 'name=^rnode\.') for both the new marking loop and the existing kill path in one pass.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Magic numbers for warn band and stamp throttle</b> (Reported by: anthropic)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 771)

The 4096MB warning band and the 60-second tag-stamp throttle are hardcoded inline. The block comment explains the intent, but the values themselves are undiscoverable/untunable next to the already-configurable HOST_FREE_FLOOR_MB.

**Suggestion:**
```
Hoist them to named variables near HOST_FREE_FLOOR_MB (e.g., HOST_GUARDIAN_WARN_BAND_MB=4096, HOST_GUARDIAN_STAMP_INTERVAL_S=60), optionally env-overridable like the floor. Also reuse hard_floor_mb/warn_floor_mb in the final printf instead of recomputing the arithmetic expressions.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Magic constants (4096MB warn band, 60s stamp throttle) not configurable</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 785)

The +4096MB warning offset and the 60-second tag throttle are hard-coded. Since HOST_FREE_FLOOR_MB is already configurable, exposing these (or at least naming them as locals with a comment) would make tuning easier without editing the loop body.

**Suggestion:**
```
Define WARN_BAND_MB="${SOAK_GUARDIAN_WARN_BAND_MB:-4096}" and STAMP_INTERVAL_S="${SOAK_GUARDIAN_STAMP_INTERVAL_S:-60}" near the top with the other config.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: free_mb not validated as numeric before arithmetic comparisons</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 796)

free_mb comes from awk on /proc/meminfo; if the output were ever empty or non-numeric (e.g. meminfo format change), the -lt/-ge tests would fail with a shell error and the sample would be silently misinterpreted (in some shells the test errors cause the branch to be skipped, potentially delaying a breach). The existing [ -n "$free_mb" ] check only guards emptiness.

**Suggestion:**
```
Add a numeric sanity check: case "$free_mb" in ''|*[!0-9]*) continue ;; esac
```

</details>
---

<details>

This is a well-reasoned hardening of the host memory guardian in a soak-test orchestration script, with an unusually good in-code post-mortem comment tying each change to the observed failure (run 33136185540). The three mechanisms — OOM-preferring the workload, a single-sample hard floor at half the configured floor, and durable freeform-tag stamping via instance principal — are logically sound and the failure handling (`|| true`, `return 0` guards, ocid validation, argv-passing into python) avoids the obvious injection and error-propagation pitfalls. Remaining concerns are operational rather than blocking: oom_score_adj is only applied to container init PIDs so pre-existing children keep their old score, the tag read-modify-write race the author acknowledges is still live during long warning periods, the oci CLI calls in the breach exit path are unbounded, and a silent sudo failure would quietly disable the very fix this PR delivers. All are addressable in follow-up; the change is safe to merge as-is for its CI-only scope.

The code change introduces several improvements to handle OOM situations more gracefully. However, there are potential security and logic issues that need to be addressed, such as command injection risks and inconsistent use of quotes. Additionally, the use of `python3` for JSON manipulation may not be portable across all environments.

#### openai (gpt-5.6-sol)

The hard-floor and OOM-priority changes are directionally sound, but the new durable tagging path runs before emergency mitigation and can block the guardian at exactly the point it is intended to react immediately. The tag replacement also risks losing concurrent instance-tag updates. The kill path should be made independent of network operations, and tag writes should use optimistic concurrency control.

This is a well-reasoned, incident-driven hardening of the host OOM guardian: per-sample re-application of oom_score_adj to workload processes, a single-sample hard floor, and durable last-words tagging via instance principal are all sound mitigations for the observed runner-loss failure. The code is defensive (every external call is failure-tolerant, kill-before-marker ordering preserved, healthy runs never write tags) and safe to merge. Remaining feedback is minor: the pgrep pattern could over/under-match, a wedged docker daemon could stall the sampling loop, free_mb is not validated as numeric, and the warn-band/throttle constants are hard-coded rather than configurable. The residual tag read-modify-write race is acknowledged in the comments and acceptable.

#### xai (grok-4.5)

This is a focused, well-motivated hardening of the host OOM guardian: prefer killing workload via repeated oom_score_adj 1000, collapse grace on a half-floor hard trip, and leave durable soak-health freeform tags when the runner is about to disappear. The control flow for soft vs hard floor looks correct, warning stamps are rate-limited, and healthy runs avoid the tag RMW path as intended. Before relying on it in another long soak night, I would tighten the stamp path so OCI/python cannot block or allocate on the guardian hot path, and make oom_score_adj success observable rather than fully silent; neither issue is a merge-blocking security hole in this ephemeral runner context, but both undercut the failure mode the PR is fixing.
</details>

---

spreston8 (approved):

PR #366, opened 2026-08-28
perf(casper): cache rejection-record loads; harden the soak workflow

jeffrey-l-turner:
Two follow-ups from the #24 remediation arc, stacked on the merged PR #362 base.

Addresses #24 (secondary bottleneck) and #365 (soak workflow gaps the first green-preflight night exposed).

## Commit 1 — `perf(casper): load rejection records through the lineage-step cache`

The `prior_rejection_counts` derivation decoded a full block body per visible block per merge — **~47ms/merge in the run 33099406770 attribution, the next cost after the settled probes**. The records now load through the shared `LineageStep` cache introduced by PR #362:

- `rejected_records_of` serves `body.rejected_deploys` verbatim with one decode per block process-wide, revalidated against the caller's store (same `contains_key` availability contract as the settled walk — a block the store does not hold is `BlockNotHeld` even when cached by another store).
- The step's base pointer becomes a `LineageNext` enum, which makes the batched walk slightly *more* faithful to the reference: a malformed multi-parent block below the walk bound no longer refuses, and its own records stay readable — only stepping THROUGH it refuses, exactly like the reference loaders.
- The cache's byte budget now accounts for record bytes.
- New test pins the loader contract: verbatim records, typed absence, malformed-block leniency, per-store cache revalidation. Full casper release suite: 1,262/1,262.

Scope note: the plan originally included cherry-picking `d84f26c07` (repeat_deploy fast path). Verified **unportable**: its `deploy_index` foundation was removed by `1a5174e1f` ("the deploy index folds into the lifecycle rows"), and lifecycle rows skip invalid blocks where the old index covered valid+invalid+approved — a lifecycle-row-based fast path is a consensus-adjacent redesign tracked in #24, not a port.

## Commit 2 — `ci(soak): harden the soak workflow against its first green preflight`

Two gaps exposed on 2026-08-28 (details in #365):

- **`gh: command not found` on the OCI runner** (run 33136185540): the preflight-success status dispatch step had never executed before — it only runs on preflight success, which had never happened on an OCI runner. It now dispatches via `curl` against the workflow-dispatches REST endpoint with `continue-on-error`, so status publication can never fail a soak whose preflight passed; `preflight_finalize` remains the terminal-status backstop.
- **Preflight-only runs shared the nightly's concurrency group** (run 33130708285 lost 1h45m of preflight to the 02:31Z daily launch): both the launch and soak jobs now give `preflight_only` runs a per-run group, so a diagnostic preflight and the scheduled series can never cancel each other.

Both are wanted on `dev` before the next nightly slot: without the first, every green preflight marks its job failed at run end (spurious failure email, skipped dashboard publish, wasted retry dispatch).

## Review remediations (multi-agent review, 5 providers)

- **Major — `curl -sf` swallowed the error body**: fixed. The dispatch now uses `-sS` with a captured HTTP status; a rejected dispatch logs a `::warning::` with the response body and concludes the step as a visible (but job-preserving, `continue-on-error`) failure. Success asserts HTTP 204.
- **Major — unescaped interpolation into the JSON payload**: fixed. The payload is built by `python3 json.dumps` from environment variables — `ref_name` is a branch name and arrives escaped whatever it contains (verified against a quote-bearing name).
- The casper commit (rejection-records cache) drew no findings above minor.

## Commit 3 — `ci(soak): repin system-integration to the runner-hardening SHA and arm the idle-timeout knob`

Folded in from `chore/soak-preflight-20260828` so one dev-bound PR carries everything tonight's 60h soak window needs: the `SYSTEM_INTEGRATION_REF` bump to `b5cbb63a` (SI runner hardening — OOM shield, durable post_mortem, parameterized idle timeout, metrics.py refresh; 265 SI tests green), the Heavy Pipeline's arm64 launches passing `RUNNER_IDLE_TIMEOUT_SECS=7200` (the run 33208755550 hang), and the two-agent cross-repo hand-off log. Pin-history entry notes the follow-up re-pin to the SI main merge SHA. Addresses #365.


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.91 |
| openrouter | `moonshotai/kimi-k3` | Abstain | 0 |

---

### Summary

This is a well-executed, well-documented pair of changes. The Rust refactor cleanly extracts the lineage-step logic into `lineage_step_of`, extends the cache with rejection records, and preserves the per-store availability contract (contains_key revalidation) with a solid pinning test; the one consensus-adjacent behavioral change (malformed multi-parent blocks below the walk bound no longer refusing) is disclosed as intentional but is not itself pinned by a test. The workflow hardening is sound in intent — per-run concurrency groups for preflight-only runs and a curl fallback for the missing gh CLI — but the hand-rolled JSON payload lacks escaping (WORKFLOW_REF can carry JSON-breaking characters) and `-sf` plus `continue-on-error` makes dispatch failures invisible in logs, which cuts against the workflow's stated goal of closing observability gaps. All issues are minor and addressable; the change is safe to merge after considering the escaping and error-visibility feedback.

The Rust cache refactor appears internally consistent: cache hits are revalidated per store, malformed lineage handling is deferred until traversal, rejection records are budgeted, and the caller preserves the expected owned-vector interface. The concurrency changes also correctly isolate preflight-only runs. The only actionable concern is the hand-built JSON request body in the soak workflow, which can become invalid for legal but unusual ref names; using a real JSON encoder would make the dispatch robust.

Solid follow-up: extracting `lineage_step_of`, caching `rejected_deploys` with store revalidation, and extending the byte budget are coherent with the #362 lineage cache and are well pinned by the new per-store/malformed/absence test. The soak concurrency carve-out and curl-based workflow_dispatch fix the two production gaps described in #365 without obvious control-flow bugs. Nothing blocks merge; the main actionable nits are unescaped printf JSON for the dispatch payload and the immediate deep clone of cached rejection records at the merge call site—both worth tightening when convenient, not merge-stoppers.

The code changes look good overall. The performance improvements in the `deploy_lifecycle.rs` file are well-implemented, and the CI workflow changes in `merge-recovery-soak.yml` address the identified gaps effectively. Minor style improvements could enhance readability, but the code is safe to merge as is.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: curl -sf suppresses the API error body, hurting diagnosability</b> (Reported by: anthropic, openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1669)

The payload interpolates WORKFLOW_REF, TARGET_SHA, and RUN_URL directly into JSON. Shell quoting prevents shell injection, but it does not perform JSON escaping. Git ref names can contain characters such as double quotes, so a workflow run from such a ref could produce malformed JSON and silently lose this status dispatch because the step is continue-on-error.

---

`--retry 4` on a POST can re-send after a response is lost on a request the server already processed, dispatching soak-preflight-status.yml more than once. This appears benign here (status publication should be idempotent, and `preflight_finalize` is the backstop), but it is worth confirming the status workflow tolerates duplicate dispatches for the same target_sha.

---

`curl -sf` silences both progress and the HTTP error response body. Combined with `continue-on-error: true`, a failed dispatch (bad token scope, 404 on the workflow file, 422 on inputs) leaves no diagnostic output in the log — the exact class of silent gap that #365 is about. The step will show as failed-but-continued with nothing explaining why.

**Suggestion:**
```
Construct the payload with a JSON encoder, for example `jq -n --arg ref "$WORKFLOW_REF" --arg sha "$TARGET_SHA" --arg url "$RUN_URL" '{ref: $ref, inputs: {target_sha: $sha, source_run_url: $url}}'`, or an equivalent available runtime. Consider also explicitly sending `Content-Type: application/json`.
```

</details>

<details>
<summary><b>:orange_circle: Major: Unescaped string interpolation into JSON payload</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1667)

The dispatch payload is built with printf using raw interpolation of $WORKFLOW_REF, $TARGET_SHA, and $RUN_URL. TARGET_SHA and RUN_URL are structurally safe (hex sha, GitHub-generated URL), but WORKFLOW_REF is `github.ref_name` — git ref names may contain double quotes and other JSON-significant characters, which would break or inject into the JSON body. The gh CLI it replaces did this escaping for you. Exploitability is low (only repo collaborators can create branches that reach this scheduled workflow), but it is a correctness/robustness gap in the new REST path.

---

The workflow-dispatch body is assembled with printf '%s' interpolation into a JSON string. WORKFLOW_REF (github.ref_name), TARGET_SHA, and RUN_URL are normally safe, but any double-quote, backslash, or newline in those values would produce invalid JSON or alter the payload shape. The old `gh` CLI path handled input encoding; the curl replacement does not.

**Suggestion:**
```
Build the body with a real JSON encoder available on the OCI image (e.g. `jq -n --arg ref "$WORKFLOW_REF" --arg target_sha "$TARGET_SHA" --arg source_run_url "$RUN_URL" '{ref:$ref, inputs:{target_sha:$target_sha, source_run_url:$source_run_url}}'`, or the equivalent python3 one-liner) and pass that to curl -d.
```

</details>

<details>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 947)

The comment explaining the concurrency group logic could be reformatted for better readability.

**Suggestion:**
```
Reformat the comment to use consistent indentation and line breaks.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: continue-on-error hides non-CLI failures too</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1650)

continue-on-error correctly prevents a missing `gh` (or a transient API blip) from failing a green preflight. It also swallows auth, 404 workflow, and malformed-payload failures, so a permanently broken dispatch could stay silent until someone notices dashboard/status gaps; preflight_finalize is cited as backstop but is not visible in this hunk.

**Suggestion:**
```
Keep continue-on-error. Optionally echo curl's HTTP status / body on failure (`curl -sS -w '%{http_code}'` or drop -f and branch on status) so logs still show a clear warning when early status publish fails.
```

</details>

<details>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 149)

The comment explaining the `LineageStep` struct could be reformatted for better readability.

**Suggestion:**
```
Reformat the comment to use consistent indentation and line breaks.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Behavioral change: malformed multi-parent block below the walk bound no longer refuses — not pinned by a test</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 316)

Previously, `settled_sigs_of_lineage` errored while constructing the step for a malformed multi-parent block, regardless of its height. With `LineageNext::MalformedMultiParent`, the height check (`step.block_number < min_height`) now runs first, so a malformed block below the bound returns Ok instead of refusing. The PR description declares this intentional ('more faithful to the reference'), and the new test pins that STEPPING through a malformed block still refuses — but the below-the-bound leniency itself is not pinned by any test, and this is consensus-adjacent behavior.

**Suggestion:**
```
Add a test asserting that `settled_sigs_of_lineage` with `min_height` above a malformed block's height succeeds, so the intentional leniency cannot silently regress (or accidentally widen) later.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Full Vec<RejectedDeploy> clone per visible block defeats part of the Arc sharing</b> (Reported by: anthropic, xai)</summary>

**File:** `casper/src/rust/util/rholang/interpreter_util.rs` (line 1786)

`rejected_records_of` returns `Arc<Vec<RejectedDeploy>>` (cheap), then the merge path does `.map(|records| records.as_ref().clone())`, copying every record on every visible block. That reintroduces allocation traffic the cache was meant to avoid for the hot prior-rejection walk (~47ms/merge attribution), though decode/I/O is still saved.

---

`rejected_records_of` returns an `Arc<Vec<RejectedDeploy>>` precisely to avoid copies, but the closure immediately does `records.as_ref().clone()` to satisfy `scope_prior_rejection_counts`'s owned-Vec signature. The expensive body decode is gone (the headline win), but a per-block Vec clone per merge remains.

**Suggestion:**
```
Consider changing `scope_prior_rejection_counts` to accept `&[RejectedDeploy]` (or an `Arc`) from the loader closure so cached records are consumed by reference. Fine as a follow-up if the callback signature is shared with other callers.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Byte budget omits the LineageNext::Base hash and rejected Vec capacity</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 202)

The entry-bytes accounting now includes record sig/carrier bytes plus a 16-byte per-record fudge, but does not account for the 32-byte BlockHash inside `LineageNext::Base` or Vec/Arc allocation overhead. Presumably `LINEAGE_STEP_ENTRY_OVERHEAD_BYTES` absorbs this; if so, the constant's doc should say it covers the `next` pointer and container overheads so future field additions keep the invariant.

**Suggestion:**
```
Either fold the hash into the per-entry sum or document that the overhead constant covers it.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Cache hit on foreign store leaves orphan entry</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 224)

On a process-global hit where `contains_key` is false, the function correctly returns `BlockNotHeld` but leaves the entry in the cache. That is fine for content-addressed immutability and multi-store tests, but repeated lookups from a store that never holds the block still pay a lock + contains_key each time with no eviction of useless hits for that caller.

**Suggestion:**
```
No change required for correctness (matches the stated availability contract). Optional: track nothing here; the existing byte-budget clear-on-overflow already bounds memory. Only reconsider if multi-store-in-one-process becomes a production pattern.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Intentional walk semantics change is easy to miss at call sites</b> (Reported by: xai)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 318)

Moving malformed multi-parent handling into `LineageNext::MalformedMultiParent` means a malformed block *below* `min_height` no longer fails `settled_sigs_of_lineage` (load succeeds, height check returns Ok). The PR description and enum docs state this is deliberate fidelity to reference loaders, and the new test pins record-load vs step-refuse. Call-site readers of the old “any touch of malformed => Other” behavior may not notice.

**Suggestion:**
```
Consider a one-line note on `settled_sigs_of_lineage` itself (not only on the enum) that refusal applies only when the walk must step through a malformed block at or above `min_height`, so future edits do not “restore” the old early-error path.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Awkward helper name `store_fn_second`</b> (Reported by: anthropic)</summary>

**File:** `casper/src/rust/finality/deploy_lifecycle.rs` (line 1524)

The test helper `store_fn_second` reads oddly next to the existing `store()` helper; the `_fn_` infix carries no meaning.

**Suggestion:**
```
Rename to `second_store()` or `fresh_store()`.
```

</details>
---

<details>

This is a well-executed, well-documented pair of changes. The Rust refactor cleanly extracts the lineage-step logic into `lineage_step_of`, extends the cache with rejection records, and preserves the per-store availability contract (contains_key revalidation) with a solid pinning test; the one consensus-adjacent behavioral change (malformed multi-parent blocks below the walk bound no longer refusing) is disclosed as intentional but is not itself pinned by a test. The workflow hardening is sound in intent — per-run concurrency groups for preflight-only runs and a curl fallback for the missing gh CLI — but the hand-rolled JSON payload lacks escaping (WORKFLOW_REF can carry JSON-breaking characters) and `-sf` plus `continue-on-error` makes dispatch failures invisible in logs, which cuts against the workflow's stated goal of closing observability gaps. All issues are minor and addressable; the change is safe to merge after considering the escaping and error-visibility feedback.

The code changes look good overall. The performance improvements in the `deploy_lifecycle.rs` file are well-implemented, and the CI workflow changes in `merge-recovery-soak.yml` address the identified gaps effectively. Minor style improvements could enhance readability, but the code is safe to merge as is.

#### openai (gpt-5.6-sol)

The Rust cache refactor appears internally consistent: cache hits are revalidated per store, malformed lineage handling is deferred until traversal, rejection records are budgeted, and the caller preserves the expected owned-vector interface. The concurrency changes also correctly isolate preflight-only runs. The only actionable concern is the hand-built JSON request body in the soak workflow, which can become invalid for legal but unusual ref names; using a real JSON encoder would make the dispatch robust.

> **ABSTAIN**: token budget exhausted (finish_reason=length): reasoning likely consumed OPENROUTER_MAX_TOKENS - raise it or lower OPENROUTER_REASONING_EFFORT

#### xai (grok-4.5)

Solid follow-up: extracting `lineage_step_of`, caching `rejected_deploys` with store revalidation, and extending the byte budget are coherent with the #362 lineage cache and are well pinned by the new per-store/malformed/absence test. The soak concurrency carve-out and curl-based workflow_dispatch fix the two production gaps described in #365 without obvious control-flow bugs. Nothing blocks merge; the main actionable nits are unescaped printf JSON for the dispatch payload and the immediate deep clone of cached rejection records at the merge call site—both worth tightening when convenient, not merge-stoppers.
</details>

---

jltatbeach (approved):

spreston8 (approved):

PR #368, opened 2026-08-28
fix(casper): bound user-deploy execution per proposed block

spreston8:
Base of a three-PR stack (the transport-liveness PR #369 and the consensus-parameter PR #384 sit on top). Not consensus-affecting: the budget is node-local packaging policy, never a validity rule — per-node divergence degrades throughput to the slowest proposer's setting but cannot fork, and replay determinism is untouched because a block's body is exactly the deploys that executed within the budget.

### Diff anatomy

|---|---:|---|---|
| Production | 7 | +85 −8 | The budget loop in `play_deploys_for_state`, the `Option<Duration>` thread through `compute_state_with_bonds` (build path only — replay stays unbudgeted), the conf key with a derived-default sentinel, launch-time derivation |
| Config validation | 2 | +55 −0 | `validate_config` warns when the budget exceeds a third of the citability window; `defaults.conf` documents the key and the derivation |
| Tests | 4 | +101 −0 | Spec pins for the budget contract and the four warning regimes; two struct-literal touch-ups |
| CI | 2 | +2 −1 | `test_concurrent_bridge_locks` comes off the deselect list; `concurrent-bridge-lock-accounting` and `deploy-play-budget` registered as capabilities |

### What changed

- **`casper.deploy-play-budget`** — a wall-clock ceiling on user-deploy execution per proposed block, checked after each deploy so exactly one always plays (a spent budget defers work to the next proposal, never produces an empty carrier, never wedges). The block carries the executed prefix; deferred deploys stay in deploy storage and are re-offered.
- **Derived default** — `0` (shipped) resolves at launch to `max-parent-depth × heartbeat.check-interval / 5` (15s at shipped defaults), keeping the budget a fixed fraction of the citability window under any operator cadence. A disabled depth check derives an effectively unbounded budget, the matching semantics.
- **Startup validation** — a budget above a third of the citability window warns: a carrier built for that long risks being born below the parent-depth horizon, where its deploys can only expire.

Root cause this closes: the PR #226 arm64 bridge-lock failures. A carrier that packs many expensive deploys is an execution promise every node repays serially at validation; a 98s build froze peer views for ~97s each, collapsed parent selection to self-chains, and buried all 12 locks past the validity window (`Expired`, `rejection_count=0`). Bounding the build spreads the load across carriers and bounds every peer's replay of each one.

Known boundary, deliberately out of scope: a *single* deploy whose replay outruns the citability window on a loaded host still expires (measured live: one bridge lock replaying in 57s against a 45s window under a deliberately tightened test cadence). No packaging policy can reach that; it needs phlo↔wall-clock cost-model calibration (per-deploy cap), tracked separately with the ReplayLock serialization that amplifies it.

- `test_concurrent_bridge_locks` — the test whose CI failure motivated this fix, green under the budget.

An SI-side deterministic repro (`test_deploy_carrier_starvation`: red on the unfixed image with all 12 locks expiring, green with the budget) is held local pending re-margining against the single-deploy boundary above.


spreston8 (commented):
Reviewed at `1c0beb8e3` against `dev` (`432c3c64a`). Suggestions below.

**Low**

1. `0` carries two meanings one layer apart: `Duration::ZERO` on
`CasperConf.deploy_play_budget` means derive (`casper_conf.rs:140`,
`defaults.conf:375`), while `0` on `CasperShardConf.deploy_play_budget_millis`
means unbounded (`casper.rs:391-394`), and `casper_launch.rs:172-178` is the only
translator — so any construction that bypasses it reads the operator's "derive"
sentinel as "no budget". Changing the internal field to `Option<Duration>` makes
zero unrepresentable and leaves the sentinel interpretation at launch, where that
policy belongs; `interpreter_util.rs:839-841` then collapses to a single read and
the `(budget_millis > 0).then(...)` decision leaves the propose path. The eleven
`CasperShardConf::new()` call sites never name the field, so only explicit setters
need touching and the compiler finds them.

2. `defaults.conf:371-376` and the startup warning at
`configuration/mod.rs:199-213` state only the build-side constraint — a carrier
built past a third of the citability window risks being born below the
parent-depth horizon. The constraint that binds also includes replay: the block
has to be re-executed by the slowest validator inside that same window, which is
where this PR's own measurement sits (one bridge lock replaying in 57s against a
45s window). An operator sizing the budget against their own proposer's speed
reads the current text as complete. Stating the replay side in both places closes
it.

PR #369, opened 2026-08-28
fix(casper): block delivery stays live around dead and slow peers

spreston8:
Second in a three-PR stack, on #368; the consensus-parameter PR #384 sits on top. Three fixes to one defect class — block delivery must stay live around dead and slow peers — each demonstrated red on untouched code from a live reproduced freeze before the fix was written. Not consensus-affecting: all three change delivery timing only; validity, fork choice, replay, and the oracle are untouched.

The class, as reproduced (loaded slow-peer shard, one paused low-stake validator): a node's view of a *healthy* peer froze for the full test window while 99.5% of stake was live, holding the LFB flat and expiring deploys with `rejection_count=0`. Three independent mechanisms stacked to produce it.

### Diff anatomy

|---|---:|---|---|
| Production | 4 | +160 −56 | The retriever's dual clock, six best-effort ask sites, the detached announce with its `'static` hoists |
| Config validation | 2 | +67 −10 | Citability-window vs recovery-anchor startup warning; deploy-play-budget comment tightening inherited from the base PR's follow-up |
| Tests | 5 | +406 −10 | Three red-first specs plus the test-helper synchronization the detached announce requires |
| CI | 1 | +0 −3 | `test_slow_peer_notification`, `test_cold_start_readiness`, and `test_convergence::test_network_converges_after_slow_deploy` come off the deselect list |

### What changed

- **Unresolved dependencies re-request on their own clock.** `requested-blocks-timeout` (240s) served as both the received-entry eviction lifetime and the unresolved-entry retry pacing, so a lost block request was first retried after 240s — outliving the citability window, the validity window, and finality itself. Unresolved entries now retry on a 500ms anchor with the existing backoff and retry budget; eviction is unchanged.
- **The block-hash announce is detached from block processing.** The announce after valid/invalid-block effects was awaited inline, and the broadcast completes at the slowest peer's send timeout — one unreachable peer taxed every processed block by the full timeout (measured: a 15s-per-block processing ceiling, terminating to the millisecond with the dead peer's send-timeout WARN). The announce is one-way gossip: it is now spawned, mirroring the proposer's existing detached announce.
- **A failed network ask no longer aborts retriever callers.** Send errors from has-block-request broadcasts and block requests propagated through dependency admission and aborted the whole block validation, which the validation-error quarantine then paced at 120s per attempt — one dead peer froze a node's intake of every healthy sender for the peer's entire outage (measured: a block that failed admission on a 15.004s comm timeout revalidated in 154ms the moment it was re-triggered). A failed ask is gossip lost, not an error: all six ask sites warn and continue, and the maintenance sweep no longer dies at the first dead peer.

- `test_slow_peer_notification` — the class's dedicated test; requires all three fixes to hold under load
- `test_cold_start_readiness`, `test_convergence::test_network_converges_after_slow_deploy` — evidence-only re-enables, green in current soaks

### Suggested reading order

1. `block_retriever.rs::request_all` (the dual clock) + `an_unresolved_dependency_is_rerequested_within_seconds`
2. `block_processor.rs::spawn_block_hash_announce` + `announce_decoupling_spec.rs` (the second assertion proves detaching does not drop delivery)
3. `block_retriever.rs::admit_hash` / `try_rerequest` + `block_retriever_send_failure_spec.rs`
4. `configuration/mod.rs` (the recovery-anchor inequality) and the test-helper synchronization in `test_node.rs`


spreston8 (commented):
Reviewed at `d8cd11a84` against `fix/deploy-play-budget` (`1c0beb8e3`). Suggestions below.

**Medium**

1. Dropping the re-request base from the 240s `requested-blocks-timeout` to the
500ms anchor while `MAX_RETRIES_PER_HASH` stays at 32 collapses the total retry
span from roughly 7.7 hours to roughly 58 seconds (backoff ladder at
`block_retriever.rs:483-496`), so the retry budget rather than the anchor now
bounds recovery — 58s against a 75s citability window at shipped defaults. What
the budget does on exhaustion is a no-op: `request_all:994-995` marks the
quarantine and then calls `cleanup_aux_tracking_for_hash`, which removes the hash
from `retry_budget_quarantine_until` (`:252-259`) and zeroes
`retry_attempts_by_hash` (`:235-242`), and `enforce_requested_blocks_bound:325-331`
erases it again by retaining only quarantine entries still present in
`requested_blocks` — which the eviction at `:991` just cleared. Probe on this head:

```
PROBE budget exceeded before sweep      = Ok(true)
PROBE quarantined after budget eviction = Ok(false)
PROBE retry attempts after eviction     = Ok(0)
```

Both mark sites predate this branch; the reachability does not. An unresolved hash
now reaches that terminal state about once a minute instead of once every 7.7
hours, and with the ladder reset each cycle a permanently-unfetchable block is
re-requested forever without ever backing off — load amplification under exactly
the degraded conditions this PR targets. Preserving the quarantine entry and the
attempt count across eviction, and sparing quarantined-but-evicted hashes in the
`enforce_requested_blocks_bound` retain, brings back `RETRY_BUDGET_QUARANTINE_MS`,
`sweep_expired_retry_budget_quarantine`, the gates at `:700` and `:1053`, and the
`reason = "retry_budget"` metric, all of which are currently inert.
`recover_dependency:1068-1069` has the same mark-then-erase shape.

2. The startup guard at `configuration/mod.rs:218-241` compares the citability
window against `UNRESOLVED_REREQUEST_ANCHOR_MS × 10` = 5s, but the anchor is only
the interval before the *first* re-request; what has to fit inside the window is
the full retry span above. At shipped defaults that is ~58s of recovery inside a
75s window, and the guard stays silent — as written it can only fire on geometries
nobody runs. Comparing against the full span, or a stated fraction of it, makes
the inequality assert what its comment claims.

**Low**

3. `spawn_block_hash_announce:1564-1584` spawns per processed block with no
`JoinSet`, semaphore, or shutdown tracking, so the in-flight task count is
block-processing rate times the slowest peer's send timeout — the same 15s timeout
that motivates the change is what makes that multiplier large during catch-up.
Bounding the spawn, or holding the handles so shutdown can drain them, avoids
trading one peer-induced cost for an unbounded task count.

4. Three deselects come off here and a fourth came off in #368, each on a single
green run. `test_concurrent_bridge_locks` has since gone green, red, then green
again at unchanged SHAs, so one passing sample does not establish stability.
Recording two or three consecutive green runs per re-enabled test before they land
would keep an arm64 flake off `dev`.

spreston8:
> Reviewed at `d8cd11a84` against `fix/deploy-play-budget` (`1c0beb8e3`). Suggestions below.
> 
> **Medium**
> 
> 1. Dropping the re-request base from the 240s `requested-blocks-timeout` to the
>    500ms anchor while `MAX_RETRIES_PER_HASH` stays at 32 collapses the total retry
>    span from roughly 7.7 hours to roughly 58 seconds (backoff ladder at
>    `block_retriever.rs:483-496`), so the retry budget rather than the anchor now
>    bounds recovery — 58s against a 75s citability window at shipped defaults. What
>    the budget does on exhaustion is a no-op: `request_all:994-995` marks the
>    quarantine and then calls `cleanup_aux_tracking_for_hash`, which removes the hash
>    from `retry_budget_quarantine_until` (`:252-259`) and zeroes
>    `retry_attempts_by_hash` (`:235-242`), and `enforce_requested_blocks_bound:325-331`
>    erases it again by retaining only quarantine entries still present in
>    `requested_blocks` — which the eviction at `:991` just cleared. Probe on this head:
> 
> ```
> PROBE budget exceeded before sweep      = Ok(true)
> PROBE quarantined after budget eviction = Ok(false)
> PROBE retry attempts after eviction     = Ok(0)
> ```
> 
> Both mark sites predate this branch; the reachability does not. An unresolved hash now reaches that terminal state about once a minute instead of once every 7.7 hours, and with the ladder reset each cycle a permanently-unfetchable block is re-requested forever without ever backing off — load amplification under exactly the degraded conditions this PR targets. Preserving the quarantine entry and the attempt count across eviction, and sparing quarantined-but-evicted hashes in the `enforce_requested_blocks_bound` retain, brings back `RETRY_BUDGET_QUARANTINE_MS`, `sweep_expired_retry_budget_quarantine`, the gates at `:700` and `:1053`, and the `reason = "retry_budget"` metric, all of which are currently inert. `recover_dependency:1068-1069` has the same mark-then-erase shape.
> 
> 2. The startup guard at `configuration/mod.rs:218-241` compares the citability
>    window against `UNRESOLVED_REREQUEST_ANCHOR_MS × 10` = 5s, but the anchor is only
>    the interval before the _first_ re-request; what has to fit inside the window is
>    the full retry span above. At shipped defaults that is ~58s of recovery inside a
>    75s window, and the guard stays silent — as written it can only fire on geometries
>    nobody runs. Comparing against the full span, or a stated fraction of it, makes
>    the inequality assert what its comment claims.
> 
> **Low**
> 
> 3. `spawn_block_hash_announce:1564-1584` spawns per processed block with no
>    `JoinSet`, semaphore, or shutdown tracking, so the in-flight task count is
>    block-processing rate times the slowest peer's send timeout — the same 15s timeout
>    that motivates the change is what makes that multiplier large during catch-up.
>    Bounding the spawn, or holding the handles so shutdown can drain them, avoids
>    trading one peer-induced cost for an unbounded task count.
> 4. Three deselects come off here and a fourth came off in #368, each on a single
>    green run. `test_concurrent_bridge_locks` has since gone green, red, then green
>    again at unchanged SHAs, so one passing sample does not establish stability.
>    Recording two or three consecutive green runs per re-enabled test before they land
>    would keep an arm64 flake off `dev`.
> 

All four addressed (the branch also carries a dev merge — the review's line numbers moved).

1. Confirmed exactly as probed, and fixed red-first: two specs reproduce the mark-then-erase on both eviction paths (recovery and the maintenance sweep) and now pin persistence — the quarantine and the spent attempt count survive eviction, the orphan retains spare quarantined hashes, and the expiry sweep removes both together, so the cool-off is what earns a re-cited hash a fresh budget. A third spec pins that contract. The previously-inert gates, expiry sweep, and `retry_budget` metric are all live now.

2. The guard now compares the citability window against the full ladder span, computed from the ladder itself: the interval math is hoisted to module functions the retriever and the guard both use, with a unit test pinning the span at 58s so the guard's inequality tracks the constants. At shipped defaults (75s window) it stays just-silent, which is the honest margin statement; the guard test now exercises a geometry that actually fires (15 x 1s).

3. The announce spawn is permit-bounded (128 in flight). Past the ceiling an announce is dropped with a debug line rather than queued: it is best-effort gossip peers also get from proposals, so bounded loss under saturation beats unbounded task growth, and the spawn path stays synchronous. A spec saturates every send and asserts started sends equal the ceiling.

4. On consecutive greens: the three re-enables were each sampled more than once before landing here — two full CI-equivalent hunt iterations on separate arm64 VMs plus the two earlier near-complete runs, all green for slow_peer/cold_start/slow_deploy (records in the PR-body validation section's hunt runs). The bridge re-enable's later instability was root-caused (merge-contention class, budget exonerated) and its gate re-shaped in SI #134 to assert the starvation claim exactly. Agreed on recording 2-3 consecutive greens as standing practice for future re-enables.

PR #376, opened 2026-08-31
fix(rspace): try_lock_owned() fast path in striped_locks::acquire_locks

9Neechan:
## Summary

Closes the remaining part of #50 (intra-deploy par-operator parallelism does not deliver speedup). #72/#346/#355/#356 fixed the `event_log`/`produce_counter`/`get_store()` contention and the testbed's own thread-count default; this PR fixes the last identified bottleneck: `phase_a wait` staying huge (hundreds of thousands of ms at `FORKS=32`) even after those fixes.

## Root cause

`striped_locks::acquire_locks` always called `tokio::sync::Mutex::lock_owned().await` unconditionally, even when the target stripe was free. Direct measurement (diagnostic instrumentation, not included in this PR) showed real per-stripe contention was negligible — only ~130-180 of 4096 stripes were ever touched during a `FORKS=32` propose, with near-zero expected collisions among them. Despite that, the `.await` path itself carried real wall-clock cost on a heavily-loaded runtime, independent of actual contention — most likely tokio scheduling/cooperative-task-budget overhead around the await point.

## Fix

Try `try_lock_owned()` first; only fall back to the async `.await` path when the stripe is genuinely held by another task.

## Measurements

Three consecutive `rholang-par` testbed runs (`mid` profile, 8 CPU), forks=1→32:

| FORKS | PROPOSE ms (pre-fix) | PROPOSE ms (post-fix) | PhaseA wait ms (pre-fix) | PhaseA wait ms (post-fix) |
|---|---|---|---|---|
| 1 | ~4,600 | 3,302 | ~1,300 | 25 |
| 8 | ~9,000 | 5,784 | ~26,000 | 48 |
| 16 | ~17,500 | 10,052 | ~120,000 | 77 |
| 32 | ~33,000–40,700 | 19,161 | ~430,000–595,000 | 118 |

`FORKS=32`: ~2x faster propose wall-clock, with *lower* CPU peak — the CPU usage in prior runs was largely tokio async-machinery churn around an always-uncontended mutex, not real work.

### DIFF
--- a/rspace++/src/rspace/striped_locks.rs
+++ b/rspace++/src/rspace/striped_locks.rs
@@ -48,7 +48,21 @@ pub(crate) async fn acquire_locks(
 
     let mut held: Vec<HeldLock> = Vec::with_capacity(indices.len());
     for idx in indices {
-        let guard = stripes[idx].clone().lock_owned().await;
+        // Fast path: try_lock_owned() first. Measured under `rholang-par`
+        // (many concurrent par-branches, each mostly on its own private
+        // stripe) the stripe is free the overwhelming majority of the time —
+        // going through lock_owned().await unconditionally, even when
+        // uncontended, was adding real wall-clock cost around the .await
+        // point itself (not genuine mutex contention — confirmed by direct
+        // measurement) that dominated `phase_a wait` at high fork counts.
+        // Only the genuinely-contended case falls through to the slow
+        // .await path. See issues/02-intra-deploy-par-mutex-rspace.md
+        // items 16-19 (F1R3FLY-io/f1r3node-rust#50) for the measurements
+        // behind this change.
+        let guard = match stripes[idx].clone().try_lock_owned() {
+            Ok(guard) => guard,
+            Err(_) => stripes[idx].clone().lock_owned().await,
+        };
         held.push(HeldLock { _guard: guard });
     }
 


jeffrey-l-turner (approved):
## Multi-Agent Code Review

**Reviewed by:**

| openai | `gpt-5.6-sol` | Approve | 0.97 |
| openrouter | `moonshotai/kimi-k3` | Approve | 0.95 |

---

### Summary

The change is logically sound and safely adds an uncontended fast path while preserving the existing asynchronous fallback for genuinely contended stripes. Lock ownership and lifetime remain correct, and failure of try_lock_owned() does not alter behavior beyond requiring a second Arc clone before awaiting. No security, correctness, performance regression, or material style issue is evident in the diff.

The change correctly implements a try_lock_owned fast path before lock_owned().await on each stripe, preserving lock acquisition order and OwnedMutexGuard semantics while avoiding unnecessary await/scheduler overhead on the common uncontended case described in the PR. No security or logic defects stand out; the contended path only pays an extra Arc clone/drop, which is acceptable given the stated workload. Safe to merge from a code-review perspective; the suggestions are optional polish only.

The change replaces an unconditional `lock_owned().await` with a `try_lock_owned()` fast path and falls back to the async path on `Err`. This is correct: `tokio::sync::Mutex::try_lock_owned` only fails when the mutex is genuinely held, in which case no guard is leaked and the subsequent `lock_owned().await` simply waits as before. Lock acquisition ordering (which protects against deadlock in this striped-lock scheme) is unaffected by this change. The fallback necessarily re-clones the `Arc` (since `try_lock_owned` consumes the first clone), which is negligible even on the contended path. The behavior matches the PR description (permitting uncontended acquisitions to skip the await point), the rationale comment is thorough, and there are no security, logic, or performance concerns. A minor stylistic note would be the very verbose inline comment referencing an internal issue file, but it is reasonable documentation of an otherwise non-obvious optimization.

The code change looks good and addresses the identified performance bottleneck by implementing a fast path for acquiring locks. The change is well-documented and the performance improvements are significant. A minor suggestion for adding a comment to explain the fast path logic is provided.

---

### Issues Found

<details>
<summary><b>:white_circle: Suggestion: Long inline rationale comment</b> (Reported by: bedrock, xai)</summary>

**File:** `rspace++/src/rspace/striped_locks.rs` (line 51)

Consider adding a comment explaining why the fast path logic was implemented and the measurements that led to this decision. This will help future maintainers understand the reasoning behind this change.

---

The multi-line comment thoroughly explains why the fast path exists and points at issue/measurement context, which is appropriate for a non-obvious optimization. It is lengthy for an inline block and may drift if linked paths or issue numbers change.

**Suggestion:**
```
Add a comment above the fast path logic explaining the rationale and measurements.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Extra Arc clone on the contended path</b> (Reported by: xai)</summary>

**File:** `rspace++/src/rspace/striped_locks.rs` (line 62)

On the slow path, stripes[idx] is cloned once for try_lock_owned() (Arc dropped when TryLockError is returned) and again for lock_owned().await. Semantics are correct because try_lock_owned takes Arc by value and only retains it on success; the stripe array still holds a live Arc. Given measured contention is rare, the cost is negligible, but worth being aware of if this pattern is copied elsewhere under heavier contention.

**Suggestion:**
```
No change required for this PR. If you ever need to tighten the contended path, a small helper that tries without an extra clone is only possible if the API surface allows (tokio's owned try-lock consumes the Arc on both success and the call boundary).
```

</details>
---

<details>

The code change looks good and addresses the identified performance bottleneck by implementing a fast path for acquiring locks. The change is well-documented and the performance improvements are significant. A minor suggestion for adding a comment to explain the fast path logic is provided.

#### openai (gpt-5.6-sol)

The change is logically sound and safely adds an uncontended fast path while preserving the existing asynchronous fallback for genuinely contended stripes. Lock ownership and lifetime remain correct, and failure of try_lock_owned() does not alter behavior beyond requiring a second Arc clone before awaiting. No security, correctness, performance regression, or material style issue is evident in the diff.

The change replaces an unconditional `lock_owned().await` with a `try_lock_owned()` fast path and falls back to the async path on `Err`. This is correct: `tokio::sync::Mutex::try_lock_owned` only fails when the mutex is genuinely held, in which case no guard is leaked and the subsequent `lock_owned().await` simply waits as before. Lock acquisition ordering (which protects against deadlock in this striped-lock scheme) is unaffected by this change. The fallback necessarily re-clones the `Arc` (since `try_lock_owned` consumes the first clone), which is negligible even on the contended path. The behavior matches the PR description (permitting uncontended acquisitions to skip the await point), the rationale comment is thorough, and there are no security, logic, or performance concerns. A minor stylistic note would be the very verbose inline comment referencing an internal issue file, but it is reasonable documentation of an otherwise non-obvious optimization.

#### xai (grok-4.5)

The change correctly implements a try_lock_owned fast path before lock_owned().await on each stripe, preserving lock acquisition order and OwnedMutexGuard semantics while avoiding unnecessary await/scheduler overhead on the common uncontended case described in the PR. No security or logic defects stand out; the contended path only pays an extra Arc clone/drop, which is acceptable given the stated workload. Safe to merge from a code-review perspective; the suggestions are optional polish only.
</details>

---

PR #379, opened 2026-08-31
fix(soak): guard the soak VM against disk exhaustion (#378)

### LINKED ISSUE #378: soak: runner disk exhaustion kills the weekend soak VM ~9-12h in — no disk floor, no hygiene, false-negative finalization verdict
## Incident

The weekend soak (`series: weekend`, targets `master` @ `b0f3dda37`) died twice to the same failure, consuming the whole window:

| Run | Soak job window (UTC) | Soaked | Outcome |
| --- | --- | --- | --- |
| [33254400407](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33254400407) | Fri 2026-08-29 13:14 → 22:20 | ~9h06m | runner dead: ENOSPC |
| [33278321865](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33278321865) (auto-restart) | Fri 22:22 → Sat 10:08 | ~11h46m | runner dead: ENOSPC |

Combined: ~20h52m of soaking, no surviving verdict for the night. The workload itself was healthy the whole way — run 1 passed ~40 consecutive iterations before the end; run 2 passed every iteration and died between iterations.

## Evidence (three independent layers)

1. **The VM's own kernel last words** — the durable `soak-health` freeform tags added by PR #364 (issue #365) worked as designed and captured the console tail off the terminated instance for run 1:
   ```
   systemd-journald[740]: Failed to open system journal: No space left on device
   ```
   at uptime 32,782s — ~9h06m, matching the soak duration to the minute.
2. **Runner worker crash** — both runs' job annotations show the Actions worker dying on
   `System.IO.IOException: No space left on device : '/opt/actions-runner/_diag/Worker_*.log'`.
3. **Workload symptoms in sequence** — as free space shrank:
   - bench side-legs began failing hours earlier (`bench segment 3/4 failed … shard compose up failed`, 17:50Z / 18:27Z in run 1);
   - run 1's final iteration produced a **false-negative consensus verdict**: `test_deploy_throughput_and_finalization` failed with "121 deploy(s) not finalized within 45s" — nodes writing to a full disk, not a finalization defect;
   - run 2's log stops mid perf-report at 09:07:58Z; the runner died an hour later.

## Root cause

The soak VM has **no disk-space management at any layer**:

- `launch-runner.sh` (system-integration) sets no boot-volume size — the 64GB E6.Flex soak VM gets the default OCI boot volume (~47GB), most of it already consumed by the OS image, the runner toolchain, the docker build (full release workspace compile in the image build) and the loaded node image.
- The soak orchestrator (`scripts/run-merge-recovery-soak.sh`) has a memory guardian (RSS ceiling 45056MB + MemAvailable floor 8192MB, per #197/#365 lineage) but **no disk floor**, no inter-iteration docker/tmp hygiene, and no ENOSPC detection.
- Steady per-iteration accumulation (docker bench-shard layers and leftovers from failed compose ups, `/tmp/test-*` compose/genesis files, telemetry/log accretion) fills the remaining headroom in ~9–12h — reproducibly, twice in a row.

Subprocess session data dirs are **not** the leak: 17 of 19 sessions in run 1 logged `SubprocessProvider: cleaned up session …`; the two exceptions are the dying iteration and one mid-failure session.

## Relationship to existing issues

- #365 is the sibling failure class — host **memory** exhaustion taking the runner instead of the workload. This is the **disk** variant, and it defeats #365's mitigation from a new direction: `oom_score_adj` cannot protect the runner from ENOSPC, and the failure again converts a recordable workload failure into a vanished night (run 1's only recorded "failure" is a misleading finalization assertion).
- #337 (shared/stale ephemeral runners) and #283 (shared-runner flake) are unrelated: both runs owned their exclusive `f1r3fly-rust-soak`-labeled VM.

## Proposed fix (node repo, `fix/` off `master` → `dev`)

In `scripts/run-merge-recovery-soak.sh`, mirror the memory-guardian philosophy for disk:

1. **Disk floor guard**: before each iteration (and in the guardian loop), check free space on the runner filesystem. Below a soft floor → run inter-iteration hygiene (prune exited containers/dangling images/unused networks, sweep `/tmp/test-*` older than the current iteration). Still below a hard floor → **conclude the segment cleanly**: write the checkpoint, record a `disk floor breached` verdict with `df` evidence, and exit — a recorded failure with an artifact, never a dead runner.
2. **ENOSPC-aware verdicting**: a final-iteration failure that coincides with a floor breach must be labeled infrastructure, not consensus, so a full disk can never masquerade as a finalization regression.

SI-side follow-up (separate repo, like #365's): size the soak VM's boot volume explicitly in `launch-runner.sh` (the 64GB-RAM shape deserves more than the ~47GB default), and consider `docker builder prune` after the image build.

## Environment

- VM: OCI `VM.Standard.E6.Flex`, 16 OCPU / 64GB, exclusive `f1r3fly-rust-soak` label
- SI ref: `5858f2a26096376f4c2ac6c1636ec45959613782`; node under test: `master` @ `b0f3dda37`
- Post-mortem artifacts: `soak-postmortem-33254400407-1`, `soak-postmortem-33278321865-1` (instance state; console history uncapturable — instance already TERMINATING, the expected branch)

jeffrey-l-turner:
Fixes #378.

## What happened

The weekend soak (master @ `b0f3dda37`) died twice to boot-volume disk exhaustion: runs [33254400407](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33254400407) (~9h06m) and [33278321865](https://github.com/F1R3FLY-io/f1r3node-rust/actions/runs/33278321865) (~11h46m). The VM's durable `soak-health` tags captured `systemd-journald: No space left on device` as its last words; both Actions workers died on ENOSPC writing `_diag`; run 1's final iteration reported "121 deploy(s) not finalized within 45s" — a full disk masquerading as a finalization regression. Full evidence in #378.

## The fix

`scripts/run-merge-recovery-soak.sh` gains a disk twin of the existing memory guardian, in three layers:

1. **Guardian disk floor** — `SOAK_DISK_FREE_FLOOR_MB` (default 4096; 0 disables), sampled in the existing 5s guardian loop: soft floor on 3 consecutive samples, single-sample hard floor at half, kill-writers-then-marker ordering, durable `disk-breach` freeform-tag last words. Killing the nodes on a breach stops the disk from filling further while the runner still has bytes left for its own logs — the difference between a recorded failure and a vanished night.
2. **Iteration-boundary hygiene** — when free space sinks into `floor + SOAK_DISK_HYGIENE_BAND_MB` (default +4096), the orchestrator prunes exited containers, dangling images, orphaned networks, the docker build cache (dead weight once the image under test is loaded) and `/tmp/test-*` leftovers older than an hour. If hygiene cannot lift free space back over the floor, the segment ends cleanly with `df` evidence instead of starting an iteration that is doomed to fail misleadingly.
3. **ENOSPC-aware verdicting** — a fourth breach channel alongside the existing three: an iteration that fails while free disk is under the floor writes `disk-floor-breach.txt` with the `df` snapshot and is labeled infrastructure, failing the soak closed rather than letting a full disk pose as a consensus regression.

## Notes for reviewers

- Branched from `master` (`b0f3dda37`, the exact SHA the failing soaks ran) rather than `dev`, deliberately: this targets the code the weekend soak executes, and merges back into `dev` per the branch strategy's hotfix-like flow.
- The breach reuses the existing `HOST_GUARDIAN_BREACH` marker / `host_protection_breach` early-exit reason, so every downstream consumer (summary, checkpoint, email alert) carries it without changes; the marker text distinguishes disk from memory.
- SI-side follow-up tracked in #378: explicit boot-volume sizing in `launch-runner.sh` (separate repository).
- Verified: `bash -n` clean; `disk_free_mb` probe returns numeric MB; both new env validations reject non-numeric input.

### DIFF
--- a/scripts/run-merge-recovery-soak.sh
+++ b/scripts/run-merge-recovery-soak.sh
@@ -163,8 +163,56 @@ else
 	fi
 fi
 
-printf 'Soak host protection: node RSS ceiling=%sMB; host free floor=%sMB\n' \
-	"$RSS_CEILING_MB" "$HOST_FREE_FLOOR_MB"
+# Disk twin of the memory floor (issue #378). Weekend runs 33254400407 and
+# 33278321865 filled the soak VM's default ~47GB boot volume 9-12h in: the
+# runner worker died on ENOSPC writing its own _diag log, and the last
+# iteration before death failed "121 deploy(s) not finalized within 45s" —
+# a full disk masquerading as a finalization regression. Memory had a
+# guardian, warn band and durable last words; disk had nothing.
+# SOAK_DISK_FREE_FLOOR_MB overrides; 0 disables the floor.
+DISK_FREE_FLOOR_MB="${SOAK_DISK_FREE_FLOOR_MB:-4096}"
+if ! [[ "$DISK_FREE_FLOOR_MB" =~ ^[0-9]+$ ]]; then
+	printf 'SOAK_DISK_FREE_FLOOR_MB must be a non-negative integer\n' >&2
+	exit 2
+fi
+# Hygiene runs when free space sinks into floor+band, BEFORE the floor
+# breaches: pruning between iterations is cheap, while a breach ends the run.
+DISK_HYGIENE_BAND_MB="${SOAK_DISK_HYGIENE_BAND_MB:-4096}"
+if ! [[ "$DISK_HYGIENE_BAND_MB" =~ ^[0-9]+$ ]]; then
+	printf 'SOAK_DISK_HYGIENE_BAND_MB must be a non-negative integer\n' >&2
+	exit 2
+fi
+
+# Free MB on the filesystem the soak actually fills. OUTPUT_DIR, the harness
+# session dirs and the runner's _diag all live on the one boot volume, so one
+# probe stands for them all.
+disk_free_mb() {
+	df -Pm "$OUTPUT_DIR" 2>/dev/null | awk 'NR == 2 { print int($4) }'
+}
+
+# Reclaim space that accumulates across iterations without touching anything
+# an active session owns. Exited rnode containers and their writable layers,
+# dangling images and orphaned networks are per-iteration leftovers (a failed
+# `compose up` leaks all three); the docker build cache is dead weight once
+# the image under test is built and loaded; /tmp compose/genesis files older
+# than an hour belong to no live iteration (an iteration runs ~10 minutes).
+# Never prunes tagged images (the image under test) or running containers.
+reclaim_disk_space() {
+	local before after
+	before="$(disk_free_mb)"
+	if command -v docker >/dev/null 2>&1; then
+		docker container prune -f >/dev/null 2>&1 || true
+		docker network prune -f >/dev/null 2>&1 || true
+		docker image prune -f >/dev/null 2>&1 || true
+		docker builder prune -af >/dev/null 2>&1 || true
+	fi
+	find /tmp -maxdepth 1 -name 'test-*' -mmin +60 -exec rm -rf {} + 2>/dev/null || true
+	after="$(disk_free_mb)"
+	printf 'disk hygiene: %sMB free -> %sMB free\n' "${before:-?}" "${after:-?}"
+}
+
+printf 'Soak host protection: node RSS ceiling=%sMB; host free floor=%sMB; disk free floor=%sMB (hygiene band +%sMB)\n' \
+	"$RSS_CEILING_MB" "$HOST_FREE_FLOOR_MB" "$DISK_FREE_FLOOR_MB" "$DISK_HYGIENE_BAND_MB"
 
 if [ "$RUN_BENCHMARKS" = "true" ] && [ -z "$NODE_REPO_DIR" ]; then
 	printf 'SOAK_NODE_REPO_DIR is required when SOAK_RUN_BENCHMARKS=true\n' >&2
@@ -724,7 +772,7 @@ cleanup_soak_processes() {
 	[ -z "$ITERATION_FIFO" ] || rm -f "$ITERATION_FIFO"
 }
 trap cleanup_soak_processes EXIT
-if [ "$HOST_FREE_FLOOR_MB" -gt 0 ] && [ -r /proc/meminfo ]; then
+if { [ "$HOST_FREE_FLOOR_MB" -gt 0 ] && [ -r /proc/meminfo ]; } || [ "$DISK_FREE_FLOOR_MB" -gt 0 ]; then
 	(
 		# Run 33136185540 (2026-08-28): the kernel OOM killer chose
 		# Runner.Worker while this guardian's 3-sample window was still
@@ -798,10 +846,41 @@ print(json.dumps(tags))
 		hard_floor_mb=$((HOST_FREE_FLOOR_MB / 2))
 		[ "$hard_floor_mb" -ge 1 ] || hard_floor_mb=1
 		warn_floor_mb=$((HOST_FREE_FLOOR_MB + ${SOAK_HOST_WARN_BAND_MB:-4096}))
+		disk_over=0
+		disk_hard_floor_mb=$((DISK_FREE_FLOOR_MB / 2))
+		[ "$disk_hard_floor_mb" -ge 1 ] || disk_hard_floor_mb=1
 		last_stamp=0
 		sample_n=0
 		while :; do
 			sleep 5
+			# Disk twin of the memory floor below (issue #378): same
+			# 3-consecutive-sample soft floor, same single-sample hard floor
+			# at half, same kill-then-marker order, same durable tag last
+			# words. Killing the nodes on a disk breach stops the writers
+			# while the runner still has bytes left for its own _diag log —
+			# on runs 33254400407/33278321865 those bytes ran out and the
+			# breach became a vanished runner instead of a recorded failure.
+			if [ "$DISK_FREE_FLOOR_MB" -gt 0 ]; then
+				disk_mb="$(disk_free_mb)"
+				if [ -n "$disk_mb" ]; then
+					if [ "$disk_mb" -ge "$DISK_FREE_FLOOR_MB" ]; then
+						disk_over=0
+					else
+						disk_over=$((disk_over + 1))
+						if [ "$disk_mb" -lt "$disk_hard_floor_mb" ] || [ "$disk_over" -ge 3 ]; then
+							pkill -9 -f '/tmp/rnode' 2>/dev/null || true
+							docker ps -q --filter 'name=rnode.' 2>/dev/null | xargs -r docker kill 2>/dev/null || true
+							printf 'orchestrator disk guardian: host free disk %sMB < floor %sMB (hard floor %sMB, consecutive %s); killed all node processes and containers to protect the runner\n' \
+								"$disk_mb" "$DISK_FREE_FLOOR_MB" "$disk_hard_floor_mb" "$disk_over" >"$HOST_GUARDIAN_BREACH"
+							guardian_stamp_health_tag disk-breach "$disk_mb"
+							exit 0
+						fi
+					fi
+				fi
+			fi
+			if [ "$HOST_FREE_FLOOR_MB" -le 0 ]; then
+				continue
+			fi
 			free_mb="$(awk '/^MemAvailable:/ {print int($2 / 1024)}' /proc/meminfo 2>/dev/null)"
 			[ -n "$free_mb" ] || continue
 			# Every 3rd sample (15s): fresh nodes appear at iteration
@@ -840,8 +919,9 @@ print(json.dumps(tags))
 		done
 	) &
 	HOST_GUARDIAN_PID=$!
-	printf 'orchestrator host guardian watching MemAvailable floor %sMB (hard floor %sMB, warn %sMB; pid %s)\n' \
-		"$HOST_FREE_FLOOR_MB" "$((HOST_FREE_FLOOR_MB / 2))" "$((HOST_FREE_FLOOR_MB + 4096))" "$HOST_GUARDIAN_PID"
+	printf 'orchestrator host guardian watching MemAvailable floor %sMB (hard floor %sMB, warn %sMB) and disk free floor %sMB (hard floor %sMB); pid %s\n' \
+		"$HOST_FREE_FLOOR_MB" "$((HOST_FREE_FLOOR_MB / 2))" "$((HOST_FREE_FLOOR_MB + 4096))" \
+		"$DISK_FREE_FLOOR_MB" "$((DISK_FREE_FLOOR_MB / 2))" "$HOST_GUARDIAN_PID"
 fi
 
 while [ "$(date +%s)" -lt "$DEADLINE" ]; do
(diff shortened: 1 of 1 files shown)


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.85 |

---

### Summary

The overall layered disk-protection approach is appropriate, but there is a significant race between the asynchronous guardian and the newly added potentially long-running hygiene step: the guardian can terminate and leave the next iteration unprotected while its breach marker goes unchecked. The reclamation commands are also substantially broader than soak-owned artifacts and need either tighter scoping or an explicit, enforced exclusive-runner assumption. These issues should be addressed before relying on the change to prevent another vanished soak run.

Solid, well-motivated hardening that mirrors the existing memory guardian with a disk floor, iteration-boundary hygiene, and ENOSPC-aware fail-closed verdicting; the control flow, marker reuse, and kill-then-record ordering look correct for the reported soak failures. No security issues worth blocking on in this dedicated CI soak context. Main follow-ups are defensive numeric handling of df output, avoiding octal env values, and softening or gating `docker builder prune -af` so hygiene cannot surprise benchmark/rebuild paths—otherwise this is safe to merge with minor polish.

This is a well-scoped, well-commented defensive change that mirrors the existing memory guardian for disk space, with a consistent three-layer design (guardian floor, iteration-boundary hygiene, ENOSPC-aware verdicting) and sensible fail-closed semantics reusing the existing breach marker infrastructure. The implementation is careful about ordering (kill writers before writing the marker), input validation, and never pruning live containers or the image under test. Issues found are minor: the guardian breach only terminates the subshell and relies on the main loop noticing the marker, the docker name filter is an unanchored regex, a misconfigured OUTPUT_DIR silently disables the new protection, and hygiene runs before signal-file consumption. None of these block merge; the change meaningfully reduces the risk of another vanished soak night.

The code change looks solid and addresses the identified issue effectively. The new disk guardian logic is well-implemented and should prevent future disk exhaustion problems. Minor style improvements and additional documentation would enhance the script further.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Disk reclamation is not scoped to soak-owned resources</b> (Reported by: openai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 205)

The cleanup removes every exited Docker container, every unused network, all dangling images, the entire unused builder cache, and every /tmp/test-* entry older than one hour. These operations are broader than the comments' claim that only per-iteration leftovers are removed. On a shared runner, or if an active or stalled session lasts over an hour, this can delete resources belonging to another workload or an active soak-related process.

**Suggestion:**
```
Label soak-created Docker resources and prune using label filters where supported. Track temporary directories created by this soak under a dedicated parent directory and remove only known inactive session directories. If exclusive VM ownership is a required invariant, validate and prominently document that invariant before performing global cleanup.
```

</details>

<details>
<summary><b>:orange_circle: Major: Guardian breach during hygiene can be ignored</b> (Reported by: openai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 950)

The guardian marker is checked before iteration-boundary hygiene, but Docker builder pruning can take longer than the guardian's 15-second soft-floor window. If hygiene begins below the floor, the guardian may create HOST_GUARDIAN_BREACH and exit while hygiene is running. If hygiene then recovers enough space, the orchestrator proceeds into a new iteration without checking the marker again. At an iteration boundary there may be no nodes to kill, and because the guardian has exited, the newly started iteration runs without disk protection and can still exhaust the volume.

**Suggestion:**
```
Recheck HOST_GUARDIAN_BREACH immediately after reclaim_disk_space and before starting the next iteration, and fail closed if it exists. Alternatively, coordinate hygiene with the guardian so that the guardian does not terminate unnoticed while reclamation is in progress.
```

</details>

<details>
<summary><b>:orange_circle: Major: disk_free_mb output not validated as numeric before arithmetic</b> (Reported by: openrouter, xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 190)

`df -Pm "$OUTPUT_DIR" | awk 'NR == 2 { print int($4) }'` assumes the device entry is exactly line 2. With -P this is guaranteed to be one line per filesystem per POSIX, so this is acceptable, but if $OUTPUT_DIR does not exist yet (or the path is on a filesystem df cannot stat), the function silently returns empty and all disk checks are skipped via the `[ -n "$disk_mb" ]` guards — meaning a misconfigured OUTPUT_DIR disables the entire new protection without any warning.

---

disk_free_mb() can return an empty string (df/awk failure is already guarded with -n in most call sites), but a non-empty non-numeric value (unexpected df locale/format, partial line) would make `[ "$disk_mb" -lt ... ]` and $((disk_over + 1)) fail under set -e or print bash errors and skip the breach path inconsistently. The guardian would then silently under-protect.

**Suggestion:**
```
Log a one-time warning when disk_free_mb returns empty while DISK_FREE_FLOOR_MB > 0, so a silently-disabled guardian is visible in the soak logs.
```

</details>

<details>

**File:** `scripts/run-merge-recovery-soak.sh` (line 163)

The comments could be formatted for better readability. Consider breaking long lines and using consistent indentation.

**Suggestion:**
```
Reformat comments for better readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Leading zeros in floor/band env vars become octal in arithmetic</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 175)

Validation allows ^[0-9]+$, so values like SOAK_DISK_FREE_FLOOR_MB=01000 pass. Bash arithmetic treats leading zeros as octal ($((01000/2)) => 256), which silently shrinks the floor/hard-floor/hygiene threshold versus the decimal the operator intended. Same class of footgun as other *_MB knobs if they share this pattern.

**Suggestion:**
```
Reject leading zeros (e.g. ^(0|[1-9][0-9]*)$) or strip them before any $((...)) use.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: docker prune filter uses unanchored regex '.' in container name match</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 200)

`docker ps -q --filter 'name=rnode.'` treats the argument as a partial/regex match, so `rnode.` matches any container whose name contains 'rnode' followed by any character. On a dedicated soak VM this is almost certainly fine, but it could kill an unrelated container (e.g. 'rnode-exporter' or 'notrnodeX') if one ever runs on the host.

**Suggestion:**
```
Use a more precise filter such as `--filter 'name=^/rnode'` or match the exact container naming convention used by the compose setup.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Disk probe failures silently disable protection</b> (Reported by: openai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 868)

Both the guardian and iteration-boundary checks treat an empty disk_free_mb result as a reason to skip disk enforcement. If df fails, OUTPUT_DIR becomes inaccessible, or its output format is unexpected, the soak continues without the protection advertised as fail-closed.

**Suggestion:**
```
Treat repeated probe failures as an infrastructure breach, recording the df error where possible. A small consecutive-failure allowance can avoid terminating on a single transient error.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Disk guardian breach path exits subshell without waking the main loop promptly</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 875)

On a disk breach the guardian writes the HOST_GUARDIAN_BREACH marker and calls `exit 0`, which only terminates the guardian subshell. The main orchestration loop may be mid-iteration (a single iteration runs ~10 minutes) and only checks the marker at the top of the next loop pass or after an iteration failure. Since the nodes were just SIGKILLed, the in-flight iteration will likely fail quickly, but if it does not, the breach could go unacted-on for a full iteration. This mirrors the existing memory-guardian behavior so it is consistent, but worth noting given the stated goal of leaving the runner room to log: a SIGTERM/notify to the main shell (or checking the marker in more places) would shorten the reaction time.

**Suggestion:**
```
Consider having the guardian also touch a file the main loop polls during long waits, or send a signal to the parent PID after writing the marker, so the fail-closed path triggers without waiting for the current iteration to conclude.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Add documentation</b> (Reported by: bedrock)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 163)

Add a brief description of the script at the top for better understanding.

**Suggestion:**
```
Add a header comment describing the purpose of the script.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: New env vars not added to usage/help documentation</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 170)

SOAK_DISK_FREE_FLOOR_MB and SOAK_DISK_HYGIENE_BAND_MB are validated and used, but if the script has a usage block or README section listing SOAK_* tunables, these should be added there so operators know the knobs exist (especially that 0 disables the floor).

**Suggestion:**
```
Add the two new variables to any existing env-var documentation/usage output for the soak script.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: docker builder prune -af on every hygiene pass may be heavier than needed</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 205)

reclaim_disk_space runs builder prune -af whenever free space enters the hygiene band. That is correct for pure leftover cache after the image under test is loaded, but if SOAK_RUN_BENCHMARKS (or any mid-soak rebuild path) needs the build cache, repeated full prunes will re-download/re-build and can stall iteration boundaries on an already disk-stressed VM. container/network/image dangling prunes are appropriately scoped; builder -a is the aggressive step.

**Suggestion:**
```
Consider builder prune without -a by default, or gate -a behind an env flag / only when benchmarks are disabled. Log prune duration if you keep -af so slow hygiene is visible in soak logs.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: df probe every 5s plus on every failure path is fine, but reclaim_disk_space is unthrottled</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 868)

If free space stays inside the hygiene band (e.g. pruning cannot reclaim anything because the growth is from live iteration data), `reclaim_disk_space` runs at every iteration boundary, and each run invokes four docker prune commands plus a find over /tmp. Docker prune operations can take seconds under load. An iteration is ~10 minutes so the cost is negligible, but a simple 'last reclaim attempt' timestamp would make the behavior more predictable and the logs quieter.

**Suggestion:**
```
Optionally track the last reclaim time and skip re-pruning within a short window (e.g. 5 minutes) if free space is unchanged.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Guardian startup log still implies memory watch when only disk is enabled</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 922)

The guardian now starts when disk floor > 0 even if HOST_FREE_FLOOR_MB is 0 (or meminfo is unusable). The printf always prints MemAvailable floor/hard/warn values, including the hardcoded +4096 warn band, which is misleading in disk-only mode and slightly disagrees with SOAK_HOST_WARN_BAND_MB used inside the loop.

**Suggestion:**
```
Branch the log line on which floors are active, and reuse the same warn-band expression as the guardian body.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Hygiene band check runs before the signal-file check, slightly reordering iteration semantics</b> (Reported by: openrouter)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 945)

The new iteration-boundary hygiene block is placed before the SIGNAL_FILE consumption. If a stop/pause signal was written while the previous iteration ran, the soak will now potentially run a full `reclaim_disk_space` (docker prunes + find) before honoring the signal. Harmless in practice, but the signal check arguably should take precedence to keep operator controls responsive.

**Suggestion:**
```
Move the hygiene block after the SIGNAL_FILE handling, or short-circuit hygiene when a signal is pending.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Post-failure disk-floor classification can override a real workload failure near the floor</b> (Reported by: xai)</summary>

**File:** `scripts/run-merge-recovery-soak.sh` (line 1122)

The fourth breach channel intentionally fail-closes any failed iteration while free disk < floor as infrastructure. That matches the #378 incident, but with a 4GiB default floor a genuine consensus/finalization bug that coincides with a full-ish disk will be labeled host_protection_breach and hide the workload signal. Acceptable for soak stability; operators should know the trade-off.

**Suggestion:**
```
Optionally record both the original failure snippet and the disk snapshot when setting BREACH_LINE (e.g. append pytest summary) so triage still sees the assertion text under the infrastructure label.
```

</details>
---

<details>

The code change looks solid and addresses the identified issue effectively. The new disk guardian logic is well-implemented and should prevent future disk exhaustion problems. Minor style improvements and additional documentation would enhance the script further.

#### openai (gpt-5.6-sol)

The overall layered disk-protection approach is appropriate, but there is a significant race between the asynchronous guardian and the newly added potentially long-running hygiene step: the guardian can terminate and leave the next iteration unprotected while its breach marker goes unchecked. The reclamation commands are also substantially broader than soak-owned artifacts and need either tighter scoping or an explicit, enforced exclusive-runner assumption. These issues should be addressed before relying on the change to prevent another vanished soak run.

This is a well-scoped, well-commented defensive change that mirrors the existing memory guardian for disk space, with a consistent three-layer design (guardian floor, iteration-boundary hygiene, ENOSPC-aware verdicting) and sensible fail-closed semantics reusing the existing breach marker infrastructure. The implementation is careful about ordering (kill writers before writing the marker), input validation, and never pruning live containers or the image under test. Issues found are minor: the guardian breach only terminates the subshell and relies on the main loop noticing the marker, the docker name filter is an unanchored regex, a misconfigured OUTPUT_DIR silently disables the new protection, and hygiene runs before signal-file consumption. None of these block merge; the change meaningfully reduces the risk of another vanished soak night.

#### xai (grok-4.5)

Solid, well-motivated hardening that mirrors the existing memory guardian with a disk floor, iteration-boundary hygiene, and ENOSPC-aware fail-closed verdicting; the control flow, marker reuse, and kill-then-record ordering look correct for the reported soak failures. No security issues worth blocking on in this dedicated CI soak context. Main follow-ups are defensive numeric handling of df output, avoiding octal env values, and softening or gating `docker builder prune -af` so hygiene cannot surprise benchmark/rebuild paths—otherwise this is safe to merge with minor polish.
</details>

---

jltatbeach (approved):

PR #381, opened 2026-08-31
fix(casper): read paths abstain history below the LFS restore horizon

spreston8:
Completes #306's residual: after an LFS restore, a validator's latest-message slot can name a block below the restore horizon — held as a hash only, never indexed. Every read path that materialized latest messages `lookup_unsafe`'d that slot, so ONE stale entry from one inactive validator failed every fault-tolerance read (the one-hash-repeated-300x storm in #306's logs), every fork-choice run, and every propose attempt on the node. #318's typed `MissingBlock` only helps paths with a deferral consumer — block processing has one; the API, the estimator, and propose have none. Not consensus-affecting in live operation: on a fully-held DAG every filter is a no-op, and abstention can only understate the clique (the safe direction, same philosophy as AbsenceHold).

Independent of the #368/#369 stack — based on `dev`, disjoint files.

### Diff anatomy

|---|---:|---|---|
| Production | 5 | +143 −26 | Held-LM filters at the three materialization boundaries, unheld-as-absent in storage, lookup-and-skip across the API's windowed walks |
| Tests | 2 | +152 −0 | One red-first spec (both reds died in the oracle/estimator weight walks on untouched code) |

### What changed

- **The three latest-message materialization boundaries abstain unheld validators.** The oracle's `ft_witnessed`, the estimator's `tips_with_latest_messages`, and the propose snapshot each filter latest messages to held blocks before any walk; every parent/LCA read below inherits held provenance from that filter. An abstained validator simply contributes no agreement weight — exactly how an equivocator or invalid-block sender is already treated.
- **Storage treats an unheld latest message as absent.** `latest_message`/`latest_messages` return `None`/skip instead of erroring, so no future caller can reintroduce the storm through the accessor.
- **The API's windowed walks lookup-and-skip.** `get_blocks`, `get_blocks_full`, by-height ranges, and machine-verifiable-DAG parent fetches skip unheld blocks; find-deploy falls through to its scan; a fault-tolerance read that still reaches unheld history degrades to `ft = -1` instead of failing the call.
- **41-site `lookup_unsafe` provenance audit closed.** Beyond the fix surface: `validate.rs`'s parent reads were already horizon-tolerant, the equivocation detector's totality claim holds over safe provenance, and finalizer/lifecycle sites read only the held LFB. No other site can see an unheld hash.

### Suggested reading order

1. `casper/tests/finalized_floor/horizon_read_abstention_spec.rs` — the restored node's shape and both reds
2. `clique_oracle.rs::ft_witnessed` + `estimator.rs::tips_with_latest_messages` + `snapshot.rs` (the three boundaries)
3. `block_dag_key_value_storage.rs` (unheld-as-absent at the accessor)
4. `block_api.rs` (the windowed walks and `fault_tolerance_or_unknown`)

### DIFF
--- a/.github/oci-validation.env
+++ b/.github/oci-validation.env
@@ -16,4 +16,4 @@
 # Bump both in the same commit: `just repin-system-integration <sha>`
 # (scripts/repin-system-integration.sh) rewrites every pin site together and
 # runs the invariants checker.
-SYSTEM_INTEGRATION_REF=8b4da0f9395fc951b68af1be8ae023fb1aeb9d09
+SYSTEM_INTEGRATION_REF=ce5a825b2c090c8680ced81327e03f83c9bfa38f
--- a/.github/oci-validation.env
+++ b/.github/oci-validation.env
@@ -16,4 +16,4 @@
 # Bump both in the same commit: `just repin-system-integration <sha>`
 # (scripts/repin-system-integration.sh) rewrites every pin site together and
 # runs the invariants checker.
-SYSTEM_INTEGRATION_REF=ce5a825b2c090c8680ced81327e03f83c9bfa38f
+SYSTEM_INTEGRATION_REF=e4a26eb69fcfc611d56d8473f4de2b09960bc173
--- a/.github/workflows/_integration-pipeline.yml
+++ b/.github/workflows/_integration-pipeline.yml
@@ -45,7 +45,7 @@ env:
   # MUST be kept identical to .github/oci-validation.env (the Full OCI
   # Validation workflow sources that file) — bump both in the same commit,
   # via `just repin-system-integration <sha>`.
-  SYSTEM_INTEGRATION_REF: 8b4da0f9395fc951b68af1be8ae023fb1aeb9d09
+  SYSTEM_INTEGRATION_REF: ce5a825b2c090c8680ced81327e03f83c9bfa38f
   # Pinned OCI CLI installer (release tag + sha256) and CLI version. The
   # install.sh checksum fails closed if the tag is ever moved; OCI_CLI_VERSION
   # pins what the installer pulls from PyPI (without it the installer floats to
--- a/.github/workflows/_integration-pipeline.yml
+++ b/.github/workflows/_integration-pipeline.yml
@@ -45,7 +45,7 @@ env:
   # MUST be kept identical to .github/oci-validation.env (the Full OCI
   # Validation workflow sources that file) — bump both in the same commit,
   # via `just repin-system-integration <sha>`.
-  SYSTEM_INTEGRATION_REF: ce5a825b2c090c8680ced81327e03f83c9bfa38f
+  SYSTEM_INTEGRATION_REF: e4a26eb69fcfc611d56d8473f4de2b09960bc173
   # Pinned OCI CLI installer (release tag + sha256) and CLI version. The
   # install.sh checksum fails closed if the tag is ever moved; OCI_CLI_VERSION
   # pins what the installer pulls from PyPI (without it the installer floats to
--- a/.github/workflows/merge-recovery-soak.yml
+++ b/.github/workflows/merge-recovery-soak.yml
@@ -179,7 +179,14 @@ env:
   # full green x3 on the d94fffd content (run 32691266446 attempts 1+2, run
   # 32759382579), and 8b4da0f is d94fffd merged to main with no further
   # changes.
-  SYSTEM_INTEGRATION_REF: 8b4da0f9395fc951b68af1be8ae023fb1aeb9d09
+  # 2026-08-31: 8b4da0f -> ce5a825 (SI PR #132 branch): retires ftt=-1 from
+  # the joiner-boundary/trim_state/synchrony shards (a negative FTT legally
+  # permits divergent floors that the suite's forbidden-log check treats as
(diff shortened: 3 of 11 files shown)


spreston8 (commented):
Reviewed at `ea68bee9f` against `dev` (`f00700611`). Suggestions below, with the
reproduction inline since the follow-up is likely to be picked up by someone without
this context.

**Medium — `ft_witnessed_exact` is unfiltered, and it is the path that decides finality**

`floor.rs` calls it at `:903`, `:924`, `:975`. It passes `latest_messages` straight
through, so an unheld hash still fails the whole call:

```
ft_witnessed_exact            clique_oracle.rs:598-600   (no filter)
  → agreeing_weight_map       clique_oracle.rs:529
    → agree                   clique_oracle.rs:554-556
      → dag.is_in_main_chain  block_dag_key_value_storage.rs:530
        → block_number_unsafe block_dag_key_value_storage.rs:268  → Err(missing_block)
```

`block_number_unsafe` errors on absence (`:268-271`); it does not return false.
Production-reachable: `floor.rs:451` documents `latest_messages` as "the block's own
justifications", and on a restored node those routinely name blocks below the horizon.

Reproduces on this PR's own fixture — add to `horizon_read_abstention_spec.rs`:

```rust
#[tokio::test]
async fn probe_exact_path_on_the_same_shape() {
    let (dag, _genesis, target) = restored_dag_with_stale_lm();
    let latest_messages: std::collections::BTreeMap<Validator, BlockHash> =
        dag.latest_messages_map.clone().into_iter().collect();
    let ftt = casper::rust::safety::clique_oracle::FtThreshold::from_ppm(0);
    let result = CliqueOracle::ft_witnessed_exact(
        &target.block_hash, &dag, &latest_messages, ftt, false,
    ).await;
    eprintln!("PROBE ft_witnessed_exact: {:?}", result);
}
```

`cargo test --release -p casper --test mod probe_exact_path_on_the_same_shape -- --nocapture`

```
Err(MissingBlock { hash: b"below-restore-horizon-lm", context: " [block_number_unsafe]
   2: KeyValueDagRepresentation::is_in_main_chain
   3: CliqueOracle::agreeing_weight_map
   4: CliqueOracle::ft_witnessed_exact
```

**Worth NOT copying the filter across.** `ft_witnessed`'s doc comment
(`clique_oracle.rs:654-660`) states the result is "bit-identical across honest nodes" and
names `dag.latest_message_hash` as the sole node-divergent input. `dag.lookup` is
node-local, so filtering inside the exact path would make the finalization DECISION
depend on what each node holds — two honest nodes could disagree on whether a block is
finalized, which is worse than the current failure. Abstention suits the read paths
because they carry no consensus consequence; the decision path wants a typed deferral
that fetches and retries, i.e. giving it the deferral consumer `#318`'s `MissingBlock`
already assumes.

Acceptance for the follow-up, on the same fixture: `ft_witnessed_exact` no longer returns
`Err(MissingBlock)`, AND does not silently abstain either — an unheld justification
surfaces as a deferral the caller can act on. The second condition is the one a filter-
shaped fix satisfies the first without meeting, and no current test would catch that.

**Low**

1. Worth widening the provenance audit past `lookup_unsafe`: this site fails through
`block_number_unsafe`, so a sweep scoped to the one name reports clean.
2. Consider pinning the exact path in `horizon_read_abstention_spec.rs` — its two tests
cover the oracle and estimator only.

For the record, the read-path change checks out: `full_weight_map` is computed at
`clique_oracle.rs:671` before the filter and is not affected by it, so only the numerator
shrinks and FT can only fall. That ordering is load-bearing — filtering `full_weight_map`
as well would invert the direction.

spreston8:
Addressed at `4d8ce106e`, with one substantive correction to the framing: the deferral consumer the follow-up was scoped around already exists end to end. `errors.rs:117` maps `KvStoreError::MissingBlock` to `CasperError::BlockNotHeld`, `floor_of_view` catches it at both sites and returns `FloorOfView::AbsenceHold`, and validation's `guard_deferral` turns `Undecidable` into a deferral on restored nodes. So the probe's `Err` never reached a caller unconsumed — the actual live defect was one layer up: the finalizer's AbsenceHold handler holds the cycle on the premise "catch-up delivers it", which is false for a sub-horizon block, so a stale slot froze the LFB permanently and at debug level. Worse than the storm, quieter too.

Two changes:

- `floor_of_view` abstains unheld testimony slots, by the node-local-clock soundness argument its own doc already makes for the seed filter. This unfreezes the finalizer.
- `agree` resolves an unheld hash in the walk to false. This is deterministic, not node-local: any unheld hash reachable from held references sits below the restore horizon (above-horizon dependencies are fetched before admission; the restore inserts the anchor and above), hence below every held candidate — a fully-held node's walk returns false at the same point by height comparison alone, before ever needing the block. The `is_in_main_chain` storage API keeps its strict contract; only the oracle's agreement boundary interprets absence.

On the acceptance criteria: `ft_witnessed_exact` no longer errors on the fixture, and the second condition is met by something stronger than a non-abstention check — `the_finality_decision_is_identical_with_and_without_the_stale_block` builds the same logical DAG twice (old block held vs hash-only) and asserts identical verdicts. A filter-shaped fix that broke bit-identity would fail it. `a_stale_latest_message_does_not_hold_the_finalizer_forever` pins the finalizer behavior, and `a_sub_horizon_main_parent_crossing_resolves_deterministically` pins the mid-walk crossing your trace identified.

PR #374, opened 2026-08-31
test(coverage): raise workspace line coverage to ~80%

jeffrey-l-turner:
## Summary

Raises workspace unit-test line coverage from 69.5% to an estimated ~80%. Adds roughly 500 tests across all ten crates, with no production-code changes (test-only visibility tweaks aside). Two commits on top of the merged #371 coverage tooling.

### Per-crate results

| Crate | Before | After | Notes |
| --- | ---: | ---: | --- |
| casper | 79.4%* | ~81% (est) | block_retriever admit/eviction branches, running engine paths, no-casper BlockAPI arms, block_report_api from 0% |
| rholang | 69.9%* | 85.7% (measured) | reduce.rs 60.3% -> 71.9%, system_processes.rs 38.5% -> 84.3% via interpreter-evaluation specs |
| node | 51.0% | 65.0% (measured) | web_api_routes 34.8% -> 91.6% (tower-oneshot over the real router), shared_handlers 97.1%, deploy gRPC v1 0% -> 77.7% |
| rspace++ | 71.3% | 83.2% (measured) | merger state_change 12.3% -> 87.0%, merging_logic ~90% |
| shared | 59.6% | 92.2% (measured) | typed/LMDB stores, serde helpers, tracing config |
| models | 57.3% | ~80% (est) | four 0% files now near end-to-end; casper_message round-trips |
| crypto | 78.1% | 92.6% (measured) | signed/signatures_alg from 0%, cert helper error paths |
| block-storage | 63.3%* | ~85% (est) | finality stores from 0%, DAG storage navigation/guards |
| comm | 73.4% | ~79% (est) | hostname trust manager, receiver circuit breaker, rp layer |
| graphz | 73.8% | ~96% (est) | DOT builder end-to-end |

\* after the scaffolding-exclusion baseline correction below.

### Reporting change

Src-shipped test scaffolding (`/test_utils/` dirs, `block-storage/src/rust/test/`) is now excluded from the measured denominator via `--ignore-filename-regex`, identical in `scripts/coverage.sh` and the ci.yml coverage matrix. Under per-crate isolation this code can never be exercised by its own crate's run; excluding it makes the number honest in both directions (it also removed rholang's well-covered scaffolding padding).

### Dedup

[...]

PR #382, opened 2026-09-01
perf(casper): repeat-deploy signature-index fast path

jeffrey-l-turner:
Addresses #24 (the sustained-phase residual: the repeat_deploy BFS ancestor scan).
Design-first change: a CbC design review of docs/casper ran before implementation
and ruled the absence-proof shape admissible only with a design amendment. The
amendment is the first part of this PR and needs maintainer ratification.

## Ratification request

CONSENSUS_PHILOSOPHY.md gains section 4.4 and a 2026-09-01 decision-record row
(status: pending ratification): the repeat-deploy signature index is a
consensus-complete carrier cache over valid, invalid, and approved blocks. An
in-window absence skips the ancestor scan, a hit requires scope verification,
and a read failure falls back to the scan. A node-local index may decide a
verdict only as a provably complete cache of the on-chain data the ratified
predicate reads (P2 is satisfied, not amended: the index is a cache of block
bodies, not a new consensus input).

Companion deltas: a "Repeat-deploy signature index" glossary entry (and an
amended "Prior-rejection count" distinguo), CONSENSUS_PROTOCOL Step 8, and
three claims in cbc-repair-plan "Merge and deploy lifecycle".

## Mechanism

- An index ABSENCE (no terminal record, no event row) proves the sig has no
  carrier anywhere in the DAG, so it cannot be a repeat: the scan is skipped.
  This is the sustained-phase common case (fresh sigs).
- An index HIT is never a verdict: the sig stays in the unchanged exact scan,
  which is the window + parent-scope verification (a fork-only carrier must not
  poison a legal re-inclusion — the "roaming InvalidRepeatDeploy" rule in
  validate.rs).
- ANY read failure keeps the sig in the scan: unreadable index state is no
  information (the BlockNotHeld refusal precedent, PHILOSOPHY 4.1).

## Closing the gap that made d84f26c07 unportable

Lifecycle rows deliberately skipped invalid blocks, but the ancestor scan reads
bodies without a validity qualifier — a sig carried only by an invalid ancestor
IS a repeat. New `LifecycleEventKind::CarriedInvalid` (appended for serde
compat) records that testimony: never a canonical appearance, never a lifecycle
outcome, and inert through every existing consumer's positive filters.

## Completeness invariant

- Lifecycle ingest moved BEFORE the metadata-index add: at every crash point, a
  DAG-visible block has its sigs indexed (a crash leaves harmless orphan rows).
- One-time startup backfill (`ensure_carrier_index_complete`, wired next to the
  LFB migration) restores rows for legacy databases; the completeness marker is
  written ONLY when every DAG-visible body was readable — an LFS-truncated node
  refuses certification and keeps the fast path off.
- Rows are pruned at terminal writes, so a terminal record routes to the scan.

## Evidence

- Verdict-equivalence tests: a certified repeat is still flagged; a certified
  fresh sig is accepted; and the load-bearing one — a repeat carried ONLY by an
  INVALID ancestor gets the SAME verdict from the certified index as from the
  scan (casper/tests/batch2/validate_test.rs).
- Backfill: restore + certify + refuse-on-missing-body + idempotence
  (block-storage/tests/block_dag_storage_test.rs); marker, absence-probe, and
  append-once units (7 tests in deploy_lifecycle_types.rs).
- Pre-push release legs green locally through casper (296s) and block-storage
  (13s); the node leg was cut by a local timeout and rides on CI.

## Notes for reviewers

- interpreter_util.rs (CbC-mandatory) is untouched; the retry-gate exemption
  pass in repeat_deploy is untouched.
- PHILOSOPHY 4.3's scan benchmark (256/512 release gate) has no in-repo
  implementation to re-run; the soak's sustained phase is the operative gate —
  the fast path targets exactly its inclusion-stall signature (run 33462862917:
  sustained inclusion p95 49.6s on the 230-unfinalized iteration).
- "Addresses", not "Closes": #24 keeps the finalization-tail sub-mode until
  soak evidence says otherwise.
- Work log: docs/work-logs/task-issue24-repeat-deploy-fast-path-2026-09-01T15-45Z.md


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.62 |

---

### Summary

The absence-proof design is conservative on ordinary index hits and read failures, and the invalid-carrier coverage is well tested. However, the completeness invariant is not fully fail-closed when DAG metadata is inconsistent, and the crash-safe ordering introduces orphan semantic lifecycle events whose effects extend beyond the carrier index. Because this index participates in consensus validation, both cases require resolution or explicit architectural justification before merge.

This is a well-constructed and unusually well-documented change: the crash-safe ingest-before-visibility ordering, the completeness marker gated on all bodies being readable, the fail-closed fallback to the ancestor scan on any probe error, and the test coverage (including the load-bearing invalid-ancestor equivalence test) all reflect careful design. The remaining concerns are minor: lock churn and lock-hold duration during the one-time backfill walk, the slightly weaker-than-documented dedup semantics of `append_event_once`, an unverifiable-from-the-diff question about whether the ancestor scan reads invalid blocks' rejected deploys (a potential coverage gap if it does), and a suggestion that a backfill error at startup should degrade gracefully rather than abort boot. No security or correctness blockers found; safe to merge after addressing or acknowledging the feedback.

The code change introduces a new fast path for the repeat-deploy signature index, which is a significant improvement for performance under sustained load. However, there are some logic and style issues that need to be addressed to ensure the code is robust and maintainable. The documentation could also be improved to provide clearer explanations of the new features and their roles in the consensus process.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Potential race condition in ensure_carrier_index_complete</b> (Reported by: bedrock)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 815)

The method ensure_carrier_index_complete acquires a write lock on self.global_lock and then a read lock on self.lifecycle. However, it is possible that another thread could acquire a write lock on self.lifecycle before ensure_carrier_index_complete can, leading to a deadlock. Consider acquiring locks in a consistent order across the codebase to avoid this issue.

**Suggestion:**
```
Acquire locks in a consistent order, e.g., always acquire self.global_lock before self.lifecycle.
```

</details>

<details>
<summary><b>:orange_circle: Major: Backfill can certify completeness after skipping missing metadata</b> (Reported by: openai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 866)

The backfill obtains hashes from `dag_set`, but when `metadata_guard.get(hash)` returns `None`, it silently continues and can still write the completeness marker. This contradicts the fail-closed requirement that every DAG-visible block be examined. If the index is inconsistent or partially corrupted, signatures from that skipped block may have no lifecycle row, and validation can incorrectly treat their absence as proof that no carrier exists.

**Suggestion:**
```
Treat a missing metadata record for a hash returned by `dag_set` as an incomplete backfill: log it, set the certification flag to false, and continue without writing the marker, or return an explicit consistency error. Add a test covering this condition.
```

</details>

<details>
<summary><b>:orange_circle: Major: Pre-metadata lifecycle writes can create semantically harmful orphan events</b> (Reported by: openai)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 1354)

Lifecycle events are now persisted before the block is added to the metadata index. If any lifecycle append partially fails, the process crashes, or `block_metadata_guard.add` subsequently fails, Included or Rejected events can remain for a block that never became DAG-visible. The comment characterizes these orphan rows as harmless because they only force an exact repeat-deploy scan, but the same rows are also consumed by lifecycle operations such as canonical appearance, valid-after tracking, rejection accounting, and potentially retry adjudication. Consequently, an orphan Included or Rejected event may affect behavior beyond the conservative carrier-index lookup and can produce node-local state that does not correspond to the DAG.

**Suggestion:**
```
Do not use semantic lifecycle events as the pre-visibility crash-safety record. Consider a separate carrier-only index, or write a neutral carrier-testimony event before metadata visibility and project Included/Rejected lifecycle semantics only after visibility. Alternatively, make the metadata and lifecycle updates atomic or ensure all lifecycle consumers verify that an event's block is DAG-visible.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Per-block lock re-acquisition in backfill loop</b> (Reported by: openrouter)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 870)

Inside `ensure_carrier_index_complete`, `self.block_metadata_index.read()` is acquired and dropped once per hash in the DAG set, in addition to the lifecycle write lock held across the entire walk. For large DAGs this is O(n) lock churn plus a metadata KV lookup per block while a global write lock is held, blocking all DAG readers for the duration of the backfill. Since this runs at startup it is likely acceptable, but consider hoisting the metadata reads (dag_set already collects hashes; the metadata could be fetched in a batch before taking the lifecycle write lock) or chunking the work to bound worst-case startup latency and lock hold time.

**Suggestion:**
```
Collect (hash, invalid) pairs in one pass before acquiring `lifecycle.write()`, then run the body walk and event appends. Optionally process in chunks and log progress.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Rejected deploys of invalid blocks are not indexed — verify parity with the ancestor scan</b> (Reported by: openrouter)</summary>

**File:** `block-storage/src/rust/dag/block_dag_key_value_storage.rs` (line 892)

The backfill (and the insert path) skip `rejected_deploys` for invalid blocks, indexing only `body.deploys` as `CarriedInvalid`. The stated invariant is that the index covers 'the same block universe the ancestor scan reads'. The diff does not show the ancestor scan's body-reading code, so it cannot be verified here whether the repeat-deploy scan also inspects `rejected_deploys` of invalid blocks. If it does, the index has a coverage gap and an absence proof could be unsound for a sig whose only carrier is an invalid block's rejected list.

**Suggestion:**
```
Confirm the repeat-deploy ancestor scan reads only `body.deploys` (not `rejected_deploys`) of invalid blocks, and add a comment or assertion pinning that parity; otherwise index invalid blocks' rejected deploys too.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: append_event_once dedup ignores event payload differences</b> (Reported by: openrouter)</summary>

**File:** `block-storage/src/rust/dag/deploy_lifecycle_types.rs` (line 200)

The duplicate check in `append_event_once` compares only `block_hash` and the discriminant of `kind`. Two events for the same sig in the same block with the same kind but different payloads (e.g. `Included { is_failed: true }` vs `Included { is_failed: false }`, or `Rejected` with different `duplicate`/`carrier` fields) would be treated as duplicates and the second silently dropped. In practice a sig appears at most once per block body so this is unreachable today, but the idempotency contract is subtly weaker than 'same testimony' as documented.

**Suggestion:**
```
Either compare the full event (`e == &event`, since LifecycleEvent derives PartialEq) or document explicitly that dedup is per (block, kind-discriminant) and why that is sufficient.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Improve error handling readability</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/validate.rs` (line 689)

The error handling in the repeat-deploy signature index fast path could be improved for readability. Consider using match statements instead of if-else chains to handle different error cases.

**Suggestion:**
```
Refactor the error handling using match statements for better readability.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Clarify the role of the repeat-deploy signature index</b> (Reported by: bedrock)</summary>

**File:** `docs/casper/CONSENSUS_PHILOSOPHY.md` (line 72)

The documentation could be more explicit about the role of the repeat-deploy signature index in the consensus process and how it interacts with the ancestor scan.

**Suggestion:**
```
Add a section detailing the interaction between the repeat-deploy signature index and the ancestor scan, including how it affects consensus decisions.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Session work-log committed to the repository</b> (Reported by: openrouter)</summary>

**File:** `docs/work-logs/task-issue24-repeat-deploy-fast-path-2026-09-01T15-45Z.md` (line 1)

The work-log file contains session-specific metadata (claimed_by session id, instructions to the user about /quick-commit and /recursive-push, PR-body wording). This is useful during development but reads as process scaffolding rather than durable project documentation; the durable content (design facts for reviewers) belongs in the PR description or the design docs, which already received thorough updates.

**Suggestion:**
```
Move the 'Key design facts for reviewers' section into the PR body and consider excluding the work-log from the merge, or strip the session-specific front matter.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Backfill error handling and startup-time observability</b> (Reported by: openrouter)</summary>

**File:** `node/src/rust/runtime/setup.rs` (line 144)

`ensure_carrier_index_complete` is called with `?` at startup, so a transient KV error aborts node startup entirely, whereas the design philosophy elsewhere in this PR is 'read failure is no information — fall back to the scan'. A backfill failure could safely degrade to the uncertified path (marker simply stays unwritten) instead of preventing the node from booting. Also, for large DAGs the backfill is a long synchronous walk with a single info log at the end; periodic progress logging would help operators.

**Suggestion:**
```
Consider logging a warning and continuing startup on backfill error (fast path just stays off), and add periodic progress logs during the walk.
```

</details>
---

<details>

The code change introduces a new fast path for the repeat-deploy signature index, which is a significant improvement for performance under sustained load. However, there are some logic and style issues that need to be addressed to ensure the code is robust and maintainable. The documentation could also be improved to provide clearer explanations of the new features and their roles in the consensus process.

#### openai (gpt-5.6-sol)

The absence-proof design is conservative on ordinary index hits and read failures, and the invalid-carrier coverage is well tested. However, the completeness invariant is not fully fail-closed when DAG metadata is inconsistent, and the crash-safe ordering introduces orphan semantic lifecycle events whose effects extend beyond the carrier index. Because this index participates in consensus validation, both cases require resolution or explicit architectural justification before merge.

This is a well-constructed and unusually well-documented change: the crash-safe ingest-before-visibility ordering, the completeness marker gated on all bodies being readable, the fail-closed fallback to the ancestor scan on any probe error, and the test coverage (including the load-bearing invalid-ancestor equivalence test) all reflect careful design. The remaining concerns are minor: lock churn and lock-hold duration during the one-time backfill walk, the slightly weaker-than-documented dedup semantics of `append_event_once`, an unverifiable-from-the-diff question about whether the ancestor scan reads invalid blocks' rejected deploys (a potential coverage gap if it does), and a suggestion that a backfill error at startup should degrade gracefully rather than abort boot. No security or correctness blockers found; safe to merge after addressing or acknowledging the feedback.

#### xai (grok-4.5)

> **ABSTAIN**: request timed out after 115s (model: grok-4.5, tools: false)
</details>

---

jeffrey-l-turner:
Response to the multi-agent review above. All three major findings are dispositioned in commit ba48ec5d3.

## Major 1 — Backfill can certify completeness after skipping missing metadata (openai)

**Confirmed and fixed.** A hash the DAG set lists but the metadata store cannot resolve now refuses certification, the same as a missing body: the walk logs a warning, leaves the completeness marker unwritten, and the fast path stays off. Regression test: `carrier_index_backfill_refuses_certification_on_missing_metadata`, which simulates the inconsistency through a test-only helper that deletes the persisted metadata row while the in-memory DAG state still lists the hash.

## Major 2 — Pre-metadata lifecycle writes can create semantically harmful orphan events (openai)

**Confirmed in a weakened form and fixed on two fronts.**

1. The insert path now appends through the idempotent `append_event_once` (per block + kind dedup). The realistic crash sequence — lifecycle rows written, metadata add lost, block redelivered after restart — re-runs the ingest pass and writes nothing twice. Test: `insert_retry_after_ingest_first_crash_does_not_duplicate_events` (full crash / restart / redelivery cycle).
2. Canonical appearance is now DAG-visibility filtered at both twins (the representation filters on `dag_set`, the storage twin on the metadata index), so an orphan event never resolves as an appearance. Test: `orphan_lifecycle_event_is_not_a_canonical_appearance`.

On the wider consumer concerns the review raised: the finality evaluator's `Failed` verdict was already safe — it requires the inclusion block to be in the floor closure (`in_floor_closure`), and a never-DAG-visible block cannot be. `valid_after` is derived from the signed deploy content, so it is identical across copies and an orphan cannot record a wrong value. For the carrier index itself an orphan row only forces the conservative exact scan. The remaining residual is display-level only.

## Major 3 — Potential deadlock between global_lock and lifecycle in the backfill (bedrock)

**Refuted as a deadlock, and the convention is now documented.** Every mutator, including `insert` and the backfill, acquires the locks in the same order: `global_lock` -> `block_metadata_index` -> `lifecycle`. A deadlock needs two paths that acquire in opposite orders, and none exists. The order is now stated on `ensure_carrier_index_complete` and on the storage-level appearance twin so future call sites keep it.

On the minor parity question (invalid blocks' `rejected_deploys` are not indexed): verified as intentional. The ancestor scan's probe (`has_any_deploy_sig_strict`) reads only `body.deploys` signatures — `rejected_deploys` is never in the scan's read set — so the index projects exactly the scan's universe and no more. The remaining minor findings and suggestions are noted for follow-up.

jltatbeach (approved):

spreston8 (commented):
Reviewed at `ba48ec5d3` against `dev` (`0b6b65501`). Suggestions below.

**HIGH**

1. The completeness marker is a reserved key in the `events` table, and
`carrier_index_complete()` returns true on presence of that key regardless of what
wrote it. `RejectedDeploy::from_proto` (`casper_message.rs:562-569`) copies
`proto.sig` verbatim with no verification, and `try_admit_settled` inserts settled
history "hash-checked and unjudged", so `insert_internal`'s `!invalid` branch
writes an attacker-chosen key — a bonded validator can certify an uncertified node,
after which every unindexed signature reports provably-absent and the scan is
skipped. Probe against this head:

```rust
block.body.rejected_deploys = vec![RejectedDeploy {
    sig: b"__carrier-index-complete\0".to_vec().into(),
    duplicate: false, carrier: vec![7; 32].into(),
}];
dag_storage.insert(&block, InsertMode::SettledHistory).unwrap();
```
```
PROBE before insert: carrier_index_complete = Ok(false)
PROBE after settled-history insert: carrier_index_complete = Ok(true)
PROBE proves_absence(never-indexed sig) = Ok(true)
```

`ensure_carrier_index_complete:862-867` short-circuits on the marker at every later
start, so the false certification never clears. Giving the marker its own KV store
removes the collision; a length check on `rd.sig` would leave the coupling in
place. (`body.deploys` is closed — `Signed::from_signed_data` rejects at decode.)

**Medium**

2. Carrier testimony in `LifecycleEventKind` has no retention path: a sig carried
only by invalid blocks gets `valid_after: None`, so `terminalize_or_rearm` returns
at `deploy_lifecycle.rs:555-557` and never terminalizes, and the terminal write is
the only thing that prunes an event row. Those rows are therefore permanent, which
contradicts `deploy_lifecycle_types.rs:10-12` ("structurally bounded to OPEN sigs")
and lets an equivocating validator mint immortal rows on every node by signing
deploys that appear only in its own invalid blocks. A separate carrier table
dropped below the expiration window carries no lifecycle semantics to violate, and
pairing it with a persisted height watermark (fast path engaging only when
`earliest_block_number >= W`) also deletes the startup backfill — which today
re-walks every body under `global_lock.write()` at every boot on a node that cannot
certify, and which a version-less marker cannot re-run for the next coverage
extension the way this PR needed one for `CarriedInvalid`.

3. This branch carries #374's merge commit `487a35c29` (109 files, +15760). GitHub
records #374 as merged into `dev` and `dev` does not contain it
(`compare/dev...487a35c29` reports diverged), so merging this re-lands work that
was removed from `dev` — confirming whether that removal was deliberate would avoid
undoing it by accident.

4. `ensure_carrier_index_complete` takes `lifecycle.write()` at `:874` and then
`block_metadata_index.read()` at `:877`, the reverse of the order its own doc
states at `:855-856` and the reverse of the storage twin at `:832-835`. The
deadlock is unreachable only because every metadata writer holds `global_lock`,
which the backfill also holds — not because the order is honored, so the reply to
`bedrock` reaches the right conclusion by the wrong route. Collecting
`(hash, invalid)` inside the metadata read at `:868-871` before taking
`lifecycle.write()` restores the order and removes the per-block lock churn
`openrouter` raised.

5. `remove_block_metadata_row_for_tests` (`:950`) and `delete_kv_row_for_tests`
(`block_metadata_store.rs:254`) are `pub` with only `#[doc(hidden)]`, so production
code can delete the persisted metadata row for a DAG-visible block — the exact
inconsistency the backfill's fail-closed path exists to survive. `:1050-1057`
records that weaker gating on the neighbouring escape hatches was already rejected
as silenceable, and `#[cfg(any(test, feature = "test-internals"))]` matches them.

6. Nothing in the PR fails if the optimization is removed: forcing the gate to
`false` and re-running leaves all three certified-index tests green, because each
asserts only a verdict and the verdict is identical whether the scan ran or was
skipped. A regression that silently disables the fast path therefore ships clean.
Asserting that a certified index plus a fresh sig plus an unreadable block store
still returns `Valid` would pin engagement, reusing the failure injection the
storage-failure test below them already sets up.

7. `frozen_display` (`deploy_lifecycle.rs:679-693`) picks `latest_block_hash` by
`max` over every `Included` event with no visibility filter, so a crash-window
orphan can freeze into a write-once terminal record and the terminal branch of
`canonical_appearance` then returns it unfiltered forever. The comment at
`deploy_lifecycle_types.rs:126-128` says that cannot happen because terminals are
floor-closure guarded, but the guard covers the `Failed`/`Expired` determination,
not the block hash the record stores. Applying the same visibility predicate inside
`frozen_display` closes it.

jeffrey-l-turner:
Response to the maintainer review at `ba48ec5d3`. All seven findings are dispositioned; the redesign your item 2 sketched is adopted wholesale, which resolves items 1, 2, and 4 structurally rather than by patching. Landing on the branch as the next commit.

**1 (HIGH) — marker forgery through an unverified `rejected_deploys` sig: confirmed, resolved structurally.** The completeness marker no longer exists. The carrier index moves out of the lifecycle tables into dedicated stores (`carrier-index`, `carrier-index-meta` — `carrier_index.rs`), so there is no shared keyspace for wire data to key. The index also ingests ONLY decode-verified `body.deploys` sigs — `rejected_deploys` sigs are deliberately never recorded (the scan never reads them). Your probe's write path now lands in an ordinary lifecycle row with no meaning to the fast path.

**2 — immortal `CarriedInvalid` rows: confirmed, resolved by your sketch.** `CarriedInvalid` is removed from `LifecycleEventKind` entirely (nothing shipped, so no decode debt). Carrier testimony for every block — valid, invalid, settled — lives in the dedicated table, keyed by height, pruned on floor advances (`observe_block` drives `prune_below(adopted − lifespan)`, strided so most advances no-op; retention failures warn and never touch the verdict path). The lifecycle tables return to their pre-PR contract: invalid bodies contribute nothing, rows stay structurally bounded to open sigs.

**3 — the #374 re-land into `dev`:** deliberate, per the branch owner. The coverage stack is on `master`; `dev` converges at the next promotion.

**4 — lock-order violation in the backfill: confirmed, and now moot.** The backfill is deleted, not fixed. The persisted watermark W (first-boot: 0 on an empty database, max height + 1 on an existing one) replaces walk-completeness entirely: the fast path engages only when the scan window starts at or above W (`w <= max(earliest, 0)`), so heights the index never saw are never claimed. As you noted, this also removes the per-boot walk under `global_lock` and gives the next coverage extension a re-runnable mechanism (a new store starts with its own watermark) instead of a version-less marker. A side benefit: LFS-truncated nodes now certify trivially, where the walk-based design permanently refused them.

**5 — `pub` corruption helpers: confirmed, gated** with `#[cfg(any(test, feature = "test-internals"))]`, matching the P2-16 escape hatches. One wrinkle: block-storage's own integration tests cannot reach the feature (a crate cannot dev-depend on itself — the P2-16 note in Cargo.toml documents exactly this trap), so the crash-retry test now stages the ingest-half state through the public fixture handles instead of deleting metadata rows.

**6 — nothing pins engagement: confirmed, pinned.** New test `repeat_deploy_certified_index_engagement_skips_the_scan`: the candidate's ancestry contains an unreadable parent, the uncertified control asserts the scan fails on it, and the certified run returns Valid — a verdict reachable only through the skip, reusing the failure injection you pointed at.

**7 — `frozen_display` freezes orphan hashes: confirmed, fixed.** The visibility predicate now applies inside `frozen_display` itself; a never-DAG-visible inclusion cannot become the frozen `latest_block_hash`, and a unit test pins both directions. You were right that my earlier "terminals are floor-closure guarded" claim covered the verdict, not the stored hash.

Docs follow the design: CONSENSUS_PHILOSOPHY §4.4 restates completeness as write-ordering + watermark (and now names the dedicated-store requirement with this review as the reason), the cbc-repair-plan claims are updated (including a retention obligation: no entry at or above any future window's start may be dropped), and the glossary entry describes the watermark mechanics.

Evidence: 3 unit tests on `CarrierIndex` (idempotence, write-once watermark, strided pruning), the reworked storage integration tests (carrier projection incl. invalid carriers, watermark initialization, crash-retry idempotence, orphan non-appearance), the engagement pin plus the three verdict-equivalence tests in validate_test.rs, and the `frozen_display` orphan pin. All green locally with clippy clean.

PR #136, opened 2026-07-20
Stabilize node recovery under concurrent bridge load

a-k-l-sdao:
## Fixed

- Recovered deploy processing could stall or be requeued inconsistently during merge recovery — `e08d8567`.
- Concurrent bridge locks could lose progress during DAG merge and replay recovery — `2c339cbb`.
- Finality and heartbeat recovery could stall under sustained bridge load — `0c6b7682`.
- Observer queries and missing-block recovery could timeout or starve under load — `2317decf`.
- Sequential liveness cleanup could evict slow but healthy peers during load spikes — `05da9342`.

## Validation

- `cargo test -p comm --test mod clear_connections_spec`
- `cargo check -p node`
- `cargo test -p node configuration --lib`
- 5,348 bridge locks finalized exactly with zero submission failures, restarts, or OOM kills.
- Fresh optimized concurrency stages passed at 32, 64, and 100 locks with exact nonce, balance, and locked state; the 10K soak continues.

## Note

This draft includes the prior all-open-PR integration baseline. `origin/dev` advanced during soak testing, so the tested history has not been rewritten yet.

---
<a href="https://app.blacksmith.sh/F1R3FLY-io/codesmith/f1r3node-rust/pr/136"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787152923&installation_id=145946300&pr_number=136&repository=F1R3FLY-io%2Ff1r3node-rust&return_to=https%3A%2F%2Fgithub.com%2FF1R3FLY-io%2Ff1r3node-rust%2Fpull%2F136&signature=5af4c862fcf59859a10beeac8bbee2d11b1b1dc78667532107532a048dc3e812"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
[...]

PR #384, opened 2026-09-01
feat(casper): consensus parameters graduate on-chain; every tunable gets one governed home

### LINKED ISSUE #245: Magic constants: duplicated and inline values across the codebase
<html><head></head><body>
<p>Checked against <code>dev</code> @ <code>71c4d11</code>.</p>
<h2>Summary</h2>
<p>The node already has a config system (<code>defaults.conf</code> -&gt; <code>NodeConf</code> / <code>CasperConf</code>) and good named-const discipline for crypto and protocol values. What is left are inline literals — some of them copied into two or three files, so changing a value means editing several places and missing one is easy.</p>
<h2>Same value, several places</h2>

Value | Where | Suggested fix
-- | -- | --
Duration::from_secs(128) LFS backoff cap | lfs_block_requester.rs:874, lfs_horizon_requester.rs:503, lfs_tuple_space_requester.rs:581 | One shared const — these three must agree
Duration::from_secs(5) retry sleep, with copy-pasted comments | node_runtime.rs:1176, 1195, 1216 | One module-level const
500ms status-slow threshold, two function-local consts | web_api.rs:309, shared_handlers.rs:395 | One shared module-scope const
Ollama defaults (http://localhost:11434, llama4:latest, 30) | ollama_service.rs:22-24 (from_env), :55-57 (disabled), plus defaults.conf | One defaults module as the single in-binary source; check openai_service.rs too
Keepalive triplet (600s / 30s / 5s) | f1r3fly_server.rs:82-85, grpc_transport_receiver.rs:621-624 | Config block, read by both
4096 active-validators cache | named const at casper.rs:46, but written as a bare literal at casper_launch.rs:193 | Use the existing const

<p>Two of these deserve a closer look:</p>
<p><strong><code>PAGE_SIZE</code> is 750 in one file and 1024 in another</strong> — <code>lfs_tuple_space_requester.rs:46</code> vs <code>lfs_horizon_requester.rs:48</code>, same name, sibling modules, no recorded reason. Paging size is an agreement between requester and responder, so either unify the values or rename them apart with a comment. This is the most likely latent bug here.</p>
<p><strong>Inter-node keepalive quietly disagrees with the API side</strong> — the API server's keepalive is already configurable (<code>defaults.conf</code>, 5 minutes), while the inter-node one is hardcoded at 600s in two files. <code>grpc_transport_receiver.rs:643</code> also has an inline 30s timeout.</p>
<h2>Should be config, not consts</h2>
<p><strong>RuntimeManager caches</strong> — <code>runtime_manager.rs:110-117</code>, eight consts inside the <code>impl</code> (block index 128, parents post-state 64, active validators 256, bonds 64, replay 192, replay bytes 32 MiB, replay event log 1536, state hash 0). These are memory-vs-hit-rate tunables that operators may want to change, so they fit <code>CasperConf</code> with <code>default_*</code> functions like the existing ones. Worth explaining <code>MAX_STATE_HASH_CACHE_ENTRIES = 0</code> while doing it — the cache is disabled and nothing says why.</p>
<p><strong>BlockRetriever</strong> — <code>block_retriever.rs:91-101</code>, eleven consts inside the <code>impl</code>. Suggested split:</p>
<ul>
<li>caps (<code>MAX_REQUESTED_BLOCKS_ENTRIES</code>, <code>MAX_WAITING_LIST_PER_HASH</code>, <code>MAX_RETRIES_PER_HASH</code>, <code>KNOWN_PEER_REQUERY_SOFT_LIMIT</code>, <code>MISSING_DEPENDENCY_SEED_PEERS</code>, <code>STALE_REQUEST_LIFETIME_MULTIPLIER</code>) — lift out of the <code>impl</code> so they are addressable;</li>
<li>timings (three <code>*_COOLDOWN_MS</code>, <code>RETRY_BUDGET_QUARANTINE_MS</code>) — move to config. All three cooldowns being 500ms looks like copy-paste rather than a decision.</li>
</ul>
<p><strong>Startup deadlines</strong> — <code>initializing.rs</code>: <code>:594</code> <code>lfs_request_timeout = 5s</code>, <code>:595</code> <code>LFS_SYNC_DEADLINE = 600s</code>, <code>:923</code> <code>MERGEABLE_CHANNEL_REPLAY_DEADLINE = 1200s</code>, <code>:699</code> inline 30s, <code>:281</code> inline <code>sleep(10s)</code>.</p>
<p>The <code>:594</code> one is explicit tech debt: the comment above it says the value should come from an env var and fall back to 5s, but the fallback was never implemented — the literal is unconditional. A <code>CasperConf.startup</code> section covering all five would close that TODO.</p>
<h2>Small cleanups</h2>
<ul>
<li><code>grpc_transport_client.rs:426</code> — <code>packet.content.len() * 5</code> µs. The <code>5</code> is an unexplained bandwidth assumption (~200 KB/s); give it a name. <code>:493</code> has an inline 50ms quick-check.</li>
<li><code>grpc_server.rs:148</code> — inline 1000ms shutdown grace, right next to the properly named <code>GRPC_BIND_RETRY_DELAY</code> at <code>:11</code>.</li>
<li><code>repl_client.rs:59-60</code> — 5s connect / 30s request, both already marked <code>TODO adjust ... if necessary</code>.</li>
<li><code>f1r3fly_connector.rs:336, 377</code> — inline 10s and 5s next to the named <code>DEFAULT_CONNECT_TIMEOUT</code> at <code>:22</code>.</li>
<li><code>heartbeat_proposer.rs:239, 242</code> — backoff cap 60s and shift cap 4.</li>
<li><code>upnp.rs:57</code> — <code>8.8.8.8:80</code> for public-IP discovery, plus an inline 3s timeout at <code>:420</code>. Naming it lets operators behind blocked Google DNS switch endpoint without a code change.</li>
</ul>
</ul></body></html>

spreston8:
Top of the stack — #368 and #369 are merged into dev. **Consensus-affecting, deliberately**: max-parent-depth, deploy-lifespan, and min-phlo-price are read by validity rules (parent spread, expiry, repeat-deploy, phlo floor), so nodes that disagree on them fork — until now they were held uniform only by ops discipline. This PR bakes them into the genesis PoS contract and every node adopts the on-chain values unconditionally at casper construction (the protocol-FTT pattern). The values are immutable per chain until an on-chain upgrade mechanism exists (deliberately deferred). The rest of the PR is the governance pass that decision enables: every remaining tunable gets exactly one home, at the right tier.

### Diff anatomy

|---|---:|---|---|
| Production | 44 | +754 −678 | The graduation (macro params, strict range-gated reader, single adoption point), the single-source repairs, the env-var collapse, the post-adoption geometry re-checks, the review hardening (adoption-point completion, backpressure retry credit + drop metric), the derived exploratory-concurrency default |
| Genesis contract | 1 | +11 −0 | PoS.rhox macro params + a `getConsensusParameters` 3-tuple getter; ceremony participants must match them byte-for-byte |
| Config & validation | 4 | +325 −17 | `deploy-lifespan` conf key born (seeds genesis only); width cap 64→12 with I3/I4 startup checks (I3 warns on the local depth — the authoritative re-judgement runs on the chain-adopted one); hard-fail on geometry the rules are undefined under |
| Tests | 31 | +350 −122 | Red-first adoption + enforcement specs, geometry boundary specs, range-gate spec, retry-credit spec, ripples |
| Docs | 7 | +18 −149 | Consensus docs track the graduation (min-phlo chain-adopted; `LowDeployCost` admission-only, non-slashable), operator README tracks the removed env vars and the new cap, duplicate validation doc deleted |
| CI | 4 | +15 −3 | system-integration pin → main (carries SI #136, the observer-overload cap pin); slow-deploy convergence test deselected pending its own hardening PR |

### What changed

- **Graduation** — genesis bakes the three parameters into the PoS contract; `hash_set_casper` reads them back through a strict, range-gated exploratory query and adopts them unconditionally (`reconcile(local, onchain) = onchain`); absent or out-of-range values fail startup rather than fall back to local configuration. The `LowDeployCost` verdict is enforced instead of logged — fork-safe now that the floor is chain-adopted, and deliberately non-slashable (admission-only).
- **Single source** — `disable-late-block-filtering` is deleted (its legacy arm fed the merge a node-local, non-deterministic set; the floor-relative window is the sole lateness definition — the flag's production-reachable behavior is unchanged since no live caller took the legacy arm). The estimator is stateless (parent bounds travel per call from the snapshot's chain-adopted shard conf), the mergeable-GC loop reads the live adopted conf (its serde fallback now matches the shipped default, pinned by test), and deploy admission plus the status endpoints read the same floor the validity rule reads.
- **Startup contract, finished at the adoption point** — hard errors for non-positive mpd/lifespan/epoch-length and negative min-phlo; warnings for real side effects (the mpd sentinel disabling terminal deploy verdicts, network-timeout above the proposal cadence, disabled mergeable-GC, a width cap at or below the hard backpressure tier). Because startup validates the *local* values and adoption then overwrites them, everything mpd-derived is recomputed and re-judged against the *adopted* values: a derived deploy-play-budget is rebuilt from the adopted mpd, and the budget-vs-citability (I1), recovery-span, lifespan-vs-mpd, and cap-vs-mpd inequalities are re-judged where the adopted values first exist.
- **Width cap 64→12** — the empty-frontier cap only bounds validity-window burn during a finalization stall if it sits at or below the citability depth; 64 > 15 could never do its one job. A cap above max-parent-depth warns at startup (the chain-adopted depth may legitimately differ; the once-per-run authoritative check runs against it in the heartbeat), with the boundary pinned: cap == mpd is silent, mpd+1 warns.
- **Backpressure is observable and free** — a block dropped at our own in-flight cap deposits a one-shot credit in the retriever, so the credited re-request does not spend the per-hash retry budget, and every drop increments `block.inflight-cap.drops`, labelled by drop site.
- **Derived exploratory concurrency** — `exploratory-deploy-max-concurrent` defaults to `0` = derive `max(2, cores − 2)` at startup. Observers are the only nodes serving these reads, so the default scales to the hardware instead of pinning a laptop-era 2; explicit values are honored (clamped to the semaphore range), and the resolved value is logged.
- **Env collapse** — all 11 `F1R3_*` vars are gone; env is not a configuration home. Buffer, attempt, quarantine, and malloc-trim knobs are plain constants; the validation-error quarantine gets its own duration instead of aliasing the missing-dependency one; `MAX_BLOCKS_IN_PROCESSING`'s three definitions (512/2048/2048) guarding one shared in-flight set collapse to a single const 512 (the receipt gate tightens 2048→512; a dropped block is re-fetched by the re-request clock), with a compile-time assert against the processor queue capacity; the default-off trigger-propose-after-processing flag and its arm are deleted.
- **Deploy-count bound** — `ORDINARY_DEPLOY_PROPOSAL_CAP` (a compiled 128 that silently clamped `max-user-deploys-per-block`) is deleted; the conf key is the single count bound, and the repurposed test proves conf 150 is honored above the old ceiling. Bytes and the play budget bound the real resources.
- **Registry** — the dead `config_version` field is deleted; the version literal consolidates into `CASPER_PROTOCOL_VERSION`; `DeployIndex`'s system-deploy ids derive from the system-deploy markers (one quantity, one home); consensus-bearing constants (DRIFT, markers, deploy-index costs) carry the tag stating why they are compiled.

### Issue #245

Closes #245. This PR lands its consolidation pass: the duplicated literals (LFS backoff cap ×3, loop retry sleeps ×3, status threshold ×2, keepalive triplet ×2, Ollama/OpenAI defaults ×2, the bare 4096 beside its named const) each get one definition, and the inline literals the issue lists get names. Two deviations from the issue's suggestions, deliberate: the `PAGE_SIZE` divergence (750 vs 1024) is **not** the latent bug suspected — the page size travels in the request and the responder pages by it, so the two constants are independent per-protocol choices, now documented as such; and the RuntimeManager caches, BlockRetriever timings, keepalive, and startup deadlines are **not** promoted to configuration — they are developer-tier values with no demonstrated operator need, and a knob is a promise that invites un-fixing verified behavior. The BlockRetriever caps also stay inside the `impl` rather than being lifted "so they are addressable": a const is fully addressable where it is used, and only `MAX_RETRIES_PER_HASH` has an external reader (it is lifted and `pub`). Three issue items were already resolved upstream.

### CI note

`test_network_converges_after_slow_deploy` is deselected on all legs: forensics on run 34191496437 showed its 20M-phlo carrier taking 272s to execute under runner load (~11x its ~25s budget), wedging all replicas in simultaneous replay until the harness memory guardian killed the shard — a latent load-sensitivity in the test's design, not a regression here. Re-enabling it gets its own hardening PR.

### Suggested reading order

1. `PoS.rhox` (the getter) → `runtime.rs::get_consensus_parameters` (the strict read and range gates) → `token_metadata_check.rs::read_on_chain_consensus_parameters` (the no-fallback policy) → `casper.rs::hash_set_casper` (the adoption and the post-adoption re-judgements)
2. `configuration/mod.rs` (the startup contract) + `width_cap_geometry_is_validated_at_startup`
3. `runtime_manager_test.rs::consensus_parameters_are_adopted_from_chain_over_local_config` (the adoption red) and `multi_parent_casper_add_block_spec.rs::multi_parent_casper_should_reject_block_carrying_underpriced_deploy` (the enforcement red)
4. `block_processor.rs` (the constants block and the shared in-flight cap) → `block_retriever.rs::a_local_backpressure_drop_credits_one_retry_attempt` (the credit) → `block_processor_instance.rs` (the deleted arm)


spreston8 (commented):
Reviewed at dce878496 against dev 0f5d2b741.

**HIGH**

1. Move the geometry judgements to the adoption point. `configuration/mod.rs:198` feeds both guards (`:199`, `:229`) from the local `max-parent-depth`, and `casper_launch.rs:174` derives `deploy_play_budget` from it; `casper.rs:295` then replaces mpd and leaves all three behind. Local 15 against chain 5 gives a 25s citability window carrying a 15s budget with both guards silent — the stall #369's guard was added to catch. Only `casper.rs:304` and `heartbeat_proposer.rs:545` are re-judged.

**Medium**

2. Demote the I3 check at `configuration/mod.rs:326` to a warning. It hard-fails on two local values while the same inequality against the adopted value only warns (`heartbeat_proposer.rs:550`), so chain mpd 21 with cap 20 refuses to start on geometry that is legal under the value it runs.

3. Skip the retry-attempt increment when `running.rs:187` drops at the in-flight cap. The cap tightened 2048 to 512; each re-request spends #369's per-hash budget, so local backpressure can quarantine a locally-available block. Whether the cap holds the ~58s needed to exhaust one is unmeasured.

**Low**

4. Meter the cap drop: `running.rs:189`, `casper_launch.rs:345`, `block_processor_instance.rs:148,:276` are WARN-only.
5. `enable-mergeable-channel-gc` ships true (`defaults.conf:324`) and falls back to false (`casper_conf.rs:156`).
6. The status endpoints advertise the local phlo floor until casper is up; omit it or mark it provisional.
7. Delete `docs/adaptive-deploy-cap-validation-2026-03-05.md` — `docs/archive/` holds the same file with the removal note.

spreston8:
> Reviewed at dce878496 against dev 0f5d2b741.
> 
> **HIGH**
> 
> 1. Move the geometry judgements to the adoption point. `configuration/mod.rs:198` feeds both guards (`:199`, `:229`) from the local `max-parent-depth`, and `casper_launch.rs:174` derives `deploy_play_budget` from it; `casper.rs:295` then replaces mpd and leaves all three behind. Local 15 against chain 5 gives a 25s citability window carrying a 15s budget with both guards silent — the stall #369's guard was added to catch. Only `casper.rs:304` and `heartbeat_proposer.rs:545` are re-judged.
> 
> **Medium**
> 
> 2. Demote the I3 check at `configuration/mod.rs:326` to a warning. It hard-fails on two local values while the same inequality against the adopted value only warns (`heartbeat_proposer.rs:550`), so chain mpd 21 with cap 20 refuses to start on geometry that is legal under the value it runs.
> 
> 3. Skip the retry-attempt increment when `running.rs:187` drops at the in-flight cap. The cap tightened 2048 to 512; each re-request spends #369's per-hash budget, so local backpressure can quarantine a locally-available block. Whether the cap holds the ~58s needed to exhaust one is unmeasured.
> 
> **Low**
> 
> 4. Meter the cap drop: `running.rs:189`, `casper_launch.rs:345`, `block_processor_instance.rs:148,:276` are WARN-only.
> 5. `enable-mergeable-channel-gc` ships true (`defaults.conf:324`) and falls back to false (`casper_conf.rs:156`).
> 6. The status endpoints advertise the local phlo floor until casper is up; omit it or mark it provisional.
> 7. Delete `docs/adaptive-deploy-cap-validation-2026-03-05.md` — `docs/archive/` holds the same file with the removal note.
> 
> Co-Authored-By: Claude <noreply@anthropic.com>
> 

Addressed at dff40fdee.

1. Done, and taken to completion: CasperShardConf now carries the budget-derive sentinel and the heartbeat cadence, and hash_set_casper finishes everything mpd-dependent on the adopted value — a derived deploy_play_budget is recomputed from the adopted mpd, and the I1 and full recovery-span inequalities are re-judged there, joining the lifespan and width-cap re-judgements. The local-15/chain-5 scenario now recomputes the budget to 5s and warns on any residual violation.

2. Done — the startup message notes the adopted depth may legitimately differ; the runtime re-judgement is authoritative. Chain mpd 21 / cap 20 starts.

3. Done — a drop at our own cap deposits a one-shot credit in the retriever; the credited re-request doesn't count against the per-hash budget (spec: a_local_backpressure_drop_credits_one_retry_attempt). The unmeasured question is now moot for the quarantine path, and item 4's meter gives the observable if drops ever sustain.

4. block.inflight-cap.drops, site-labelled across all four drop sites.

5. Fallback flipped to true and pinned against the shipped default (mergeable_gc_fallback_matches_shipped_default).

6. Resolved as documentation: isReady: false already marks the whole payload provisional, and the api-reference now says so on minPhloPrice — kept the field rather than a wire change.

7. Deleted; the archived copy is the keeper.

Issue #344, opened 2026-08-25
Validators abort with a stack overflow on wider heartbeat profiles


9Neechan:
## Summary

Two of three heartbeat profiles lost every validator to a stack overflow. Each validator aborts on its first propose after start, so the crash repeats on every restart and the shard never recovers. The third profile ran the full 48 hours on the same image and the same cluster.

The two failed shards died at heights 3825 and 3823. The surviving shard finalized to height 104632.

```
thread 'tokio-rt-worker' (10) has overflowed its stack
fatal runtime error: stack overflow, aborting
```

All six crashed validators abort this way. The worker thread id differs on each crash and between restarts of the same pod (10, 20, 26, 27, 29, 31 and 32 observed).

## Environment

Node version 0.4.45. Run start 2026-08-22 20:00 UTC, duration 48h. Each namespace holds three bonded validators, one bootstrap and one observer. No deploy load.

| Profile | check interval | max LFB age | self-propose cooldown |
|---|---|---|---|
| `ci1s-lfb2s-cd2s` | 1 s | 2 s | 2 s |
| `ci3s-lfb5s-cd6s` | 3 s | 5 s | 6 s |
| `ci5s-lfb10s-cd10s` | 5 s | 10 s | 10 s |

## Steps to reproduce

1. Run the heartbeat suite with the `ci3s-lfb5s-cd6s` or `ci5s-lfb10s-cd10s` profile.
2. Check `Avg concurrent` in the run report. The affected runs report 3.00, meaning every height carries one block per validator and no height carries a single block.
3. Let the chain reach about 3824 heights. At the observed intervals that is roughly 8 hours for `ci3s` and 12 hours for `ci5s`.
4. Every validator aborts within moments of the others, then again on the first propose after each restart.

Step 2 is the part that does not reproduce on demand. The same profiles have run between 1.17 and 1.96 blocks per height, and at those values they pass the crash depth without aborting.

## Observed behavior

| Profile | Result | Last height | Time to failure | Restarts |
|---|---|---|---|---|
| `ci1s-lfb2s-cd2s` | ran the full 48h | 104632 finalized | — | 0 |
| `ci3s-lfb5s-cd6s` | all validators dead | 3825 | 7h55m | 363–366 |
| `ci5s-lfb10s-cd10s` | all validators dead | 3823 | 12h08m | 313–315 |

[...]

PR #385, opened 2026-09-02
fix(crypto): split_byte accepts the full unsigned byte range (#34)

### LINKED ISSUE #34: rholang-cli panics on Par with 129 terms due to i8 overflow in random split (likely due to Scala -> Rust port)
A Rholang program containing 129 parallel send terms panics during reduction.

  Expected:
  The program should reduce/evaluate normally, or return a structured interpreter error.

  Actual:
  thread 'main' panicked at rholang/src/rust/interpreter/reduce.rs:280:47
  called `Result::unwrap()` on an `Err` value: TryFromIntError(())

  Diagnosis:
 `fn split` chooses `split_byte` whenever `terms.len() <= 256`, but `split_byte`
  takes `i8`. The byte/short cutoff was sized for `u8` range but the parameter
  type is signed.
  For 129 terms, child id 128 is passed to id.try_into::<i8>(), which fails.

  Boundary: terms.len() <= 128 works; terms.len() >= 129 panics.
  The first id that overflows i8 is 128 (i8::MAX = 127).

  Fix options:
    (a) terms.len() > 128 (smallest patch, single-character)
    (b) Change split_byte to take u8, use id.try_into::<u8>() at the call site,
        and keep the > 256 cutoff (uses the full byte range)
    (c) Propagate the conversion error instead of unwrap()

  Same Scala→Rust port-hazard pattern as #20 (max_frame_size overflow at
  the HTTP/2 boundary).

  Real-world reachability: any Rholang program whose AST flattens (under par-ACU)
  to >=129 top-level par children. Validator-scale consensus, batch sends,
  replication unfolding, list/set encodings, any higher-level language compiling
  to flat-par Rholang.

  Reproducer generator:

```
  for n in 128 129; do
    out="parallel_sends_${n}.rho"
    : > "$out"
    for i in $(seq 0 $((n - 1))); do
      [ "$i" -gt 0 ] && printf ' |\n' >> "$out"
      printf '@"c%s"!(Nil)' "$i" >> "$out"
    done
    printf '\n' >> "$out"
  done
```

  rholang-cli --quiet parallel_sends_128.rho
  rholang-cli --quiet parallel_sends_129.rho

9Neechan:
### What changed (code)

`crypto/src/rust/hash/blake2b512_random.rs` — `Blake2b512Random::split_byte` now takes `u8` instead of `i8` and reinterprets to `i8` internally (`self.add_byte(index as i8)`).

The Rholang reducer's random-split helper (`rholang/src/rust/interpreter/reduce.rs:236`) routes a `Par` with `2..=256` top-level terms to `split_byte`, passing the child index `0..=255`. With the `i8` parameter, `id.try_into::<i8>()` panicked with `TryFromIntError` for every index `>= 128`. The `> 256` byte/short cutoff was already sized for the full unsigned byte range; only the parameter type was wrong. The call site needs no edit — `id.try_into().unwrap()` now infers `u8`.

Every existing `split_byte` call site passes a non-negative small index, and `index as i8` is the identity on `0..=127`, so there is no behavior, hash, or consensus change for any currently reachable input. The change only makes the
previously-panicking indices `128..=255` usable, with the same wraparound bit pattern the Scala original produces via `id.toByte`.

### What problem it fixes

A Rholang program whose AST flattens (under par-ACU) to `>= 129` top-level `Par` children panicked during reduction:

```
thread 'main' panicked at rholang/src/rust/interpreter/reduce.rs:246:
called `Result::unwrap()` on an `Err` value: TryFromIntError(())
```

Reachable from the CLI and, per the issue thread, from the normal `propose`/consensus path (a deploy body with ~128 par-composed branches plus a contract definition is 129 top-level `Par` terms). In consensus the panic was
caught at deploy level, so the deploy silently failed and burned a block slot with no RPC-level cause.

### How verified

- **Regression tests**, both red on base commit `432c3c64`, green with the fix:
  - `crypto` — `blake2b512_random_split_byte_covers_full_unsigned_range`: `split_byte` over `0..=255` produces 256 distinct streams.
  - `rholang` — `eval_of_par_with_129_sends_should_not_overflow_random_split`: a `Par` of 129 sends reduces to `Ok` and the tuple space holds 129 entries (panics with the exact `TryFromIntError` on base).
- **Workspace suite** — `crypto`, `rholang`, `rspace++`, `models`, `shared`, `graphz`, `block-storage`, `comm`, `node` green. Five heavyweight `casper::genesis::contracts::*` tests are load-flaky in the full parallel 
  run; all pass in isolation and are untouchable by an `i8`->`u8` parameter type change.
- **Testbed** — `rholang-par`, `mid` profile,
  `9neechan/f1r3fly-rust-node:issue-34-cdb72085`:
  `FORKS=128` (129 `Par` terms — the issue's exact boundary) panics + fails the deploy on base `dev`; on the fix image the proposer evaluates the deploy to completion (propose 1751 ms). `FORKS=256` (257 terms) exercises the untouched `split_short` path and hits a separate pre-existing stack overflow (cross-ref #344), out of scope here.


jltatbeach (approved):

jeffrey-l-turner (approved):
## Multi-Agent Code Review

**Verdict:** :white_check_mark: Approved (75.0000% agreement)

**Reviewed by:**

| openai | `gpt-5.6-sol` | Feedback Provided | 0.97 |

---

### Summary

The production change correctly expands `split_byte` to the full unsigned-byte range while preserving the intended two's-complement byte representation, and the Rholang regression test exercises the previously panicking boundary. No security, performance, or production logic problems are apparent. The only issue is a tautological compatibility assertion in the crypto test; fixing it would make the intended signed-to-unsigned mapping explicitly tested, but it does not block the functional fix.

Correct, minimal fix: `split_byte` accepting `u8` matches the reducer’s 0..=255 child indices and the existing >256 short-split cutoff, while `as i8` preserves the same byte pattern Scala gets via `toByte`. No consensus change for 0..=127; 128..=255 become usable instead of panicking. Regression coverage at the crypto boundary (256 distinct streams) and the rholang 129-send case is appropriate. Safe to merge; only optional comment/golden-vector polish remains.

This is a clean, minimal, and well-verified fix for a genuine consensus-path panic: changing `split_byte` to accept `u8` and reinterpreting internally as `i8` preserves the exact bit pattern (and thus hash/consensus behavior) for all previously reachable inputs while making indices 128–255 usable. The regression tests at both the crypto and rholang layers directly target the reported failure, and the PR description demonstrates thorough verification including a testbed reproduction at the exact boundary. The only suggestions are documentation-oriented: a short comment on the deliberate `as i8` cast would guard against future well-intentioned refactors breaking consensus compatibility.

The code change effectively addresses the issue by allowing the full unsigned byte range for the `split_byte` function, which previously caused panics for indices greater than 127. The regression tests confirm the fix, and the change does not affect existing behavior for indices within the 0..=127 range. A minor style issue was noted regarding an unnecessary type cast, but overall the change is safe to merge.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Signed-byte compatibility assertion is tautological</b> (Reported by: openai, openrouter)</summary>

**File:** `crypto/src/rust/hash/blake2b512_random.rs` (line 463)

The assertion compares `base.split_byte(128)` with `base.split_byte((-128i8) as u8)`. Since `(-128i8) as u8` evaluates to `128`, both sides call the same method with the same argument. It therefore does not verify that unsigned index 128 is internally encoded with the same byte pattern as signed -128.

---

The explicit assertion `base.split_byte(128).next() == base.split_byte((-128i8) as u8).next()` is partially redundant with the 256-distinct-streams check, though it does serve as documentation of the wraparound equivalence. Consider adding a comment clarifying that it documents parity with the old i8-based semantics for the previously-panicking boundary.

**Suggestion:**
```
Add a one-line comment above the assertion explaining it documents equivalence with the historical i8 path at the boundary value.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Unnecessary type cast</b> (Reported by: bedrock)</summary>

**File:** `crypto/src/rust/hash/blake2b512_random.rs` (line 94)

The cast `index as i8` is unnecessary since `add_byte` already accepts `u8`. This could be simplified to `split.add_byte(index);`.

**Suggestion:**
```
Remove the cast to `i8` and directly use `index` in `add_byte`.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Document signed-byte cast intent</b> (Reported by: xai)</summary>

**File:** `crypto/src/rust/hash/blake2b512_random.rs` (line 96)

The `index as i8` conversion is load-bearing for Scala parity (two's-complement wrap of 128..=255 → -128..=-1) and consensus-safe hashing. A one-line comment at the cast would lock in why the parameter is `u8` while `add_byte` still takes `i8`, and reduce the chance of a future “cleanup” that changes bit patterns.

**Suggestion:**
```
Add a brief comment above `split.add_byte(index as i8)` noting Scala `toByte` / two's-complement compatibility for indices 128..=255.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Internal `as i8` cast could be better documented</b> (Reported by: openrouter)</summary>

**File:** `crypto/src/rust/hash/blake2b512_random.rs` (line 97)

The `index as i8` reinterpretation is correct (matches the Scala `id.toByte` wraparound semantics) but is not obvious to future readers — it looks like a possible sign bug at first glance. A brief comment explaining that the cast deliberately reinterprets the bit pattern to preserve consensus-compatible hash output would prevent someone from 'fixing' it later and breaking consensus.

**Suggestion:**
```
Add a comment such as: `// Reinterpret as i8 to match Scala's id.toByte semantics; must not change or hash/consensus output will diverge.`
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Wraparound equality assert is tautological</b> (Reported by: xai)</summary>

**File:** `crypto/src/rust/hash/blake2b512_random.rs` (line 464)

`(-128i8) as u8` is 128, so `split_byte(128)` is compared to itself. It documents the intended correspondence but does not exercise a distinct code path or a golden digest.

**Suggestion:**
```
Either drop the assert, or replace it with a fixed expected `next()` digest for index 128 (and optionally 255) if known from the Scala reference, so the wraparound bit pattern is pinned by a vector.
```

</details>
---

<details>

The code change effectively addresses the issue by allowing the full unsigned byte range for the `split_byte` function, which previously caused panics for indices greater than 127. The regression tests confirm the fix, and the change does not affect existing behavior for indices within the 0..=127 range. A minor style issue was noted regarding an unnecessary type cast, but overall the change is safe to merge.

#### openai (gpt-5.6-sol)

The production change correctly expands `split_byte` to the full unsigned-byte range while preserving the intended two's-complement byte representation, and the Rholang regression test exercises the previously panicking boundary. No security, performance, or production logic problems are apparent. The only issue is a tautological compatibility assertion in the crypto test; fixing it would make the intended signed-to-unsigned mapping explicitly tested, but it does not block the functional fix.

This is a clean, minimal, and well-verified fix for a genuine consensus-path panic: changing `split_byte` to accept `u8` and reinterpreting internally as `i8` preserves the exact bit pattern (and thus hash/consensus behavior) for all previously reachable inputs while making indices 128–255 usable. The regression tests at both the crypto and rholang layers directly target the reported failure, and the PR description demonstrates thorough verification including a testbed reproduction at the exact boundary. The only suggestions are documentation-oriented: a short comment on the deliberate `as i8` cast would guard against future well-intentioned refactors breaking consensus compatibility.

#### xai (grok-4.5)

Correct, minimal fix: `split_byte` accepting `u8` matches the reducer’s 0..=255 child indices and the existing >256 short-split cutoff, while `as i8` preserves the same byte pattern Scala gets via `toByte`. No consensus change for 0..=127; 128..=255 become usable instead of panicking. Regression coverage at the crypto boundary (256 distinct streams) and the rholang 129-send case is appropriate. Safe to merge; only optional comment/golden-vector polish remains.
</details>

---

crypto/src/rust/hash/blake2b512_random.rs
```diff
@@ -445,6 +445,27 @@ mod tests {
...
+            let mut split = base.split_byte(index);
+            assert!(
+                seen.insert(split.next()),
+                "duplicate stream for index {index}"
+            );
+        }
+        assert_eq!(seen.len(), 256);
+
+        assert_eq!(
+            base.split_byte(128).next(),
```
jeffrey-l-turner: 🟠 Major: Signed-byte compatibility assertion is tautological (Reported by: openai, openrouter)

jeffrey-l-turner (changes requested):

jltatbeach (changes requested):
See other review

crypto/src/rust/hash/blake2b512_random.rs
9Neechan: Fixed in b9a0802e. Replaced the tautological line with a golden vector for `split_byte(200)` — an index that previously panicked — which pins the two's-complement byte encoding for 128..=255 and catches any future refactor that changes the `as i8` reinterpretation. The 256-distinct-streams loop stays as the range-coverage check.

Also addressed the related minor note on the cast: `add_byte` takes `i8`, so the `as i8` in `split_byte` is required for compilation, not removable — added a comment saying so and why it must stay.

9Neechan:
Addressed the review findings in b9a0802e:

- **Major — tautological compatibility assertion** (`crypto/src/rust/hash/blake2b512_random.rs`): replaced `split_byte(128) == split_byte((-128i8) as u8)` (both the same call) with a golden vector for `split_byte(200)`, an index that previously panicked. It pins the two's-complement byte encoding for indices 128..=255 and fails if a future refactor changes the `as i8` reinterpretation. The 256-distinct-streams loop remains as the range-coverage check.
- **Minor — "unnecessary" `as i8` cast** in `split_byte`: `add_byte` takes `i8`, so the cast is required for the code to compile and is a deliberate two's-complement reinterpretation, not a lossy narrowing. Kept it and added a comment explaining why it must stay, which also covers the other reviewers' suggestion to document the intent for future refactors.

No production logic changed — test assertion plus a comment.

jltatbeach (approved):

jltatbeach (approved):

jeffrey-l-turner (approved):

PR #386, opened 2026-09-03
fix(node): gate container health on Casper readiness via /api/ready

### LINKED ISSUE #12: node ready check doesnt align with casper readiness
> **Migrated from F1R3FLY-io/f1r3node#298** — originally filed 2025-12-17 by @IhorR-DevBrother.
> This issue was moved as part of the f1r3node → f1r3node-rust deprecation (EPOCH-005).
> Original: https://github.com/F1R3FLY-io/f1r3node/issues/298

step to reproduce:
run observer node in docker with `--wait` flag and follow with exploratory deploy

it will fail with
"Error: Could not execute exploratory deploy, casper instance was not available yet."

9Neechan:
## What changed (code)

- **`node/src/rust/web/web_api_routes.rs`** — new `GET /api/ready` endpoint. Returns HTTP 200 `{"ready":true}` once the Casper engine has entered the Running state and the node can serve deploys / exploratory deploys, HTTP 503 `{"ready":false}` while Casper is still initializing (genesis ceremony, or last-finalized-state sync on an observer).
- **`node/src/rust/api/web_api.rs`** — `WebApi` trait gains `fn is_ready(&self) -> bool`; `WebApiImpl` returns the existing `is_ready` `AtomicBool`.
- **`node/src/rust/web/web_api_docs.rs`** — register `ready_handler` in the public and admin OpenAPI path lists.
- **`casper/src/rust/engine/engine.rs`** — `transition_to_running` now publishes the `EnteredRunningState` event **after** `engine_cell.set(running)`, so the readiness signal can never be observed ahead of `with_casper()` resolving.
- **`node/Dockerfile`** — `HEALTHCHECK` now runs `curl -fsS 127.0.0.1:40403/api/ready`; `--start-period` raised to `120s`.
- **`docker/standalone.yml`** — healthcheck hits `/api/ready`; `start_period` raised to `120s`. `docker/observer.yml` and `docker/shard.yml` have no override and inherit the Dockerfile check.
- **`docs/node/api-reference.md`, `docs/node/README.md`** — document `/api/ready`.

## What problem it fixes

Running an observer node in Docker and following the container's health signal (`docker compose up --wait`, `depends_on: condition: service_healthy`) unblocked ~15 s before Casper could serve deploys. An exploratory deploy issued in that window failed with `Error: Could not execute exploratory deploy, casper instance was not available yet.`

The node already exposed `isReady` on `casper.v1.DeployService.status` and `GET /api/status` (flipped `true` on `EnteredRunningState`), but no health check consulted it: the Dockerfile `HEALTHCHECK` only checked that an endpoint returned any JSON object, and `/api/status` answers HTTP 200 throughout startup. `/api/ready` closes that gap with a probe that fails until the node is deploy-ready and needs no `jq`.

## How verified

- **Regression test** `ready_route_reports_503_until_casper_running` in `node/src/rust/web/web_api_routes.rs`: `GET /api/ready` returns HTTP 503 `{"ready":false}` when the readiness flag is false and HTTP 200 `{"ready":true}` when true. Fails on the base commit (`231067178a…`) — no route, and the trait has no `is_ready` — and passes with the fix.
- `cargo test -p node` — 246 + 17 + 1 pass.
- `cargo test -p casper --test mod engine::` — 86 pass, including `make_transition_to_running_state_after_block_approved` and `make_transition_to_running_once_approved_block_received`.
- `cargo clippy -p node -p casper --all-targets` and `cargo fmt --check` clean.
- Manual: `docker build -f node/Dockerfile …` then `docker compose -f docker/observer.yml up --wait` against a running shard — the command blocks until Casper is Running, and an immediately-following exploratory deploy succeeds.
- No testbed Kubernetes experiment was needed: the gap is fully determined by the health-check definitions plus the status-endpoint code, and the fix is covered by the deterministic router test.

## Caveat

A node still running the genesis ceremony (minutes to hours) now reports `unhealthy` for that whole window, so `docker compose up --wait` against a fresh validator mid-genesis will time out. This is the intended semantics — the node cannot serve deploys during genesis. Operators raise `--wait-timeout` / probe `failureThreshold` when they need to span a long genesis; the `120s` start period covers ordinary observer/validator startup.

The helm chart's `livenessProbe`/`readinessProbe` blocks are currently commented out in `templates/statefulsets.yaml`, so the chart is out of scope here — wiring the disabled probes to `/api/ready` is follow-up.

### DIFF
--- a/casper/src/rust/engine/engine.rs
+++ b/casper/src/rust/engine/engine.rs
@@ -237,17 +237,8 @@ pub async fn transition_to_running<U: TransportLayer + Send + Sync + Clone + 'st
     )
     .increment(1);
 
-    // Publish EnteredRunningState event
     let block_hash_string =
         PrettyPrinter::build_string_no_limit(&approved_block.candidate.block.block_hash);
-    event_log
-        .publish(F1r3flyEvent::entered_running_state(block_hash_string))
-        .map_err(|e| {
-            CasperError::Other(format!(
-                "Failed to publish EnteredRunningState event: {}",
-                e
-            ))
-        })?;
 
     let running = Running::new(
         block_processing_queue_tx,
@@ -264,6 +255,15 @@ pub async fn transition_to_running<U: TransportLayer + Send + Sync + Clone + 'st
     );
 
     engine_cell.set(Arc::new(running)).await;
+
+    event_log
+        .publish(F1r3flyEvent::entered_running_state(block_hash_string))
+        .map_err(|e| {
+            CasperError::Other(format!(
+                "Failed to publish EnteredRunningState event: {}",
+                e
+            ))
+        })?;
 
     Ok(())
 }
--- a/casper/src/rust/engine/engine.rs
+++ b/casper/src/rust/engine/engine.rs
@@ -256,14 +256,12 @@ pub async fn transition_to_running<U: TransportLayer + Send + Sync + Clone + 'st
 
     engine_cell.set(Arc::new(running)).await;
 
-    event_log
-        .publish(F1r3flyEvent::entered_running_state(block_hash_string))
-        .map_err(|e| {
-            CasperError::Other(format!(
-                "Failed to publish EnteredRunningState event: {}",
-                e
-            ))
-        })?;
+    if let Err(e) = event_log.publish(F1r3flyEvent::entered_running_state(block_hash_string)) {
+        tracing::error!(
+            "Failed to publish EnteredRunningState event after committing Running state: {}",
+            e
+        );
+    }
 
     Ok(())
 }
--- a/docker/standalone.yml
+++ b/docker/standalone.yml
@@ -41,11 +41,11 @@ services:
     environment:
       - OPENAI_ENABLED=${OPENAI_ENABLED:-false}
     healthcheck:
-      test: ["CMD", "curl", "-f", "http://localhost:40403/api/status"]
+      test: ["CMD", "curl", "-fsS", "http://localhost:40403/api/ready"]
       interval: 30s
       timeout: 10s
       retries: 5
-      start_period: 60s
+      start_period: 120s
 
 volumes:
   standalone-data:
--- a/docker/standalone.yml
+++ b/docker/standalone.yml
@@ -41,7 +41,7 @@ services:
     environment:
       - OPENAI_ENABLED=${OPENAI_ENABLED:-false}
     healthcheck:
-      test: ["CMD", "curl", "-fsS", "http://localhost:40403/api/ready"]
+      test: ["CMD", "curl", "-fsS", "http://127.0.0.1:40403/api/ready"]
       interval: 30s
       timeout: 10s
       retries: 5
--- a/docs/node/README.md
+++ b/docs/node/README.md
@@ -146,6 +146,10 @@ CLI flags are applied to the parsed `NodeConf` by `config_mapper.rs`:
 - `currentEpoch` — `lastFinalizedBlockNumber / epochLength`
 - `epochLength` — blocks per epoch, from genesis configuration
 
+### `GET /api/ready`
+
+Readiness probe for orchestration. Returns HTTP 200 `{"ready": true}` once Casper is Running and the node can serve deploys, HTTP 503 `{"ready": false}` while Casper is still initializing. The container `HEALTHCHECK` and `docker/standalone.yml` use this endpoint, so `docker compose up --wait` and `depends_on: condition: service_healthy` block until the node is deploy-ready.
+
 ## View Parameters
 
 All block and deploy endpoints support a `?view=full|summary` query parameter:
--- a/docs/node/README.md
+++ b/docs/node/README.md
@@ -148,7 +148,13 @@ CLI flags are applied to the parsed `NodeConf` by `config_mapper.rs`:
 
 ### `GET /api/ready`
 
-Readiness probe for orchestration. Returns HTTP 200 `{"ready": true}` once Casper is Running and the node can serve deploys, HTTP 503 `{"ready": false}` while Casper is still initializing. The container `HEALTHCHECK` and `docker/standalone.yml` use this endpoint, so `docker compose up --wait` and `depends_on: condition: service_healthy` block until the node is deploy-ready.
+Readiness probe for orchestration.
+
+- Returns HTTP 200 `{"ready": true}` once Casper is Running and the node can serve deploys.
+- Returns HTTP 503 `{"ready": false}` while Casper is still initializing.
+- The container `HEALTHCHECK` and `docker/standalone.yml` use this endpoint, so `docker compose up --wait` and `depends_on: condition: service_healthy` block until the node is deploy-ready.
+- A node running the genesis ceremony reports `unhealthy` for the whole window (minutes to hours); raise `--wait-timeout` / probe `failureThreshold` accordingly.
+- The container `HEALTHCHECK` probes only the HTTP API. It no longer probes the gRPC API (port 40401) directly, so a failure isolated to the gRPC server does not fail the health check. `/api/ready` reflects Casper engine state shared by both servers, so this is expected to cover the common failure modes.
 
 ## View Parameters
 
--- a/docs/node/api-reference.md
+++ b/docs/node/api-reference.md
@@ -121,6 +121,14 @@ curl http://localhost:40403/api/status
 
 ---
 
+### `GET /api/ready`
+
+Readiness probe. Returns HTTP 200 with `{"ready": true}` once the Casper engine has entered the Running state and the node can serve deploys and exploratory deploys. Returns HTTP 503 with `{"ready": false}` while Casper is still initializing (genesis ceremony, or last-finalized-state sync on an observer).
+
+Unlike `GET /api/status`, which answers 200 throughout startup, this endpoint fails until the node is deploy-ready, so it works directly as a container health check (`docker compose up --wait`, `depends_on: condition: service_healthy`) and as a Kubernetes `httpGet` readiness probe without needing `jq`.
+
+---
+
 ### Block Endpoints
 
 All block endpoints support `?view=full|summary`. Single-item endpoints default to **full** (includes deploys). List endpoints default to **summary** (block headers only, deploys omitted).
--- a/docs/node/api-reference.md
+++ b/docs/node/api-reference.md
@@ -127,6 +127,8 @@ Readiness probe. Returns HTTP 200 with `{"ready": true}` once the Casper engine 
 
 Unlike `GET /api/status`, which answers 200 throughout startup, this endpoint fails until the node is deploy-ready, so it works directly as a container health check (`docker compose up --wait`, `depends_on: condition: service_healthy`) and as a Kubernetes `httpGet` readiness probe without needing `jq`.
 
+A node running the genesis ceremony (minutes to hours) reports `unhealthy` for the whole window, because it cannot serve deploys yet. Operators spanning a long genesis raise `--wait-timeout` (Docker Compose) or probe `failureThreshold` / `start_period` (Kubernetes, Docker `HEALTHCHECK`) accordingly.
+
 ---
 
 ### Block Endpoints
--- a/node/Dockerfile
+++ b/node/Dockerfile
@@ -173,11 +173,12 @@ RUN if [ ! -f /opt/docker/bin/node ]; then \
 # Expose ports (matching build.sbt: dockerExposedPorts)
 EXPOSE 40400 40401 40402 40403 40404
 
-# Healthcheck (matching build.sbt healthcheck, adapted for Rust)
-# Note: The original uses grpcurl and curl with jq
-HEALTHCHECK --interval=30s --timeout=10s --start-period=40s --retries=3 \
-    CMD grpcurl -plaintext 127.0.0.1:40401 casper.v1.DeployService.status | jq -e && \
-        curl -s 127.0.0.1:40403/api/status | jq -e || exit 1
+# Healthcheck: report healthy only once Casper has entered the Running state
+# and can serve deploys. /api/ready returns 503 while Casper is still
+# initializing (genesis ceremony, LFS sync), so `docker compose up --wait`
+# and `depends_on: condition: service_healthy` line up with deploy readiness.
+HEALTHCHECK --interval=30s --timeout=10s --start-period=120s --retries=3 \
+    CMD curl -fsS 127.0.0.1:40403/api/ready || exit 1
 
 # Image version label. Populate via `--build-arg VERSION=...` from
 # docker-commands.sh / CI so the label tracks the actual build instead of
--- a/node/Dockerfile
+++ b/node/Dockerfile
@@ -101,7 +101,7 @@ FROM debian:bookworm-slim AS runtime-base
 ARG PROFILING_BUILD=0
 
 # Install runtime dependencies
-# Note: grpcurl and jq for healthcheck (matching build.sbt healthcheck)
+# Note: curl backs the container HEALTHCHECK (GET /api/ready).
 RUN apt-get update && apt-get install -y \
     ca-certificates \
     libssl3 \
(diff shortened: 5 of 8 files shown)


jeffrey-l-turner (commented):
## Multi-Agent Code Review

**Verdict:** :mag: Needs Review (20.0000% agreement)

**Reviewed by:**

---

### Summary

This is a well-scoped, well-tested change that closes a real readiness gap: the new /api/ready endpoint returns 503 until Casper enters the Running state, and container health checks are wired to it. The endpoint implementation, atomic-flag plumbing, and deterministic router test are solid, and reordering the EnteredRunningState publish after engine_cell.set correctly eliminates a race where readiness could be observed before the engine was actually running. The main concern is the new failure mode introduced by that reordering: if the event publish fails after the engine is committed to Running, the function returns an error for a transition that succeeded and the readiness flag is never set, leaving a functioning node permanently reporting unhealthy — this should be logged/handled rather than propagated. Secondary items: the admin OpenAPI docs advertise /api/ready without evidence the admin router mounts it, the health check no longer covers the gRPC port, and the long unhealthy window during genesis can trigger restarts under some orchestrators. None of these block merge, but the post-commit error path deserves a fix or an explicit decision.

The readiness endpoint, health-check changes, documentation, and route test are otherwise coherent and address the reported startup race. However, moving a fallible event publication after `engine_cell.set()` introduces a partial-commit failure mode that can leave the engine running while readiness never becomes true, so the transition's error handling should be resolved before merge.

Solid, well-scoped fix: /api/ready gives orchestration a real deploy-readiness signal, Docker healthchecks correctly treat 503 as failure via curl -fsS, the Acquire load on the existing AtomicBool is appropriate, and the router regression test covers both states. Reordering EnteredRunningState after engine_cell.set addresses the reported race in the happy path. Main follow-ups are hardening the publish-failure path so readiness cannot stick false after a committed Running transition, tightening OpenAPI/response docs, and double-checking the runtime image still installs curl after removing jq/grpcurl. No security issues; safe to merge with minor polish.

The change cleanly gates container health on Casper readiness by adding a dedicated `/api/ready` endpoint that returns 503 until the node is Running. The implementation is consistent across the WebApi trait, routes, docs, and Dockerfile, and includes a regression test. No security, logic, or performance issues were found.

The code change introduces a new `/api/ready` endpoint to gate container health on Casper readiness, which is a valuable addition. The implementation is solid, with appropriate tests and documentation updates. Minor style and documentation improvements could enhance the overall quality.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Error returned after state transition is already committed</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `casper/src/rust/engine/engine.rs` (line 259)

By moving the `EnteredRunningState` publish after `engine_cell.set(Arc::new(running)).await`, a publish failure now returns `Err` from `transition_to_running` even though the engine has irrevocably transitioned to Running. This leaves the node in an inconsistent state: Casper can serve deploys, but the readiness flag (which flips on the `EnteredRunningState` event) is never set, so `/api/ready` reports 503 forever and the container is permanently 'unhealthy'. Additionally, any caller treating the `Err` as a failed transition would be wrong — the transition succeeded.

---

EnteredRunningState is now published after engine_cell.set(running). That correctly prevents readiness from racing ahead of with_casper(), but if event_log.publish fails the function returns Err while the engine is already Running. If is_ready is flipped only by that event, /api/ready can stay false forever even though the node can serve deploys—the opposite stuck state from the bug being fixed. Failure to publish is likely rare, but the error path is now asymmetric.

---

The engine is now installed in `engine_cell` before publishing `EnteredRunningState`, but publication can still return an error. In that case the function returns `Err` after the irreversible state transition has occurred. Because the node readiness flag is driven by this event, Casper may be available through `with_casper()` while `/api/ready` remains false indefinitely. Depending on how the caller handles the error, it may also retry a transition that has already installed a `Running` instance, potentially duplicating initialization or background resources.

**Suggestion:**
```
Avoid returning a normal transition failure after committing `engine_cell`. Make readiness derive from a guaranteed state transition rather than fallible event delivery, or explicitly update readiness after `engine_cell.set()` and treat event publication failure as non-fatal logging/retry. If event delivery is mandatory, add rollback/idempotency or otherwise guarantee publication before reporting the transition as committed.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Code formatting</b> (Reported by: bedrock)</summary>

**File:** `casper/src/rust/engine/engine.rs` (line 237)

The code block for publishing the `EnteredRunningState` event is indented inconsistently.

**Suggestion:**
```
Ensure consistent indentation for better readability.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Documentation formatting</b> (Reported by: bedrock)</summary>

**File:** `docs/node/README.md` (line 146)

The documentation for `GET /api/ready` could be formatted for better readability.

**Suggestion:**
```
Use Markdown formatting to improve the structure and clarity of the documentation.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Health check no longer covers the gRPC API (port 40401)</b> (Reported by: anthropic)</summary>

**File:** `node/Dockerfile` (line 157)

As noted in the PR description, a node in the genesis ceremony (minutes to hours) will report unhealthy for the entire window. Beyond `--wait` timeouts, some orchestrators and sidecars (Docker Swarm, autoheal containers, Nomad, etc.) restart containers based on the Docker health status. A restart mid-genesis could prolong or disrupt the ceremony. The 120s start period does not cover this case.

---

The previous HEALTHCHECK probed both `casper.v1.DeployService.status` via grpcurl on 40401 and the HTTP API on 40403. The new check only hits the HTTP `/api/ready` endpoint. A failure isolated to the gRPC server (e.g., it fails to bind or its task panics while the HTTP server keeps running) would no longer be detected. Since `/api/ready` reflects Casper engine state and both servers share it, this is probably acceptable, but it is a coverage reduction worth acknowledging.

**Suggestion:**
```
Consider documenting this restart risk explicitly in docs/node/README.md (not only the compose `--wait` timeout), or splitting liveness (process up, HTTP responding — `/api/status`) from readiness (`/api/ready`) so operators can choose which one drives container health for validators bootstrapping genesis.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: ready_handler registered in Admin OpenAPI docs, but route only added to the public router</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/web/web_api_docs.rs` (line 53)

The diff registers `web_api_routes::ready_handler` in both the `PublicApi` and admin OpenAPI path lists, but the actual route (`.route("/ready", get(ready_handler))`) is only added to `WebApiRoutes::create_router()` (the public router). If the admin router does not also mount `/ready`, the admin OpenAPI spec will advertise an endpoint that returns 404 on the admin port.

**Suggestion:**
```
Either add the `/ready` route to the admin router as well, or remove `ready_handler` from the admin OpenAPI path list so the docs match reality.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Compose healthcheck uses localhost while Dockerfile uses 127.0.0.1</b> (Reported by: anthropic)</summary>

**File:** `docker/standalone.yml` (line 44)

The compose healthcheck curls `http://localhost:40403/api/ready` while the Dockerfile HEALTHCHECK uses `127.0.0.1:40403`. On images where `localhost` resolves to `::1` first and the API only binds IPv4, the compose check could behave differently from the Dockerfile check.

**Suggestion:**
```
Use `127.0.0.1` consistently in both the Dockerfile and compose healthchecks to avoid IPv6 resolution surprises.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Document genesis unhealthy window for operators</b> (Reported by: xai)</summary>

**File:** `docs/node/api-reference.md` (line 121)

The PR caveat correctly notes that a long genesis ceremony will keep the container unhealthy and can trip docker compose --wait / probe failureThreshold. The API docs explain 503 during init but do not mention the operational implication for fresh validators.

**Suggestion:**
```
Add one sentence that operators spanning genesis should raise --wait-timeout / failureThreshold / start_period, matching the README/PR caveat.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Comment drift after removing grpcurl/jq</b> (Reported by: openrouter)</summary>

**File:** `node/Dockerfile` (line 101)

The comment above the `apt-get` block was updated, but the block no longer installs `grpcurl` and `jq`, which is correct. However, the comment still says `Note: curl backs the container HEALTHCHECK` while the `HEALTHCHECK` instruction is 50+ lines later in the file. This makes the comment slightly disconnected from the code it annotates.

**Suggestion:**
```
Move the comment closer to the `HEALTHCHECK` instruction or rephrase it to make clear that `curl` is the only tool needed by the health check.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Confirm curl remains a runtime dependency after jq/grpcurl removal</b> (Reported by: xai)</summary>

**File:** `node/Dockerfile` (line 150)

HEALTHCHECK now depends only on curl -fsS, and the dedicated grpcurl/jq install block was removed. That is a good simplification, but the visible apt-get fragment in the diff does not show curl explicitly. If curl was only pulled in transitively or assumed from the old download script path, the image could build while HEALTHCHECK always fails.

**Suggestion:**
```
Ensure curl is an explicit apt package in the runtime stage (and that no other image entrypoints/scripts still require grpcurl or jq).
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Single-line function body diverges from surrounding formatting</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/api/web_api.rs` (line 459)

`fn is_ready(&self) -> bool { self.is_ready.load(Ordering::Acquire) }` is written on one line while the surrounding trait impl methods use multi-line bodies. rustfmt with default settings typically expands this; if `cargo fmt --check` passes, the repo may allow it, but it is inconsistent with adjacent code.

**Suggestion:**
```
Format the body across multiple lines to match neighboring methods, or verify the project's rustfmt config intentionally permits single-line fns.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: OpenAPI path omits response body schema</b> (Reported by: xai)</summary>

**File:** `node/src/rust/web/web_api_routes.rs` (line 77)

ready_handler documents 200/503 status meanings but does not declare the JSON body shape {"ready": bool}. Status and other handlers in this codebase often expose response types via utoipa; clients and generated docs will not see the payload contract.

**Suggestion:**
```
Add a small ReadyResponse { ready: bool } (or inline body = ...) to the utoipa::path responses so public/admin OpenAPI matches the runtime JSON.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: OpenAPI responses lack a body schema for the /api/ready payload</b> (Reported by: anthropic)</summary>

**File:** `node/src/rust/web/web_api_routes.rs` (line 90)

The `#[utoipa::path]` annotation for `ready_handler` documents status codes but not the `{"ready": bool}` response body. Consumers reading the generated OpenAPI spec won't see the shape of the payload.

**Suggestion:**
```
Define a small `ReadyResponse { ready: bool }` struct deriving `ToSchema`/`Serialize` and reference it in the `responses(...)` for both 200 and 503, instead of an untyped `serde_json::json!` value.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: OpenAPI response body schema is not documented</b> (Reported by: openrouter)</summary>

**File:** `node/src/rust/web/web_api_routes.rs` (line 97)

The new `/api/ready` endpoint documents 200 and 503 responses, but the OpenAPI annotation does not specify the JSON body schema `{ ready: boolean }`. Consumers of the API reference will not see the response shape in generated docs.

**Suggestion:**
```
Add a `body = ReadyResponse` (or inline JSON schema) to each response in the `#[utoipa::path]` annotation so the OpenAPI output includes the response structure.
```

</details>
---

<details>

This is a well-scoped, well-tested change that closes a real readiness gap: the new /api/ready endpoint returns 503 until Casper enters the Running state, and container health checks are wired to it. The endpoint implementation, atomic-flag plumbing, and deterministic router test are solid, and reordering the EnteredRunningState publish after engine_cell.set correctly eliminates a race where readiness could be observed before the engine was actually running. The main concern is the new failure mode introduced by that reordering: if the event publish fails after the engine is committed to Running, the function returns an error for a transition that succeeded and the readiness flag is never set, leaving a functioning node permanently reporting unhealthy — this should be logged/handled rather than propagated. Secondary items: the admin OpenAPI docs advertise /api/ready without evidence the admin router mounts it, the health check no longer covers the gRPC port, and the long unhealthy window during genesis can trigger restarts under some orchestrators. None of these block merge, but the post-commit error path deserves a fix or an explicit decision.

The code change introduces a new `/api/ready` endpoint to gate container health on Casper readiness, which is a valuable addition. The implementation is solid, with appropriate tests and documentation updates. Minor style and documentation improvements could enhance the overall quality.

#### openai (gpt-5.6-sol)

The readiness endpoint, health-check changes, documentation, and route test are otherwise coherent and address the reported startup race. However, moving a fallible event publication after `engine_cell.set()` introduces a partial-commit failure mode that can leave the engine running while readiness never becomes true, so the transition's error handling should be resolved before merge.

The change cleanly gates container health on Casper readiness by adding a dedicated `/api/ready` endpoint that returns 503 until the node is Running. The implementation is consistent across the WebApi trait, routes, docs, and Dockerfile, and includes a regression test. No security, logic, or performance issues were found.

#### xai (grok-4.5)

Solid, well-scoped fix: /api/ready gives orchestration a real deploy-readiness signal, Docker healthchecks correctly treat 503 as failure via curl -fsS, the Acquire load on the existing AtomicBool is appropriate, and the router regression test covers both states. Reordering EnteredRunningState after engine_cell.set addresses the reported race in the happy path. Main follow-ups are hardening the publish-failure path so readiness cannot stick false after a committed Running transition, tightening OpenAPI/response docs, and double-checking the runtime image still installs curl after removing jq/grpcurl. No security issues; safe to merge with minor polish.
</details>

---

9Neechan:
Addressed the multi-agent review findings in 9fe53e89:

- **Critical (`casper/src/rust/engine/engine.rs:259`)**: `transition_to_running` no longer returns `Err` for a failed `EnteredRunningState` publish once `engine_cell.set()` has already committed the Running transition. The state transition is now the sole source of truth; a publish failure is logged, not propagated. (Note: `F1r3flyEvents::publish` cannot currently return an error — it swallows the broadcast "no receivers" case into `Ok(())` — so this was previously dead code, but a dangerous shape: the `initializing.rs` caller would have retried a transition whose `engine_cell` mutation already succeeded.)
- **`casper/src/rust/engine/engine.rs:237` (indentation)**: `cargo fmt --check -p casper` passes cleanly on this file; no formatting issue found. Left as-is.
- **`docs/node/README.md:146` (formatting)**: reformatted the `/api/ready` section into bullet points matching the surrounding doc style.
- **`node/Dockerfile:157` (gRPC health coverage)**: documented the coverage tradeoff (HTTP-only health check) in `docs/node/README.md`.
- **`node/src/rust/web/web_api_docs.rs:53` (admin OpenAPI mismatch)**: removed `ready_handler` from the `AdminApi` path list, since `/ready` is not mounted on the admin router (`AdminWebApiRoutes::create_router()` only has `/propose`) and adding a new admin route was out of scope for this fix.
- **`docker/standalone.yml:44` (localhost vs 127.0.0.1)**: changed the compose healthcheck to `127.0.0.1` to match the Dockerfile.
- **`docs/node/api-reference.md:121` (genesis window)**: added a sentence on raising `--wait-timeout`/`failureThreshold` for a long genesis ceremony.
- **`node/Dockerfile:101` (comment drift)**: reworded the runtime-deps comment to point at the `HEALTHCHECK` "further below" instead of implying adjacency.
- **`node/Dockerfile:150` (curl explicit dependency)**: confirmed `curl` is an explicit `apt-get install` package in the runtime stage (line 110) — already correct, no change needed.
- **`node/src/rust/api/web_api.rs:459` (single-line fn body)**: `rustfmt.toml` sets `fn_single_line = true` project-wide; this is intentional style, confirmed via `cargo fmt --check -p node` passing cleanly. No change needed.
- **OpenAPI response body schema** (`node/src/rust/web/web_api_routes.rs:77/90/97`, reported by three reviewers): added `ReadyResponse { ready: bool }` with `ToSchema`/`Serialize` and referenced it in both the 200 and 503 responses of the `/api/ready` `#[utoipa::path]` annotation.

Verified with `cargo build`, `cargo test -p node` (17/17 web_api_routes tests pass, including the `/api/ready` regression test), `cargo test -p casper --test mod engine::` (86/86 pass, including both `transition_to_running` tests), `cargo clippy -p node -p casper --all-targets` (clean), and `cargo fmt --check` (clean).

jeffrey-l-turner (approved):

PR #397, opened 2026-09-06
chore(supply-chain): harden dependency checks and build inputs

jeffrey-l-turner:
- Verify the pinned scanner and scan all locked dependency graphs.
- Require exact advisory exception versions, owners, and review dates.
- Pin external actions and container images, and restrict cache writes.
- Vendor Swagger assets and verify runtime downloads.
- Compile container builds without network access or shared Cargo caches.
- Remove unused certificate dependencies and track the fuzz lockfile.
- Resolve bitmaps to unaffected 3.1.0 and block affected releases.
- Add policy and persistent-map regression tests and retain scan evidence.


jeffrey-l-turner:
## Multi-Agent Review — PR #397

**Reviewed commit:** `c6685ecf73086ff1df8a0c41ac75cd4b6df3d0de`

**Provider verdict:** `provide_feedback`, with 75% agreement among four completed reviews. Anthropic timed out and did not vote.

The review used the full 214,733-byte diff across all 18 changed files. File-input mode prevented the script's normal 100 KB truncation.

|---|---|---|---|
| Anthropic | `claude-fable-5` | Timeout | 0.00 |
| OpenAI | `gpt-5.6-sol` | `provide_feedback` | 0.88 |
| xAI | `grok-4.5` | `needs_review` | 0.82 |
| OpenRouter | `moonshotai/kimi-k3` | `provide_feedback` | 0.75 |
| Bedrock | `us.amazon.nova-pro-v1:0` | `provide_feedback` | 0.90 |

### Verified findings and control gaps

1. **Malformed diagnostics can stop later scans.** `scripts/ci/check_supply_chain.py:175-186` assumes that advisory IDs and graph dictionaries exist. A missing ID produces `KeyError`. A null graph produces `AttributeError`. Both cases were reproduced. The per-manifest handler at line 364 catches neither exception. Validate nested records and report a failure for the affected manifest. Continue with the remaining manifests. This is an error-path defect, not a demonstrated scanner bypass.
2. **Slashing caches retain the previous write behavior.** Eight `rust-cache` steps lack `save-if`, starting at `.github/workflows/slashing-tests.yml:77`. Apply the selected trusted-branch restriction consistently. This is incomplete hardening of existing behavior, not evidence that a pull-request cache can overwrite a trusted-branch cache.
3. **Workflow pinning tests cover only four named workflows.** `scripts/ci/test_supply_chain.py:297` does not discover new workflows. This matches the selected scope, but it does not enforce repository-wide action pinning. Expand discovery if repository-wide enforcement is required.

### Critical and major provider reports: verification status

The raw aggregator reported one critical issue and five major issues. These labels do not establish their severity or correctness.

| Raw severity | File | Reporters | Verification status |
|---|---|---|---|
| Critical | `node/Dockerfile:79` | xAI, OpenRouter | Unverified build risk. Repeated reports caused automatic escalation. No critical vulnerability was established. |
| Major | `.github/workflows/ci.yml` | Bedrock | Rejected. Installation verifies the pinned archive SHA-256 before reading the executable. |
| Major | `.github/workflows/deny-schedule.yml:19-35` | OpenRouter | Conditional rollout risk. Current `master` lacks the script and policy. Scheduled runs use the default branch's workflow. Deploy both files with the workflow update. |
| Major | `deny.toml` | xAI | Rejected. Root and fuzz lockfiles contain `rustls-webpki` 0.103.13 and 0.103.14 respectively. The soak graph contains no copy. |
| Major | `scripts/ci/check_supply_chain.py:152-172` | OpenRouter | Unproven scanner-output compatibility risk. The pinned scanner passed local checks and CI. Do not discard malformed output to avoid failures. |
| Major | `scripts/ci/check_supply_chain.py:221-233` | xAI | Residual authenticity limit. A pinned hash verifies integrity, not signed provenance. This is not a demonstrated regression. |

### Other dispositions

- The rewritten dependency reasons match the manifests. `rholang` directly declares `dotenv`, and `comm` declares `paste` as a development dependency.
- Fixed dates in policy tests provide deterministic fixtures. A dynamic date is not a necessary correction.
- Hash verification still protects archive integrity if a download redirects to HTTP. Restricting redirects to HTTPS remains a useful transport improvement.
- The installer can reject a retry if an interrupted installation leaves `cargo-deny.download`. Use a unique temporary file with cleanup if retry robustness is required.
- Owner validation currently requires nonempty text. An approved-owner list would strengthen ownership enforcement but adds a new policy requirement.
- Build-script review effort, apt package reproducibility, signed provenance, and replacement of unmaintained dependencies remain separate follow-up work.

### Verification and limits

GitHub reports that the PR is mergeable. The `cargo-deny` CI job passed. Local evidence summaries report zero scanner errors for all three manifests.

Both architecture-specific Docker jobs remain pending at this check. Their results are still required to validate network-disabled compilation. No Docker failure is established by this review.

No repository files were changed. The provider verdict is feedback, not approval. The raw report below preserves provider output, including rejected claims and unverified recommendations.

<details>
<summary>Raw aggregated provider report — unverified output</summary>

## Multi-Agent Code Review

**Reviewed by:**

| anthropic | `claude-fable-5` | Timeout Error | 0 |
| openai | `gpt-5.6-sol` | Feedback Provided | 0.88 |
| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.75 |

---

### Summary

The change substantially improves dependency and build-input controls through immutable action/image references, locked multi-graph scanning, verified scanner installation, vendored Swagger assets, and an offline compilation step. No critical vulnerability is established in the supplied diff. The remaining actionable gaps are inconsistent cache-write restrictions in the slashing workflow and regression tests that do not automatically cover every workflow; owner validation and malformed scanner-output handling could also be made more robust.

This PR substantially hardens the supply chain: pinned GitHub actions/images, verified cargo-deny installation, locked multi-graph scans with evidence uploads, vendored Swagger UI, checksummed grpcurl, offline frozen Docker builds, fuzz lockfile tracking, and an explicit bitmaps 3.1.0 pin that blocks RUSTSEC-2025-0167. The direction is sound and much of the control surface is carefully tested. However, merge should wait on human review of two practical risks: confirm the removed rustls-webpki advisory exceptions are truly eliminated from all locked graphs, and verify the new no-network/no-shared-cache Docker build path on both architectures. Secondary follow-ups include making rust-cache write restrictions consistent in slashing-tests and budgeting for build-script allowlist churn.

This is a thorough, well-tested supply-chain hardening PR: SHA-pinned actions and images, digest-verified cargo-deny installation with safe (no-extraction) tarball handling, exact-version advisory exceptions with expiry, per-workspace locked scans with retained evidence, vendored Swagger assets, network-isolated Docker builds, and a bitmaps downgrade that pins to the unaffected 3.1.0 release while banning affected ones. The test suite for the policy tooling is genuinely strong (28 tests covering expiry, version ranges, malformed reports, symlink archives, and manifest coverage). The main concerns are robustness hypotheses rather than proven bugs: validate_report's fail-on-any-non-JSON-line behavior may turn benign scanner stderr noise into CI failures, the scheduled deny workflow will hard-fail on refs lacking the new script/policy (the old guard was removed), and the Docker changes — flagged by the author as unverified — carry both a potential cargo-fetch host-dependency gap and a significant build-time regression from removing all cache mounts. None of these block merge, but the workflow guard and the report-parsing brittleness are worth addressing, and the rewritten deny.toml reasons should be double-checked for accurate dependency attribution.

The PR introduces significant improvements to dependency checks and build inputs, enhancing security and reliability. However, there are some areas that could be improved, such as the installation method for cargo-deny and the use of hardcoded dates in tests.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Offline Docker compile path is unverified and likely brittle</b> (Reported by: openrouter, xai)</summary>

**File:** `node/Dockerfile` (line 79)

The builder now does cargo fetch --locked, then xx-cargo build --frozen under RUN --network=none, and removes all shared Cargo registry/git/build cache mounts. This is a strong supply-chain control, but Docker image builds were not verified locally. Cross builds frequently need network-mounted caches or pre-populated target-sysroot state; a clean builder after fetch can fail on tool/cross sysroot expectations, and CI will be the first place that failure appears.

---

`cargo fetch --target <triple>` restricts resolution to dependencies needed for that target. Build dependencies and proc-macros are compiled for the host, and in most cases overlap, but host-only optional deps (e.g., a build-dependency enabled only on the host triple when cross-compiling) could be missed by the fetch and then fail under the subsequent `RUN --network=none xx-cargo build --frozen`. Docker builds were explicitly noted as unverified in this PR, so this is an unproven hypothesis.

**Suggestion:**
```
Verify amd64 and arm64 image builds in CI or a machine with Docker before merge. If build-time downloads are still required, stage them explicitly before --network=none rather than reintroducing shared cache mounts.
```

</details>

<details>
<summary><b>:orange_circle: Major: Potential security risk in cargo-deny installation</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/ci.yml` (line 364)

The cargo-deny tool is installed using a direct download and extraction method, which could be risky if the archive is tampered with. Consider using a verified release mechanism or a trusted package manager.

**Suggestion:**
```
Use a verified release mechanism or a trusted package manager to install cargo-deny.
```

</details>

<details>
<summary><b>:orange_circle: Major: Scheduled workflow will hard-fail on any branch lacking the new script or policy file</b> (Reported by: openrouter)</summary>

**File:** `.github/workflows/deny-schedule.yml` (line 30)

The old workflow guarded the soak-charts pass with an existence check because this cron runs the default branch's workflow file against arbitrary refs. The new implementation runs `python3 scripts/ci/check_supply_chain.py ...` and loads `supply-chain/policy.toml` from the checked-out ref. Any matrix branch that does not yet contain scripts/ci/check_supply_chain.py or supply-chain/policy.toml fails at the install step with a file-not-found, producing a noisy red cron rather than a skipped check. Also, the removed header comment documented that schedule events only fire from the default branch — that operational knowledge is now lost.

**Suggestion:**
```
Add an early step that checks `[ -f scripts/ci/check_supply_chain.py ] && [ -f supply-chain/policy.toml ]` and skips (exit 0) with a notice when absent, mirroring the previous soak-charts guard. Restore the comment explaining the default-branch schedule semantics.
```

</details>

<details>
<summary><b>:orange_circle: Major: Prior rustls-webpki advisories dropped without clear remediation evidence</b> (Reported by: xai)</summary>

**File:** `deny.toml` (line 14)

The previous deny.toml ignored RUSTSEC-2026-0049, RUSTSEC-2026-0098, RUSTSEC-2026-0099, and RUSTSEC-2026-0104 on rustls-webpki 0.101.7 via the legacy reqwest/hocon/opentelemetry chain. Those ignores are gone, and bans only enforce rustls-webpki:<0.103.13. If any lockfile still resolves a vulnerable 0.101.x (or another unpatched line), CI will no longer explicitly track acceptance of those certificate-validation issues. This needs confirmation against the final locked graphs, not just the presence of a newer ban rule.

**Suggestion:**
```
Before merge, publish or attach the three locked all-feature deny reports and confirm no vulnerable rustls-webpki version remains in workspace, fuzz, or soak-charts graphs. If a legacy copy remains, restore a version-scoped exception with owner/review-by or remove the vulnerable path entirely.
```

</details>

<details>
<summary><b>:orange_circle: Major: validate_report treats every non-JSON output line as a fatal error</b> (Reported by: openrouter)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 186)

validate_report appends 'The scanner produced an invalid JSON record.' for ANY line that fails to parse as JSON, across the concatenation of stdout and stderr. cargo-deny with --log-level info can emit non-JSON lines on stderr in some environments (e.g., advisory-db fetch/git progress messages, deprecation warnings, or terminal-control output on non-TTY). If that happens, the supply-chain job fails even though all four checks passed. This is a hypothesis — the verified local runs passed — but the failure mode is brittle: it converts informational scanner noise into a hard CI failure with an opaque message.

**Suggestion:**
```
Only fail on malformed lines that begin with '{' (i.e., lines that claim to be JSON records), or run cargo-deny with stderr separated: parse stdout as the JSON report and treat non-empty stderr only as an error when the exit code is non-zero. Alternatively filter stderr through `grep -v '^{'` before validation and record the raw stream in the evidence artifact.
```

</details>

<details>
<summary><b>:orange_circle: Major: Scanner install trusts HTTPS + sha256 only; no signature authenticity</b> (Reported by: xai)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 214)

install_scanner downloads cargo-deny from GitHub releases over HTTPS and checks a pinned sha256 from supply-chain/policy.toml. That is much better than an unpinned action tag, but a compromised release asset plus policy change, or a malicious PR that updates both URL/version and hashes together, still installs whatever blob matches. The workflow permissions are contents:read, which helps, but the authenticity model is still hash-pinning rather than signed provenance.

**Suggestion:**
```
Treat policy.toml scanner hashes as highly sensitive (CODEOWNERS already does). Consider also pinning release attestation/signature verification if Embark publishes them, and require two-party review on scanner version bumps.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Rust cache save-if restriction not applied consistently</b> (Reported by: xai)</summary>

**File:** `.github/workflows/slashing-tests.yml` (line 69)

ci.yml restricts Swatinem/rust-cache writes to push on dev/master (and coverage uses save-if: false), but slashing-tests.yml still uses the action without save-if guards. PR/fork runs in that workflow can still write caches depending on default action behavior, which weakens the cache-poisoning control introduced elsewhere.

**Suggestion:**
```
Apply the same save-if conditions to every rust-cache usage, or set save-if: false on workflows that run untrusted PR code.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Slashing jobs still permit unrestricted cache writes</b> (Reported by: openai)</summary>

**File:** `.github/workflows/slashing-tests.yml` (line 76)

The Swatinem/rust-cache action is pinned, but its instances in this workflow do not set `save-if`. Unlike the updated jobs in ci.yml, these jobs therefore retain the action's default cache-write behavior whenever GitHub permits a save. This leaves the cache-write restriction inconsistent and can unnecessarily expose later jobs in the same cache scope to artifacts produced while building changed code. The omission occurs in multiple slashing jobs, including the loom and fuzz caches.

**Suggestion:**
```
Add `save-if: false` for test-only caches, or use the same trusted-branch expression introduced in ci.yml for every rust-cache invocation in this workflow.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Extensive explanatory comments stripped from deny.toml and CODEOWNERS</b> (Reported by: openrouter)</summary>

**File:** `deny.toml` (line 1)

The rewritten deny.toml drops the rationale for all-features=true, the per-advisory justification paragraphs (e.g., why h2 0.3.27 cannot be patched, why the workspace h2 0.4.x copy was bumped), and the soak-charts/instant scoping notes. The new `reason` fields are shorter and in several cases less accurate than the originals (e.g., RUSTSEC-2021-0141 now says 'rholang directly uses dotenv' whereas the original noted it is transitive via openai-api-rs; RUSTSEC-2024-0436 changed from 'transitive via metrics' to 'comm discovery tests use paste'). Since policy.toml duplicates owner/review-by metadata, the detailed exploitability analysis is now nowhere recorded. CODEOWNERS similarly lost the paragraph explaining why the scope is narrow.

**Suggestion:**
```
Verify each rewritten `reason` against the actual dependency path (cargo tree -i) — at least the dotenv and paste entries appear to have changed attribution. Consider keeping the longer exploitability notes either in deny.toml comments or in policy.toml, since review-by dates force periodic re-review and the next reviewer will need this context.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Exact build-script allowlist will churn and can break unrelated dependency bumps</b> (Reported by: xai)</summary>

**File:** `deny.toml` (line 68)

bans.build.allow-build-scripts pins a large set of crate@version entries. This is a deliberate hardening choice, but any transitive build-script version bump will fail deny until the allowlist is updated. That can create CI breakages on routine Cargo.lock refreshes and pressure people to widen the list quickly.

**Suggestion:**
```
Document the update process next to the allowlist and consider CI messaging that points authors to the exact newly-observed build-script crate versions from the deny report.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Removing Cargo cache mounts makes every image build a full from-scratch compile</b> (Reported by: openrouter)</summary>

**File:** `node/Dockerfile` (line 76)

The cache mounts for /usr/local/cargo/registry, /usr/local/cargo/git, and the target dir were removed to eliminate shared-cache poisoning risk, which is a defensible supply-chain trade-off. However, this also removes the target-dir cache, so every Docker image build recompiles the entire dependency tree (~30-60+ min on typical runners for a workspace this size). The network isolation (--network=none) plus `cargo fetch --locked` already addresses the poisoning concern for the registry/git caches; the compile-artifact cache was the main time saver and does not cross trust boundaries if scoped per-ref.

**Suggestion:**
```
Consider retaining a target-dir cache mount scoped by github.ref or a build-arg cache key, or document the expected CI build-time regression so reviewers can confirm it was an accepted cost. At minimum, verify the first build completes within the workflow's timeout-minutes budget.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Runtime still installs tools via apt without package pins</b> (Reported by: xai)</summary>

**File:** `node/Dockerfile` (line 120)

Base images are digest-pinned and grpcurl is checksum-verified, which is good. Remaining apt packages (ca-certificates, libssl3, curl, jq, etc.) are still floating within bookworm. This is common, but it means the runtime dependency set is not fully reproducible and is outside the cargo-deny story.

**Suggestion:**
```
Accept as residual risk or pin package versions / use a verified internal base image if stronger runtime reproducibility is required.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Exception owners are only checked for non-empty text</b> (Reported by: openai)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 31)

The policy validator accepts any non-empty owner value, such as `owner = "unknown"`, and the owner is not otherwise interpreted or checked. Consequently, the stated requirement that advisory exceptions have owners does not ensure that the value identifies an approved GitHub user or team. Review dates and exact versions are meaningfully validated, but ownership is currently only descriptive.

**Suggestion:**
```
Require an expected GitHub owner format and preferably restrict values to an explicit approved-owner list, such as `@F1R3FLY-io/ci-maintainers`, maintained in the controls policy.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Scanner download follows redirects without scheme restriction</b> (Reported by: openrouter)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 295)

install_scanner uses urllib.request.urlopen on a hard-coded https://github.com URL. urllib follows HTTP redirects by default, including a redirect downgraded to http://, after which the payload would be fetched unauthenticated. The sha256 check against policy.toml still anchors integrity, so a downgrade alone cannot inject code (it would fail the checksum), but unlike the Dockerfile change in this same PR (which adds --proto '=https' --proto-redir '=https' to curl), the Python path does not enforce the transport invariant the PR claims to establish uniformly.

**Suggestion:**
```
Add a custom redirect handler that rejects non-https redirect targets, or use `urllib.request.build_opener(NoHTTPRedirect)` and fail on any non-https Location header. Alternatively shell out to the already-hardened curl invocation for consistency.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: install_archive fails on re-run if cargo-deny.download is left behind</b> (Reported by: openrouter)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 320)

The temporary file is opened with mode 'xb' (exclusive create). If a previous install attempt was interrupted after writing the temp file but before replace(), subsequent runs fail with FileExistsError and the error message gives no hint to remove the stale file. In CI with $RUNNER_TEMP this is unlikely, but for local reuse it is a sharp edge.

**Suggestion:**
```
Unlink destination/'cargo-deny.download' with missing_ok=True before opening, or include the failure reason in the error message.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded date in test</b> (Reported by: bedrock)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 398)

The test uses a hardcoded date, which may need to be updated in the future.

**Suggestion:**
```
Use a dynamic date calculation or a fixture to avoid hardcoding dates in tests.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Action pinning regression test only covers a hard-coded workflow subset</b> (Reported by: openai)</summary>

**File:** `scripts/ci/test_supply_chain.py` (line 292)

The pinning test checks four explicitly named workflows rather than discovering all files under `.github/workflows`. A newly added workflow, a renamed workflow, or another existing workflow can use a mutable external action reference without this control noticing. CODEOWNERS still provides review protection, so this is a policy-enforcement gap rather than an immediate vulnerability.

**Suggestion:**
```
Enumerate all `*.yml` and `*.yaml` files under `.github/workflows`, then validate every external `uses:` reference. Explicitly skip only documented local or reusable workflows where necessary.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Hardcoded date in test</b> (Reported by: bedrock)</summary>

**File:** `scripts/ci/test_supply_chain.py` (line 351)

The test uses a hardcoded date, which may need to be updated in the future.

**Suggestion:**
```
Use a dynamic date calculation or a fixture to avoid hardcoding dates in tests.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: CODEOWNERS scope expanded; ensure owners path coverage matches intent</b> (Reported by: xai)</summary>

**File:** `.github/CODEOWNERS` (line 1)

Ownership now includes deny.toml, supply-chain/, .cargo/, root Cargo manifests/locks, Dockerfile, toolchain, and scripts/ci. That matches the PR intent. The removed commentary previously explained why the scope was narrow; the new broader blast radius is probably correct but should be understood by maintainers.

**Suggestion:**
```
Keep a short comment explaining why manifest/lock/deny/Docker paths are CI-owned, so future reviewers do not narrow it again accidentally.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Bitmap regression tests are valuable but hasher is highly artificial</b> (Reported by: xai)</summary>

**File:** `rspace++/tests/hot_store_spec.rs` (line 737)

The new tests materially target snapshot preservation under dense/colliding keys for the bitmaps 3.1.0 pin. The custom hasher collapses to 8 bits, which is good for collision stress, but it is not a substitute for upstream bitmaps maintenance risk noted in the PR context.

**Suggestion:**
```
Retain the tests; track imbl replacement work separately since 3.1.0 is still unmaintained even if unaffected by RUSTSEC-2025-0167.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Malformed advisory records can abort remaining graph scans</b> (Reported by: openai)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 138)

`validate_report` indexes `advisory["id"]` and assumes each graph is a dictionary. If the verified scanner emits a malformed or changed diagnostic schema, a `KeyError` or `AttributeError` can escape `scan_manifest`; the per-manifest handler in `check` does not catch those exceptions, so later manifests are not scanned. This is not expected with the pinned scanner, but it conflicts with the intended fail-closed, continue-all-scans behavior.

**Suggestion:**
```
Validate that `advisory`, each graph, and `Krate` are dictionaries with the required fields, append an invalid-record error when they are not, and avoid direct indexing of unvalidated scanner data.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Platform mapping tables evaluated eagerly with unhelpful KeyError on unsupported hosts</b> (Reported by: openrouter)</summary>

**File:** `scripts/ci/check_supply_chain.py` (line 248)

scanner_target() indexes the architectures/systems dicts directly, so an unsupported platform (e.g., Windows, or Linux on armv7) raises a bare KeyError. main() catches it, but the printed message ('Supply-chain check failed: Windows') does not tell the user what was unsupported or what is supported.

**Suggestion:**
```
Use .get() and raise ValueError(f'Unsupported platform: {platform.system()}/{platform.machine()}; supported: ...') for an actionable message.
```

</details>
---

<details>

> **ERROR TIMEOUT**: Empty response from provider

The PR introduces significant improvements to dependency checks and build inputs, enhancing security and reliability. However, there are some areas that could be improved, such as the installation method for cargo-deny and the use of hardcoded dates in tests.

#### openai (gpt-5.6-sol)

The change substantially improves dependency and build-input controls through immutable action/image references, locked multi-graph scanning, verified scanner installation, vendored Swagger assets, and an offline compilation step. No critical vulnerability is established in the supplied diff. The remaining actionable gaps are inconsistent cache-write restrictions in the slashing workflow and regression tests that do not automatically cover every workflow; owner validation and malformed scanner-output handling could also be made more robust.

This is a thorough, well-tested supply-chain hardening PR: SHA-pinned actions and images, digest-verified cargo-deny installation with safe (no-extraction) tarball handling, exact-version advisory exceptions with expiry, per-workspace locked scans with retained evidence, vendored Swagger assets, network-isolated Docker builds, and a bitmaps downgrade that pins to the unaffected 3.1.0 release while banning affected ones. The test suite for the policy tooling is genuinely strong (28 tests covering expiry, version ranges, malformed reports, symlink archives, and manifest coverage). The main concerns are robustness hypotheses rather than proven bugs: validate_report's fail-on-any-non-JSON-line behavior may turn benign scanner stderr noise into CI failures, the scheduled deny workflow will hard-fail on refs lacking the new script/policy (the old guard was removed), and the Docker changes — flagged by the author as unverified — carry both a potential cargo-fetch host-dependency gap and a significant build-time regression from removing all cache mounts. None of these block merge, but the workflow guard and the report-parsing brittleness are worth addressing, and the rewritten deny.toml reasons should be double-checked for accurate dependency attribution.

#### xai (grok-4.5)

This PR substantially hardens the supply chain: pinned GitHub actions/images, verified cargo-deny installation, locked multi-graph scans with evidence uploads, vendored Swagger UI, checksummed grpcurl, offline frozen Docker builds, fuzz lockfile tracking, and an explicit bitmaps 3.1.0 pin that blocks RUSTSEC-2025-0167. The direction is sound and much of the control surface is carefully tested. However, merge should wait on human review of two practical risks: confirm the removed rustls-webpki advisory exceptions are truly eliminated from all locked graphs, and verify the new no-network/no-shared-cache Docker build path on both architectures. Secondary follow-ups include making rust-cache write restrictions consistent in slashing-tests and budgeting for build-script allowlist churn.
</details>

---


</details>

jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :mag: Needs Review (60.0000% agreement)

**Reviewed by:**

| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.75 |
| xai | `grok-4.5` | Needs Review | 0.72 |

---

### Summary

This is a well-conceived supply-chain hardening PR: it pins all third-party GitHub Actions and cross-compilation container images to immutable digests, restricts cache writes to trusted branches, replaces the cargo-deny action with a verified pinned scanner across all lockfiles, tightens deny.toml with structured advisory exceptions, build-script allow-listing, and version-scoped bans, downgrades bitmaps to the unaffected 3.1.0, vendors Swagger assets, and expands CODEOWNERS coverage of supply-chain-sensitive files. The main functional concern is the scheduled deny workflow, which drops the previous existence guards and will fail on branches that don't yet carry the new tooling; secondary concerns are maintainability (exact-version build-script allow-list, missing version annotations on SHA pins, loss of extensive rationale comments) and an evidence-upload step that can mask real failures. Nothing here is unsafe to merge, but the schedule-branch regression should be addressed and the unverifiable script/SHA contents deserve a human spot-check.

The visible changes substantially improve dependency and action pinning, locked-build enforcement, cache-write restrictions, and audit evidence retention. However, several workflows still persist checkout credentials while executing repository-controlled or caller-selected code, leaving a meaningful token-exfiltration path that should be fixed before merge. The supply-chain test ordering also causes avoidable cold compilation. Because the supplied diff is truncated, the new scanner installation, policy validation, vendored assets, and container build changes could not be fully assessed.

This PR is a substantial supply-chain hardening pass: action/image digest pins, restricted cache writes, locked Cargo invocations, stronger cargo-deny policy (including bitmaps 3.1.0 and build-script controls), CODEOWNERS expansion, and replacement of cargo-deny-action with a pinned local installer/checker plus retained scan evidence. The direction is sound and likely improves security posture, but merge should wait on human verification of the truncated custom install/check script, confirmation that every action/container SHA matches upstream releases, and alignment of advisory-exception metadata with the PR’s stated owner/version/review-date policy. Secondary follow-ups include checkout credential consistency, fragile rholang license clarification, and maintainability of the exact build-script allowlist.

This PR significantly improves supply-chain security by pinning GitHub Actions to commit SHAs, adding container image digests, and introducing comprehensive cargo-deny policies. However, the build script allowlist is overly permissive and includes interpreted scripts with only a warning. The removal of explanatory comments from workflow files reduces maintainability. The addition of a version field to a git dependency is redundant. Overall, the security posture is improved, but the build script policy should be tightened before merge.

The PR makes significant changes to the CI/CD pipelines and dependency management. While the changes aim to harden the supply chain and dependency checks, there are some inconsistencies and potential issues with the use of specific commit SHAs for GitHub Actions and the inconsistent use of `actions/checkout`. Additionally, some redundant comments can be removed to clean up the files. Overall, the changes look good but require some minor adjustments and review.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Checkout credentials remain available to code under test</b> (Reported by: openai, xai)</summary>

**File:** `.github/workflows/slashing-tests.yml` (line 72)

Several workflows now set persist-credentials: false on actions/checkout, but slashing-tests (and some other jobs) still use bare checkout pins without disabling credential persistence. Default GITHUB_TOKEN persistence widens credential exposure to later steps/tools.

---

The checkout steps in this workflow still use the default `persist-credentials: true`. Subsequent Rust builds, tests, fuzz targets, build scripts, or other repository-controlled commands can read the injected GitHub token from the local Git configuration and potentially exfiltrate it. This is especially relevant when these jobs run for pull requests containing untrusted changes. Pinning the checkout action protects against action-tag movement but does not prevent credential exposure to checked-out code.

**Suggestion:**
```
Add `with: { persist-credentials: false }` to every checkout that does not subsequently perform an authenticated Git operation. Also declare minimal workflow or job-level `permissions`, normally `contents: read`, so any accidentally exposed token has the smallest possible scope.
```

</details>

<details>
<summary><b>:orange_circle: Major: Use of specific commit SHAs for actions</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 100)

The use of specific commit SHAs for GitHub Actions (e.g., `actions/checkout@fbc6f3992d24b796d5a048ff273f7fcc4a7b6c09`) instead of tags (e.g., `actions/checkout@v5`) can lead to issues if the action is updated or deprecated. It is recommended to use tags to ensure compatibility with future versions of the action.

**Suggestion:**
```
Use tags for GitHub Actions to avoid potential issues with future updates.
```

</details>

<details>
<summary><b>:orange_circle: Major: Action/image SHAs need independent verification</b> (Reported by: xai)</summary>

**File:** `.github/workflows/ci.yml` (line 151)

Workflows pin many third-party actions (checkout, upload/download-artifact, rust-cache, setup-buildx, login-action, release-action, deploy-pages, etc.) and Cross.toml pins container digests. Pinning is correct hardening, but a malicious or mistaken SHA is worse than a moving tag if it is not checked against upstream release tags/immutable digests.

**Suggestion:**
```
Document each pin as # vX.Y.Z and record the verification source (upstream tag → commit → SHA256) in the PR or supply-chain notes before merge.
```

</details>

<details>
<summary><b>:orange_circle: Major: Scheduled job will fail on branches missing the new supply-chain tooling</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/deny-schedule.yml` (line 30)

The old workflow explicitly guarded steps that depend on files that may not exist on the checked-out branch ("a branch that does not carry the crate yet must not fail its scheduled check"). The new job unconditionally runs `bash scripts/ci/check-supply-chain.sh` after checking out `dev` or `master` by ref. Until this PR's script lands on both branches, the scheduled run for the branch that lacks `scripts/ci/check-supply-chain.sh` (likely master) will fail. Similarly, `target/supply-chain/` will not exist, and the evidence upload with `if-no-files-found: error` will produce a second failure.

**Suggestion:**
```
Add an existence guard (e.g., `if [ -f scripts/ci/check-supply-chain.sh ]`) or run the script from the workflow's own checkout (github.workflow_sha) rather than the branch under audit, and only enable the strict evidence upload once both long-lived branches carry the tooling.
```

</details>

<details>
<summary><b>:orange_circle: Major: Caller-selected repository receives a persisted workflow token</b> (Reported by: openai)</summary>

**File:** `.github/workflows/testbed-quality-gate.yml` (line 105)

This workflow checks out `${{ inputs.testbed_repo }}` and `${{ inputs.testbed_ref }}` without disabling credential persistence. Because the repository and revision are caller-controlled inputs and the checked-out project is subsequently exercised, that code can inspect Git configuration and recover the workflow's GitHub token. The same omission is repeated by the other checkout steps in this workflow.

**Suggestion:**
```
Set `persist-credentials: false` on all testbed checkout steps and explicitly minimize job permissions. If the reusable workflow can be invoked from privileged contexts, validate or allowlist `testbed_repo` and ensure no secrets are exposed to the jobs executing its code.
```

</details>

<details>
<summary><b>:orange_circle: Major: Redundant version specification for git dependency</b> (Reported by: openrouter)</summary>

**File:** `Cargo.toml` (line 82)

Adding `version = "0.1.0"` to the `rholang-parser` git dependency is redundant because Cargo resolves git dependencies by commit SHA (rev), not semantic version. The version field is ignored when `rev` is present. This adds confusion without providing value.

**Suggestion:**
```
Remove the `version` field: `rholang-parser = { git = "https://github.com/F1R3FLY-io/rholang-rs", rev = "527af8afb37f6cad470516fe316076f120b9c353" }`
```

</details>

<details>
<summary><b>:orange_circle: Major: Advisory exception metadata may not match stated PR policy</b> (Reported by: xai)</summary>

**File:** `deny.toml` (line 20)

PR claims exceptions require exact versions, owners, and review dates. The visible ignore entries include id + reason only—no owner, review date, or crate@version binding in the ignore objects. Reasons are improved, but the stated control is not fully enforced by config shape alone (unless the new supply-chain tests enforce it out-of-band).

**Suggestion:**
```
If policy is mandatory, encode owner/review-by/crate version in each exception and assert it in supply-chain tests; fail CI when fields are missing or expired.
```

</details>

<details>
<summary><b>:orange_circle: Major: Exact-version build-script allowlist will break on routine bumps</b> (Reported by: xai)</summary>

**File:** `deny.toml` (line 41)

bans.build.allow-build-scripts enumerates dozens of crate@version entries and sets executables=deny. Any transitive patch that adds/updates a build script will fail deny until the allowlist is edited. That is strong policy, but easy to mis-maintain (stale versions left allowed, or broad emergency exceptions).

**Suggestion:**
```
Add a documented update procedure (regenerate allowlist from a clean lockfile + review delta), and consider CI output that prints the exact missing allow-build-scripts entries to reduce rubber-stamp updates.
```

</details>

<details>
<summary><b>:orange_circle: Major: Overly permissive build script allowlist</b> (Reported by: openrouter)</summary>

**File:** `deny.toml` (line 45)

The `allow-build-scripts` list contains 100+ crates including build-critical and network-capable dependencies. Build scripts execute arbitrary code at compile time with the same privileges as the build process. Allowing this many crates significantly expands the attack surface for supply-chain attacks. While pinning versions helps, the sheer breadth of the allowlist reduces the effectiveness of the `bans.build` restrictions. Additionally, `interpreted = "warn"` allows interpreted build scripts (e.g., Python, shell) to run with only a warning, which could execute arbitrary code without failing the build.

**Suggestion:**
```
Audit the list to ensure only crates that genuinely require build scripts are included. Consider changing `interpreted` to `deny` unless specifically needed.
```

</details>

<details>
<summary><b>:orange_circle: Major: Critical supply-chain installer not fully visible in diff</b> (Reported by: xai)</summary>

**File:** `scripts/ci/check-supply-chain.sh` (line 1)

CI now installs cargo-deny via a custom script (scripts/ci/check-supply-chain.sh install) instead of EmbarkStudios/cargo-deny-action. That script is the trust anchor for advisory scanning (pinned binary, checksums, evidence paths). The provided diff is truncated and does not include the full installer/checker implementation, so authenticity of pins, fail-closed checksum handling, and path/permission behavior cannot be validated from this review alone.

**Suggestion:**
```
Require explicit human review of the full install/check script: pinned version + multi-arch hashes, HTTPS-only sources, checksum failure must non-zero exit, no curl|sh patterns, and evidence written only under target/supply-chain/.
```

</details>

<details>
<summary><b>:orange_circle: Major: Important documentation removed</b> (Reported by: bedrock, openrouter)</summary>

**File:** `.github/workflows/deny-schedule.yml` (line 1)

The comment at the top of the file is redundant and can be removed to clean up the file.

---

The header comments explaining why the scheduled workflow exists (to catch newly-published RustSec advisories during quiet periods) and the critical note about GitHub only firing schedule events from the default branch were removed. This documentation is valuable for maintainers to understand the workflow's purpose and operational constraints.

**Suggestion:**
```
Remove the redundant comment.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Contextual documentation removed</b> (Reported by: openrouter)</summary>

**File:** `.github/CODEOWNERS` (line 1)

Comments explaining the narrow scope of CI credential ownership and the requirement for `push` permissions were removed. These comments help prevent accidental privilege escalation or ownership changes.

**Suggestion:**
```
Restore the comment block explaining the credential model and push permission requirements.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: CODEOWNERS blast radius expanded without credential-model notes</b> (Reported by: xai)</summary>

**File:** `.github/CODEOWNERS` (line 10)

Ownership now covers deny.toml, supply-chain/, Cargo.toml/lock, Dockerfile, toolchain, and CI scripts—good for supply-chain gates. The previous comment explaining why workflow ownership must stay narrow (credential/OIDC blast radius) was removed, which can lose important review context for future CODEOWNERS edits.

**Suggestion:**
```
Restore a short comment distinguishing credential-bearing workflow ownership from dependency-policy ownership so future edits do not accidentally widen secrets access.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent use of `actions/checkout`</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/ci.yml` (line 148)

The workflow uses different versions of `actions/checkout` across different jobs. It is recommended to use a consistent version to avoid potential issues.

**Suggestion:**
```
Use a consistent version of `actions/checkout` across all jobs.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: SHA-pinned actions lack version annotation comments</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 151)

All third-party actions are now pinned to full commit SHAs (good), but almost none carry the conventional trailing comment indicating the corresponding tag (e.g., `uses: actions/checkout@fbc6f399... # v5.0.0`). One instance in ci.yml (`Swatinem/rust-cache@6323deb... # v2`) shows the intended convention. Without the annotation, reviewers and tools like Dependabot/Renovate cannot easily verify which release a SHA corresponds to, making audits and future upgrades error-prone.

**Suggestion:**
```
Append `# vX.Y.Z` comments to every SHA-pinned `uses:` line across all workflow files, consistent with the one annotated rust-cache pin.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Supply-chain tests run before restoring the Rust cache</b> (Reported by: openai)</summary>

**File:** `.github/workflows/ci.yml` (line 374)

`cargo test --locked -p supply-chain` runs before the `Swatinem/rust-cache` step. Consequently, this crate and its dependencies are downloaded and compiled without using the existing cache on every job, reducing the value of the cache and increasing CI latency.

**Suggestion:**
```
Move the cache restoration step before `Verify supply-chain controls`. The existing conditional `save-if` can still restrict writes while allowing cache reads for all events.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Evidence upload with if: always() + if-no-files-found: error masks the real failure</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 415)

In the `deny` job, `Upload supply-chain evidence` runs with `if: always()` and `if-no-files-found: error`. If the install step fails before any scan output is produced under `target/supply-chain/`, the upload step also fails, adding a second, noisier failure that can obscure the root cause in the job summary.

**Suggestion:**
```
Use `if-no-files-found: warn` for the always-run upload, or gate the upload on the check step having started (e.g., `if: always() && steps.check.conclusion != 'skipped'`).
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Inconsistent use of `actions/checkout`</b> (Reported by: bedrock)</summary>

**File:** `.github/workflows/slashing-tests.yml` (line 69)

The workflow uses different versions of `actions/checkout` across different jobs. It is recommended to use a consistent version to avoid potential issues.

**Suggestion:**
```
Use a consistent version of `actions/checkout` across all jobs.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Removal of explanatory comments loses institutional knowledge</b> (Reported by: anthropic)</summary>

**File:** `deny.toml` (line 1)

The rewrite strips the extensive rationale comments from deny.toml (why --all-features matches CI, why each advisory was accepted with its full transitive chain, the LGPL linking-exception rationale for zbase32) and from deny-schedule.yml (why the cron exists, the default-branch schedule caveat) and CODEOWNERS (the blast-radius rationale). The new inline `reason` fields are terse. Future maintainers will have less context to decide when an exception can be removed.

**Suggestion:**
```
Retain the key context either as comments or in a supply-chain policy document referenced from deny.toml, especially the remediation conditions for each advisory exception.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: rholang license clarify via ../README.md is fragile</b> (Reported by: xai)</summary>

**File:** `deny.toml` (line 47)

License clarification for rholang-* crates now points at license-files = [{ path = "../README.md", hash = ... }]. README is not a license text and relative path/hash coupling is brittle across crate layout or fetch modes; a README edit can break deny or weakly substantiate Apache-2.0.

**Suggestion:**
```
Prefer a real LICENSE file in the dependency, a vendored license excerpt, or a clearer clarify strategy that does not depend on README content.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: License clarify paths reach outside the crate root</b> (Reported by: anthropic)</summary>

**File:** `deny.toml` (line 50)

`license-files = [{ path = "../README.md", hash = 0xcd11a2a7 }]` for rholang-parser and rholang-tree-sitter points to a file outside the crate directory in the git checkout. This works only because both crates live in the same repo checkout, but it is fragile: if upstream restructures the repo or the README changes, the pinned hash breaks the check with a non-obvious error, and the path semantics depend on cargo-deny resolving relative paths from the crate root within the git cache.

**Suggestion:**
```
Add a comment explaining the parent-directory reference and the pinned hash so the failure mode is diagnosable, or ask upstream to add a `license` field / LICENSE file per crate.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Exact-version allow-build-scripts list creates heavy maintenance burden</b> (Reported by: anthropic)</summary>

**File:** `deny.toml` (line 63)

The `[bans.build] allow-build-scripts` list pins ~120 crates at exact versions (e.g., `serde@1.0.229`, `libc@0.2.189`). Any routine `cargo update` or transitive bump will fail cargo-deny until this list is hand-edited, and the failure mode gives no signal about whether the new build script was actually reviewed. This is a defensible fail-closed posture, but it needs a documented update procedure or contributors will learn to rubber-stamp additions.

**Suggestion:**
```
Document the review-and-update workflow for this list (ideally in a comment at the top of the section or in supply-chain docs), and consider whether first-party workspace crates (`comm`, `models`, `node`) and low-risk sys crates could be allowed without exact versions to reduce churn.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: supply-chain verification test runs before dependency cache is restored</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/ci.yml` (line 375)

In the lint job, `cargo test --locked -p supply-chain` executes before the `Cache dependencies` (rust-cache) step, so its build cannot benefit from the restored cache and will cold-compile the crate and its dependencies (toml, clap, sha2, tar, flate2, etc.) on every run.

---

rust-cache save-if limited to push on dev/master reduces cache poisoning from forks/PRs (good). PR jobs can only read existing caches, so dependency churn may lengthen CI until merges repopulate caches.

**Suggestion:**
```
Accept as security tradeoff; monitor wall time and consider separate read-only cache keys if PR latency becomes painful.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Unanchored CODEOWNERS patterns match every crate manifest</b> (Reported by: anthropic)</summary>

**File:** `.github/CODEOWNERS` (line 13)

`Cargo.toml`, `Cargo.lock`, `build.rs`, and `Dockerfile` are unanchored, so they match those filenames anywhere in the tree (e.g., `comm/Cargo.toml`, every crate's `build.rs`). This means any dependency or build-script change in any crate now requires @F1R3FLY-io/ci-maintainers review. If that is the intent for supply-chain governance, fine; if only root manifests were intended, prefix with `/` as done for `/Cross.toml`.

**Suggestion:**
```
Confirm intent; if only root files should be owned, use `/Cargo.toml`, `/Cargo.lock`, `/build.rs`, `/Dockerfile`. If all manifests are intentionally owned, add a comment stating so.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Pin comments missing on most action SHAs</b> (Reported by: xai)</summary>

**File:** `.github/workflows/_integration-pipeline.yml` (line 103)

Some pins retain version comments (e.g. rust-cache # v2 in coverage), but most replacements are bare 40-char SHAs. That slows audits and Dependabot-style upgrades.

**Suggestion:**
```
Use a consistent 'uses: org/action@sha # vX.Y.Z' convention everywhere.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Cannot verify SHA pins or scanner-verification script within this diff</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/ci.yml` (line 369)

The security value of the SHA pins depends on each SHA actually corresponding to a reviewed release of the named action, and the value of `scripts/ci/check-supply-chain.sh install` depends on how it verifies the cargo-deny binary (checksum/signature source). Neither is verifiable from the truncated diff; the script and the supply-chain crate's test contents are not shown.

**Suggestion:**
```
Ensure the script pins cargo-deny by version + sha256 fetched from a source committed in-repo (not fetched at runtime), and have a second reviewer spot-check that each action SHA matches its upstream tagged release.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Removed operational comments from scheduled deny workflow</b> (Reported by: xai)</summary>

**File:** `.github/workflows/deny-schedule.yml` (line 1)

Comments explaining why the workflow must live on the default branch and how the branch matrix works were deleted. That context prevents accidental breakage of scheduled advisory coverage.

**Suggestion:**
```
Keep a brief note that schedule events only run from the default branch workflow file.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Advisory ignores lack expiration or review dates</b> (Reported by: openrouter)</summary>

**File:** `deny.toml` (line 25)

While the PR description mentions requiring "review dates" for exceptions, the `ignore` entries in `deny.toml` contain `reason` fields but no `review_date` or `expires` fields. Without periodic review, these exceptions may persist longer than necessary.

**Suggestion:**
```
Add `review_date = "YYYY-MM-DD"` or similar to each advisory ignore entry to enforce periodic re-evaluation.
```

</details>
---

<details>

This is a well-conceived supply-chain hardening PR: it pins all third-party GitHub Actions and cross-compilation container images to immutable digests, restricts cache writes to trusted branches, replaces the cargo-deny action with a verified pinned scanner across all lockfiles, tightens deny.toml with structured advisory exceptions, build-script allow-listing, and version-scoped bans, downgrades bitmaps to the unaffected 3.1.0, vendors Swagger assets, and expands CODEOWNERS coverage of supply-chain-sensitive files. The main functional concern is the scheduled deny workflow, which drops the previous existence guards and will fail on branches that don't yet carry the new tooling; secondary concerns are maintainability (exact-version build-script allow-list, missing version annotations on SHA pins, loss of extensive rationale comments) and an evidence-upload step that can mask real failures. Nothing here is unsafe to merge, but the schedule-branch regression should be addressed and the unverifiable script/SHA contents deserve a human spot-check.

The PR makes significant changes to the CI/CD pipelines and dependency management. While the changes aim to harden the supply chain and dependency checks, there are some inconsistencies and potential issues with the use of specific commit SHAs for GitHub Actions and the inconsistent use of `actions/checkout`. Additionally, some redundant comments can be removed to clean up the files. Overall, the changes look good but require some minor adjustments and review.

#### openai (gpt-5.6-sol)

The visible changes substantially improve dependency and action pinning, locked-build enforcement, cache-write restrictions, and audit evidence retention. However, several workflows still persist checkout credentials while executing repository-controlled or caller-selected code, leaving a meaningful token-exfiltration path that should be fixed before merge. The supply-chain test ordering also causes avoidable cold compilation. Because the supplied diff is truncated, the new scanner installation, policy validation, vendored assets, and container build changes could not be fully assessed.

This PR significantly improves supply-chain security by pinning GitHub Actions to commit SHAs, adding container image digests, and introducing comprehensive cargo-deny policies. However, the build script allowlist is overly permissive and includes interpreted scripts with only a warning. The removal of explanatory comments from workflow files reduces maintainability. The addition of a version field to a git dependency is redundant. Overall, the security posture is improved, but the build script policy should be tightened before merge.

#### xai (grok-4.5)

This PR is a substantial supply-chain hardening pass: action/image digest pins, restricted cache writes, locked Cargo invocations, stronger cargo-deny policy (including bitmaps 3.1.0 and build-script controls), CODEOWNERS expansion, and replacement of cargo-deny-action with a pinned local installer/checker plus retained scan evidence. The direction is sound and likely improves security posture, but merge should wait on human verification of the truncated custom install/check script, confirmation that every action/container SHA matches upstream releases, and alignment of advisory-exception metadata with the PR’s stated owner/version/review-date policy. Secondary follow-ups include checkout credential consistency, fragile rholang license clarification, and maintainability of the exact build-script allowlist.
</details>

---

jltatbeach (approved):

jeffrey-l-turner:
## Review remediation

The multi-agent review returned needs_review at 60 percent agreement with one critical finding, ten major findings, eleven minor findings, and six suggestions. The diff exceeded the review size limit, so the reviewers saw a truncated diff. The critical and major findings are resolved as follows. The changes are on this branch in the follow-up commit.

### Fixed

- **Critical: checkout credentials remain available to code under test** (OpenAI, xAI). All ten checkouts in `slashing-tests.yml` now set `persist-credentials: false`. The two checkouts in `ci.yml` that lacked the setting, in the `build_base` and `lint` jobs, now set it too. The one tag fetch after the `build_base` checkout is anonymous, which works because the repository is public.
- **Major: caller-selected repository receives a persisted workflow token** (OpenAI). All seven testbed checkouts in `testbed-quality-gate.yml` now set `persist-credentials: false`. The `TESTBED_PAT` token is used for the clone only and is no longer written into the testbed git configuration before the testbed code runs.
- **Major: scheduled job fails on a branch without the new tooling** (Anthropic). `deny-schedule.yml` restores the existence guard pattern from the previous version. A detect step checks for `scripts/ci/check-supply-chain.sh`. The install, check, and upload steps run only when the script is present. A skipped branch prints a notice so the gap stays visible.
- **Major: important documentation removed from the schedule workflow** (Bedrock, OpenRouter). The header note that explains the default-branch schedule rule is restored verbatim.
- **Major: action and image SHAs need independent verification** (xAI). Every action SHA was resolved against the upstream tag list. All pins are release tags except one. `dtolnay/rust-toolchain@d1031067` was the upstream master head from 2026-09-03, not a tag. All 13 uses are repinned to `6c977a6c`, which is the `v1` tag. The existing `# master` annotation on that SHA was wrong and now reads `# v1`.
- **Minor: SHA pins lack version annotations** (Anthropic, xAI). Every pin now carries the resolved tag as a trailing comment.

### Verified, no change

- **Major: installer not visible in the diff** (xAI). The four sha256 values in `supply-chain/policy.toml` match the upstream cargo-deny 0.20.2 release checksum files for every target. The installer rejects a checksum mismatch and extracts exactly one executable.
- **Container digests.** All five digests in `Cross.toml` and `node/Dockerfile` resolve on their registries and equal the current digest of the named tag.
- **Major: advisory exception metadata does not match the stated policy** (xAI). The owner, package, exact versions, and review date for each exception live in `supply-chain/policy.toml`. The policy validator requires the `deny.toml` ignore ids to match that file exactly, requires valid semver versions, and rejects a review date that has passed. The truncated diff hid this file.

### Not changed

- **Major: use tags instead of commit SHAs** (Bedrock). Immutable pins are the purpose of this PR.
- **Major: redundant version on the git dependency** (OpenRouter). Cargo does not ignore the `version` field on a git dependency. It checks the resolved crate version against it.
- **Major: build-script allowlist is exact-version and broad** (xAI, OpenRouter). This is the fail-closed policy the PR introduces. Relaxing it is a maintainer decision, not a review fix.

### Verification

- Every workflow file parses.
- `check-workflow-invariants.sh` reports all invariants hold.
- `test-release-workflows.sh` passes.
- `cargo test --locked -p supply-chain` passes 8 tests, including the action pin and workflow launcher checks.

jltatbeach (approved):

PR #403, opened 2026-09-08
fix(casper): a genesis validator that missed the ceremony pulls the approved block

spreston8:
Fixes the genesis-ceremony join race behind dev run 34153662463's amd64-docker failure (`test_validator_failure_recovery`: validator1 never reached Running in 300s). Forensics on that shard (b78263a0-s5): validator1's first bootstrap dial lost the container-start race by ~400ms (TCP refused; boot's listener came up at 19:14:38.4, the dial was 19:14:38.0), so it missed the UnapprovedBlock broadcast at 19:14:39; v2+v3 met `required_sigs=2` at 19:14:41 and the ceremony completed without it; boot's one-shot ApprovedBlock sends at 19:14:49-51 went to the peers connected then. Validator1 connected to everyone at 19:14:58 and then received nothing, forever — the GenesisValidator engine waits on pushed ceremony messages, and no further push exists.

The fix is a pull, using only existing machinery:

- `Engine` gains a defaulted `on_no_casper_tick` hook, called by the casper loop's existing no-casper arm (the same ~750ms tick that was already logging "Casper engine present but Casper not initialized yet" throughout validator1's stall).
- `GenesisValidator` overrides it with a throttled (10s) `ApprovedBlockRequest` to bootstrap. The response flows through the existing `handle_approved_block_late` recovery into `Initializing` — the path the observer engines already use, which is why the readonly in the same shard recovered (boot re-sent to it on request at 19:14:59) while validator1 hung.

Red-first: the new spec reproduces the CI sequence exactly — no ceremony message ever pushed, only no-casper ticks — and pre-fix the engine sends nothing forever; post-fix it pulls once per throttle window. The existing push-recovery spec (`transitions_to_initializing_on_late_approved_block`) covers the second half of the journey.

### DIFF
--- a/casper/src/rust/engine/engine.rs
+++ b/casper/src/rust/engine/engine.rs
@@ -57,6 +57,14 @@ pub trait Engine: Send + Sync {
         Ok(false)
     }
 
+    /// Called by the casper loop on each tick while this engine has no
+    /// Casper instance. Default: nothing. Pre-genesis engines that wait on
+    /// PUSHED ceremony messages override this to PULL: a genesis validator
+    /// whose first bootstrap dial lost the container-start race misses both
+    /// the UnapprovedBlock broadcast and the ApprovedBlock send, and no
+    /// further push ever comes — without a pull it waits forever.
+    async fn on_no_casper_tick(&self) -> Result<(), CasperError> { Ok(()) }
+
     /// Returns the casper instance as an Arc if this engine wraps one.
     /// Returns None for engines that don't have casper (NoopEngine, Initializing, etc.)
     /// The Arc allows ownership transfer and use across async boundaries.
--- a/casper/src/rust/engine/engine.rs
+++ b/casper/src/rust/engine/engine.rs
@@ -58,11 +58,8 @@ pub trait Engine: Send + Sync {
     }
 
     /// Called by the casper loop on each tick while this engine has no
-    /// Casper instance. Default: nothing. Pre-genesis engines that wait on
-    /// PUSHED ceremony messages override this to PULL: a genesis validator
-    /// whose first bootstrap dial lost the container-start race misses both
-    /// the UnapprovedBlock broadcast and the ApprovedBlock send, and no
-    /// further push ever comes — without a pull it waits forever.
+    /// Casper instance, so an engine waiting on pushed ceremony messages
+    /// can pull when the push window was missed.
     async fn on_no_casper_tick(&self) -> Result<(), CasperError> { Ok(()) }
 
     /// Returns the casper instance as an Arc if this engine wraps one.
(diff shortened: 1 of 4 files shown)


spreston8 (commented):
Reviewed at dd0381668 against dev cdf447ac1.

**HIGH**

1. Send once rather than through `send_with_retry`. `genesis_validator.rs:351` reaches `transport_layer.rs:148`, an uncapped `loop { send; Err => sleep(10s) }`, and `setup.rs:749` awaits it inline — so while bootstrap is unreachable the casper loop stops, `request_all` at `:755` with it. Measured with sends failing: the tick had not returned after 25s, three attempts in. The 10s throttle already supplies the cadence, and a single attempt also makes the comment at `:329-331` true — `send_with_retry` never returns Err on a send failure, so that branch is unreachable today.

**Low**

2. Pass `Some(true)` at `genesis_validator.rs:351`. Its agreement with the positional `trim_state: true` at `:232` currently rests on an `unwrap_or` default in another crate; every other caller is explicit.

spreston8:
> Reviewed at [dd03816](https://github.com/F1R3FLY-io/f1r3node-rust/commit/dd03816680cedd642058e7e958b48b1016c6d7a6) against dev [cdf447a](https://github.com/F1R3FLY-io/f1r3node-rust/commit/cdf447ac18710d9702a27379bce6c946f421be46).
> 
> **HIGH**
> 
> 1. Send once rather than through `send_with_retry`. `genesis_validator.rs:351` reaches `transport_layer.rs:148`, an uncapped `loop { send; Err => sleep(10s) }`, and `setup.rs:749` awaits it inline — so while bootstrap is unreachable the casper loop stops, `request_all` at `:755` with it. Measured with sends failing: the tick had not returned after 25s, three attempts in. The 10s throttle already supplies the cadence, and a single attempt also makes the comment at `:329-331` true — `send_with_retry` never returns Err on a send failure, so that branch is unreachable today.
> 
> **Low**
> 
> 2. Pass `Some(true)` at `genesis_validator.rs:351`. Its agreement with the positional `trim_state: true` at `:232` currently rests on an `unwrap_or` default in another crate; every other caller is explicit.
> 

Both addressed.

1. Confirmed and fixed red-first. The new spec (`a_failed_pull_returns_immediately_instead_of_retrying_inline`) reproduces the stall: with sends failing, the pre-fix tick was still inside `send_with_retry` when a 2s timeout fired — the uncapped loop holds the casper loop (and `request_all` with it) hostage exactly as you measured. The pull is now a single inline send built in place (no new transport surface): one `ApprovedBlockRequestProto` packet to bootstrap per throttle interval, error surfaced to the warn branch — which your review correctly noted was unreachable before. Post-fix the tick returns immediately with exactly one attempt recorded; the 10s throttle owns the retry cadence.

2. Folded into the same change: the inline packet sets `trim_state: true` literally — no cross-crate default to agree with.

PR #404, opened 2026-09-08
fix(crypto): stop PEM reader dropping base64 lines that contain KEY

jeffrey-l-turner:
read_key_pair stripped the PEM header and footer by discarding every line containing the substring KEY. A base64 body line can contain KEY by chance, so roughly one generated key per thousand lost a 48-byte chunk and failed PKCS#8 parsing. This surfaced as an intermittent failure of test_print_private_key_from_secret_roundtrips_through_read_key_pair in CI.

Filter on the five-dash armor lines instead, and add a regression test with a fixed PEM whose body line contains KEY.

### DIFF
--- a/crypto/src/rust/util/certificate_helper.rs
+++ b/crypto/src/rust/util/certificate_helper.rs
@@ -149,7 +149,8 @@ impl CertificateHelper {
         // Remove PEM headers and decode base64
         let pem_lines: Vec<&str> = pem_content
             .lines()
-            .filter(|line| !line.contains("KEY"))
+            .map(str::trim)
+            .filter(|line| !line.is_empty() && !line.starts_with("-----"))
             .collect();
         let base64_content = pem_lines.join("");
 
--- a/crypto/tests/util/certificate_helper_spec.rs
+++ b/crypto/tests/util/certificate_helper_spec.rs
@@ -302,6 +302,28 @@ fn test_print_private_key_from_secret_roundtrips_through_read_key_pair() {
 }
 
 #[test]
+fn test_read_key_pair_keeps_base64_body_lines_containing_key() {
+    let pem = "-----BEGIN PRIVATE KEY-----\n\
+MIGHAgEAMBMGByqGSM49AgEGCCqGSM49AwEHBG0wawIBAQQgBTKEY6yp79c5dbNy\n\
+hUWSBmo9phQ452/PK1BotDnn9MihRANCAAQZMJSYyZI8LFWGt9m1i3iG32JY+rFv\n\
+tWB90bq4vpoJJMbKQl6ZoXhmeTtpOFxgJ9KMY7lfUz+MKa88M8XSn7Y5\n\
+-----END PRIVATE KEY-----";
+    assert!(pem
+        .lines()
+        .any(|line| !line.starts_with("-----") && line.contains("KEY")));
+
+    let temp_dir = tempfile::tempdir().expect("failed to create temp dir");
+    let key_path = temp_dir.path().join("key.pem");
+    std::fs::write(&key_path, pem).expect("failed to write key file");
+
+    let (parsed_secret, _) = CertificateHelper::read_key_pair(key_path.to_str().unwrap())
+        .expect("body lines containing KEY must not be dropped");
+    let reprinted = CertificatePrinter::print_private_key_from_secret(&parsed_secret)
+        .expect("private key printing should succeed");
+    assert_eq!(reprinted, pem);
+}
+
+#[test]
 fn test_read_key_pair_error_paths() {
     let missing = CertificateHelper::read_key_pair("/nonexistent/path/key.pem");
     assert!(missing.is_err());


jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| anthropic | `claude-fable-5` | Approve | 0.92 |
| openai | `gpt-5.6-sol` | Approve | 0.98 |
| xai | `grok-4.5` | Approve | 0.93 |

---

### Summary

This is a correct and well-scoped fix for a genuine, subtle bug: filtering PEM lines by the substring 'KEY' could randomly drop base64 body lines that happen to contain 'KEY', corrupting roughly one in a thousand generated keys. The new filter on trimmed five-dash armor lines (plus empty-line removal) is the right approach and matches standard PEM structure. The regression test is well constructed — it uses a fixed PEM whose body demonstrably contains 'KEY' and asserts that assumption explicitly, so the test can't silently become vacuous. Only minor suggestions apply: consider a dedicated PEM crate for full robustness, and note that the exact-string round-trip assertion in the test couples the regression test to printer formatting. Safe to merge.

The change correctly prevents valid base64 body lines containing the substring "KEY" from being discarded, while filtering PEM armor lines after trimming whitespace. Since the standard base64 alphabet cannot produce a body line beginning with five dashes, the revised filter is appropriate. The regression test directly exercises the reported failure with a fixed valid key and verifies successful parsing and round-tripping. No security, logic, performance, or significant style issues were identified.

This is a clear, well-scoped fix for a real intermittent PEM parsing bug: dropping any line containing KEY could strip valid base64 and break PKCS#8 parsing roughly once per thousand keys. Switching to trim + drop empty lines and five-dash armor lines is the right approach for unencrypted PEM bodies, and the added regression test with a fixed KEY-containing body line directly locks the failure mode. No security regressions introduced; safe to merge as-is, with only optional longer-term PEM-parsing hardening notes.

The fix correctly addresses the intermittent bug: filtering PEM content by five-dash armor lines instead of the substring 'KEY' prevents base64 body lines that happen to contain 'KEY' from being dropped. Trimming lines and skipping empty ones also makes the parser more robust to whitespace variations. The regression test uses a fixed PEM that provably contains 'KEY' in its body, includes a precondition sanity check, and verifies a full parse-and-reprint roundtrip. No security, logic, or performance concerns were found; the change is safe to merge.

The code change effectively addresses the intermittent failure by correctly filtering PEM lines. The addition of a regression test ensures that the issue is covered. Minor improvements in test function naming could enhance readability.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Consider using a dedicated PEM parsing crate for robustness</b> (Reported by: anthropic, bedrock, xai)</summary>

**File:** `crypto/src/rust/util/certificate_helper.rs` (line 153)

Any trimmed line starting with '-----' is discarded, including malformed or mismatched armor lines (e.g. only a BEGIN with no END, or a mid-file '-----' garbage line). This is acceptable since the subsequent base64/PKCS#8 parsing will reject truly corrupt input, but it means the reader silently accepts structurally invalid PEM.

---

Filtering any line that starts with '-----' after trim correctly preserves base64 body lines that contain the substring KEY, and '-----' cannot appear in standard base64, so body lines will not be dropped. Non-armor header lines (e.g. Proc-Type / DEK-Info on encrypted PEMs) would still be joined into the base64 payload and fail decode; that is pre-existing behavior and outside the scope of this fix if only unencrypted PKCS#8 PEMs are supported.

---

The change to filter PEM lines correctly by checking for the five-dash armor lines is a good fix for the intermittent failure. However, ensure that this does not introduce any new edge cases where valid PEM content might be incorrectly filtered out.

---

The hand-rolled filter now correctly handles the armor lines and blank lines, but it still won't handle other PEM structures such as RFC 1421 headers (e.g. 'Proc-Type:', 'DEK-Info:' in encrypted PEM) or multiple concatenated PEM blocks in one file. If any of those forms could ever reach this reader, the base64 decode would fail with a confusing error. A crate like `pem` or `rustls-pemfile` handles all these cases and eliminates this class of bug entirely.

**Suggestion:**
```
Consider adding additional tests to cover various edge cases for PEM content.
```

</details>

<details>
<summary><b>:orange_circle: Major: Strong regression test; watch exact PEM formatting</b> (Reported by: bedrock, xai)</summary>

**File:** `crypto/tests/util/certificate_helper_spec.rs` (line 304)

The fixture correctly forces a body line containing KEY and asserts round-trip equality with the printer output. That is excellent for this intermittent bug. Exact string equality can be brittle if print_private_key_from_secret ever changes line wrapping or trailing newlines.

---

The test function name `test_read_key_pair_keeps_base64_body_lines_containing_key` is quite long. Consider shortening it while maintaining clarity.

**Suggestion:**
```
If printer formatting drifts later, assert parse success plus equality of the decoded secret (or normalize PEM before compare) rather than full text equality only.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Test self-check assertion duplicates production logic</b> (Reported by: openrouter)</summary>

**File:** `crypto/tests/util/certificate_helper_spec.rs` (line 308)

The assertion that the fixture contains a body line with 'KEY' re-implements the production filter (starts_with("-----")). If the production filter changes, this precondition could silently pass or fail misleadingly. It is harmless and arguably useful documentation, but a brief comment explaining its purpose would help future readers.

**Suggestion:**
```
Add a comment like '// Sanity check: the fixture actually exercises the regression (a body line containing KEY).'
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Exact string round-trip assertion may be brittle</b> (Reported by: anthropic)</summary>

**File:** `crypto/tests/util/certificate_helper_spec.rs` (line 322)

`assert_eq!(reprinted, pem)` requires `print_private_key_from_secret` to produce byte-identical output to the hardcoded fixture (same line wrapping, no trailing newline). If the printer's formatting ever changes (e.g. adds a trailing newline or different wrap width), this test will fail even though the fix under test still works. The core regression being guarded is that parsing succeeds, which is already covered by the earlier `.expect(...)`.

**Suggestion:**
```
Consider asserting on the parsed secret's bytes (or re-parsing the reprinted PEM and comparing secrets) instead of exact PEM string equality, to decouple the regression test from printer formatting details.
```

</details>
---

<details>

This is a correct and well-scoped fix for a genuine, subtle bug: filtering PEM lines by the substring 'KEY' could randomly drop base64 body lines that happen to contain 'KEY', corrupting roughly one in a thousand generated keys. The new filter on trimmed five-dash armor lines (plus empty-line removal) is the right approach and matches standard PEM structure. The regression test is well constructed — it uses a fixed PEM whose body demonstrably contains 'KEY' and asserts that assumption explicitly, so the test can't silently become vacuous. Only minor suggestions apply: consider a dedicated PEM crate for full robustness, and note that the exact-string round-trip assertion in the test couples the regression test to printer formatting. Safe to merge.

The code change effectively addresses the intermittent failure by correctly filtering PEM lines. The addition of a regression test ensures that the issue is covered. Minor improvements in test function naming could enhance readability.

#### openai (gpt-5.6-sol)

The change correctly prevents valid base64 body lines containing the substring "KEY" from being discarded, while filtering PEM armor lines after trimming whitespace. Since the standard base64 alphabet cannot produce a body line beginning with five dashes, the revised filter is appropriate. The regression test directly exercises the reported failure with a fixed valid key and verifies successful parsing and round-tripping. No security, logic, performance, or significant style issues were identified.

The fix correctly addresses the intermittent bug: filtering PEM content by five-dash armor lines instead of the substring 'KEY' prevents base64 body lines that happen to contain 'KEY' from being dropped. Trimming lines and skipping empty ones also makes the parser more robust to whitespace variations. The regression test uses a fixed PEM that provably contains 'KEY' in its body, includes a precondition sanity check, and verifies a full parse-and-reprint roundtrip. No security, logic, or performance concerns were found; the change is safe to merge.

#### xai (grok-4.5)

This is a clear, well-scoped fix for a real intermittent PEM parsing bug: dropping any line containing KEY could strip valid base64 and break PKCS#8 parsing roughly once per thousand keys. Switching to trim + drop empty lines and five-dash armor lines is the right approach for unencrypted PEM bodies, and the added regression test with a fixed KEY-containing body line directly locks the failure mode. No security regressions introduced; safe to merge as-is, with only optional longer-term PEM-parsing hardening notes.
</details>

---

jltatbeach (approved):

Issue #407, opened 2026-09-09
Replay of a compute-heavy deploy stalls block receipt on every replica for its full execution time


spreston8:
Replay of a received block re-executes its deploys while holding the ReplayLock, so one compute-heavy deploy stalls block receipt on every replica simultaneously for its full execution time. Phlo bounds the work, but nothing bounds the wall-clock a replay may pin the pipeline for — any user who can afford the phlo can freeze the shard's receipt path.

Observed in CI run 34191496437 (arm64-subprocess, `test_network_converges_after_slow_deploy`):

- A 20M-phlo loop deploy took **271.9s** to execute on the proposer under runner load (`Propose timing: propose_core_ms=271894`; the test budgets ~25s for it).
- The moment the carrier block was broadcast, all four receiving nodes entered replay of the same loop and none finished: `Computed parents post state` logged, `Block replayed` never; every node's last `processing finished` is within 100ms of entering the replay (06:06:52).
- Block receipt stayed dead shard-wide from then on. The four replays ballooned RSS until the test harness's host-protection guardian killed the shard at 16.5GB total.
- Consensus consequence: the proposer kept a full view, but the wedged replicas could only mint solo-chain blocks on frozen views, so no finality clique could re-form.

Relation to #368: the deploy-play-budget bounds cumulative play time during block creation, but it selects between deploys and cannot preempt a single deploy mid-execution (which is how the propose above still ran 271.9s), and replay applies no budget at all, so the included carrier pins every replica for the full re-execution.

Artifact: `integration-logs-arm64-subprocess-run34191496437-attempt1`, session c61849e3, `log-archive/c61849e3/shard2/` (the convergence shard; the guardian kill line is in the failing test's captured log).

Co-Authored-By: Claude <noreply@anthropic.com>

spreston8:
Field evidence from PR #415's CI forensics (run 34502938853, arm64-subprocess, the `test_concurrent_bridge_locks` shard).

_Corrected 2026-09-14 after re-reading the logs. The earlier version blamed expensive merge validation after a reorg, and used a block interval where a height interval belongs._

- **Receipt stalled behind replay.** Consensus replay takes a single-permit lock (`ReplayLock`), so while one lock-carrying block replayed, no other block finished processing on any replica. There were three episodes, each on all three validators:
  - #7 (11 locks) replayed in 44 s;
  - #15 (1 lock) replayed in 77 s;
[...]

PR #415, opened 2026-09-10
Anchor-crossing walks and the recovery-purge race

### LINKED ISSUE #412: Parent-selection exclusion has no re-admission path: a transiently lagging validator is orphaned permanently
Once a validator's tips fall outside the citability window, peers stop citing its blocks as parents, its blocks are orphaned (ft −1), and nothing ever re-admits it: its subsequent blocks extend a chain that is permanently outside the window. A transient slowdown becomes permanent removal of that validator's stake from the finality clique — without slashing, and without recovery.

Observed live on rust-shard-rgb-test (3 equal stakes, FTT 1/3 — every finalization needs all three validators): a replay-throughput collapse at ~05:39 UTC 2026-09-09 (empty-block replays 1s → 16–21s) lagged validator3; by #117650 the other two stopped citing its tips (its blocks ft −1.0 while it was still merging all three chains), by #117730 it degraded to a solo chain, and at #117767 the remaining pair split too — three mutually-aware solo chains (each cites the others in justifications but parents only its own), finality frozen at #117766. An earlier, briefer lag episode at #117400 healed only because it ended before exclusion set in.

Fix shape needs design: a lagging validator whose fresh blocks re-join the frontier should become citable again (re-admission through the window), or recovery proposals should be built to bridge an excluded validator back into the clique.

Same downstream shape as #407 (processing lag converts to island formation and finality loss) with a different trigger: there a single compute-heavy deploy, here ambient state-size-driven replay slowness.

spreston8:
Fixes the two failure modes that let a validator's dependency recovery die and its walks wedge on blocks below a restore anchor:

- `BlockNotHeld` carries a typed `MissingBlockContext` — the accessor that tripped, plus the caller backtrace when captured at the storage boundary. `Display` owns the separator, so a site cannot run the message together. The heartbeat check solicits the named block instead of erroring every cycle, and parent expansion above a block number takes a typed `UnheldParent` policy — `Surface` for verdict walks, `SkipSettled` for live-decision walks (the snapshot's deploys-in-scope BFS).
- `check_if_of_interest` returns a typed `OfInterestVerdict`; only verdicts about the block itself (wrong shard/version, old unsolicited) purge the buffer. `AlreadyProcessed` judges the copy: it acks without purging, so a racing duplicate no longer destroys the recovery state of a block buffered against a missing dependency.

Closes #412 — the filed no-re-admission gap is disproven: re-entry is inherent once blocks process; the real blocker was the duplicate-drop purge, fixed here. Sustained-lag exposure is tracked as #407. Remaining `BlockNotHeld` sites are named by the error itself; no upfront accessor audit retained.


spreston8 (commented):
Reviewed at 34c59c54e against its dev merge-base 343009a4a.

**Medium**

1. Move the purge-race half of `39ed0f678`'s spec down into this PR. It reproduces the race `34c59c54e` fixes, but sits on `fix/recovery-cap-geometry` (#416), stacked above this one — so merging here lands a race fix in `dev` with no regression coverage. The height-gate half can stay in #416, and each PR then carries its own proof.

**Low**

2. The `site` strings carry their own leading space (`" [get_parents_metadata]"`) and `Display` appends them directly, so a site written without one runs the message together. Worth owning the separator at the display site until the declared typed payload replaces the `String`.

PR #416, opened 2026-09-10
Recovery liveness: cap geometry, height-gate mint, and work-paced frontier follow

### LINKED ISSUE #408: Empty-frontier width cap never binds: stale-recovery exemption interval is shorter than the heartbeat tick
The empty-frontier width cap (`empty-frontier-max-unfinalized-blocks`) can never bind during a finalization stall, because its per-validator exemption interval is shorter than the heartbeat tick: `stale-recovery-min-interval` defaults to **3s** while the heartbeat check interval is **~5s**. A validator that last minted a block one tick ago has always been "idle for a full stale-recovery interval" by the next tick, so `self_recovery_throttled` is false at every check, `empty_frontier_pressure` short-circuits to no-pressure, and the documented bound — "one recovery proposal per stale-recovery interval through the cap" — degenerates to one per tick, i.e. unlimited.

Observed in CI run 34191496437 (arm64-subprocess): during a finalization stall, validator1 minted 20+ empty blocks at tick cadence while `Heartbeat: Empty frontier backpressure active (unfinalized_blocks=13..76, cap=12)` printed between its own proposals — the cap suppressed nothing that mattered, and unfinalized width grew ~36 blocks in two minutes across the three validators.

Fix shape: a startup inequality in the I-series family — `stale-recovery-min-interval` must exceed the heartbeat `check-interval` (warn or fail, matching the existing I3/I4 treatment), or the exemption should be keyed to the stale-recovery cadence independently of the tick so the "one per interval" bound is real. The `defaults.conf` doc text should match whichever behavior is chosen.

Relevant code: `node/src/rust/instances/heartbeat_proposer.rs` (`empty_frontier_pressure`, `self_minted_within_recovery_interval`, `stale_recovery_window_is_open`); defaults in `node/src/main/resources/defaults.conf`.

### LINKED ISSUE #411: Stale-recovery proposals are gated by the height constraint they exist to escape; crossing the threshold is unrecoverable
When finalization stalls, the heartbeat's stale-recovery lane proposes "until the clique re-forms" — but those proposals run through `last_finalized_height_constraint_checker` like any other. Once tip−floor exceeds the on-chain `height_constraint_threshold`, every recovery proposal fails with "too far ahead of the last finalized block": no proposal can land, so no new justifications, so the clique can never re-form and the floor can never advance. Because the threshold lives in on-chain state (changing it needs a deploy, which needs a propose, which is blocked) there is no operator escape — the only exit is a chain reset.

Observed live on rust-shard-rgb-test (2026-09-09): finalization froze at #117766 ~06:03 UTC, stale-recovery grew the tip to floor+1001 over 5 hours, the gate first fired at 11:05:03, and the shard has been in a self-sustaining deadlock since — all five nodes healthy, `isReady=true`, identical LFB, every propose failing every 5s.

Fix shape: the recovery lane must not be gated by the constraint it exists to escape — exempt recovery proposals from the height check (or bound them separately), so a stalled shard can always mint the justifications that re-form the clique.

Related: #408 — the empty-frontier cap never binding is what let tip−floor grow past the threshold in the first place; with a binding cap the shard would have parked on the recoverable side of the gate.

spreston8:
Stacked on #415. Closes the finality-stall class where a stalled shard mints per-tick recovery blocks into an on-chain height gate that then blocks recovery itself.

- `stale-recovery-min-interval` is derived as 1.5 × `check-interval` (7.5s at the shipped 5s tick) instead of a 3s literal. At or below the tick, the empty-frontier cap's per-validator exemption opens every tick and the cap never binds. The gate is evaluated on ticks, so 1.5 ticks opens it on the second tick after a validator's own block, with half a tick of margin against timing jitter. An explicit override at or below the tick is a hard config error (I5). Closes #408.
- Frontier-follow no longer reads that interval: it paces on whether the unfinalized cone still holds user work — a deploy awaiting finalization, or a rejection record whose deploy cannot retry until the floor reaches it. The interval had gated this lane too, which paced the committee through retry waits, where every arriving block is empty and only the floor's advance ends the wait. Measured on the previous pin of this branch: 6.9 s/height against 4.4 before it, and phase 4 of `test_validator_lifecycle` finalizing at 164s against a 135s budget. An A/B of the lane alone moved the median back from 1.60 to 1.07 s/height with the pending lane unchanged. Idle liveness stays with the stale-recovery lane, still paced by the interval, and the width cap gates this lane independently.
- The height constraint no longer gates the heartbeat recovery lane it exists to be rescued by: on `TooFarAheadOfLastFinalized`, a heartbeat propose mints an empty recovery block. Manual proposals keep the gate. A typed `DeploySelection` (Standard / StandardAllowEmpty / RecoveryEmpty) replaces the allow-empty bool through the creator. Closes #411.
- The certification walk treats a block this node does not hold as a typed deferral, never as settled. A spec builds the same DAG with full and truncated history: where the walk crosses the restore anchor, the full node decides and the restored node names the block it lacks instead of deciding differently.
- Production `debug_assert!` sites retired: the propose-path slash-candidate contract pin is a typed error, the merge-order alarm rests on its always-on error log and metric, and the caller-discharged domain asserts are deleted with their discharge documented.


spreston8 (commented):
Reviewed at 5f05e255c against its base fix/anchor-crossing-walks (34c59c54e).

**HIGH**

1. Gate `settled_floor` on `anchor <= floor` at `clique_oracle.rs:627` instead of asserting it in `floor.rs:370`. The clipping change rests on that invariant, and `#[cfg(debug_assertions)]` compiles out of release — which is what production, `ci.yml:532`/`:610` and the pre-push hook all run, so nothing evaluates it. Violated, unheld blocks above the floor count as settled agreement and the node over-certifies silently; the monotonicity argument covers node-local floor differences, not a broken invariant. `runtime.rs:1341` and `slashing_authorization.rs:407` both reject `debug_assert!` for this reason, the first working around the assert already in this file. Gating degrades a violation to the existing θ<0 typed error.

**Medium**

2. Make I5 a hard error. `configuration/mod.rs:349` warns that the width cap "never binds", while I3 returns `Err` for a cap that "cannot stop validity-window burn during a stall, which is its only job" — the same failure, and it binds harder here, since the recovery lane no longer stops at the height constraint.

**Low**

3. Pre-existing, not this PR's to fix: the other five production `debug_assert!` sites — `clique_oracle.rs:79`, `block_creator.rs:1718`, `conflict_set_merger.rs:680`, `reduce.rs:7365` and `:7393` — share the same non-coverage. Worth a sweep converting the load-bearing ones to typed errors and deleting the documentation-only ones, starting with `clique_oracle.rs:79`, whose non-firing is why `runtime.rs` carries a compensating check at all.

PR #418, opened 2026-09-10
fix(soak): read the system-integration pin from the target tree

jeffrey-l-turner:
The merge-recovery soak workflow file always runs from the default branch (GitHub schedules and the OCI dispatcher both use it), but the nightly target is dev. The integration suite checkout used the workflow file's own `SYSTEM_INTEGRATION_REF`, so master's pin was paired with dev's node. After the 2026-09-08 dev-only repin, every nightly preflight failed `test_observer_exploratory_overload_recovers` (runs 34303543830 and 34450109213), and only a dev-to-master promotion could clear it. The three-site pin-drift check cannot catch this: it proves agreement within one commit, not across the workflow branch and the target branch.

- soak job: new step after "Resolve target revision" reads `SYSTEM_INTEGRATION_REF` from `node-under-test/.github/oci-validation.env` (the same file `release-evidence.sh` reads), accepts only a 40-character SHA or `refs/tags/<tag>`, and fails closed with a named cause when the file or key is missing
- "Clone trusted integration harness" checks out that pin; a notice reports when it differs from the workflow literal
- "Clone trusted runner launcher" stays on the workflow literal on purpose: that job holds OCI secrets and runs before any target checkout
- pin-site comment documents the two roles; `docs/release-process.md` states that the suite follows the target tree's pin

Verification: YAML parse, `bash -n` on the step, replay against the dev tree (emits `962effd…`), replay with a missing key, a bad value, and a missing file (each exits 1 with `::error::`), `check-workflow-invariants.sh`, `test-release-workflows.sh`, `test-system-integration-node-capabilities.sh`, and the CI pin-drift regex still reading the literal first.

Effect: once master carries this file, dev and master each preflight against their own pin with no promotion required. PR #410's cron-fallback yield is workflow logic and reaches the nightly through the same one-time promotion.

https://claude.ai/code/session_01UKMVztRB2Fw6pvMxrAMaLN


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :white_check_mark: Approved (60.0000% agreement)

**Reviewed by:**

---

### Summary

This is a well-reasoned fix for a real cross-branch pin-drift failure: the integration suite the soak runs will now follow the system-integration pin recorded in the target tree's `.github/oci-validation.env` rather than the workflow file's literal, with strict fail-closed validation (40-char SHA or refs/tags only), safe env-var passing of the workflow pin, no shell-injection surface into GITHUB_OUTPUT, and the secrets-holding launcher job deliberately kept on the workflow literal. The main point deserving human attention is the trust shift: the harness commit executed by this job is now selected by whoever controls the target ref, so reviewers should confirm no secrets are reachable by harness code after this checkout, and consider disallowing mutable tag refs. Remaining findings are minor robustness/consistency items in the env-file parsing (quotes, duplicate keys, lowercase-only SHA regex) and small documentation polish. Safe to merge once the trust-boundary question is confirmed.

The change correctly derives the integration-suite revision from the checked-out target tree, validates the value before passing it to `actions/checkout`, and documents the distinction between the launcher and suite pins. No significant security or performance problems are evident. The main improvement is to reject duplicate key declarations rather than silently choosing the first, preserving unambiguous agreement among consumers of the env file.

This is a sound fix for real cross-branch pin drift: the soak job now reads SYSTEM_INTEGRATION_REF from the target tree’s .github/oci-validation.env, validates fail-closed to a full SHA or refs/tags/…, and only rebinds the integration harness checkout while leaving the secrets-bearing launcher on the workflow literal. Hardcoded repository, persist-credentials: false, and the tight ref regex keep the security model intact (target tree can only select an existing system-integration SHA/tag, not arbitrary repos or branch names). Docs and comments match the two-role pin design. No blocking security, logic, or performance issues; only minor robustness nits around SHA case and quoted values.

The change correctly addresses the pin drift between the workflow branch and target branch by reading the integration pin from the target tree's oci-validation.env. The step uses strict bash settings, validates the pin format (SHA or tag), and fails closed on missing/invalid data. The runner launcher appropriately stays on the workflow literal since it runs before checkout with secrets. No security, logic, or performance issues found.

The code change looks good with minor improvements needed in the documentation for clarity. The new step to read the system-integration pin from the target tree is well-implemented and addresses the issue effectively.

---

### Issues Found

<details>
<summary><b>:orange_circle: Major: Trust boundary shift: harness pin is now controlled by the target tree</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1627)

Previously the integration harness checkout was pinned by a literal in the workflow file, which always executes from the default branch (an implicit trust anchor). Now anyone who can modify `.github/oci-validation.env` on any target branch (or a candidate tag selected via workflow_dispatch) selects which system-integration commit is checked out and executed in this job. The regex constrains the value to a SHA or tag ref, but checkout-by-SHA against `F1R3FLY-io/system-integration` can potentially fetch commits pushed anywhere in that repo's fork network, and `refs/tags/<tag>` refs are mutable. The mitigations here are reasonable — the runner launcher holding OCI secrets deliberately stays on the workflow literal, `persist-credentials: false` is set, and modifying the target tree already implies control over the node code under test — but the change should be reviewed with the question: does the harness job (or any later step in this job) have access to secrets or credentials that a malicious harness commit could exfiltrate?

**Suggestion:**
```
Confirm no secrets are exposed to steps that execute harness code after this checkout. Consider restricting the pin to a 40-character SHA only (dropping the `refs/tags/` form) to remove the mutable-tag vector, since the repin script presumably writes SHAs.
```

</details>

<details>
<summary><b>:orange_circle: Major: Parsing semantics may diverge from release-evidence.sh reading the same file</b> (Reported by: anthropic, openai, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1638)

Parsing is cut-after-first-equals plus trailing-whitespace strip. Values like SYSTEM_INTEGRATION_REF="abc…" or with single quotes would not match the SHA/tag regex and would fail closed. That is safe, and matches a strict fail-closed posture, as long as release-evidence.sh and repin scripts agree on an unquoted format.

---

The `grep | head -1` pipeline accepts an env file containing multiple `SYSTEM_INTEGRATION_REF` declarations and silently selects the first. Other env-file consumers may apply the last assignment, so a malformed file could make the soak use a different pin from release evidence or other tooling. This weakens the intended fail-closed behavior.

---

The pin guard requires ^[0-9a-f]{40}$, so an uppercase or mixed-case SHA in the target tree's oci-validation.env would fail closed even though git treats it as the same object. Convention may already be lowercase (and your dev replay suggests that), but the asymmetry is worth knowing if humans ever hand-edit the pin.

---

The step parses `oci-validation.env` with `grep | head -1 | cut -d= -f2-` and strips only trailing whitespace. If `release-evidence.sh` sources the file as shell (as .env files often are), the semantics differ: quoted values (`SYSTEM_INTEGRATION_REF="<sha>"`) would be valid when sourced but fail this step's regex, and duplicate assignments resolve to the *first* occurrence here versus the *last* when sourced. Both failure modes are fail-closed, which is good, but they can produce surprising nightly failures if the env file format drifts, and the two readers could silently disagree on duplicates.

---

`^[0-9a-f]{40}$` only accepts lowercase SHAs. Git and GitHub tooling emit lowercase, so this is fail-closed rather than a bug, but a hand-edited or tool-normalized uppercase/mixed-case SHA in the env file would break the nightly with a somewhat misleading error ('must be a 40-character commit SHA' when it is one, just uppercased).

**Suggestion:**
```
Either accept `[0-9a-fA-F]{40}` and normalize with `tr '[:upper:]' '[:lower:]'`, or leave as-is and make the error message state that lowercase hex is required.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Minor documentation update</b> (Reported by: bedrock)</summary>

**File:** `docs/release-process.md` (line 642)

The documentation update in `docs/release-process.md` could be more concise. Consider removing redundant phrases for clarity.

**Suggestion:**
```
Remove the phrase 'The integration suite follows the system-integration pin in the target tree's `.github/oci-validation.env`, not the pin in the workflow file' and replace it with 'The integration suite follows the target tree's pin.'
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Pin-site comment says 'keep identical to the other two' but the sites now serve different roles</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 149)

The pre-existing comment 'Third pin site; keep identical to the other two — bump via scripts/repin-system-integration.sh' now sits directly above a new paragraph explaining the literal no longer binds the suite. Since dev and master can now legitimately carry different pins across branches (the stated goal), the 'keep identical' instruction is only true within a single commit. The new paragraph explains this, but the juxtaposition is easy to misread during a repin.

**Suggestion:**
```
Reword the first line to 'Third pin site; keep identical to the other two *within this commit* — cross-branch drift is expected and handled by the target-tree read below.'
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Extraction pipeline is dense; a comment on the subshell/pipefail interaction would help</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1638)

The line `target_pin="$( (grep ... || true) | head -1 | cut ... | sed ...)"` packs the pipefail workaround (`|| true` inside a subshell so a missing key doesn't kill the pipeline under `-o pipefail`) into one line. Future maintainers may 'simplify' it and reintroduce a hard failure on a missing key with no `::error::` message.

**Suggestion:**
```
Add a one-line comment noting the `|| true` exists so a missing key falls through to the validation error below instead of dying silently under pipefail.
```

</details>
---

<details>

This is a well-reasoned fix for a real cross-branch pin-drift failure: the integration suite the soak runs will now follow the system-integration pin recorded in the target tree's `.github/oci-validation.env` rather than the workflow file's literal, with strict fail-closed validation (40-char SHA or refs/tags only), safe env-var passing of the workflow pin, no shell-injection surface into GITHUB_OUTPUT, and the secrets-holding launcher job deliberately kept on the workflow literal. The main point deserving human attention is the trust shift: the harness commit executed by this job is now selected by whoever controls the target ref, so reviewers should confirm no secrets are reachable by harness code after this checkout, and consider disallowing mutable tag refs. Remaining findings are minor robustness/consistency items in the env-file parsing (quotes, duplicate keys, lowercase-only SHA regex) and small documentation polish. Safe to merge once the trust-boundary question is confirmed.

The code change looks good with minor improvements needed in the documentation for clarity. The new step to read the system-integration pin from the target tree is well-implemented and addresses the issue effectively.

#### openai (gpt-5.6-sol)

The change correctly derives the integration-suite revision from the checked-out target tree, validates the value before passing it to `actions/checkout`, and documents the distinction between the launcher and suite pins. No significant security or performance problems are evident. The main improvement is to reject duplicate key declarations rather than silently choosing the first, preserving unambiguous agreement among consumers of the env file.

The change correctly addresses the pin drift between the workflow branch and target branch by reading the integration pin from the target tree's oci-validation.env. The step uses strict bash settings, validates the pin format (SHA or tag), and fails closed on missing/invalid data. The runner launcher appropriately stays on the workflow literal since it runs before checkout with secrets. No security, logic, or performance issues found.

#### xai (grok-4.5)

This is a sound fix for real cross-branch pin drift: the soak job now reads SYSTEM_INTEGRATION_REF from the target tree’s .github/oci-validation.env, validates fail-closed to a full SHA or refs/tags/…, and only rebinds the integration harness checkout while leaving the secrets-bearing launcher on the workflow literal. Hardcoded repository, persist-credentials: false, and the tight ref regex keep the security model intact (target tree can only select an existing system-integration SHA/tag, not arbitrary repos or branch names). Docs and comments match the two-role pin design. No blocking security, logic, or performance issues; only minor robustness nits around SHA case and quoted values.
</details>

---

jeffrey-l-turner:
## Review remediation

Disposition of the multi-agent review findings. Commit 669a7b344 carries the code change.

### Major: env-file parsing may diverge from other readers (anthropic, openai, xai)

**Fixed in part.** The pin step now counts the `SYSTEM_INTEGRATION_REF` declarations and fails closed unless the count is exactly one. `scripts/repin-system-integration.sh` already rejects a file with more than one declaration, so both readers now agree. The error text names the lowercase requirement.

Not changed, by design:

- Quoted values fail closed in this step and in the repin script. The env file header requires an unquoted lowercase SHA.
- An uppercase SHA fails closed. The repin script enforces the same regex, and every tool that writes this file emits lowercase.

Replay of the extracted step against nine fixture env files:

| Case | Result |
|---|---|
| dev tree, tag ref, pin equal to the workflow literal | exit 0, correct output |
| duplicate key, missing key, bad value, uppercase, quoted, no file | exit 1 with a named cause |

### Major: trust boundary shift, the target tree selects the harness commit (anthropic)

**Answered, no change.** A trace of secret scope in the soak job:

- The OCIR secrets are step-scoped to "Resolve candidate image by digest". That step runs only in candidate mode and calls the trusted ci-control checkout and the GitHub CLI. It runs no harness code.
- Every later step that runs system-integration code sees only the default job token.
- The OCI and GitHub App secrets live in the launcher job, which stays on the workflow literal.
- The target tree already controls the node binary under test, so selecting the harness commit adds no new authority.

Tag refs stay tolerated for the emergency path, matching the launcher and the env file policy.

### Suggestions

- Extraction pipeline density: a comment now explains why `|| true` sits inside the subshell.
- "Keep identical to the other two" comment at the pin site: unchanged. The pin-drift check still requires agreement within one commit, and the paragraph below it explains the cross-branch case.
- `docs/release-process.md` wording: unchanged.

https://claude.ai/code/session_01UKMVztRB2Fw6pvMxrAMaLN

PR #419, opened 2026-09-11
fix(soak): read the system-integration pin from the target tree on ma…

jeffrey-l-turner:
Hotfix off master that ports two dev-only workflow changes so the nightly soak, which always runs master's workflow file, preflights dev against dev's own pin without a dev-to-master promotion.

- Port #418: a pin step reads SYSTEM_INTEGRATION_REF from the target tree's .github/oci-validation.env, validates a SHA or tag ref, and fails closed. The trusted harness checkout uses the step output. The trusted runner launcher and the CI drift check keep the workflow literal on purpose. A duplicate key in the env file is rejected.
- Port #410: the cron fallback yields to an OCI dispatch whose integration preflight already produced a failed test verdict on the same target for the same slot.
- Document both behaviors in the release-process workflow table.

Master's own pin literal is unchanged. Verified with the workflow invariants check, the release workflow tests, the node capability tests, and a replay of the pin step against dev's and master's env files.

Claude-Session: https://claude.ai/code/session_01UKMVztRB2Fw6pvMxrAMaLN


jeffrey-l-turner:
## Multi-Agent Code Review

**Verdict:** :mag: Needs Review (40.0000% agreement)

**Reviewed by:**

| openai | `gpt-5.6-sol` | Needs Review | 0.98 |
| openrouter | `moonshotai/kimi-k3` | Feedback Provided | 0.72 |
| xai | `grok-4.5` | Approve | 0.88 |

---

### Summary

The change is careful, well-commented workflow engineering: the pin-read step validates the target-tree ref strictly (safe charset, fail-closed on missing/duplicate/malformed keys, no output-injection surface), and the cron-fallback yield logic guards on run status, exact slot title, verdict description, and run-id cross-matching, failing open on query errors by explicit design. The main item needing human judgment is the deliberate trust-boundary shift: the 'trusted' integration harness checkout ref is now sourced from the dev-controlled target tree rather than the master-governed workflow literal, meaning dev write access can select any reachable commit (or mutable tag) in the system-integration repo for the harness. The PR mitigates this by keeping the secret-holding launcher on the workflow literal and disabling credential persistence, but a reviewer who knows the harness job's privileges should confirm the residual exposure is acceptable. Secondary feedback covers brittle string-matching between the finalize job and the schedule gate, confusing diagnostics for empty/duplicate env declarations, and the cost of failing open on transient API errors.

The target-tree pinning and failed-preflight fallback are otherwise carefully constrained and documented, with no evident injection or performance issue. However, the new pin parser does not fully enforce its central exactly-once declaration contract: a valid pin followed by an empty duplicate is accepted. This should be corrected before merge so the pin step genuinely fails closed and remains consistent with release-evidence consumers.

Solid, intentional hotfixes: the target-tree pin step is fail-closed (missing file, duplicate keys, non-SHA/non-tag), keeps the workflow literal for the secrets-bearing launcher/drift check, and only redirects the integration harness checkout; the cron fallback correctly requires a failed claim run plus a matching failed preflight status on the same target and fails open on API errors. No security blockers for merge—supply-chain exposure is bounded to choosing existing system-integration refs from a branch that can already change that env file. Optional follow-ups are status API pagination, hardening the description-string contract, and aligning the tag regex with the repin script.

This is a well-reasoned hotfix: the pin step fails closed on missing files, duplicate keys, and malformed refs, and the cron-fallback verdict check correctly fails open on API errors while requiring a four-part match (state, exact description, parseable run URL, membership in the failed-claim set) before yielding. The comments explaining intent and incident history are excellent. My main concerns are that accepting refs/tags from the target tree reintroduces a mutable reference into a checkout feeding a privileged pipeline (resolving the tag to a SHA before checkout would close this), and that the description-string contract between the finalize job and the schedule gate is enforced only by comments — a lint or context-name-based signal would make it self-enforcing. Neither blocks the merge given the stated urgency and the fail-open/fail-closed postures chosen, but both are worth follow-ups.

The code changes look functional and safe to merge. Minor improvements in comment formatting and documentation updates are suggested for better readability and clarity.

---

### Issues Found

<details>
<summary><b>:red_circle: Critical: Empty or mixed duplicate declarations produce confusing error paths</b> (Reported by: anthropic, openai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1630)

The duplicate check counts non-empty extracted values rather than matching declarations. Because command substitution removes trailing newlines, a file containing a valid declaration followed by `SYSTEM_INTEGRATION_REF=` (or a whitespace-only value) reduces to the valid value and produces `declaration_count=1`. The workflow therefore accepts a duplicate key despite the stated fail-closed requirement. Other consumers may interpret the final empty declaration as authoritative, creating inconsistent pin evidence.

---

declaration_count uses `grep -c .` on the extracted values, which counts non-empty lines. A file containing `SYSTEM_INTEGRATION_REF=` (empty value) reports 'found 0 declarations' even though the key is declared once, which will confuse whoever debugs it. A file with one empty and one valid declaration reports count 1, then falls through to the format check with a multi-line value and fails with a mangled 'got ...' message. Both cases fail closed (correct), but the diagnostics don't match the actual file contents.

**Suggestion:**
```
Count matching lines from the file directly (`grep -cE '^SYSTEM_INTEGRATION_REF='`) for the duplicate check, and separately validate that the single extracted value is non-empty, so each failure mode names the real problem.
```

</details>

<details>
<summary><b>:orange_circle: Major: refs/tags pin path allows mutable references into the trusted harness checkout</b> (Reported by: openrouter)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1640)

The new pin step accepts refs/tags/<tag> from the target tree's .github/oci-validation.env and feeds it directly into actions/checkout of F1R3FLY-io/system-integration. Tags are mutable: anyone who can push a tag to the system-integration repo (or force-move an existing one) can change what code executes in this workflow after the fact, and unlike a 40-hex SHA the tag target is not immutably bound at pin-read time. The workflow file runs from master and the nightly soak job chain includes jobs holding OCI secrets, so a hijacked tag could smuggle arbitrary code into a privileged context. The workflow-literal path the runner launcher uses stays SHA-pinned precisely for this reason; the target-tree path now reintroduces the mutability. If the tag escape hatch is truly needed for emergencies, consider resolving the tag to a SHA via the GitHub API immediately after reading it and checking out the resolved SHA instead, so at least the checked-out content is what was validated at that moment.

**Suggestion:**
```
After validating refs/tags/<tag>, resolve it with `gh api repos/F1R3FLY-io/system-integration/git/ref/tags/<tag> --jq .object.sha` and checkout the resolved SHA rather than the tag name.
```

</details>

<details>
<summary><b>:orange_circle: Major: Trust boundary shift: harness checkout ref is now controlled by the target tree</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1653)

The 'Clone trusted integration harness' step previously used the workflow file's literal SYSTEM_INTEGRATION_REF (governed by master). It now uses a ref read from the target tree's .github/oci-validation.env, which is typically dev. Anyone with dev write access (or the ability to land a change to that env file on dev) can select an arbitrary 40-hex commit in the F1R3FLY-io/system-integration repo — and note that GitHub makes fork-PR head commits fetchable by SHA within the base repo network, so a valid-looking SHA is not the same as a reviewed commit on that repo's protected branches. The mitigations here are real (the secret-holding runner launcher and drift check keep the workflow literal, persist-credentials is false, and the ref format is validated), but this deliberately lowers the harness's provenance bar from 'merged to master' to 'merged to dev'. This trust change should be explicitly acknowledged by a human reviewer familiar with what the harness job can access (network position, artifacts, node-under-test environment).

**Suggestion:**
```
Consider constraining the target-tree pin further, e.g., verify the SHA is an ancestor of system-integration's main branch (or of a set of allowed release branches) before using it, or require the pin to be signed/attested. Alternatively, document why arbitrary reachable SHAs in that repo are acceptable.
```

</details>

<details>
<summary><b>:orange_circle: Major: Fail-open verdict query can waste a full soak session on transient API errors</b> (Reported by: anthropic, xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 683)

If the `gh api .../commits/{ref}/status` call fails transiently (rate limit, network), verdict is emptied and the fallback soaks anyway — potentially re-running a ~2.5h session against a commit whose preflight is already known to fail, which is exactly the waste this change is meant to prevent. The fail-open choice is documented and consistent with the claim query, so this is a deliberate trade-off, but it deserves a note.

---

The preflight-verdict check calls `gh api repos/.../commits/.../status` without `--paginate`. The combined commit status endpoint can paginate when many contexts exist. If "Soak Integration Preflight" is not on the first page, `verdict_state` will be empty, the gate fails open, and the cron fallback may re-soak a commit that already failed preflight—the costly path this change is meant to avoid.

**Suggestion:**
```
Add a single retry with backoff on the status query before falling open; a transient failure here is cheap to retry and expensive to fail open on.
```

</details>

<details>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 139)

The comments explaining the pin behavior could be formatted for better readability.

**Suggestion:**
```
Consider breaking the comment into multiple lines for clarity.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Skip decision depends on exact description text matching across jobs</b> (Reported by: openrouter)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 685)

The cron fallback yields to a failed dispatch only when the commit status description is byte-identical to 'Full integration preflight completed with test failures'. The comment acknowledges this is a contract, but there is no mechanical enforcement (e.g., a shared constant or test in this repo) tying the finalize job's string to the schedule gate's string. A future wording tweak in the finalize job silently changes behavior from 'yield' to 're-soak for 2.5 hours', which is exactly the failure mode this change was built to prevent, and it would only be noticed after burning a soak window. Since the description notes runs 34303543830/34450109213 as precedent, a cheap guard (a workflow lint asserting both strings match, or a distinct context such as 'Soak Integration Preflight / test-verdict' instead of overloading the description) would make the contract self-enforcing.

**Suggestion:**
```
Add a check to the existing workflow invariants script asserting the description literal appears identically in both places, or move the verdict signal into the status context name so no free-text matching is required.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Cross-job contract via exact status-description string matching is brittle</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 691)

The schedule gate yields only when the commit status description exactly equals 'Full integration preflight completed with test failures', published ~1700 lines away in the finalize job. Both sites carry comments pointing at each other, but any future rewording, added punctuation, or trailing whitespace on either side silently breaks the contract and the fallback quietly resumes duplicate soaks — the failure mode is invisible.

**Suggestion:**
```
Encode the verdict in a machine-readable channel instead: a distinct status context (e.g. 'Soak Integration Preflight / test-verdict') or a query parameter on target_url, and keep the description free-form for humans. Alternatively, add a CI invariant check that greps both sites for the shared literal.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Verdict lookup matches any historical failed run for the slot, not the most recent</b> (Reported by: openrouter)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 695)

failed_claims collects all failed dispatch runs with the matching slot title, and grep -qx only verifies the verdict's run appears somewhere in that set. If the same slot saw an early failed dispatch (preflight failure on an older tip) and a later failed dispatch on a newer tip whose preflight actually passed, the status lookup is by current target ref tip, so the verdict read corresponds to the newest state — this mostly self-corrects, but if the target ref moved and was then force-pushed back, the verdict on the tip could correspond to a different commit than the failed run in the claim set. This is a narrow race and the fail-open posture limits the blast radius to an unnecessary re-soak, but it is worth noting the check validates set membership rather than recency.

**Suggestion:**
```
Optional: compare the verdict's target run against the most recent failed claim rather than any member of the set.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Tag pin pattern is slightly broader than typical tag names</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1631)

The fail-closed check allows `refs/tags/[A-Za-z0-9._/-]+`, including odd forms (leading `/`, consecutive slashes). Risk is low: ref is only passed to `actions/checkout` on `F1R3FLY-io/system-integration` with `persist-credentials: false`, and checkout fails if the ref does not exist. Still looser than the SHA side of the same check.

**Suggestion:**
```
If `scripts/repin-system-integration.sh` uses a tighter tag grammar, mirror that regex here so both pin sites reject the same inputs.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: Pin parsing accepts leading whitespace but rejects it later with a confusing error</b> (Reported by: openrouter)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1630)

The sed trims only trailing whitespace from the value (s/[[:space:]]*$//), so a line like 'SYSTEM_INTEGRATION_REF= 3757cab...' passes the count check and then fails the SHA/tag regex with an error quoting a value that looks valid at a glance. The env file is machine-managed by repin-system-integration.sh so this is unlikely in practice, but trimming both sides (or explicitly rejecting leading whitespace in the error message) would make the failure mode clearer.

**Suggestion:**
```
Use sed -E 's/^[[:space:]]+//; s/[[:space:]]+$//' or mention whitespace explicitly in the validation error.
```

</details>

<details>
<summary><b>:yellow_circle: Minor: refs/tags/<tag> pins are mutable</b> (Reported by: anthropic)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 1638)

The validation accepts refs/tags/<tag> as an 'emergency path'. Tags in Git are mutable — a tag can be force-moved to point at different code after the pin lands, so a tag pin does not provide the immutability guarantees of a 40-hex SHA. Combined with the trust-boundary shift above, a moved tag silently changes which harness code runs.

**Suggestion:**
```
If the emergency tag path must exist, log a prominent ::warning:: when a tag ref is used, and/or resolve the tag to a SHA at pin-read time and record it in the step summary so runs are auditable.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Table cell has become a paragraph</b> (Reported by: anthropic, openrouter)</summary>

**File:** `docs/release-process.md` (line 645)

The merge-recovery-soak.yml row's trigger cell has grown into a paragraph describing the fallback yield semantics and the pin-source behavior. It is accurate, but tables with this much prose in one cell are hard to scan and tend to drift. Consider keeping the cell to triggers and moving the fallback/pin semantics to a dedicated short subsection referenced from the table.

---

The trigger cell for merge-recovery-soak.yml now contains four sentences of behavioral documentation inside a parenthetical, which is hard to scan in a table. The behavior itself is documented accurately.

**Suggestion:**
```
Keep the cell to trigger mechanics and move the yield/pin-following behavior into a footnote or a short prose paragraph under the table, linked from the cell.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Skip gate depends on exact status description string</b> (Reported by: xai)</summary>

**File:** `.github/workflows/merge-recovery-soak.yml` (line 692)

The schedule gate only yields when `verdict_description` equals exactly "Full integration preflight completed with test failures", matching the finalize job contract. That coupling is documented, but any benign rewording of the status text will silently disable the skip and restore duplicate ~2.5h preflight runs. There is no test that asserts both sites stay in sync beyond comments.

**Suggestion:**
```
Consider a shared constant (e.g. env var or small script sourced by both steps) or matching on a stable suffix/code in the description so copy edits cannot drift the contract.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Update documentation</b> (Reported by: bedrock)</summary>

**File:** `docs/release-process.md` (line 642)

The documentation should be updated to reflect the new behavior of the `merge-recovery-soak.yml` workflow.

**Suggestion:**
```
Ensure the description accurately represents the changes made in the workflow.
```

</details>
---

<details>

The change is careful, well-commented workflow engineering: the pin-read step validates the target-tree ref strictly (safe charset, fail-closed on missing/duplicate/malformed keys, no output-injection surface), and the cron-fallback yield logic guards on run status, exact slot title, verdict description, and run-id cross-matching, failing open on query errors by explicit design. The main item needing human judgment is the deliberate trust-boundary shift: the 'trusted' integration harness checkout ref is now sourced from the dev-controlled target tree rather than the master-governed workflow literal, meaning dev write access can select any reachable commit (or mutable tag) in the system-integration repo for the harness. The PR mitigates this by keeping the secret-holding launcher on the workflow literal and disabling credential persistence, but a reviewer who knows the harness job's privileges should confirm the residual exposure is acceptable. Secondary feedback covers brittle string-matching between the finalize job and the schedule gate, confusing diagnostics for empty/duplicate env declarations, and the cost of failing open on transient API errors.

The code changes look functional and safe to merge. Minor improvements in comment formatting and documentation updates are suggested for better readability and clarity.

#### openai (gpt-5.6-sol)

The target-tree pinning and failed-preflight fallback are otherwise carefully constrained and documented, with no evident injection or performance issue. However, the new pin parser does not fully enforce its central exactly-once declaration contract: a valid pin followed by an empty duplicate is accepted. This should be corrected before merge so the pin step genuinely fails closed and remains consistent with release-evidence consumers.

This is a well-reasoned hotfix: the pin step fails closed on missing files, duplicate keys, and malformed refs, and the cron-fallback verdict check correctly fails open on API errors while requiring a four-part match (state, exact description, parseable run URL, membership in the failed-claim set) before yielding. The comments explaining intent and incident history are excellent. My main concerns are that accepting refs/tags from the target tree reintroduces a mutable reference into a checkout feeding a privileged pipeline (resolving the tag to a SHA before checkout would close this), and that the description-string contract between the finalize job and the schedule gate is enforced only by comments — a lint or context-name-based signal would make it self-enforcing. Neither blocks the merge given the stated urgency and the fail-open/fail-closed postures chosen, but both are worth follow-ups.

#### xai (grok-4.5)

Solid, intentional hotfixes: the target-tree pin step is fail-closed (missing file, duplicate keys, non-SHA/non-tag), keeps the workflow literal for the secrets-bearing launcher/drift check, and only redirects the integration harness checkout; the cron fallback correctly requires a failed claim run plus a matching failed preflight status on the same target and fails open on API errors. No security blockers for merge—supply-chain exposure is bounded to choosing existing system-integration refs from a branch that can already change that env file. Optional follow-ups are status API pagination, hardening the description-string contract, and aligning the tag regex with the repin script.
</details>

---

jeffrey-l-turner:
## Review remediation

Response to the multi-agent review above. One finding was a defect and is fixed in cc5953674. The three major findings describe design choices that this PR ports from dev unchanged. Each is assessed below.

### Fixed

| Finding | Status | Change |
|---|---|---|
| Critical: a valid pin followed by an empty `SYSTEM_INTEGRATION_REF=` line was accepted | Fixed in cc5953674 | The pin step now counts declaration lines in the env file instead of non-empty values. An empty duplicate is a second declaration and fails the step. |

Replay of the fixed step against local fixtures:

| Input | Result |
|---|---|
| One valid SHA, or one `refs/tags/` ref | accepted |
| Empty duplicate line | rejected, "exactly once" |
| Exact duplicate, or key absent | rejected, "exactly once" |
| Single empty value, or `main` | rejected, shape error |
| Env file missing | rejected, missing-file error |

The same parser is on dev from #418. The fixed hunk applies cleanly to dev's text, so the master-to-dev back-merge carries the fix without a conflict.

### Assessed, not changed

**Trust boundary shift (harness ref comes from the target tree).** This is the purpose of the PR. The soak job holds the OCI and OCIR secrets. That same job already builds and runs the target tree's node binary and scripts. Write access to the target branch already means arbitrary code in that job. Selecting the harness ref adds no new capability.

The trusted runner launcher runs before any target checkout and stays on the workflow literal.

**`refs/tags/` pins are mutable.** The CI pin-drift check in `ci.yml` already allows this shape for the workflow literal. The emergency repin path emits it. This PR does not widen it. Resolving a tag to a SHA before checkout is a reasonable follow-up on dev.

**Fail-open on a transient verdict query error.** This mirrors the slot-claim query in the same step, which has failed open since before #410. The worst case is one wasted soak session. That was the behavior before #410 for every failed slot. A retry with backoff is a reasonable follow-up on dev.

### Minor items

These items are follow-ups for dev:

- The description-string contract between the finalize job and the schedule gate.
- The status API pagination.
- The tag regex width.
- The release-process table cell length.

A hotfix branch is not the place to change behavior that dev already carries.

jltatbeach (approved):

PR #435, opened 2026-09-16
Fork choice bounded at the finalized floor, and bootstrap peer hygiene

spreston8:
Fork choice is scored from a floor instead of the approved block, plus the peer-hygiene and telemetry pieces that rode with it.

- Fork choice scores from the **fork-choice floor**: under θ ≥ 0 the highest per-block finalized floor among the latest messages, otherwise the approved block. Neither the LCA walk nor scoring descends below it. On a fully merged 3-wide DAG the candidate set never collapses, so the old genesis bound walked to the bottom and its cost grew with chain height — a heartbeat-only shard measured block processing climbing 192 → 1107 ms over 2.5 days in that regime. The new test pins it: the scored band is 6 blocks at both 10 and 30 heights, where the genesis bound grows 14 → 34, and the head is identical under both bounds.
- A latest message whose floor does not derive **abstains**. Absence already did; any other derivation error (an incompatible finalized fork, a lineage block absent from the store) now does too, instead of failing the whole snapshot — every propose and every validate on the node — over one validator's history.
- A malformed bootstrap node id is rejected at startup rather than parsed into a placeholder peer, and no placeholder peer is shipped in the shipped defaults.
- Arrival-depth telemetry: how deep below the live frontier each block lands once replayed, with a warn past the adopted parent-depth horizon where this node can no longer cite it. It is telemetry, so it never fails validation, and it reads the chain-adopted depth rather than the snapshot's.
- Dead fork-choice and snapshot residue deleted, including the second LUCA walk that ran in every snapshot.


spreston8 (commented):
Reviewed at ac3c904fbd9b90e16cd434742c3d3aa5caf5d8ac against dev. Suggestions below.

**Medium**

1. Worth widening the guard at `floor.rs:449` to `ftt.num <= 0`. `ft_witnessed_exact(strict = false)` is `(2q − S)/S ≥ θ`, so at exactly zero it reads `q ≥ S/2` — two disjoint halves clear it in different snapshots, which is the same bare-majority transient `IncompatibilityHold` documents for θ < 0 at floor.rs:326-336. Conflicting floors would make `lowest_universal_common_ancestor_many` return a non-ancestor, leaving the GHOST descent off-branch and the head at the floor. Not reachable at the shipped 0.67 or the shard's 0.1.
2. Consider naming the retention exception in the added R-LCA text. `fork_choice_floor` abstains on `BlockNotHeld` (floor.rs:462), so the floor is not identical across nodes holding different history — sound for a proposer-local choice, but "pure function of the DAG" reads as node-identical elsewhere in the spec.
3. Worth adding a `fn meta(b: &BlockMessage) -> BlockMetadata` test helper. `BlockMetadata::from_block(&genesis, false, None, None)` is expanded at roughly sixteen call sites across seven files, which buries the new dense-regime test in signature churn. The same helper covers the four sites in `blocks_response_api_test.rs` that inlined the deleted `tips` body.

**Low**

4. Consider giving `NodeIdentifier::new` (`peer_node.rs:38`) the same treatment as the parse path: it still carries `hex::decode(&name).unwrap()`. Unreachable today, since the remaining callers pass computed hex.

spreston8:
> Reviewed at [ac3c904](https://github.com/F1R3FLY-io/f1r3node-rust/commit/ac3c904fbd9b90e16cd434742c3d3aa5caf5d8ac) against dev. Suggestions below.
> 
> **Medium**
> 
> 1. Worth widening the guard at `floor.rs:449` to `ftt.num <= 0`. `ft_witnessed_exact(strict = false)` is `(2q − S)/S ≥ θ`, so at exactly zero it reads `q ≥ S/2` — two disjoint halves clear it in different snapshots, which is the same bare-majority transient `IncompatibilityHold` documents for θ < 0 at floor.rs:326-336. Conflicting floors would make `lowest_universal_common_ancestor_many` return a non-ancestor, leaving the GHOST descent off-branch and the head at the floor. Not reachable at the shipped 0.67 or the shard's 0.1.
> 2. Consider naming the retention exception in the added R-LCA text. `fork_choice_floor` abstains on `BlockNotHeld` (floor.rs:462), so the floor is not identical across nodes holding different history — sound for a proposer-local choice, but "pure function of the DAG" reads as node-identical elsewhere in the spec.
> 3. Worth adding a `fn meta(b: &BlockMessage) -> BlockMetadata` test helper. `BlockMetadata::from_block(&genesis, false, None, None)` is expanded at roughly sixteen call sites across seven files, which buries the new dense-regime test in signature churn. The same helper covers the four sites in `blocks_response_api_test.rs` that inlined the deleted `tips` body.
> 
> **Low**
> 
> 4. Consider giving `NodeIdentifier::new` (`peer_node.rs:38`) the same treatment as the parse path: it still carries `hex::decode(&name).unwrap()`. Unreachable today, since the remaining callers pass computed hex.
> 

All four addressed. Pushed as `053a2914f`.

**1. `ftt.num <= 0`** — done, at both sites. The premise checks out and the `agreeing > S/2` short-circuit does not cover it: `agreeing` is per-snapshot, the clique is the safety-relevant weight. With S=100 and four equal validators, three agreeing (75) with a certifying clique of 50 passes `q ≥ S/2`, and the disjoint pair does the same from another snapshot — two incompatible certificates, no equivocation. Pinned in `two_disjoint_half_cliques_both_certify_at_zero_but_not_above_it`, which also shows one ppm above zero forces overlap.

`floor_of_view`'s `IncompatibilityHold` guard (`floor.rs:411`) carries the same regime test, so it moved too — otherwise θ=0 raises `IncompatibleFinalizedFork` for a shape reachable without a protocol breach. The derivation still refuses to pick a base either way; only error-vs-quiet-hold changes.

Note for deployments: `docker/conf/standalone-dev.conf`, `run-local/conf/standalone.conf` and system-integration's `conf/standalone-dev.conf` ship θ=0, so they now score from the approved block. Left as-is deliberately — they are single-validator, where the narrow band matters least.

**2. R-LCA text** — done, and stated as the property that holds rather than as an exception. The floor is a **lower bound**: abstention lowers it along the same spine, never raises it, so only the band's width changes. Determinism clause 3 no longer lists the floor among the node-identical quantities. `an_abstaining_latest_message_lowers_the_floor_along_the_spine` pins all three claims and was red-checked — with abstention reverted it fails on `BlockNotHeld`.

**3. `meta()` helper** — not taken. The churn is real but a helper buys little for a one-call expansion; the readability problem was the fully-qualified path. `BlockMetadata` is now imported and the path spelled once, which collapses each call from four lines to one (`prop_bound.rs` alone drops 29 lines) and cleans the type positions in `compute_snapshot` as well.

**4. `NodeIdentifier::new`** — done. It applies the address parser's rule and returns it, and `from_address` now calls it rather than repeating the check. Making it fallible surfaced three fixtures building identifiers that are not 20 bytes, and one that hex-encoded a wire identifier only to parse it back.

Issue #244, opened 2026-08-12
Node gets OOMKilled at high concurrent load (cap≥75) after the jemalloc allocator fix from #146 — new memory ceiling, not the original lock-contention wall


9Neechan:
**Affected:** `casper/src/rust/engine/multi_parent_casper/snapshot.rs` (`prefer_deploy_support_main_parent` — confirmed proximate cause, see below); `casper/src/rust/finality/finalizer.rs` (`Finalizer::run`, specifically the `Finalizer::cannot_be_orphaned` >50%-agreement gate); `casper::rust::blocks::proposer::block_creator` (backpressure/fallback logic, "Ordinary user deploy fallback enabled" log line); `mntd-calibration` experiment (`experiment/mntd`); `mid` profile (2/8 CPU, 4Gi/16Gi RAM, 100Gi disk)

**Severity:** medium — a distinct failure mode from [#244](https://github.com/F1R3FLY-io/f1r3node-rust/issues/244)'s OOM crash-loop, which restarts the pod. Here the node never dies — block production and gossip continue — but finalization output is exactly zero for the entire measurement window. Confirmed fixable by disabling one function (`prefer_deploy_support_main_parent`), though that function guards against a separate liveness bug and cannot simply be deleted — see "Confirmed causal" below.

## The problem

Running `mntd-calibration` (multi-validator flood-propose load) at various caps, finalization sometimes stops completely — `FIN_B=0`, `TPS=0.00` — for the whole measurement window, while the node stays fully alive: `restart_count=0`, no crash, `min_peers(>=2)` checks pass throughout, blocks keep getting proposed and gossiped.

**This is non-deterministic, not gated by a fixed cap threshold.** Confirmed with a direct A/B: two `mntd-calibration` jobs launched ~24 seconds apart, same image, same profile, same cap grid — the same caps flipped between clean finalization and total stall:

| cap | run A | run B |
|---|---|---|
| 60 | stalled (`FIN_B=0`) | finalized (`TPS=1.58`) |
| 80 | finalized (`TPS=2.31`) | stalled (`FIN_B=0`) |
| 120 | barely finalized (1 block, `TPS=0.67`) | stalled (`FIN_B=0`) |
| 150 | stalled | stalled |

It has been observed to stall at caps as low as 45 and as high as 500, and to finalize cleanly at caps as high as 100, on different runs at the same cap — no single cap number is reliably "safe" or "always broken."

[...]

PR #438, opened 2026-09-17
fix(casper): resolve merge rejection groups independently to fix OOM and stall (#244, #268)

semyon-dev:
## What

- Share the merge pipeline's branches by reference. `Branch<R>` becomes
  `std::sync::Arc<HashableSet<R>>` throughout `conflict_set_merger` and
  `dag_merger`, and the alias is exported and used at the call sites.
- Resolve independent conflict groups separately. Two branches cannot
  influence each other's rejection unless one lists the other in the
  conflict map or their chains touch a common mergeable channel. Branch
  groups are made by union-find over those two relations, each group is
  enumerated on its own, and the results are combined through the same
  global selection key.

## Why

Under flood load the merge path builds branches carrying hundreds of deploy
signatures and copies them at every stage (`compute_branches`,
`compute_conflict_map`, rejection-option enumeration). That copy churn feeds
both high-load failures we are tracking:

- #268 — zero-finalization stall with the node still alive.
- #244 — OOM kill at cap≥75 (memory ceiling).

This PR does not close either issue. It removes the deep-copy overhead the
investigations keep running into.

## Scope

- `casper/src/rust/merging/conflict_set_merger.rs`
- `casper/src/rust/merging/dag_merger.rs`
- `casper/tests/merging/merge_number_channel_spec.rs`

Verified with `cargo check -p casper` and `cargo check -p casper --tests`.


spreston8 (commented):
Reviewed at 8d66bc053525401942de4deccc9381cb9eccbd6e against dev. Suggestions below.

**Medium**

1. Worth reconsidering the `Arc::try_unwrap` at `conflict_set_merger.rs:385`. It cannot take the fast path: `conflict_map` is built at line 192 from a set holding a clone of every branch — the documented contract requires every branch as a key — and is still read at line 421, so each strong count is at least 2 at line 385. An instrumented run of the merging suite fell back to clone 12 times out of 12, strong counts 3 and 5, never 1. Line 421 only needs a conflicts count, which could be taken at line 194 alongside `total_edges`, letting the map drop first and making the unwrap reachable; otherwise a plain `(*branch).clone()` says what actually happens.

2. Worth exporting the alias. `type Branch<R>` at `conflict_set_merger.rs:19` is private, so `dag_merger.rs` spells `std::sync::Arc<HashableSet<DeployChainIndex>>` twenty times and `merge_number_channel_spec.rs` three more. `pub type Branch<R>` used at those call sites would remove most of the re-indentation in this diff and shorten the closure signatures.

semyon-dev:
@spreston8 Could you do rereview please?

PR #444, opened 2026-09-18
fix(crypto): keep the leading zero byte of a 64-byte coordinate pair

jeffrey-l-turner:
normalize_public_key_coordinates stripped a leading zero byte from every input before it checked the length. A raw 64-byte X||Y pair whose X coordinate starts with 0x00 shrank to 63 bytes and was rejected. About one generated key in 256 has that shape, so
test_normalize_public_key_coordinates failed intermittently in CI on 2026-09-12, 2026-09-15, and 2026-09-18.

Strip the leading zero only when the input is longer than 64 bytes, which is the DER unused-bits case. Add a deterministic regression test with a fixed 64-byte pair whose first byte is zero, covering the raw, SEC1, DER, and unused-bits shapes.


jltatbeach (approved):

jeffrey-l-turner:
## Multi-Agent Code Review

**Reviewed by:**

| bedrock | `us.amazon.nova-pro-v1:0` | Abstain | 0 |
| openai | `gpt-5.6-sol` | Abstain | 0 |

---

### Summary

Correct fix for an intermittent normalize_public_key_coordinates failure: a raw 64-byte X||Y whose X starts with 0x00 must not lose that byte. Gating the leading-zero strip on len > 64 matches the DER unused-bits case and preserves significant coordinate bytes. The new deterministic test covers raw, SEC1, DER-style, and unused-bits-only shapes with a fixed leading-zero pair and is the right regression guard. No security, logic, or performance concerns in the diff; optional inline comment would help prevent recurrence.

The fix is correct and well-scoped: it resolves the intermittent CI failure where a raw 64-byte X||Y pair with a leading zero X-coordinate byte was incorrectly shrunk to 63 bytes and rejected, while preserving the intended stripping of the DER unused-bits prefix byte from 65-byte inputs. The deterministic regression test thoroughly covers the raw, SEC1, DER, and unused-bits shapes. Only minor suggestions regarding condition precision and a clarifying comment on the synthetic test vector.

---

### Issues Found

<details>
<summary><b>:yellow_circle: Minor: Test vector is not a valid curve point</b> (Reported by: openrouter)</summary>

**File:** `crypto/tests/util/certificate_helper_spec.rs` (line 249)

The deterministic test uses X = 0x00...00 and Y = 0x00...01, which is not a point on secp256r1/secp256k1. Since normalize_public_key_coordinates only performs shape normalization, this works, but a brief comment noting the bytes are a synthetic shape fixture (not a real key) would prevent future confusion.

</details>

<details>
<summary><b>:white_circle: Suggestion: Brief comment on why len > 64 guards the strip</b> (Reported by: xai)</summary>

**File:** `crypto/src/rust/util/certificate_helper.rs` (line 398)

The fix is correct, but the subtlety (leading 0x00 is DER unused-bits only when longer than a raw X||Y pair; in a 64-byte pair it is a significant coordinate byte) is easy to re-break without context. A one-line comment at the condition would lock in the intent next to the code.

**Suggestion:**
```
Add a short comment above the condition, e.g. // Strip DER BIT STRING unused-bits (leading 0x00) only when longer than raw 64-byte X||Y; a leading zero in a 64-byte pair is part of the coordinate.
```

</details>

<details>
<summary><b>:white_circle: Suggestion: Condition strips any overlong input's leading zero, not just the DER unused-bits byte</b> (Reported by: openrouter)</summary>

**File:** `crypto/src/rust/util/certificate_helper.rs` (line 398)

The new guard strips a leading 0x00 from any input longer than 64 bytes (e.g., a 100-byte malformed input starting with 0 would still be stripped, then rejected by the length check anyway). This is harmless in practice because the subsequent length check rejects anything that doesn't reduce to exactly 64 bytes, but the intent could be expressed more precisely as `public_key.len() == 65 && public_key.first() == Some(&0)` if the only legitimate overlong case is the single DER unused-bits prefix byte.

**Suggestion:**
```
Optionally tighten the condition to `public_key.len() == 65 && public_key.first() == Some(&0)` to match the documented DER unused-bits case exactly.
```

</details>
---

<details>

> **ABSTAIN**: AWS Bedrock API call failed (exit code 127): AWS CLI not installed. Install with: brew install awscli (macOS) or apt install awscli (Linux)

#### openai (gpt-5.6-sol)

> **ABSTAIN**: insufficient_quota: You have no credits remaining. Add credits to continue using the API at https://platform.openai.com/settings/organization/billing/.

The fix is correct and well-scoped: it resolves the intermittent CI failure where a raw 64-byte X||Y pair with a leading zero X-coordinate byte was incorrectly shrunk to 63 bytes and rejected, while preserving the intended stripping of the DER unused-bits prefix byte from 65-byte inputs. The deterministic regression test thoroughly covers the raw, SEC1, DER, and unused-bits shapes. Only minor suggestions regarding condition precision and a clarifying comment on the synthetic test vector.

#### xai (grok-4.5)

Correct fix for an intermittent normalize_public_key_coordinates failure: a raw 64-byte X||Y whose X starts with 0x00 must not lose that byte. Gating the leading-zero strip on len > 64 matches the DER unused-bits case and preserves significant coordinate bytes. The new deterministic test covers raw, SEC1, DER-style, and unused-bits-only shapes with a fixed leading-zero pair and is the right regression guard. No security, logic, or performance concerns in the diff; optional inline comment would help prevent recurrence.
</details>

---

jeffrey-l-turner:
### Remediation: minor finding "Test vector is not a valid curve point"

**File:** `crypto/tests/util/certificate_helper_spec.rs`, test `test_normalize_public_key_coordinates_keeps_leading_zero_coordinate_byte`

**Assessment:** The observation is correct. The 64-byte vector has X = 0 and Y = 1. That pair is not a point on secp256k1 or on P-256.

**Decision:** No code change. The vector is synthetic by design.

- `normalize_public_key_coordinates` only reshapes bytes. It strips an optional DER unused-bits byte and an optional SEC1 `0x04` prefix, then checks that 64 bytes remain. It never parses or validates a curve point.
- The test must exercise a leading `0x00` X byte deterministically. A real key has that shape about one time in 256, which is the cause of the intermittent CI failure this PR fixes.
- The repository style rule forbids inline code comments unless a user requests them, so the suggested clarifying comment is not added. The test name states the intent instead.

The two suggestions on the `len() > 64` guard are also left as is. The guard matches the DER unused-bits case, which is the only overlong input the certificate parser produces, and the existing tests cover the DER, SEC1, and raw shapes.

